PR #35, opened 2026-03-06
fix: Delegate server.sh to systemctl when systemd manages the service

nbramia:
## Summary
- `server.sh start/stop/restart` now detect when `lifeos-api` is managed by systemd and delegate to `systemctl` instead of manually spawning processes via `nohup`. This prevents ghost server processes that escape systemd's cgroup tracking and cause port conflicts + crash-loops.
- Adds a `sudoers` rule to `setup-systemd.sh` so the nightly sync can restart the server without a password prompt.

## Test plan
- [x] `server.sh restart` — delegates to systemd, no ghost processes, systemd PID matches port owner
- [x] `server.sh stop` — delegates to systemd, port freed
- [x] `server.sh start` — delegates to systemd, waits for healthy
- [x] Full test suite: 1755 passed, 12 skipped

### DIFF
--- a/scripts/server.sh
+++ b/scripts/server.sh
@@ -259,17 +259,46 @@ show_status() {
     echo ""
 }
 
+# Check if systemd is managing the service (avoid ghost processes)
+is_systemd_managed() {
+    systemctl is-active lifeos-api.service &>/dev/null ||
+    systemctl is-enabled lifeos-api.service &>/dev/null
+}
+
+# Run systemctl command with sudo (passwordless via sudoers rule from setup-systemd.sh)
+sctl() {
+    sudo systemctl "$@"
+}
+
 # Main
 case "${1:-status}" in
     start)
-        start_server
+        if is_systemd_managed; then
+            log_info "Delegating to systemd..."
+            kill_server  # Clear any ghost processes holding the port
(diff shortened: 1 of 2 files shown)


nbramia (commented):
## PR Review: fix: Delegate server.sh to systemctl when systemd manages the service

### Recommended
- **[Correctness]** `sctl start` could fail if a ghost process (from a previous manual `nohup` launch) is still holding port 8000. Added `kill_server` before `sctl start` to clear ghosts. *(Fixed)*
- **[Security]** `setup-systemd.sh` wrote the sudoers file directly without `visudo -c` validation — a syntax error could brick sudo. Added temp file + `visudo -c -f` validation before installing. *(Fixed)*

### Minor
- **[Correctness]** `is_systemd_managed()` used redundant `&>/dev/null 2>&1`. Cleaned to `&>/dev/null`. *(Fixed)*
- **[Security]** `sctl()` initially tried `systemctl` without sudo first, triggering polkit password prompts. Simplified to always use `sudo systemctl`. *(Fixed)*

### Summary
All findings addressed. The PR correctly solves the ghost process problem by delegating to systemd when the service is managed, and the sudoers rule is safely installed with validation. Merge-ready.

PR #52, opened 2026-03-08
fix: Prevent OOM crashes with memory-gated sync and swap

### LINKED ISSUE #50: fix: Prevent OOM crashes with memory management hardening
## Problem

The system suffers two classes of OOM crashes on the Ryzen AI MAX+ 395 (unified memory architecture):

1. **Daytime crashes**: 96 GB allocated to GPU in BIOS leaves only ~30 GB for system RAM. Normal dev work + LLM inference can exceed total available memory.
2. **Nightly sync crashes**: The embedding model (gte-Qwen2-1.5B) loads while llama-server still holds ~59 GB VRAM. No coordination exists — sync runs embeddings and LLM concurrently with no memory gating or model unloading.

## Context

- Hardware: Ryzen AI MAX+ 395, 96 GB unified memory (BIOS allocates GPU vs CPU split)
- Current BIOS: 96 GB → GPU, ~30 GB visible to CPU
- LLM: gpt-oss-120b MXFP4 quant (~59 GB VRAM) — not FP16/Q8, so model size isn't the issue
- Embedding model: gte-Qwen2-1.5B-instruct — can't coexist with the 60 GB LLM (noted in MEMORY.md)
- Sync pipeline (`run_all_syncs.py`) does NOT stop llama-server before loading embeddings — no `systemctl stop lifeos-llm`, no memory check, no sequencing
- No swap configured, no cgroups — OOM killer takes out the whole session

### 1. Reduce BIOS GPU allocation to 80 GB
The MXFP4 model needs ~59 GB. 80 GB gives ~21 GB headroom for GPU overhead while freeing ~16 GB back to system RAM (~46 GB total CPU-visible). 80 GB also accommodates the largest planned model (gpt-oss-120b); smaller models like Qwen3-32B (#49) would have even more headroom. This alone likely fixes daytime crashes.

### 2. Fix nightly sync model swap sequencing (memory-gated)
Before embedding phases in `run_all_syncs.py`:
- **Check available GPU memory** — only stop llama-server if insufficient for embeddings
- If stop needed: stop llama-server (`systemctl stop lifeos-llm`)
- Poll until VRAM is actually freed (check `free -h` or `/sys/class/drm/*/mem_info_vram_used`)
- Only then proceed with embedding operations
- **After sync**: restart llama-server **if it was running before sync started** (track this independently — do NOT use the #48 autostart toggle for this, since a user may have manually started the LLM with autostart off)

This approach is model-aware: with Qwen3-32B (~20 GB, #49) + embeddings (~3 GB) on 80 GB, the stop/start is skipped entirely. With gpt-oss-120b (~59 GB), it kicks in.

Consider using `ExecStartPre=` in the sync systemd service as an additional memory gate safeguard.

### 3. Set up 8 GB swap on NVMe
Safety net against OOM kills. Will be slow if actually used, but the machine degrades gracefully instead of crashing. zram is another option for ~20-30% effective RAM increase via compression.

### 4. Use cgroups to cap dev/terminal processes
Create a cgroup for dev work (e.g., 16 GB limit). Test suites and builds fail gracefully or slow down instead of triggering OOM kills on the entire session.

## Files involved

- `scripts/run_all_syncs.py` — add memory-gated LLM stop/start around embedding phases
- `config/systemd/lifeos-sync.service` — potentially add `ExecStartPre=` memory gate
- `scripts/setup-systemd.sh` — swap/zram setup, cgroup configuration
- `docs/guides/installation.md` — document BIOS allocation recommendation

## Acceptance criteria

- [ ] BIOS GPU allocation reduced to 80 GB, system shows ~46 GB CPU-visible RAM
- [ ] Nightly sync checks available memory before embedding phases
- [ ] Nightly sync only stops llama-server when memory is insufficient (model-aware)
- [ ] Nightly sync restores llama-server to its pre-sync state (running → restart, stopped → leave stopped)
- [ ] 8 GB swap file configured on NVMe
- [ ] Dev cgroup configured with memory cap
- [ ] No OOM crashes during normal daytime usage
- [ ] No OOM crashes during nightly sync

## Related

- #48 — Add toggle to disable local LLM auto-start/restart (controls systemd boot/crash behavior; sync stop/start is independent)
- #49 — Add Qwen3-32B as a local model option (smaller model may eliminate need for sync stop/start entirely)

nbramia:
## Summary

- Sync pipeline checks GPU VRAM before embedding phases; stops llama-server only if insufficient memory, restarts it afterward
- Adds `lifeos-llm` to sudoers for passwordless `systemctl stop/start` from sync scripts
- Creates 8 GB swap file as OOM safety net during `setup-systemd.sh`
- Documents recommended BIOS GPU allocation (80 GB) and optional cgroups for dev processes
- Suppresses pre-existing E402 lint errors in `scripts/` (intentional early `load_dotenv`)

Closes #50

## Test evidence

```
1085 passed, 939 deselected, 407 warnings in 297s
```

## Review focus

- Memory-gating logic in `run_all_syncs.py` — VRAM threshold, stop/start sequencing, safety-net restart
- Sudoers rule additions — minimal privilege for sync pipeline
- Swap file creation in `setup-systemd.sh` — idempotent, fstab-aware

### DIFF
--- a/scripts/run_all_syncs.py
+++ b/scripts/run_all_syncs.py
@@ -137,6 +137,10 @@ def _start_llm() -> None:
 _active_run_id: int | None = None
 _active_source: str | None = None
 
+# Track LLM state across signal handlers (module-level for SIGTERM access)
+_llm_stopped_for_sync: bool = False
+_llm_was_running_before_sync: bool = False
+
 
 def _handle_sigterm(signum, frame):
     """Clean up sync_health.db on SIGTERM so we don't leave stale RUNNING rows."""
@@ -149,6 +153,9 @@ def _handle_sigterm(signum, frame):
             )
         except Exception:
             pass  # Best-effort — DB may be locked
+    # Restore LLM if we stopped it
+    if _llm_stopped_for_sync and _llm_was_running_before_sync:
+        _start_llm()
     sys.exit(128 + signum)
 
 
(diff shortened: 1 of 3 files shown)


nbramia (commented):
## PR Review: fix: Prevent OOM crashes with memory-gated sync and swap

### Action Required

- **[Correctness + Security + Requirements]** SIGTERM/crash while LLM is stopped leaves it down permanently. The `_handle_sigterm` handler (line ~141) does not restart the LLM. If the sync process is killed, OOM-killed, or crashes during an embedding phase after `_stop_llm_for_embeddings()`, the LLM stays down until manual intervention. The safety-net at line ~964 only runs on normal exit. **Fix:** (a) Add LLM restart to `_handle_sigterm`; (b) wrap the sync loop in `try/finally` so the safety-net always fires; (c) consider a sentinel file (`data/.llm_stopped_by_sync`) so the watchdog timer can recover from SIGKILL/OOM-kill.

- **[Correctness]** Memory-gating check runs BEFORE skip guards — LLM may be stopped needlessly. The memory check at line ~881 runs before the disabled-source, monthly-check, recently-synced, and dependency checks (lines ~897–931). If `vault_reindex` triggers the LLM stop but is then skipped (disabled, recently synced, dependency-failed), and `crm_vectorstore` is also skipped, the LLM remains stopped for the entire remainder of the sync. The safety net catches this eventually, but the LLM is needlessly down. **Fix:** Move the memory-gating block to after all skip/continue guards, immediately before the `run_sync()` call.

- **[Standards]** No tests for ~83 lines of new logic. The PR adds 4 helper functions and ~35 lines of integration logic. The existing `tests/test_run_all_syncs.py` has mock infrastructure (`_run_with_patches`) that makes this straightforward to test. Minimum coverage needed: (a) LLM stopped when memory insufficient, (b) LLM NOT stopped when sufficient, (c) LLM restarted after embedding phases, (d) safety-net fires on missed restart, (e) internal `_llm_*` keys cleaned from results.

### Recommended

- **[Correctness]** Use `enumerate(sources)` in the loop header instead of `sources.index(source)` at line ~942 — avoids O(n) lookup and removes the redundant `if source in sources else -1` guard.
- **[Correctness]** `_get_available_gpu_memory_mb()` returns the first `/sys/class/drm` entry with VRAM files. Use `sorted(Path(...).glob("card[0-9]*"))` and skip `card0-DP-1` etc. for robustness.
- **[Correctness]** Use local variables (`llm_stopped`, `llm_was_running`) instead of stuffing state into the `results` dict — avoids leaking internal keys if an exception fires between set and cleanup.
- **[Security]** Sudoers `Runas` spec is `ALL=(ALL)` — should be `ALL=(root)` since these commands only need root.
- **[Standards]** Log message at line ~887 prints `None MB` when sysfs read fails — use `available or 'unknown'` for cleaner logs.

### Minor

- **[Standards]** VRAM poll loop magic numbers (10 retries × 2s) could be named constants.
- **[Standards]** One-shot guard `"_llm_stopped" not in results` should have an inline comment explaining why re-evaluation is intentionally skipped.
- **[Security]** Pre-existing: `eval echo "~$REAL_USER"` in setup-systemd.sh (line 33) is an injection vector. Not introduced by this PR, but now higher-impact since `$REAL_USER` flows into sudoers. Consider replacing with `getent passwd "$REAL_USER" | cut -d: -f6`.
- **[Correctness]** `_stop_llm_for_embeddings()` returns `True` even if VRAM isn't freed within 20s — acceptable (service IS stopped), but worth noting.
- **[Requirements]** Issue suggested `ExecStartPre=` in sync service as defense-in-depth — not implemented, but the in-process check is the primary mechanism.

### Summary

The core approach is sound — checking actual VRAM availability before deciding whether to stop the LLM is the right design. The three Action Required items need fixing before merge: the SIGTERM/crash recovery gap is the most critical (LLM down for hours unattended), the skip-guard ordering is a likely real-world scenario, and the new logic needs test coverage. The Recommended items are quality improvements that would make the code more robust.

PR #89, opened 2026-05-23
refactor: delete dead LLM-routing scaffolding

### LINKED ISSUE #86: refactor: delete dead LLM-routing scaffolding
## Problem

Three parallel code paths look like they implement features (Ollama-based query routing, per-query model-tier selection, lifeos-llm.service swap-out for embeddings) but none are actually wired up in production. They make the code harder to read and lead to documentation drift — PR #85 had to correct several user-facing docs that had been built around those imaginary features.

## Context

This was surfaced while tracing the live Telegram chat path (see PR #85's investigation). PR #85 corrected the docs to match reality; this issue cleans up the code so the docs and code agree.

The three offenders:

1. **`api/services/chat_helpers.py::_classify_via_ollama`** — defined at line 698 but never called by the live path. `classify_action_intent` (line 657) calls `_classify_via_haiku` and falls back to regex patterns, not to Ollama. Only `tests/e2e/test_telegram_reminders.py` references it (via mocks).

2. **`api/services/model_selector.py` tier exports** — `get_claude_model_name` and the `CLAUDE_MODELS` dict (which still has a stale comment about Haiku 4.5 not being available). `agent_loop.py:91` and `synthesizer.py:116` both explicitly document `model_tier` as ignored. Only `tests/test_model_selector.py` exercises these.

3. **`scripts/run_all_syncs.py` `lifeos-llm.service` branch** in `_stop_llm_for_embeddings` and `_start_llm` — the systemd unit is disabled and inactive on production, the API service no longer `Wants=` it (commit referenced in [MEMORY.md](../master/MEMORY.md) Key Lessons), and the user explicitly stopped relying on it. The ollama unload/warm logic added in PR #83 is what actually runs.

## Suggested approach

One PR, three independent commits or one squashed:

- Delete `_classify_via_ollama` and its mocks. Audit whether `api/services/ollama_client.py` has any non-chat callers worth keeping (e.g. inside `summarizer.py` or `person_facts.py`); remove the module if not.
- Delete `get_claude_model_name`, `CLAUDE_MODELS`, and `classify_query_complexity` (or, alternatively, actually wire `classify_query_complexity`'s result through to the Anthropic client — but the spirit of this issue is "remove what isn't load-bearing"). Update `tests/test_model_selector.py` to drop the deleted symbols' tests; keep any tests that still cover live behavior.
- Strip the `lifeos-llm.service` branch from `_stop_llm_for_embeddings`/`_start_llm`. Keep the ollama unload/warm logic intact. Remove the now-unused systemctl `sudo` NOPASSWD entry from the user's sudoers file (separate manual step, document in PR description).

After the cleanup, document in the README and AGENTS.md that `LIFEOS_ANTHROPIC_MODEL` is the single knob for orchestrator-model choice (currently implied but not stated).

## Files involved

- `api/services/chat_helpers.py`
- `api/services/ollama_client.py` (audit, possibly delete)
- `api/services/model_selector.py`
- `scripts/run_all_syncs.py` (functions `_stop_llm_for_embeddings`, `_start_llm`, and the `_lifeos_llm_was_stopped_by_us` state flag)
- `tests/test_model_selector.py`
- `tests/e2e/test_telegram_reminders.py`
- `tests/test_chat_helpers.py` (if it exercises the deleted classifier)
- `README.md` / `AGENTS.md` for the `LIFEOS_ANTHROPIC_MODEL` documentation

## Acceptance criteria

- [ ] `_classify_via_ollama` is removed; no production call site for `OllamaClient` remains in the chat path
- [ ] `get_claude_model_name` and `CLAUDE_MODELS` are removed (or, if kept, are actually used by the agent loop)
- [ ] `_stop_llm_for_embeddings` and `_start_llm` no longer reference `lifeos-llm.service`
- [ ] Full unit test suite passes (`./scripts/test.sh`)
- [ ] A manual sync run completes cleanly (`run_all_syncs.py --source vault_reindex --execute`)
- [ ] README/AGENTS.md state that `LIFEOS_ANTHROPIC_MODEL` is the single orchestrator-model knob

Estimated effort: ~3 hours including tests.

nbramia:
## Summary

Closes #86. Three parallel code paths looked like they implemented features but none were wired into production. PR #85 corrected the docs to match reality; this PR cleans up the code so the docs and code agree.

Net: **-631 lines** (89 added, 720 removed).

## Changes

1. **`_classify_via_ollama` removed** from `api/services/chat_helpers.py`. The live intent classifier is `_classify_via_haiku` with a regex fallback; the Ollama variant was only mocked in tests. Its `model_tier="haiku"` argument is dropped too since `synthesizer.py:116` documents it as ignored.

2. **`api/services/model_selector.py` deleted** — `CLAUDE_MODELS`, `get_claude_model_name`, `classify_query_complexity`, `ModelSelector` class, `get_model_selector`, `reset_model_selector`. The agent loop (`agent_loop.py:91`) and synthesizer both ignore `model_tier`, so the tier classifier never actually changed the model. `tests/test_model_selector.py` deleted. `tests/reset_singletons.py` no longer resets the gone-singleton. `api/routes/chat.py` drops the `model_select` perf span and labels the routing event with `LIFEOS_ANTHROPIC_MODEL` directly. `api/services/query_router.py` drops `recommended_model` / `complexity_score` fields on `RoutingResult` (no external readers) and stops calling `classify_query_complexity`.

3. **`lifeos-llm.service` branch removed** from `_stop_llm_for_embeddings` and `_start_llm` in `scripts/run_all_syncs.py`. The service is disabled on production and `lifeos-api.service` no longer `Wants=` it. The ollama unload/warm logic (PR #83) is what actually runs and stays intact. Also drops the `_lifeos_llm_was_stopped_by_us` flag and simplifies `_is_llm_running` to just check ollama VRAM.

4. **E2E tests retargeted** in `tests/e2e/test_telegram_*.py` from `_classify_via_ollama` to `_classify_via_haiku`. They were mocking the dead path, so the tests were effectively making real Anthropic calls — now they actually mock the live classifier.

5. **README + AGENTS.md** state that `LIFEOS_ANTHROPIC_MODEL` is the single orchestrator-model knob with no per-query tiering.

6. **Test import retargeting + ruff cleanups** on touched files: `tests/test_adaptive_retrieval.py` and `tests/test_chat_api.py` now import `extract_search_keywords` / `detect_compose_intent` from `api.services.chat_helpers` (their actual source) instead of via the `api.routes.chat` re-export that ruff (correctly) removed. Pre-existing pre-`pytestmark` imports cleaned up in `test_chat_api.py`.

## Manual follow-up

The sudoers `NOPASSWD` entry for `systemctl start/stop lifeos-llm.service` on the production machine can be removed manually — this PR doesn't touch sudoers since that's a per-installation concern.

## Review focus

- `scripts/run_all_syncs.py` — confirm the simplified `_stop_llm_for_embeddings` / `_start_llm` still cooperates correctly with the SIGTERM handler at line ~226 (the safety-net restart logic in the orchestrator's main loop at lines ~1080–1094 still uses the same `_llm_stopped_for_sync` / `_llm_was_running_before_sync` flags, which I left intact).
- `api/services/query_router.py` — verify that removing `recommended_model`/`complexity_score` from `RoutingResult` doesn't break any consumers I missed (grep showed only `chat.py:1084` which is in the same PR).

## Test plan

- [x] Pre-push test suite: 1140 passed, 1039 deselected
- [x] Manual import check: `_is_llm_running()`, `_stop_llm_for_embeddings`, `_start_llm`, `_ollama_unload_all`, `_ollama_warm` all importable and callable
- [x] Touched-file ruff check: clean


nbramia (commented):
## Adversarial review — Round 1

### Action Required

- **[Doc drift introduced by this PR]** `docs/specs/technical/observability.md:28` still lists the `model_select` perf span as instrumented, but this PR removed the `with trace_span("model_select"):` block from `api/routes/chat.py`. The current cell text — "records a recommended tier in the trace; the agent loop currently uses the configured `LIFEOS_ANTHROPIC_MODEL` regardless" — is now flat-out wrong; the span no longer emits at all. Either drop that row from the table or rewrite it to reflect the new world where `model_tier` is just a plain assignment, not a traced span. (This is exactly the docs/code drift the PR claims to prevent.)

### Recommended

- **[Behavior regression for `LIFEOS_LOCAL_LLM_AUTOSTART=true` users]** `scripts/setup-systemd.sh:101-108` still installs and enables `lifeos-llm.service` whenever `LLM_AUTOSTART=true`, but `_is_llm_running()` (line 139) and `_stop_llm_for_embeddings()` (line 178) no longer notice that service. A fresh downloader who sets `LIFEOS_LOCAL_LLM_AUTOSTART=true` will get llama-server pinning ~60 GB VRAM, then on the first nightly sync `_is_llm_running()` returns False (it only sees ollama), the GPU-memory check at line 1054 is skipped, and embeddings race llama-server for VRAM. Either (a) actually delete the service file + autostart support in this PR, (b) re-add a minimal `is-active lifeos-llm.service` probe inside `_is_llm_running()` so the existing GPU-memory check still fires (without the sudo stop/start branch), or (c) document loudly in the PR + README that `LIFEOS_LOCAL_LLM_AUTOSTART=true` is now unsupported and the service file should be removed in a follow-up. Per the project's "open-source: nothing broken-by-default for a fresh downloader" rule (MEMORY.md), I'd push for (a) or (b).

- **[`_is_llm_running` reduced fidelity]** Same root cause as above: the function used to short-circuit-True when `lifeos-llm.service` was active. Now it only checks ollama. Any non-ollama VRAM hog (lifeos-llm.service, or someone running llama-server manually for `LIFEOS_LLM_BACKEND=local`) is invisible. The GPU-memory probe at line 1054 only runs if `_llm_was_running_before_sync` is True, so for these users the sync silently proceeds with insufficient VRAM and falls back to the CPU path documented in MEMORY.md (10x slower). Consider replacing the body with a direct GPU-memory check (`_get_available_gpu_memory_mb() < threshold`) so the LLM-detection is implicit and source-agnostic.

### Minor

- **[Semantic change in `model_tier` value]** `api/routes/chat.py:1068` now assigns the full model id (e.g. `claude-haiku-4-5`) to `model_tier`, where it used to be a tier label (`haiku`/`sonnet`/`opus`). This value is persisted to `perf_traces.db` and emitted in the SSE `routing` event payload (`'reasoning': f'Agentic loop ({model_tier})'`). No live consumer filters on the value (frontend only reads `data.sources`, perf API doesn't query by tier), so this is non-breaking — but historical perf-trace data will have mixed tier-label vs model-id values. Worth a one-line note in the PR description, or rename the local var to `model_name` for clarity since "tier" no longer fits.

- **[Stale comment in `_start_llm`]** `scripts/run_all_syncs.py:209-211` says "We don't use settings.ollama_model because the actually-loaded model may differ (e.g. a heavier summarizer model running alongside a smaller router)." With the lifeos-llm branch gone, the function only does the ollama warm — the comment is still correct but reads oddly given how short the function is now. Optional cleanup.

### Verified clean

- All deleted symbols (`_classify_via_ollama`, `classify_query_complexity`, `get_claude_model_name`, `CLAUDE_MODELS`, `ModelSelector`, `get_model_selector`, `reset_model_selector`, `_lifeos_llm_was_stopped_by_us`, `recommended_model`, `complexity_score`) have zero surviving references across `*.py`/`*.md` (excluding `__pycache__`).
- `OllamaClient` is still used by `api/services/person_facts.py` (lines 1057, 1285), `api/services/query_router.py` (line 19), and `scripts/sync_entity_cleanup.py` (line 331) — all intact.
- `_classify_via_haiku(prompt)` exists at `api/services/chat_helpers.py:698` with the new signature; the call site at line 685 passes only `prompt`; `synthesizer.get_response(prompt, max_tokens=64)` accepts that — `model_tier` is an optional kwarg.
- SIGTERM handler at `scripts/run_all_syncs.py:226-240` still uses `_llm_stopped_for_sync` and `_llm_was_running_before_sync` (both still defined at lines 222-223) and calls `_start_llm()` (still defined at line 205). Safety net at lines 1090-1094 also still well-formed.
- `_stop_llm_for_embeddings()` return-type contract preserved: returns `True` on at least one successful unload (line 192 / line 202), `False` if nothing was loaded (line 191). The main-loop branch at line 1058-1059 still reads it correctly.
- All e2e test mock targets (`_classify_via_haiku`) resolve to the live code path. `tests/test_run_all_syncs.py` and `tests/e2e/test_telegram_*` all pass (22+214 tests green locally).
- `tests/test_adaptive_retrieval.py` and `tests/test_chat_api.py` retargeted imports resolve correctly — `extract_search_keywords` lives at `chat_helpers.py:84`, `detect_compose_intent` at `chat_helpers.py:186`.
- Frontend `web/index.html:3551` only reads `data.sources` from the routing SSE event, not `reasoning`/`model_tier` — safe.

### Summary

Solid dead-code cleanup; call-site audit is clean and tests pass. Two real concerns that need attention before merge: (1) the `model_select` span entry in observability.md needs to be removed in this same PR — it's literally the type of docs/code drift this PR exists to fix; (2) `LIFEOS_LOCAL_LLM_AUTOSTART=true` is still a documented, installable code path, and the simplified `_is_llm_running`/`_stop_llm_for_embeddings` no longer cooperate with it — either remove that path entirely or restore minimal detection so the existing memory check fires.

nbramia:

| 1 | `docs/specs/technical/observability.md:28` still lists `model_select` perf span | Action Required | **Accept** | The PR's stated purpose is to make docs match code; missing this is the exact failure mode the PR is meant to prevent. |
| 2 | `LIFEOS_LOCAL_LLM_AUTOSTART=true` regression — setup-systemd.sh installs `lifeos-llm.service`, but the PR removed the code that coordinated with it | Recommended | **Accept (upgrade to Action Required)** | Open-source friendliness contract (MEMORY.md): "no hardcoded personal values, machine-specific tuning, or anything broken-by-default for a fresh downloader." A downstream user enabling autostart + `LIFEOS_LLM_BACKEND=local` would have llama-server racing the embedding model for VRAM. |
| 3 | `_is_llm_running` reduced fidelity — only sees ollama, misses manually-started llama-server | Recommended | **Accept** | Will combine fix with #2: replace LLM-detection with a direct GPU-memory threshold check (source-agnostic), so it covers ollama, llama-server, and any other GPU-resident LLM. |
| 4 | `model_tier` variable at chat.py:1068 now stores model id, not tier label | Minor | **Accept** | Rename to `orchestrator_model` for accuracy. |
| 5 | Stale comment in `_start_llm` about "summarizer model running alongside a smaller router" | Minor | **Accept** | Comment is misleading after the function was simplified. Quick cleanup. |

**Findings forwarded to addresser:** 5 (4 Action Required after upgrade, 1 Minor cleanup bundled with the others).

nbramia:
## Review Round 1 — Feedback Addressed

| 1 | `docs/specs/technical/observability.md:28` stale `model_select` perf-span entry | Applied | Removed the row from the Instrumented Stages table. Commit `8279afd`. |
| 2 | `LIFEOS_LOCAL_LLM_AUTOSTART` regression for downstream users | Applied (combined with #3) | See #3. |
| 3 | `_is_llm_running` reduced fidelity (only saw ollama) | Applied | Rewrote `_is_llm_running` to check available GPU memory against `_EMBEDDING_MEMORY_THRESHOLD_MB` directly — source-agnostic, detects ollama, llama-server (lifeos-llm.service), or anything else holding VRAM. Orchestrator's existing post-stop check handles the case where `_stop_llm_for_embeddings` can only unload ollama: if VRAM is still pinned, embeddings fall back to CPU (existing behavior). Also tidied the stale "summarizer/router" comment in `_start_llm`. Commit `4ccd7fd`. |
| 4 | `model_tier` variable misnamed (stores model id now, not tier label) | Applied | Renamed to `orchestrator_model` and updated the routing-event label string. The `perf_traces.db` column name stays `model_tier` (schema concern) but the comment now documents the value-shape change. Commit `737240e`. |
| 5 | Stale comment in `_start_llm` | Applied | Bundled into commit `4ccd7fd`. |

**Tests:** `./scripts/test.sh` (via pre-push hook) — 1140 passed, 1039 deselected
**Commits:** `8279afd`, `4ccd7fd`, `737240e`

nbramia:
## Review Round 2 — Verification

Independently verified Round 1 fixes against the current branch HEAD:

| # | Finding | Verified fixed? | Evidence |
|---|---------|-----------------|----------|
| 1 | observability.md `model_select` row | ✓ | `grep -nE "model_select" docs/specs/technical/observability.md` → empty |
| 2+3 | `_is_llm_running` source-agnostic | ✓ | New impl checks `_get_available_gpu_memory_mb() < _EMBEDDING_MEMORY_THRESHOLD_MB`; covers ollama, llama-server, anything else |
| 4 | `model_tier` → `orchestrator_model` | ✓ | Renamed at chat.py:1069 and propagated through agent_loop call (line 1099), routing event (line 1091), and assistant message routing metadata (line 1155). Perf-trace column name `model_tier` retained (schema concern); comment documents the value-shape change. |
| 5 | `_start_llm` stale comment | ✓ | Old "summarizer/router" wording gone; only remaining `lifeos-llm.service` reference is in `_is_llm_running` docstring as an illustrative example, which is appropriate. |

No new findings. No regressions introduced by the fixes. **Review loop complete.**

Issue #98, opened 2026-05-26
epic: external agent worker for #agent-tagged tasks (A–F)


nbramia:
> **Status:** Epic / tracking issue. Implementation is split across Issues A–F below.

## Goal

Delegate tasks to AI agents by tagging them `#agent` in the LifeOS task list. An external worker picks them up, executes via Claude Opus (Anthropic Managed Agents) or local Gemma (llama-server), and notifies completion via Telegram with a one-paragraph summary. Agents can spawn each other, coordinate, ask the user clarifying questions, and respect per-task + per-day budgets.

This unlocks "drop something into the task list, walk away, come back to a result" — research, drafting, multi-step lookups, anything tedious but non-urgent. With managed agents, "non-urgent" can mean hours or days of agent work.

## Implementation plan

Sequential, each issue adding observable value:

- [x] **A — #99** Infrastructure prep: Gemma swap + LifeOS MCP HTTP transport + Cloudflare Tunnel + bearer auth
- [x] **B — #100** Worker scaffolding + atomic claim: systemd unit, agent_sessions.db, transcript_store, spend_tracker (daily $100 cap)
- [x] **C — #101** Haiku preflight + local Gemma executor — first end-to-end path (`#agent #local`)
- [x] **D — #102** Managed Agents driver + Vault setup — Claude path end-to-end with cloud connectors
- [x] **E — #103** Inter-agent tools (`lifeos_agent_*`): spawn, send, check, yield-and-resume, lineage budgets, concurrency caps
- [x] **F — #104** Telegram clarifications via return-and-resume — close the human-in-the-loop dimension

A and B are sequential. C and D can land in parallel once A+B are merged (they share preflight from C). E and F are independent and can land in either order after D.

## Key design decisions (frozen)

| Decision | Choice | Rationale |
|---|---|---|
| Claude-path hosting | Anthropic Managed Agents | They host the loop, store session state, run our MCP attached. $0.08/session-hour + tokens. No loop/sleep/crash recovery to build. |
| Local-path hosting | Self-hosted loop via `llm_client.py` | Managed Agents is Claude-only. Local tasks are typically shorter; simpler loop is fine. |
| Coordination primitive | **Yield-and-resume**, not block-and-wait | Tool-call timeouts (~60s industry default) make long blocking calls infeasible; idle sessions still bill $0.08/hr. Parent yields → session ends → resumes when children done with their output injected. |
[...]

PR #105, opened 2026-05-26
feat: agent-worker prep — Gemma swap + MCP HTTP transport (A of F)

### LINKED ISSUE #99: feat: agent-worker prep — Gemma swap + LifeOS MCP HTTP transport (A of F)
## Problem

Two prerequisites for the agent-worker feature (epic #98) that have no dependencies on each other and gate everything else:

1. The local LLM service runs gpt-oss-120b, but the agent worker's `#local` path needs Gemma.
2. The LifeOS MCP server is stdio-only, which Anthropic Managed Agents cannot attach to.

This issue does both — pure infrastructure prep, no agent code yet.

## Context

Part of the agent-worker series (epic **#98**). See the epic for the full design. This issue is **Issue A** in the A→F sequence and blocks every issue that follows.

### 1. Swap local LLM to Gemma

`/etc/systemd/system/lifeos-llm.service` currently runs:
```
llama-server -hf ggml-org/gpt-oss-120b-GGUF -ngl 99 -fa on --jinja --no-mmap -c 32768 ...
```

Swap the `-hf` flag to load Gemma instead (already cached at `~/.cache/llama.cpp/unsloth_gemma-4-26B-A4B-it-GGUF_*`):
```
llama-server -hf unsloth/gemma-4-26B-A4B-it-GGUF ...
```

Verify:
- Gemma loads in VRAM and accepts requests on port 8080
- `--jinja` template works with Gemma's chat format (Gemma uses a different template than gpt-oss — may need `--chat-template` override or a `--jinja` adjustment)
- Context length still 32768 (Gemma 4 supports it)
- Existing `api/services/llm_client.py` calls against `LIFEOS_LLM_BACKEND=local` still produce sensible responses

Side effect (user-acknowledged): LifeOS chat orchestrator quality may drop on `LIFEOS_LLM_BACKEND=local`. Default backend is Anthropic, so this is acceptable.

### 2. Add HTTP transport to `mcp_server.py`

Currently `mcp_server.py` exposes tools over stdio only. Add a parallel streamable-HTTP transport using the MCP Python SDK's HTTP server primitives. Both transports run from the same process based on a `--transport` flag (or run both simultaneously if SDK supports).

- New env vars in `config/settings.py`:
  - `LIFEOS_MCP_HTTP_PORT` (default `8765`)
  - `LIFEOS_MCP_BEARER_TOKEN` (no default — must be set explicitly)
- HTTP transport requires `Authorization: Bearer <token>` header on every request; returns `401` otherwise.
- stdio transport ignores the bearer check (local trust).
- Stdio entry point (`~/.claude.json`'s `lifeos` mcpServer command) must continue to work unchanged.

### 3. Cloudflare Tunnel route + bearer token setup

Expose the HTTP MCP endpoint via Cloudflare Tunnel. User already has Cloudflare infrastructure for QuickStage — piggyback on the same `cloudflared` setup or add a sibling tunnel config.

- Choose a hostname (e.g., `mcp.<user-domain>` or similar — leave the exact name to user preference, document in setup guide).
- Tunnel routes `https://<hostname>/` → `http://localhost:8765`.
- Outbound-only (no inbound firewall changes).
- Generate the bearer token once (e.g., `openssl rand -hex 32`), document storage in `.env`.

### 4. Setup documentation

Create `docs/guides/agent-worker-setup.md` (will be expanded in later issues, but stub it now). For this issue, the doc covers:
- Bearer token generation and `.env` placement
- Cloudflare Tunnel config snippet for the MCP route
- How to verify the endpoint responds (`curl -H "Authorization: Bearer <token>" https://<hostname>/...`)
- How to verify Gemma is loaded (`curl http://localhost:8080/v1/models`)

### Out of scope

- Any agent-worker code (Issues B–F)
- Managed Agents Vault provisioning — that's Issue D
- Updating systemd `Wants=`/`After=` for agent worker — that's Issue B

## Files involved

**New:**
- `docs/guides/agent-worker-setup.md` (stub)

**Modified:**
- `/etc/systemd/system/lifeos-llm.service` — swap `-hf` flag to Gemma; verify chat-template flags
- `mcp_server.py` — add HTTP transport + bearer auth middleware
- `config/settings.py` — `LIFEOS_MCP_HTTP_PORT`, `LIFEOS_MCP_BEARER_TOKEN`
- `.env.example` — document new env vars
- Cloudflare Tunnel config (location depends on existing setup; document in setup guide)

## Testing

Lean, scoped to this issue's surface — no broad integration tests:

- [ ] **Unit**: HTTP transport returns `401` when bearer header is missing
- [ ] **Unit**: HTTP transport returns `401` when bearer header is wrong
- [ ] **Unit**: HTTP transport returns `200` + valid MCP response when bearer header matches
- [ ] **Unit**: stdio transport works with no bearer token configured (regression)
- [ ] **Smoke** (manual, documented in setup guide): `curl` against the Cloudflare Tunnel hostname returns the MCP tool list
- [ ] **Smoke** (manual): `llama-server` reports Gemma model loaded; one `/v1/chat/completions` request returns a sensible response
- [ ] **Regression**: existing `tests/test_*` suite passes after Gemma swap (anything that hits the local LLM in tests should still work)

## Acceptance criteria

- [ ] `systemctl restart lifeos-llm && systemctl status lifeos-llm` shows Gemma loaded successfully
- [ ] LifeOS chat with `LIFEOS_LLM_BACKEND=local` produces coherent responses (manual spot check)
- [ ] `mcp_server.py` started with `--transport http` listens on `LIFEOS_MCP_HTTP_PORT` and enforces bearer auth
- [ ] `mcp_server.py` started with `--transport stdio` (or default) still works for local Claude Code; verified by Claude Code being able to call `lifeos_search`
- [ ] Cloudflare Tunnel hostname returns the MCP tool list when called with the correct bearer token; returns `401` otherwise
- [ ] `docs/guides/agent-worker-setup.md` documents bearer generation, tunnel config, and verification steps

nbramia:
## Summary
Prerequisites for the external agent worker (epic #98) — first issue in the A→F sequence. Local LLM default swapped to Gemma 4 26B (smaller VRAM footprint, leaves room for embeddings to coexist). `mcp_server.py` gains a bearer-token-protected HTTP transport behind a new `lifeos-mcp-http.service` systemd unit so Anthropic Managed Agents can reach LifeOS tools over the internet via Cloudflare Tunnel. Local Claude Code stdio transport is unchanged.

Closes #99 · Relates to #98

## Test evidence
- `pytest tests/test_mcp_http_transport.py` — **12/12 passed** (bearer rejection, dispatcher parity stdio↔http, JSON-RPC notification semantics, malformed-JSON 400, fail-fast on missing token)
- `pytest -m unit` via pre-push hook — **1187/1187 passed** in 86s
- Pre-existing failure in `test_summarizer.py::test_real_summary_generation` confirmed on `main` (Ollama summarizer returns empty content). Per AGENTS.md "Tests Are Sacred" — documented, not fixed here.

## Review focus
- **Bearer auth hardening** in `mcp_server.py:build_http_app` — uses `hmac.compare_digest` (constant-time), case-insensitive scheme parsing, refuses to start without a token.
- **Dispatcher refactor** in `mcp_server.py:dispatch` — single function shared by stdio + http, preserves original method handling (initialize / notifications/initialized / tools/list / tools/call / unknown). Original `'request_id' in dir()` guard replaced with explicit `is_notification` check.
- **Open-source hygiene** — no real hostnames or tokens committed. `.env.example` documents the env vars with empty defaults. `setup-systemd.sh` only enables the HTTP unit when `LIFEOS_MCP_BEARER_TOKEN` is set in `.env`. `docs/guides/agent-worker-setup.md` uses `mcp.example.com` as placeholder throughout.
- **No new third-party deps** — HTTP transport reuses `fastapi` + `uvicorn` already in `requirements.txt`. The `mcp` SDK was avoided per the "ask first on new deps" rule in AGENTS.md.
- **`hmac` import lives inside `build_http_app`** so the stdio path never imports `fastapi`/`hmac` (keeps stdio startup minimal for Claude Code subprocess spawning).

After merging, operator runs:
1. `sudo ./scripts/setup-systemd.sh` to install the new `lifeos-mcp-http.service` and swap the `lifeos-llm.service` to Gemma.
2. Generate bearer token with `openssl rand -hex 32`, add to `.env` as `LIFEOS_MCP_BEARER_TOKEN`.
3. Add Cloudflare Tunnel route per `docs/guides/agent-worker-setup.md`.
4. `curl -H "Authorization: Bearer \$TOKEN" https://<tunnel-hostname>/mcp ...` to confirm public reachability + auth.

These are deferred from PR scope because they touch production infra and the tunnel hostname is operator-specific.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -25,15 +25,32 @@
 # Local LLM model (HuggingFace GGUF ID for llama-server)
 # Also used by setup-systemd.sh for the systemd service
 # Available models:
-#   ggml-org/gpt-oss-120b-GGUF  — ~59 GB VRAM (MXFP4), highest quality
-#   Qwen/Qwen3-32B-GGUF        — ~20 GB VRAM (Q4_K_M), good quality, leaves headroom for embeddings
+#   unsloth/gemma-4-26B-A4B-it-GGUF — ~16 GB VRAM (Q4_K_M), default; pairs well with embedding model
+#   Qwen/Qwen3-32B-GGUF             — ~20 GB VRAM (Q4_K_M), strong general-purpose option
+#   ggml-org/gpt-oss-120b-GGUF      — ~59 GB VRAM (MXFP4), highest quality but starves embeddings
 # To switch models: change this value, then run: sudo ./scripts/setup-systemd.sh
-# LIFEOS_LLM_MODEL=ggml-org/gpt-oss-120b-GGUF
+# LIFEOS_LLM_MODEL=unsloth/gemma-4-26B-A4B-it-GGUF
 
 # Auto-start local LLM on boot and restart on crash (default: false)
 # When false, llama-server must be started manually: sudo systemctl start lifeos-llm
 # Run sudo ./scripts/setup-systemd.sh after changing this value
 # LIFEOS_LOCAL_LLM_AUTOSTART=false
+
+# =============================================================================
+# MCP HTTP TRANSPORT (for remote agent platforms; local Claude Code keeps stdio)
+# =============================================================================
+
+# Bearer token required by the MCP HTTP transport. Empty disables the HTTP
+# transport entirely. Generate one with: openssl rand -hex 32
+# Anthropic Managed Agents (and any other remote MCP caller) must send this
+# token in the `Authorization: Bearer ...` header.
+# LIFEOS_MCP_BEARER_TOKEN=
+
+# Bind address + port for the MCP HTTP transport. Default binds to localhost
+# only — pair with a Cloudflare Tunnel (or similar) for public exposure rather
+# than binding to 0.0.0.0 directly.
+# LIFEOS_MCP_HTTP_HOST=127.0.0.1
+# LIFEOS_MCP_HTTP_PORT=8765
 
 # =============================================================================
 # EMBEDDING MODEL — sentence-transformers (cached locally)
--- a/config/settings.py
+++ b/config/settings.py
@@ -70,9 +70,29 @@ class Settings(BaseSettings):
     local_llm_url: str = Field(default="http://localhost:8080", alias="LIFEOS_LOCAL_LLM_URL")
     local_llm_timeout: int = Field(default=90, alias="LIFEOS_LOCAL_LLM_TIMEOUT")
     local_llm_model: str = Field(
-        default="ggml-org/gpt-oss-120b-GGUF",
+        default="unsloth/gemma-4-26B-A4B-it-GGUF",
         alias="LIFEOS_LLM_MODEL",
         description="HuggingFace GGUF model ID for llama-server"
+    )
+
+    # MCP HTTP transport (used by remote agent platforms; local Claude Code keeps stdio)
+    mcp_http_port: int = Field(
+        default=8765,
+        alias="LIFEOS_MCP_HTTP_PORT",
+        description="Port for the MCP HTTP transport (lifeos-mcp-http systemd unit)"
+    )
+    mcp_http_host: str = Field(
+        default="127.0.0.1",
+        alias="LIFEOS_MCP_HTTP_HOST",
+        description="Bind address for the MCP HTTP transport. Default 127.0.0.1 so only "
+                    "the local Cloudflare Tunnel daemon can reach it; the tunnel handles "
+                    "public exposure."
+    )
+    mcp_bearer_token: str = Field(
+        default="",
+        alias="LIFEOS_MCP_BEARER_TOKEN",
+        description="Bearer token required by the MCP HTTP transport. Generate with "
+                    "`openssl rand -hex 32`. Empty disables the HTTP transport."
     )
     local_llm_autostart: bool = Field(
         default=False,
--- a/config/systemd/lifeos-mcp-http.service
+++ b/config/systemd/lifeos-mcp-http.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=LifeOS MCP HTTP Transport
+After=network.target lifeos-api.service
+Wants=lifeos-api.service
+
+[Service]
+Type=simple
+User=__USER__
+WorkingDirectory=__LIFEOS_DIR__
+EnvironmentFile=__LIFEOS_DIR__/.env
+ExecStart=__VENV__/bin/python __LIFEOS_DIR__/mcp_server.py --transport http
+Restart=on-failure
+RestartSec=10
+StandardOutput=append:__LIFEOS_DIR__/logs/mcp-http.log
+StandardError=append:__LIFEOS_DIR__/logs/mcp-http.log
+
+[Install]
+WantedBy=multi-user.target
--- a/docs/guides/AGENTS.md
+++ b/docs/guides/AGENTS.md
@@ -13,6 +13,7 @@ This directory contains operational guides — how to set up, configure, and run
 - `scripts.md` — Available scripts and their usage
 - `troubleshooting.md` — Common issues and solutions
 - `claude-code-orchestration.md` — Claude Code multi-agent orchestration patterns
+- `agent-worker-setup.md` — External agent worker prerequisites (Gemma swap, MCP HTTP transport, Cloudflare Tunnel, bearer token)
 
 ## Key Principles
 
--- a/docs/guides/agent-worker-setup.md
+++ b/docs/guides/agent-worker-setup.md
@@ -1,0 +1,152 @@
+# Agent Worker Setup
+
+> **Status:** Stub — expanded by later issues in the agent-worker series (#98)
+> **Last Updated:** 2026-05-26
+> **Audience:** Operators
+
+One-time setup for the external agent worker that picks up `#agent`-tagged tasks and executes them via Claude Opus (Anthropic Managed Agents) or a local Gemma model. This guide covers **prerequisites only** — the worker itself ships in later issues.
+
+---
+
+## What this issue (#99) sets up
+
+- **Local LLM** swapped from `gpt-oss-120b` to `unsloth/gemma-4-26B-A4B-it-GGUF`. Smaller VRAM footprint, leaves headroom for the embedding model to coexist.
+- **MCP HTTP transport** on `mcp_server.py` so a remote agent platform (Anthropic Managed Agents) can call LifeOS tools without stdio access to the host.
+- **Bearer-token auth** required by the HTTP transport. The stdio transport (used by local Claude Code) is unchanged and has no token check.
+- **Cloudflare Tunnel** exposes the bearer-protected HTTP endpoint to the public internet.
+
+---
+
+## Step 1 — Swap the local LLM to Gemma
+
+The `LIFEOS_LLM_MODEL` env var controls which GGUF `llama-server` loads. The default in this repo is now Gemma:
+
+```bash
+# .env (or .env.example to see the documented options)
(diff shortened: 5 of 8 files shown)


nbramia (commented):
## PR Review: feat: agent-worker prep — Gemma swap + MCP HTTP transport (A of F)

### Action Required
_None — no blockers._

### Recommended
- **[Correctness/Spec]** JSON-RPC parse errors return bare HTTP 400 with no body instead of a `-32700 Parse error` JSON-RPC envelope (`mcp_server.py:1730-1733`). The MCP Streamable HTTP spec inherits JSON-RPC 2.0 semantics, which says malformed requests should return a JSON-RPC error object with code `-32700`. Returning a plain 400 is acceptable for now (Managed Agents tolerates it), but for spec parity consider returning `JSONResponse({"jsonrpc":"2.0","id":null,"error":{"code":-32700,"message":"Parse error"}}, status_code=400)`.
- **[Security/Operational]** `setup-systemd.sh` reads the bearer token from `.env` via `_read_env`, which pipes the value through `tr -d '[:space:]'` (`scripts/setup-systemd.sh:42`). The token itself is only used as a boolean signal ("enable the unit?"), so the contents are never re-emitted — fine in practice. But the inline comment that says "the MCP HTTP transport is only enabled when a bearer token is configured" could mislead a future contributor into thinking the script *passes* the token to systemd. A one-line clarifying comment ("token value is only used to decide whether to enable the unit; the unit itself reads `.env` at runtime") would prevent confusion.
- **[Security]** `_check_auth` parses the header with `header.partition(" ")` (`mcp_server.py:1721-1722`). A header like `"Bearer  abc"` (two spaces) leaves `token=" abc"` with a leading space, which `hmac.compare_digest` will reject. Not exploitable, but operators will hit a confusing 401 if they ever paste `Authorization: Bearer  $TOKEN` with an extra space (a common copy-paste pitfall). `token = token.strip()` after `partition` would be more forgiving without weakening auth.
- **[Operational/Standards]** New `lifeos-mcp-http.service` writes to `logs/mcp-http.log` (`config/systemd/lifeos-mcp-http.service:14-15`). Logrotate config `config/logrotate-lifeos.conf` globs `*.log` so this is covered — confirmed. No action; just flagging that the implicit coupling is fine.

### Minor
- **[Standards]** `agent-worker-setup.md` lacks the "Status / Audience / Last Updated" frontmatter convention used by other guides (it has them as bullet points, not YAML frontmatter — actually consistent with other guides in `docs/guides/`, so this is fine). Verified against `docs/guides/AGENTS.md` standards.
- **[Standards]** `agent-worker-setup.md` correctly uses `mcp.example.com` as a placeholder throughout — no real hostname leak. Bearer token is referenced as `$LIFEOS_MCP_BEARER_TOKEN`, no real token committed. Open-source hygiene clean.
- **[Correctness]** `dispatch()` now wraps the entire request handler in a single try/except (`mcp_server.py:1647-1680`). The previous code only caught errors in the outer loop with the buggy `'request_id' in dir()` guard. The new code is a strict improvement — exceptions in `tools/list` or `initialize` now return a proper `-32000` error instead of being swallowed silently. Good change.
- **[Correctness]** `tools/list` and `initialize` are now treated as notification-suppressible (`mcp_server.py:1654, 1660`). MCP clients shouldn't send them as notifications, but the early-return is defensive and harmless.
- **[Correctness]** Batched JSON-RPC handling in `_handle` (`mcp_server.py:1735-1739`) returns `202` when every request in the batch is a notification — matches the streamable-HTTP spec. The test suite covers single-notification 202 but not batched-all-notifications 202; consider adding a test if batched requests are ever expected from Managed Agents.
- **[Performance]** `import hmac` and FastAPI imports are inside `build_http_app` (`mcp_server.py:1713-1716`) so the stdio path doesn't pay for them. Stated rationale in PR description matches the code. Good.
- **[Standards]** No new third-party deps added; `fastapi>=0.109.0` and `uvicorn[standard]>=0.27.0` already in `requirements.txt:2-3`. Verified.
- **[Standards]** Sudoers line in `setup-systemd.sh:154` is now 600+ chars on a single line. Already was — this PR just appended to it. Not your problem to fix here, but it's becoming hard to audit; consider splitting per-service in a future cleanup.

### Verification performed
- Checked out `feat/agent-worker-prep` and ran `pytest tests/test_mcp_http_transport.py` → **12/12 passed** (matches PR claim).
- Ran full unit suite (`pytest -m unit`) → **1194 passed, 1 failed**. The single failure is `test_summarizer.py::TestSummarizerIntegration::test_real_summary_generation` — confirmed pre-existing on `main` (the test file was added in `96e56a9` "Initial release" and the failure is unrelated to this PR). PR description correctly flags this. ✔
- Confirmed `unsloth/gemma-4-26B-A4B-it-GGUF` model is cached at `~/.cache/llama.cpp/unsloth_gemma-4-26B-A4B-it-GGUF_*` and HuggingFace repo responds 200. Gemma swap is operationally feasible.
- Confirmed the new `lifeos-mcp-http.service` template references match the `setup-systemd.sh` substitution variables (`__USER__`, `__LIFEOS_DIR__`, `__VENV__`).
- Confirmed `LIFEOS_ANTHROPIC_MODEL=claude-haiku-4-5` default unchanged — Anthropic-backend behavior is not affected by the local-LLM swap.
- Confirmed no API/sync/MCP-tool surface code changed — scope claim ("infra prep only") holds.

### Summary
**Merge-ready.** Scope is tight, tests are comprehensive and pass, security posture is sound (constant-time compare, fail-fast on missing token, localhost-only bind by default, conditional systemd enablement). The recommended items are all polish — no blockers for landing issue A of the agent-worker series.

nbramia:

| 1 | Malformed JSON returns bare 400 instead of `-32700 Parse error` envelope | Recommended | **Accept** | Cheap spec-compliance win; one extra branch in `_handle`. |
| 2 | Comment in `setup-systemd.sh` about `MCP_BEARER_TOKEN` could mislead future contributor | Recommended | **Accept** | Doc clarity, costs nothing. |
| 3 | `header.partition(" ")` doesn't strip the token, so double-space silently 401s | Recommended | **Accept** | Operator ergonomic; `.strip()` is one line. |
| – | Batched-all-notifications 202 path lacks test | Minor | **Accept** | Small additional test for a real path; trivial to add. |

**Findings forwarded:** 4 (all applied inline — they are 1-3 line changes each, so spawning an addresser subagent would cost more context than the fixes themselves).

nbramia:
## Round 1 Addresser Summary

| 1 | Malformed JSON → bare 400 | **Applied** | `_handle` now returns a JSON-RPC `-32700 Parse error` envelope (id: null, status 400). Test updated to assert envelope shape. |
| 2 | Misleading comment in `setup-systemd.sh` | **Applied** | Comment clarifies the bearer check there is install-time gating only; the unit reads `.env` at runtime. |
| 3 | `partition(\" \")` doesn't strip token | **Applied** | `_check_auth` now `.strip()`s the token. New test covers `Bearer   <token>  ` ergonomic case. |
| 4 | Batched-all-notifications 202 path untested | **Applied** | Added two batch tests: all-notifications → 202, mixed → only responses for entries with id. |

**Tests:** `pytest tests/test_mcp_http_transport.py` — 15/15 passed.
**Commit:** `b984176 fix: address review round 1 — spec-compliant parse error, token strip, batch test`

Issue #102, opened 2026-05-26
feat: Managed Agents driver + Vault setup (D of F)


nbramia:
## Problem

`#agent` tasks (without `#local`) currently route to `claude` but are parked unimplemented. This issue wires up the Anthropic Managed Agents driver and the one-time Vault setup so Claude-path tasks actually run on Anthropic's infrastructure with the LifeOS MCP server attached over HTTP plus cloud connectors (Gmail/Calendar/Drive/Superhuman/Zapier).

After this issue, **both `#agent` and `#agent #local` work end-to-end**, with appropriate model selection and tool access. Inter-agent coordination (E) and Telegram clarifications (F) come next.

## Context

Part of the agent-worker series (epic **#98**). **Issue D**, depends on **#99** (Issue A) and **#100** (Issue B). Shares preflight code with **#101** (Issue C) — can land in parallel with C after the shared preflight is merged.

### 1. `managed_driver.py`

New module that wraps the Anthropic Managed Agents API:

- **Create session**: `create_session(system_prompt, tools, mcp_servers, vault_id) -> session_id`. Includes:
  - Our LifeOS MCP server URL (from Issue A's Cloudflare Tunnel) + bearer header
  - All cloud connector MCPs (Gmail/Calendar/Drive/Superhuman/Zapier) via the Vault (Issue D's setup)
  - Built-in tools: Bash, Read, Write, Edit, WebFetch, WebSearch
- **Stream session events** (SSE): consume the session's event stream, write each event to `data/agent_transcripts/{session_id}.jsonl` in the unified format, update token / dollar / session-hour totals on the session row.
- **Resume session**: `resume_session(session_id, new_user_turn) -> ...` (used by Issue F for clarification replies; the API surface exists in this issue, callers come later).
- **Kill session**: `kill_session(session_id, reason)`.
- **Get session state**: `get_session(session_id) -> {status, last_event, ...}` for the worker's main loop.

The driver translates between our internal session model and Anthropic's session model. `sessions.managed_agent_session_id` is populated from the API response.

### 2. Session-hour accounting

Add session-hour overhead to `pricing.py` and the budget enforcer:
- `$0.08/hour` while a managed session is alive
- Tracked by `(now - session.started_at)` for live sessions, plus reserve in preflight estimates
- Counts toward `total_dollars` and the per-task budget and the daily $100 cap

### 3. Vault provisioning (manual, documented in setup guide)

[...]

Issue #101, opened 2026-05-26
feat: Haiku preflight + local Gemma executor (C of F)


nbramia:
## Problem

Replace the no-op dispatcher from Issue B with a real two-step pipeline:

1. **Haiku preflight** — cheap structured-output call that parses budget, infers routing, flags ambiguity, classifies expected output, and sanity-checks the task.
2. **Local executor** — for `#agent #local` tasks (or tasks Haiku routed to `local`), run a real agent loop on Gemma via `llm_client.py` with tool dispatch and external budget enforcement.

After this issue, **`#agent #local` tasks run end-to-end** and produce real Telegram completion notifications with one-paragraph summaries. The Claude path is still stubbed; Issue D adds that.

## Context

Part of the agent-worker series (epic **#98**). **Issue C**, depends on **#99** (Issue A) and **#100** (Issue B). Blocks Issue D mostly only in the sense that Issue D shares the preflight; otherwise D can proceed in parallel after this lands.

### 1. Haiku preflight (`preflight.py`)

Single Anthropic API call to `claude-haiku-4-5` with structured-output schema:

```json
{
  "budget": { "wall_seconds": 14400, "max_tokens": 500000, "max_dollars": 5.00 },
  "routing": "local | claude | ask",
  "routing_reason": "string",
  "expected_output": "text | file | external_action | structured",
  "ambiguity": null,
  "sane": true,
  "sane_reason": "string"
}
```

Responsibilities:
- **Budget parsing** from natural-language task title (e.g., "spend 10 min on this", "max $0.50", "up to 10k tokens"). Defaults from settings when missing.
- **Routing inference** in priority order:
  1. `#local` tag → `local`
  2. Title cues like "with local agent", "using gemma", "use claude", "with opus" → infer
  3. Neither → `routing: "ask"` (Issue F handles the Telegram round-trip; for this issue, treat `ask` as a hard failure with a logged reason — Telegram clarification comes in F)
- **Ambiguity check** — underspecified tasks ("reply to John") return a clarifying question (held but not sent until Issue F)
- **Expected output shape** — for notification phrasing
- **Sanity flag** — garbage titles, accidental `#agent` tags, obviously destructive shapes

For this issue, `ambiguity != null` and `routing == "ask"` both result in:
- Tag swap to `#agent-blocked`
- Telegram notification: `"⏸ task blocked, needs clarification: <question>"` (one-way notification only; reply handling lands in F)
- Session marked `blocked` in the DB

[...]

PR #106, opened 2026-05-26
feat: agent-worker scaffolding + atomic claim (B of F)

### LINKED ISSUE #100: feat: agent-worker scaffolding + atomic claim (B of F)
## Problem

Build the worker plumbing — long-running process, persistent state, atomic task claiming, transcript storage, spend tracking — without yet adding any LLM execution. After this issue, the worker can claim `#agent` tasks, swap their tags atomically, mark them no-op-complete, and enforce the $100/day global cap. No tokens are spent. This proves the scaffolding before later issues stack execution on top.

## Context

Part of the agent-worker series (epic **#98**). **Issue B**, blocks Issues C/D/E/F. Depends on **#99** (Issue A).

### 1. Worker package skeleton

New package `api/services/agent_worker/`:

- `worker.py` — main loop. Polls `/api/tasks?status=todo&tag=agent` every `LIFEOS_AGENT_WORKER_POLL_SECONDS` (default 60). For each unclaimed task: check global daily spend cap → claim (atomic tag swap) → invoke dispatcher (no-op stub for this issue) → mark complete with a placeholder summary → notify via existing Telegram. Resumable across restart by querying `agent_sessions.db` on startup.
- `session_store.py` — SQLite wrapper for `data/agent_sessions.db`. Schemas listed below.
- `transcript_store.py` — append-only JSONL writer/reader for `data/agent_transcripts/{session_id}.jsonl`. Each line is one event (`{ts, kind, payload}`).
- `spend_tracker.py` — daily $-cap enforcement. Tracks per-day cumulative spend in `daily_spend` table. Method `can_start_task(estimated_dollars) -> bool` consulted by worker before claim.

### 2. SQLite schemas (`data/agent_sessions.db`)

```sql
CREATE TABLE sessions (
  task_id TEXT PRIMARY KEY,
  session_id TEXT UNIQUE NOT NULL,           -- our internal id; later: also tracks managed_agent_session_id
  status TEXT NOT NULL,                       -- claimed | running | yielded | completed | failed | budget_exceeded | blocked
  routing TEXT,                               -- local | claude | null (not yet decided)
  budget_json TEXT,                           -- JSON: {wall_seconds, max_tokens, max_dollars}
  started_at INTEGER NOT NULL,
  last_activity_at INTEGER NOT NULL,
  total_input_tokens INTEGER DEFAULT 0,
  total_output_tokens INTEGER DEFAULT 0,
  total_dollars REAL DEFAULT 0.0,
  expected_output TEXT,
  parent_session_id TEXT,                     -- for future lineage; nullable
  managed_agent_session_id TEXT               -- nullable; populated by Issue D
);

CREATE INDEX idx_sessions_status ON sessions(status);
CREATE INDEX idx_sessions_parent ON sessions(parent_session_id);

CREATE TABLE daily_spend (
  date TEXT PRIMARY KEY,                      -- ISO date in local TZ
  total_dollars REAL NOT NULL DEFAULT 0.0
);
```

Other tables (`messages`, `pending_questions`, `sleeps`) are introduced in later issues — not in this one.

### 3. Atomic tag-swap helper in `task_manager.py`

The current task model supports tag mutation via `PUT /api/tasks/{id}`. Add a method `swap_tag(task_id, from_tag, to_tag) -> bool` that performs the read → check → write atomically (single SQLite transaction or compare-and-swap pattern), returning `False` if the `from_tag` is not present (task already claimed). Worker uses this to claim:

```python
if not task_manager.swap_tag(task_id, "agent", "agent-running"):
    continue  # someone else got it
```

Surface as `POST /api/tasks/{id}/swap-tag?from=agent&to=agent-running` for the worker to call from outside the API process.

### 4. systemd unit + setup script

New file: `/etc/systemd/system/lifeos-agent-worker.service` (delivered via setup script, not committed to repo since paths are user-specific):

```ini
[Unit]
Description=LifeOS Agent Worker
After=network.target lifeos-api.service
Requires=lifeos-api.service

[Service]
Type=simple
User=nathanramia
EnvironmentFile=/home/nathanramia/Code/LifeOS/.env
ExecStart=/home/nathanramia/.venvs/lifeos/bin/python -m api.services.agent_worker.worker
WorkingDirectory=/home/nathanramia/Code/LifeOS
Restart=on-failure
RestartSec=10
StandardOutput=append:/home/nathanramia/Code/LifeOS/logs/agent-worker.log
StandardError=append:/home/nathanramia/Code/LifeOS/logs/agent-worker.log

[Install]
WantedBy=multi-user.target
```

Setup helper: `scripts/setup-agent-worker.sh` (mirrors pattern of `scripts/setup-systemd.sh`).

### 5. Configuration

`config/settings.py` additions:
- `LIFEOS_AGENT_WORKER_POLL_SECONDS` (default 60)
- `LIFEOS_AGENT_DEFAULT_BUDGET_DOLLARS` (default 5.00)
- `LIFEOS_AGENT_DEFAULT_WALL_SECONDS` (default 14400 — 4h)
- `LIFEOS_AGENT_DEFAULT_MAX_TOKENS` (default 500000)
- `LIFEOS_AGENT_DAILY_CAP_DOLLARS` (default 100.00)
- `LIFEOS_AGENT_CLARIFICATION_TIMEOUT_HOURS` (default 72 — 3 days; used in Issue F)

### 6. No-op dispatcher

For this issue only: the dispatcher just writes a "no-op completion" event to the transcript, marks the session `completed`, posts a Telegram notification like `"✅ no-op completion for task: <title> (worker scaffolding test)"`, and adds $0.00 to daily spend.

### Out of scope

- Haiku preflight (Issue C)
- Local executor (Issue C)
- Managed Agents driver (Issue D)
- Inter-agent tools (Issue E)
- Telegram clarifications (Issue F)
- Anything that spends actual money

## Files involved

**New:**
- `api/services/agent_worker/__init__.py`
- `api/services/agent_worker/session_store.py`
- `api/services/agent_worker/transcript_store.py`
- `api/services/agent_worker/spend_tracker.py`
- `scripts/setup-agent-worker.sh`
- `data/agent_sessions.db` (created on first worker startup)
- `data/agent_transcripts/` (created on first worker startup)

**Modified:**
- `api/services/task_manager.py` — atomic `swap_tag` method
- `api/routes/tasks.py` — `POST /api/tasks/{id}/swap-tag` endpoint
- `config/settings.py` — new env vars listed above
- `.env.example` — document new env vars
- `docs/guides/agent-worker-setup.md` — append systemd setup instructions

## Testing

Tests are scoped to what this issue actually builds — no LLM mocking, no managed-agent fixtures yet.

- [ ] **Unit**: `transcript_store` appends events and reads them back in order; survives reopen
- [ ] **Unit**: `session_store` round-trips a session row; rejects duplicate `session_id`
- [ ] **Unit**: `spend_tracker.can_start_task(amount)` returns `False` once daily total + estimate would exceed `LIFEOS_AGENT_DAILY_CAP_DOLLARS`; returns `True` at exactly the cap boundary
- [ ] **Unit**: `spend_tracker` records a new day with $0 at local midnight
- [ ] **Race test**: two concurrent `swap_tag(task_id, "agent", "agent-running")` calls on the same task — exactly one returns `True`, exactly one returns `False`. Use threading or asyncio.gather against a real SQLite file.
- [ ] **Integration**: spin up worker against a test DB with one `#agent` task; verify within 2 poll cycles it claims, no-op completes, and the task is gone from `#agent` list and present in `#agent-running` then removed. Mock Telegram send.
- [ ] **Integration**: worker restart mid-cycle — kill worker after claim but before completion; restart; verify it picks up and completes the claimed-but-incomplete task

## Acceptance criteria

- [ ] `sudo ./scripts/setup-agent-worker.sh` installs and enables the systemd unit; `systemctl status lifeos-agent-worker` shows running
- [ ] Creating a task with `#agent` tag is picked up within ~60s, claimed atomically, no-op completed, Telegram notification sent
- [ ] `swap_tag` is verified atomic against concurrent racers in tests
- [ ] `agent_sessions.db` and `data/agent_transcripts/` are created on first run
- [ ] Daily spend cap of $100 is enforced; setting cap to $0.00 prevents any new claims (verifiable via env override)
- [ ] Worker survives `systemctl restart lifeos-agent-worker` mid-task and resumes

nbramia:
## Summary
Issue B of the agent-worker series (epic #98). Builds the worker plumbing — long-running poll loop, atomic task claiming, SQLite-backed session store, append-only JSONL transcripts, daily \$-cap ledger — without spending any LLM money. A no-op dispatcher proves the round-trip: poll → claim (atomic tag-swap) → record session + transcript → no-op complete → Telegram notification. Real execution lands in Issue C (#101) for Gemma and Issue D (#102) for managed Opus.

Closes #100 · Relates to #98

## Test evidence
- `pytest tests/test_agent_worker_*.py tests/test_task_manager_swap_tag.py` — **25/25 passed** (stores, tracker, swap_tag race, worker loop)
- `pytest -m unit` via pre-push hook — **1215/1215 passed** in 86s (up from 1187 on \`main\`; 28 new tests, no regressions)
- Pre-existing \`test_summarizer.py::test_real_summary_generation\` still fails on \`main\` (Ollama-side, unrelated)

## Review focus
- **\`TaskManager.swap_tag\`** uses the existing in-process lock (\`self._lock\`) for the check-and-swap. The race test (\`tests/test_task_manager_swap_tag.py::test_swap_tag_is_atomic_under_race\`) launches 8 threads at a barrier and asserts exactly one winner. The worker calls this via \`POST /api/tasks/{id}/swap-tag?from=&to=\` so two worker processes serialize through the API's singleton.
- **\`SpendTracker\` boundary semantics** — the cap is an inclusive ceiling (\`total + estimate <= cap\` is allowed). Documented in the method docstring and tested at the boundary. \`record(0)\` is a no-op; negative values rejected.
- **\`Worker\` is signal-safe** — \`stop()\` is settable from a SIGTERM handler; \`run()\` sleeps in 1s slices so shutdown is prompt. \`tick()\` is the testable unit; \`run()\` is the daemon wrapper.
- **systemd unit is opt-in** via \`LIFEOS_AGENT_WORKER_AUTOSTART\` (default \`false\`). A fresh clone running \`sudo ./scripts/setup-systemd.sh\` will install the template but leave the unit disabled — no surprise polling. The sudoers rule was extended to cover start/stop/restart on the new unit.
- **Open-source hygiene** — no hardcoded paths or tokens in templates. Settings expose env-var overrides for every knob. Setup guide uses \`mcp.example.com\` placeholders. Worker survives missing Telegram config (\`send_message\` import falls back to a noop).
- **No LLM spend in this PR** — \`spend_tracker.record(0)\` is the only path called. Real spend accounting comes in C/D.

## Architectural notes
- \`SessionStore\` uses WAL mode and short-lived connections per call — safe for multi-thread / multi-process readers. SQLite serializes writers, which is the only guarantee we need for the daily-cap UPSERT.
- \`TranscriptStore\` is intentionally dumb append-only JSONL so multiple processes can write simultaneously without a coordinator (each session has a unique file).
- The worker hits the API over HTTP rather than importing \`TaskManager\` directly. This keeps the worker restartable without touching FastAPI startup and lets the existing in-process lock serialize \`swap_tag\` across worker + API requests.

1. \`LIFEOS_AGENT_WORKER_AUTOSTART=true\` in \`.env\`
2. \`sudo ./scripts/setup-systemd.sh\`
3. Create an \`#agent\`-tagged task via \`POST /api/tasks\`
4. Within ~60s, task should be marked \`done\` with a Telegram "no-op completed" message
5. Inspect \`data/agent_transcripts/<session_id>.jsonl\` for \`claim\` → \`noop_dispatch\` → \`noop_complete\` events

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -53,6 +53,34 @@
 # LIFEOS_MCP_HTTP_PORT=8765
 
 # =============================================================================
+# AGENT WORKER (external worker that picks up #agent-tagged tasks)
+# =============================================================================
+
+# Enable systemd auto-start for the agent worker. Off by default — the worker
+# polls the task list and claims #agent-tagged tasks. Run sudo ./scripts/setup-systemd.sh
+# after changing.
+# LIFEOS_AGENT_WORKER_AUTOSTART=false
+
+# How often the worker polls for new #agent tasks (seconds).
+# LIFEOS_AGENT_WORKER_POLL_SECONDS=60
+
+# Default per-task budget when the task title doesn't specify one. Haiku
+# preflight (Issue C) parses natural-language budget hints from titles like
+# "spend 5 min" or "max $0.50" and falls back to these defaults.
+# LIFEOS_AGENT_DEFAULT_BUDGET_DOLLARS=5.00
+# LIFEOS_AGENT_DEFAULT_WALL_SECONDS=14400
+# LIFEOS_AGENT_DEFAULT_MAX_TOKENS=500000
+
+# Global daily $-cap across ALL agent tasks. New tasks are not claimed once
+# today's cumulative spend would exceed this. In-flight tasks finish their
+# own budgets. Set to 0 to pause all new claims.
+# LIFEOS_AGENT_DAILY_CAP_DOLLARS=100.00
+
+# How long a clarification (#agent-blocked) task waits for a Telegram reply
+# before being abandoned permanently. Used by Issue F.
+# LIFEOS_AGENT_CLARIFICATION_TIMEOUT_HOURS=72
+
+# =============================================================================
 # EMBEDDING MODEL — sentence-transformers (cached locally)
 # =============================================================================
 
--- a/.gitignore
+++ b/.gitignore
@@ -135,5 +135,8 @@ scripts/clear-caches.sh
 !.claude/settings.json
 !.claude/skills/
 
-# Duplicate of CLAUDE.md
-AGENTS.md
+# Per-subdirectory AGENTS.md files (api/, docs/, etc.) are project docs and
+# are tracked. This rule only ignores ad-hoc AGENTS.md files that some Claude
+# Code tooling may write at the repo root — the tracked root-level AGENTS.md
+# overrides this since gitignore doesn't apply to files already in the index.
+/AGENTS.md
--- a/api/routes/tasks.py
+++ b/api/routes/tasks.py
@@ -6,7 +6,7 @@ CRUD endpoints for tasks stored in Obsidian-compatible markdown.
 import logging
 from typing import Optional
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, HTTPException, Query
 from pydantic import BaseModel, Field
 
 from api.services.task_manager import get_task_manager, Task
@@ -140,6 +140,31 @@ async def list_tags():
     """List all distinct tags across all tasks (any status) with usage counts."""
     manager = get_task_manager()
     return TagListResponse(tags=[TagUsage(**t) for t in manager.list_tags()])
+
+
+class SwapTagResponse(BaseModel):
+    swapped: bool
+    reason: Optional[str] = None
+
+
+@router.post("/{task_id}/swap-tag", response_model=SwapTagResponse)
+async def swap_tag(
+    task_id: str,
+    from_tag: str = Query(..., alias="from", description="Tag to remove (with or without '#')"),
+    to_tag: str = Query(..., alias="to", description="Tag to add in its place"),
+):
+    """Atomically replace one tag with another on a task.
+
+    Used by the external agent worker to claim `#agent` tasks. Returns
+    `{swapped: false}` (with a `reason`) when the task does not exist or
+    `from` is not currently among the task's tags — both indicate the worker
+    should move on to the next candidate rather than retry.
+    """
+    manager = get_task_manager()
+    if manager.get(task_id) is None:
+        return SwapTagResponse(swapped=False, reason="task not found")
+    ok = manager.swap_tag(task_id, from_tag, to_tag)
+    return SwapTagResponse(swapped=ok, reason=None if ok else f"tag '{from_tag}' not present")
 
 
 @router.get("/{task_id}", response_model=TaskResponse)
--- a/api/services/agent_worker/AGENTS.md
+++ b/api/services/agent_worker/AGENTS.md
@@ -1,0 +1,51 @@
+# Agent Worker
+
+External long-running worker that picks up `#agent`-tagged tasks, executes them via Claude Opus (Managed Agents) or local Gemma (llama-server), and notifies via Telegram. Lives outside the FastAPI process — talks to LifeOS over HTTP via `/api/tasks`.
+
+> **Status:** Scaffolding only (Issue B / #100). No LLM execution yet.
+> **Owning epic:** [#98](https://github.com/nbramia/LifeOS/issues/98)
+
+## Files in this package
+
+| File | Responsibility |
+|------|---------------|
+| `worker.py` | Main poll loop, claim/dispatch, startup resume, signal-safe stop |
+| `session_store.py` | SQLite-backed `sessions` + `daily_spend` tables; `Session` dataclass |
+| `transcript_store.py` | Append-only JSONL per `session_id` at `data/agent_transcripts/` |
+| `spend_tracker.py` | Daily $-cap ledger (inclusive ceiling; cap ≤ 0 pauses claims) |
+
+## Lifecycle (current scope)
+
+```
+poll → can_start_task(default_budget)?
+     → list /api/tasks?status=todo&tag=agent
+     → atomic swap #agent → #agent-running
+     → session row + transcript "claim" event
+     → no-op dispatcher (replaced in Issues C/D)
+     → mark task complete
+     → transcript "noop_complete" + Telegram notify
+```
+
+On startup, `resume_pending()` scans non-terminal sessions and either marks them complete or rolls the tag back to `#agent` for retry. This makes the worker SIGKILL-safe.
+
+## Where future issues plug in
+
+| Issue | Adds |
+|-------|------|
+| C ([#101](https://github.com/nbramia/LifeOS/issues/101)) | Haiku preflight + local Gemma executor → replaces `_dispatch_noop` for `#local`-tagged tasks |
+| D ([#102](https://github.com/nbramia/LifeOS/issues/102)) | Managed Agents driver → replaces `_dispatch_noop` for Claude-routed tasks |
+| E ([#103](https://github.com/nbramia/LifeOS/issues/103)) | `lifeos_agent_*` MCP tools, lineage budgets, yield-and-resume |
+| F ([#104](https://github.com/nbramia/LifeOS/issues/104)) | Telegram clarification round-trip via reply-threading |
+
+## Open-source guardrails
+
+- Worker is **opt-in** via `LIFEOS_AGENT_WORKER_AUTOSTART=true` (default `false`). Fresh clones don't start polling.
+- All operational knobs are env vars (see `config/settings.py` `agent_*` fields and `.env.example`).
+- Public-internet exposure (the MCP HTTP transport) requires `LIFEOS_MCP_BEARER_TOKEN` — empty disables the HTTP unit entirely.
+- Cap of 0 (or negative) pauses all new claims as a kill-switch.
+
+## Related Documents
+
+- [`docs/guides/agent-worker-setup.md`](../../../docs/guides/agent-worker-setup.md) — operator setup
+- [`api/services/AGENTS.md`](../AGENTS.md) — sibling services
+- [Epic #98](https://github.com/nbramia/LifeOS/issues/98) — full design
--- a/api/services/agent_worker/__init__.py
+++ b/api/services/agent_worker/__init__.py
@@ -1,0 +1,8 @@
+"""External agent worker for `#agent`-tagged tasks.
+
+This package is the scaffolding for the agent worker described in epic #98.
+Issue B (this issue) lands the plumbing: a long-running poll loop, atomic
+task claiming, a session/transcript store, and a daily spend cap. Real LLM
+execution arrives in Issues C and D; inter-agent coordination in E; Telegram
+clarifications in F.
+"""
--- a/api/services/agent_worker/session_store.py
+++ b/api/services/agent_worker/session_store.py
@@ -1,0 +1,224 @@
+"""SQLite-backed session store for the agent worker.
+
+One row per agent session (which maps 1:1 to a claimed task in this issue).
+Schema is deliberately permissive — later issues add columns and tables
+(`messages`, `pending_questions`, `sleeps`, lineage fields). The store
+intentionally exposes thin CRUD; orchestration lives in `worker.py`.
+"""
+from __future__ import annotations
+
+import json
+import sqlite3
+import time
+import uuid
+from dataclasses import dataclass
+from pathlib import Path
+
+
+DEFAULT_DB_PATH = Path("data/agent_sessions.db")
+
+
+# Status vocabulary used across the agent worker. Kept here so other modules
+# can import the constants instead of stringly-typed values.
+STATUS_CLAIMED = "claimed"
+STATUS_RUNNING = "running"
+STATUS_YIELDED = "yielded"
+STATUS_COMPLETED = "completed"
+STATUS_FAILED = "failed"
+STATUS_BUDGET_EXCEEDED = "budget_exceeded"
+STATUS_BLOCKED = "blocked"
+
+TERMINAL_STATUSES = frozenset({
+    STATUS_COMPLETED,
+    STATUS_FAILED,
+    STATUS_BUDGET_EXCEEDED,
+})
+
+
+@dataclass
+class Session:
+    """Mirrors one row in the `sessions` table.
+
+    Only fields needed by Issue B are populated; later issues fill in routing,
+    budget, token counts, etc. Stored timestamps are unix epoch seconds (int).
+    """
+
+    task_id: str
+    session_id: str
+    status: str
+    started_at: int
+    last_activity_at: int
+    routing: str | None = None
+    budget: dict | None = None
+    expected_output: str | None = None
+    parent_session_id: str | None = None
+    managed_agent_session_id: str | None = None
+    total_input_tokens: int = 0
+    total_output_tokens: int = 0
+    total_dollars: float = 0.0
+
+
+_SCHEMA = """
+CREATE TABLE IF NOT EXISTS sessions (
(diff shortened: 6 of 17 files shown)


nbramia (commented):
## PR Review: feat: agent-worker scaffolding + atomic claim (B of F)

### Action Required

- **[Correctness]** `Worker.tick()` cap-zero guard is broken. `tick()` (worker.py:113) calls `spend_tracker.can_start_task(0.0)` to detect "cap reached." With `daily_cap_dollars=0.0` and `today_total=0.0`, `0 + 0 <= 0` returns `True`, so claiming proceeds. The PR description and `.env.example:30` both state "Set to 0 to pause all new claims" — currently false. The passing test `test_worker_pauses_at_daily_cap` (test_agent_worker_loop.py:165) only works because it pre-records `0.01`. Either change the guard to `<` (strict), special-case `daily_cap_dollars == 0`, or pass a tiny positive estimate. Without this fix, operators who set the cap to 0 will still get claims.

- **[Correctness / Requirements]** No startup resume implementation. Issue #100 explicitly specifies "Resumable across restart by querying `agent_sessions.db` on startup." `Worker.run()` (worker.py:86) never calls `session_store.list_non_terminal()`. If the worker is SIGKILL'd between `update_status(STATUS_RUNNING)` (worker.py:214) and `_complete_task` (worker.py:217), the task is stuck at `#agent-running` with a `running` session row and no recovery path — even though `list_non_terminal()` is implemented (session_store.py:198). At minimum, the worker should log/notify orphans on startup so the operator knows to intervene.

- **[Correctness]** Orphaned task on `_complete_task` failure. In `_dispatch_noop` (worker.py:217–225), if the HTTP call to `/api/tasks/{id}/complete` fails (network, 5xx, timeout), `_complete_task` returns `False`, status is set to `STATUS_FAILED`, and the task remains at `#agent-running` forever. There is no retry, no operator-visible recovery, and no tag rollback. A failed completion is silently turned into permanent task limbo. Either retry, swap tag back to `#agent-failed`, or surface to the operator beyond a single Telegram notify.

### Recommended

- **[Correctness]** `_claim` rollback can fail silently. worker.py:193–199: if `session_store.create` raises, the worker tries to swap `RUNNING_TAG → AGENT_TAG`. If that second swap also fails, the task is stuck at `#agent-running` with no session row — logged at ERROR but no Telegram alert, no metric. Worth notifying because the operator otherwise has no signal.

- **[Correctness / Standards]** `record(0)` is not a no-op despite docstring. spend_tracker.py:73–92: `record(0.0)` still issues an `INSERT … ON CONFLICT … DO UPDATE`. It creates a `daily_spend` row with `0.0` for today if none exists. The PR description says "`record(0)` is a no-op" — it isn't; it has a side effect. Either short-circuit when `dollars == 0` or update the description. Minor functional impact, but the docstring/PR claim is incorrect.

- **[Security]** `swap-tag` endpoint has no authentication. routes/tasks.py:150 follows the existing pattern (other task routes are also unauthenticated), so this is consistent — but the LifeOS API binds 0.0.0.0 for Tailscale access. Any caller on the network can now race-claim agent tasks. With Issue D/managed agents wired in, this becomes a real spending vector. Worth tracking as a follow-up if not addressing now.

- **[Standards / Tests]** Failure paths untested. No test exercises `_dispatch_noop`'s `_complete_task = False` branch (worker.py:222–225), nor the `_claim` rollback path (worker.py:193–199). Both are listed in `_notify` calls so the user will see them in production — but a regression in either branch wouldn't fail CI.

- **[Standards]** Bare `Exception` in tests instead of `sqlite3.IntegrityError`. test_agent_worker_stores.py:53 uses `pytest.raises(Exception)`, but `session_store.create`'s docstring (session_store.py:132) commits to raising `sqlite3.IntegrityError`. Pin the test to the specific exception so a regression that swallows the IntegrityError still fails.

### Minor

- **[Standards]** Worker startup logs the cap in dollars but `_dispatch_noop` Telegram messages use emojis (worker.py:221, 225). AGENTS.md: "Only use emojis if the user explicitly requests it." User-facing Telegram is arguably exempt (existing convention in `api/services/telegram.py` callers), but worth a moment of confirmation.

- **[Correctness]** Path-traversal check in `TranscriptStore._path` (transcript_store.py:29) refuses `/` and `..` but accepts `\`. Not exploitable on Linux/macOS, but cheap to fix: `if any(c in session_id for c in ("/", "\\", "..")): raise ValueError(...)`.

- **[Security]** Reflected user input in error reason. routes/tasks.py:167: `f"tag '{from_tag}' not present"` echoes the untrusted query param. Not exploitable (JSON response, no rendering), but worth sanitizing or omitting.

- **[Correctness]** Stale `session_store.get(task_id) is not None` check in `tick()` (worker.py:125) is mostly redundant — claimed tasks have tag `#agent-running` and won't match the `tag=agent` filter. The guard only matters if the operator manually re-tags `#agent` after claim. Harmless but worth a comment.

- **[Standards]** `session_store.py` defines `list_non_terminal()` and `list_by_status()` but neither is referenced anywhere in this PR. Either wire one up for restart-resume (see Action Required #2) or note in the docstring that these are for upcoming issues.

- **[Standards]** New package `api/services/agent_worker/` could use an `AGENTS.md` per CLAUDE.md's "When creating new directories or modules, check if an AGENTS.md + CLAUDE.md pair is appropriate." Optional but consistent.

- **[Correctness]** `agent_default_budget_dollars`, `agent_default_wall_seconds`, `agent_default_max_tokens`, and `agent_clarification_timeout_hours` settings are defined (config/settings.py:834–860) but unused anywhere in this PR. Marked as "Used by Issue F" in the docstring, which is fine — but they will show up in `LIFEOS_*` env audits as orphaned settings until C/D/F land.

- **[Correctness]** `record()` and `today_total()` use `date.today()` which is the local-timezone date. The PR description says "ISO date in local TZ" — explicit, fine. Just noting that a daemon running across midnight UTC vs local will see the bucket change at different wall-clock times. Acceptable for $-cap purposes.

- **[Standards]** `_dispatch_noop` reads `task["id"]` directly (worker.py:206) after `tick()` already validated it. Pass `task_id` through as a parameter to make the contract explicit.

### Summary

Solid scaffolding PR — small surface, focused scope, well-tested core paths (25/25 in targeted tests, 1215/1215 in unit suite). Three correctness gaps prevent merge as-is: (1) cap=0 doesn't actually block new claims, (2) no startup resume of orphaned `running` sessions as required by the issue spec, (3) failed `_complete_task` leaves tasks at `#agent-running` with no recovery. Once those are addressed (or explicitly deferred with comments + follow-up issues), the rest is merge-ready.

nbramia:

| 1 | Cap-zero guard broken (claims still proceed at cap=0) | Action Required | **Accept** | Genuine bug — docs claim cap=0 pauses claims, code doesn't. Fix: short-circuit `daily_cap_dollars <= 0` in `can_start_task` + pass actual estimate from `tick()`. |
| 2 | Missing startup resume (orphans on SIGKILL mid-claim) | Action Required | **Accept** | Explicitly required by issue #100. Adding `resume_pending()` called from `run()`. |
| 3 | Orphan task on complete-failure (stuck `#agent-running`) | Action Required | **Accept** | Real lifecycle gap. Roll tag back to `#agent` so next tick can retry. |
| 4 | `_claim` rollback silent failure | Recommended | **Accept** | Cheap to add Telegram alert on this rare failure. |
| 5 | `record(0)` not a no-op (UPSERTs zero-dollar row) | Recommended | **Accept** | Trivial early-return. |
| 6 | `swap-tag` endpoint unauthenticated | Recommended | **Reject** | Out of scope. Consistent with other task routes; no new attack surface (auth is project-wide concern). |
| 7 | Failure paths untested | Recommended | **Accept** | Adding tests for completion-failure rollback, claim-rollback failure, resume_pending. |
| 8 | `pytest.raises(Exception)` too broad | Recommended | **Accept** | Pin to `sqlite3.IntegrityError`. |
| 9 | Emojis in Telegram notifications | Minor | **Reject** | Issue #100 spec literally uses ✅ in the example. User-requested in this UI surface. |
| 10 | `_path` doesn't reject backslash | Minor | **Accept** | Defense in depth — costs nothing. |
| 11 | Reflected query input in error reason | Minor | **Reject** | Operator is the only caller; no rendering context where reflection matters. |
| 12 | Redundant `session_store.get` check in `tick()` | Minor | **Reject** | Load-bearing for the documented "re-tagged after completion" scenario tested in `test_worker_skips_already_claimed_tasks`. |
| 13 | `list_*` defined but unused | Minor | **Reject** | Used by the resume fix landing in this same round. |
| 14 | Missing AGENTS.md in agent_worker package | Minor | **Accept** | Hygiene — matches project conventions. |
| 15 | Orphan settings (agent_default_*, clarification_timeout) | Minor | **Reject** | Intentional forward-declaration; consumed in #101/#102/#104. Co-locating them documents the env-var surface up front. |

**Findings forwarded:** 9 (3 Action Required, 4 Recommended, 2 Minor — all applied inline since each is a focused 5-30 line change).

nbramia:
## Round 1 Addresser Summary

| 1 | Cap=0 didn't actually pause claims | **Applied** | `SpendTracker.can_start_task` short-circuits to False when `daily_cap_dollars <= 0`. `Worker.tick` now passes `settings.agent_default_budget_dollars` as the estimate. Tests updated: cap=0 explicitly denied; negative cap denied. |
| 2 | Missing startup resume (issue #100 requirement) | **Applied** | New `Worker.resume_pending()` runs before `tick()` in `run()`. Iterates `list_non_terminal()`, retries `/complete`, finalizes session or rolls tag back to `#agent` for retry. 2 tests cover success + failure paths. |
| 3 | Orphan task on complete-fail (stuck `#agent-running`) | **Applied** | `_dispatch_noop` rolls tag back when API completion fails. Telegram message reflects whether the rollback succeeded. New test `test_worker_rolls_back_tag_on_complete_failure` covers this. |
| 4 | `_claim` rollback silent | **Applied** | Now sends a Telegram alert when both the session-create and the tag-rollback fail (rare double-failure, but operator-actionable). |
| 5 | `record(0)` UPSERTed a row | **Applied** | Early-return. New test `test_spend_tracker_record_zero_is_noop`. |
| 6 | `pytest.raises(Exception)` too broad | **Applied** | Pinned to `sqlite3.IntegrityError`. |
| 7 | TranscriptStore path-traversal allows `\\` | **Applied** | Now rejects `/`, `\\`, `..`, and empty session_ids. |
| 8 | Missing AGENTS.md in agent_worker package | **Applied** | `api/services/agent_worker/AGENTS.md` documents package scope, lifecycle, future-issue plug-in points, open-source guardrails. Required tightening `.gitignore` (the broad `AGENTS.md` pattern was ignoring subdirectory docs even though `docs/**/AGENTS.md` was tracked from earlier commits). Now scoped to `/AGENTS.md` (root only) with a clarifying comment. |

**Tests:**
- New: 30/30 pass (was 25; +5 for cap=0, negative cap, record(0), complete-fail rollback, resume success + resume failure)
- Pre-push hook full unit suite: 1220/1220 pass (was 1215 on `main`)

**Commit:** `c3a5bc1 fix: address review round 1 — cap=0 pause, startup resume, complete-fail rollback`

Issue #104, opened 2026-05-26
feat: Telegram clarifications via return-and-resume (F of F)


nbramia:
## Problem

Issues C and D send one-way Telegram notifications when an agent needs clarification — but the user's reply isn't routed back to the parked task. This issue closes the loop: replying (via Telegram reply-thread) to a clarification question wakes the parked agent session with the answer, resumes it, and the agent continues.

After this issue, the human-in-the-loop dimension is complete: ambiguous tasks, routing-ask cases, and any future agent-initiated clarifications all round-trip cleanly.

## Context

Part of the agent-worker series (epic **#98**). **Issue F**, depends on **#101** (Issue C — preflight produces clarification questions) and **#102** (Issue D — managed sessions need resume support). Independent from Issue E.

### 1. `pending_questions` table

```sql
CREATE TABLE pending_questions (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  session_id TEXT NOT NULL,
  task_id TEXT NOT NULL,
  question TEXT NOT NULL,
  sent_message_id INTEGER NOT NULL,       -- Telegram message_id of the question we sent
  sent_at INTEGER NOT NULL,
  answer TEXT,                            -- populated when user replies
  answered_at INTEGER,
  processed INTEGER DEFAULT 0,            -- worker sets to 1 once answer consumed
  timed_out INTEGER DEFAULT 0             -- worker sets to 1 after timeout fires
);
CREATE INDEX idx_pending_q_open ON pending_questions(answered_at, processed, timed_out)
  WHERE answered_at IS NULL AND timed_out = 0;
CREATE INDEX idx_pending_q_ready ON pending_questions(answered_at, processed)
  WHERE answered_at IS NOT NULL AND processed = 0;
```

### 2. Sending a clarification (worker side)

When a session needs to ask the user (from preflight `ambiguity`, `routing: "ask"`, or any agent-initiated clarification via a new `lifeos_user_ask(question)` tool):

1. Worker calls existing Telegram `send_message_async(question_text)`; captures returned `message_id`
2. Insert `pending_questions` row
3. Swap task tag → `#agent-blocked`
4. Mark session `blocked`
5. If parent session is managed: call `kill_session` to stop session-hour billing (consistent with yield pattern from Issue E)

### 3. Receiving a clarification (Telegram listener)

Modify `api/services/telegram.py`'s `_handle_update` — add a hook **at the top** before the chat pipeline:

```python
if update.message.reply_to_message:
[...]

Issue #103, opened 2026-05-26
feat: inter-agent tools — spawn, yield-and-resume, lineage budgets (E of F)


nbramia:
## Problem

Single-agent execution works after Issues C and D, but agents can't yet collaborate — there's no way for one agent to spawn another, send messages between live sessions, read each other's transcripts, or coordinate via the yield-and-resume pattern required to avoid blocking tool calls and idle billing.

This issue adds the `lifeos_agent_*` tool family that turns the worker into a real multi-agent orchestration system. A managed Opus session can fan out an army of local Gemma workers (or other Opus children) and coordinate with them.

## Context

Part of the agent-worker series (epic **#98**). **Issue E**, depends on **#100** (Issue B), **#101** (Issue C), and **#102** (Issue D). Independent from Issue F.

### 1. New MCP tool family

Added to `mcp_server.py` (so both managed and local agents can call them through the same MCP interface):

| Tool | Behavior |
|------|----------|
| `lifeos_agent_spawn(prompt, model: "claude" \| "local", budget?, expected_output?)` | Creates a new child task row + session, inherits caller's `session_id` as `parent_session_id`. Budget drawn from parent's remaining. Returns `{child_session_id}` immediately (async — child runs independently). |
| `lifeos_agent_send(session_id, message)` | Appends a user turn to a live child or peer session and wakes it. For yielded (terminated) sessions, queues the message for the next resume. |
| `lifeos_agent_check(session_id)` | Non-blocking status: `{status, last_turn_summary, tokens_used, dollars_used, last_activity_at}`. Use for short polling. |
| `lifeos_agent_yield_until(children: [session_id, ...], reason?)` | **Primary coordination primitive.** Tells the worker "terminate my session now and resume me when ALL listed children reach a terminal state." Worker: marks parent session `yielded`, kills it (managed) or returns control (local), tracks dependency. When all children done → creates fresh resumed session with children's outputs injected as a user turn. Returns from the tool call immediately; the agent's current turn completes and the session ends. |
| `lifeos_agent_kill(session_id, reason)` | Terminate a child. Only allowed on descendants of the caller. |
| `lifeos_agent_transcript_read(session_id, since_turn?)` | Read full or incremental transcript of any session. Backed by `data/agent_transcripts/{session_id}.jsonl`. Same format for managed and local. |
| `lifeos_agent_sessions_list(filters?)` | List sessions matching `{status?, model?, parent_session_id?, task_tag?}`. Returns metadata for each. |

[...]

PR #107, opened 2026-05-26
feat: Haiku preflight + local Gemma executor (C of F)

### LINKED ISSUE #101: feat: Haiku preflight + local Gemma executor (C of F)

nbramia:
## Summary
Issue C of the agent-worker series (epic #98). Replaces Issue B's no-op dispatcher with a real two-stage pipeline: a cheap Claude Haiku preflight call classifies each task (budget / routing / ambiguity / sanity / expected output), then a real agent loop on the local Gemma model executes via \`llm_client.LocalLLMClient\` with budget-enforced tool use. The Claude path is intentionally deferred to Issue D — it routes correctly but rolls the tag back to \`#agent\` and notifies. The \`#agent #local\` path is now end-to-end functional.

Closes #101 · Relates to #98

## Test evidence
- \`pytest tests/test_agent_worker_*.py tests/test_task_manager_swap_tag.py\` — **70/70 passed** (preflight 12, tools 17, executor 9, worker 9, stores 18, swap_tag 5)
- Pre-push hook full unit suite — **1260/1260 passed** (was 1220 on \`main\`; +40 new tests, no regressions)
- Pre-existing \`test_summarizer.py::test_real_summary_generation\` still fails on \`main\` (Ollama side, unrelated)

## Review focus
- **Preflight hardening** — `preflight.py:parse_preflight_response` defends against \`\`\`json\`\`\` fences, partial schemas, exceptions, and invalid routing/expected_output values. Failure modes always return \`sane=False routing=ask\` so the worker parks the task rather than running on junk.
- **Budget enforcement is external** — \`local_executor.execute\` checks token + wall + dollar at the top of each turn against the DB row, not via LLM self-report. Kill is graceful: status set to \`STATUS_BUDGET_EXCEEDED\`, transcript event recorded, tag swapped.
- **Sleep yield** — \`sleep(seconds)\` writes a \`sleeps\` row, returns \`ExecutorOutcome(status=STATUS_YIELDED, wake_at=...)\`, and \`Worker._wake_sleeping_sessions()\` resumes the executor at the right time. No idle billing in the cloud-managed case (Issue D will use the same primitive).
- **LifeOS MCP proxy via \`LifeOSMCPServer\`** — \`ToolRegistry\` instantiates the existing MCP server class and re-uses its \`tools\` list + \`_call_api\` + \`_format_response\`, so there's no duplication of the curated endpoint catalog. Standard tools (Read/Write/Edit/Bash/WebFetch) are minimal but functional.
- **WebSearch is a structured stub** — LifeOS doesn't ship a search backend. The tool is exposed so the agent can plan around it; calling it returns a clear \"not configured\" error rather than silently failing.
- **Pricing table is conservative** — \`pricing.cost_for\` falls through to Opus rates for unknown models so a typo can't accidentally suppress enforcement.
- **Worker outcome dispatch** — \`_handle_outcome\` centralizes terminal-state Telegram notifications and tag swaps. Sleep yields produce no Telegram (operator only hears about terminal states).
- **\`resume_pending\` semantics changed** — sleeping sessions are left alone (main loop wakes them); blocked sessions stay blocked (Issue F closes the loop); only RUNNING/CLAIMED get rolled back to \`#agent\` for retry. We can't safely re-enter a partial LLM conversation without risking duplicate side effects.

## What this enables
- \`#agent #local\` tasks now run end-to-end on Gemma with Telegram completion summaries
- Title cues without tags route correctly: \"with claude\" → claude (parked for D); \"using gemma\" → local
- Ambiguous tasks (\"reply to John\") land in \`#agent-blocked\` with the question via Telegram (one-way until F)
- Routing-ask tasks (no \`#local\`, no cue) also blocked with model-choice question
- Sleeping sessions correctly yield and resume across worker restarts
- Daily \$-cap kill-switch (\`LIFEOS_AGENT_DAILY_CAP_DOLLARS=0\`) works end-to-end

## Out of scope (covered in later issues)
- Managed Agents driver / Claude path execution (#102 — D)
- Inter-agent \`lifeos_agent_*\` tools, lineage budgets (#103 — E)
- Telegram clarification round-trip via reply-threading (#104 — F)
- WebSearch backend (depends on operator-side API key; not in v1 scope)

1. \`sudo systemctl restart lifeos-llm\` (Gemma must be running)
2. \`sudo systemctl restart lifeos-agent-worker\` (after rebuild)
3. Create a task: \`POST /api/tasks {"description": "echo the current date", "tags": ["agent", "local"]}\`
4. Within ~60s, expect a Telegram message like \"✅ Agent worker: completed 'echo the current date' (text) — N tokens, \$0.00, Ns wall.\\n\\nMay 26, 2026.\"

### DIFF
--- a/api/services/agent_worker/local_executor.py
+++ b/api/services/agent_worker/local_executor.py
@@ -48,6 +48,57 @@ MAX_TOOL_CALLS_PER_TURN = 16
 PER_TURN_MAX_TOKENS = 4096
 
 
+def _normalize_tool_calls(raw_calls) -> list[dict]:
+    """Normalize tool_calls into Anthropic-shape dicts.
+
+    `LocalLLMClient` returns raw OpenAI-format calls
+    (`{"id": ..., "type": "function", "function": {"name": ..., "arguments": "<json>"}}`).
+    `AnthropicLLMClient` returns Anthropic-shape `_ToolUseBlock` instances
+    or dicts with `name`/`input`. The executor wants a single shape so the
+    dispatcher and persistence layer stay simple. This function detects the
+    shape and produces `{"id", "name", "input"}` dicts in all cases.
+    """
+    if not raw_calls:
+        return []
+    normalized: list[dict] = []
+    for call in raw_calls:
+        # Anthropic-shape: object/dict with a `.name` and `.input`.
+        anth_name = getattr(call, "name", None)
+        anth_input = getattr(call, "input", None)
+        if anth_name and anth_input is not None:
+            normalized.append({
+                "id": getattr(call, "id", "") or f"call_{uuid.uuid4().hex[:12]}",
+                "name": anth_name,
+                "input": anth_input or {},
+            })
+            continue
+        if isinstance(call, dict):
+            if "function" in call and isinstance(call["function"], dict):
+                # OpenAI-shape — `arguments` is a JSON string.
+                func = call["function"]
+                args_raw = func.get("arguments", "{}")
+                if isinstance(args_raw, str):
+                    try:
+                        args = json.loads(args_raw)
+                    except json.JSONDecodeError:
+                        args = {}
+                else:
+                    args = args_raw or {}
+                normalized.append({
+                    "id": call.get("id") or f"call_{uuid.uuid4().hex[:12]}",
+                    "name": func.get("name", ""),
+                    "input": args,
+                })
+            elif "name" in call:
+                # Anthropic-shape dict.
+                normalized.append({
+                    "id": call.get("id") or f"call_{uuid.uuid4().hex[:12]}",
+                    "name": call["name"],
+                    "input": call.get("input") or {},
+                })
+    return normalized
+
+
 @dataclass
 class ExecutorOutcome:
     """What happened during a single `execute(session)` invocation."""
@@ -123,49 +174,53 @@ class LocalExecutor:
         """
         sid = session.session_id
         budget = session.budget or {}
-        wall_deadline = self._wall_deadline(session, budget)
         self.session_store.update_status(session.task_id, STATUS_RUNNING)
 
         if not self.session_store.get_messages(sid):
             self._seed_conversation(session, task, budget)
 
         while True:
-            # Wall-clock budget check.
-            if wall_deadline and time.time() > wall_deadline:
-                return self._finalize_budget_exceeded(session, "wall_seconds")
-
-            # Token budget check.
+            # Wall-clock budget check — uses cumulative active seconds, not
+            # wall-from-start (so sleeps don't eat into the run budget).
             updated = self.session_store.get(session.task_id)
+            if budget.get("wall_seconds"):
+                if (updated.total_active_seconds or 0) >= budget["wall_seconds"]:
+                    return self._finalize_budget_exceeded(session, "wall_seconds")
             tokens_used = (updated.total_input_tokens or 0) + (updated.total_output_tokens or 0)
             if budget.get("max_tokens") and tokens_used >= budget["max_tokens"]:
                 return self._finalize_budget_exceeded(session, "max_tokens")
-            # Dollar budget (local is $0 but the path is exercised for parity).
             if budget.get("max_dollars") is not None and (updated.total_dollars or 0) >= budget["max_dollars"]:
                 return self._finalize_budget_exceeded(session, "max_dollars")
 
+            turn_start = time.time()
             try:
                 response = self._call_llm(sid)
             except Exception as exc:
                 logger.exception("local executor LLM call failed: %s", exc)
+                # Charge the time we spent trying so the budget reflects real
+                # work even on failure.
+                self.session_store.record_active_seconds(session.task_id, time.time() - turn_start)
                 return self._finalize_failed(session, f"LLM call failed: {exc}")
 
-            self._persist_assistant_turn(sid, response)
-            self._record_spend(session, response)
+            # Normalize tool_calls to a single shape (OpenAI ↔ Anthropic).
+            normalized_calls = _normalize_tool_calls(response.tool_calls)
+            truncated_calls = normalized_calls[:MAX_TOOL_CALLS_PER_TURN]
 
-            tool_calls = response.tool_calls or []
-            if not tool_calls:
+            self._persist_assistant_turn(sid, response, truncated_calls)
+            self._record_spend(session, response)
+            self.session_store.record_active_seconds(session.task_id, time.time() - turn_start)
+
+            if not truncated_calls:
                 # Final answer — no more tool calls expected.
                 return self._finalize_completed(session, response.text)
 
             yielded_seconds: int | None = None
             tool_results: list[dict] = []
-            for i, call in enumerate(tool_calls[:MAX_TOOL_CALLS_PER_TURN]):
-                name = getattr(call, "name", None) or call.get("name", "")
-                args = getattr(call, "input", None) or call.get("input", {})
-                call_id = (getattr(call, "id", None)
-                           or call.get("id")
-                           or f"call_{uuid.uuid4().hex[:12]}")
-                result: ToolResult = self.tools.dispatch(name, args or {})
+            for call in truncated_calls:
+                name = call["name"]
+                args = call["input"]
+                call_id = call["id"]
+                result: ToolResult = self.tools.dispatch(name, args)
                 self.transcript_store.append(
                     sid,
                     "tool_call",
@@ -186,6 +241,20 @@ class LocalExecutor:
                     yielded_seconds = result.yield_seconds
                     # Stop dispatching further tools this turn — we're going to yield.
                     break
+
+            # If the agent yielded mid-turn, we may have fewer tool_results
+            # than persisted tool_use blocks. Pad with synthetic results so
+            # the next turn satisfies Anthropic's 1:1 invariant.
+            if yielded_seconds is not None and len(tool_results) < len(truncated_calls):
+                already_handled = {tr["tool_use_id"] for tr in tool_results}
+                for call in truncated_calls:
+                    if call["id"] not in already_handled:
+                        tool_results.append({
+                            "type": "tool_result",
+                            "tool_use_id": call["id"],
+                            "content": "skipped — sibling tool requested a sleep yield first",
+                            "is_error": False,
+                        })
 
             # Persist the tool_results as a user-role turn (Anthropic convention).
             self.session_store.append_message(sid, "user", tool_results)
@@ -226,22 +295,19 @@ class LocalExecutor:
             tools=self.tools.definitions(),
         )
 
-    def _persist_assistant_turn(self, session_id: str, response) -> None:
-        # Build an Anthropic-style content list mixing text + tool_use blocks.
+    def _persist_assistant_turn(self, session_id: str, response, normalized_calls: list[dict]) -> None:
+        """Persist the assistant turn using the *truncated* normalized call list
+        so tool_use and tool_result block counts stay 1:1 in later turns.
+        """
         content_blocks: list[dict] = []
         if response.text:
             content_blocks.append({"type": "text", "text": response.text})
-        for call in (response.tool_calls or []):
-            name = getattr(call, "name", None) or call.get("name", "")
-            args = getattr(call, "input", None) or call.get("input", {})
-            call_id = (getattr(call, "id", None)
-                       or call.get("id")
-                       or f"call_{uuid.uuid4().hex[:12]}")
+        for call in normalized_calls:
             content_blocks.append({
                 "type": "tool_use",
-                "id": call_id,
-                "name": name,
-                "input": args,
+                "id": call["id"],
+                "name": call["name"],
+                "input": call["input"],
             })
         usage = getattr(response, "usage", None)
         tokens_in = getattr(usage, "input_tokens", 0) if usage else 0
@@ -263,12 +329,6 @@ class LocalExecutor:
     # ------------------------------------------------------------------
     # Finalizers
     # ------------------------------------------------------------------
-
-    def _wall_deadline(self, session, budget: dict) -> float | None:
-        wall = budget.get("wall_seconds")
-        if not wall:
-            return None
-        return float(session.started_at) + float(wall)
 
     def _finalize_completed(self, session, final_text: str) -> ExecutorOutcome:
         self.session_store.update_status(session.task_id, STATUS_COMPLETED)
--- a/api/services/agent_worker/local_executor.py
+++ b/api/services/agent_worker/local_executor.py
@@ -1,0 +1,299 @@
+"""Local agent loop for #agent #local tasks.
+
+The executor drives one session of conversation with the local llama-server
+(Gemma in this LifeOS) through `llm_client.LocalLLMClient`. Each turn:
+
+  1. Load the session's prior messages from the DB
+  2. Call the LLM with the tool catalog
+  3. Persist the assistant turn + token usage
+  4. Check wall + token budgets — kill the loop on breach
+  5. If the model produced tool calls, dispatch each, append tool_result
+     blocks to the conversation, and loop
+  6. If the model produced a final text answer, mark the session complete
+
+Sleep is a "yield": the executor writes a `sleeps` row, returns
+`ExecutorOutcome(status="sleeping", ...)`, and the worker's main loop wakes
+the session at the requested time by calling `execute(session)` again.
+"""
+from __future__ import annotations
+
+import json
+import logging
+import time
+import uuid
+from dataclasses import dataclass
+
+from api.services.agent_worker.pricing import cost_for
+from api.services.agent_worker.session_store import (
+    STATUS_BUDGET_EXCEEDED,
+    STATUS_COMPLETED,
+    STATUS_FAILED,
+    STATUS_RUNNING,
+    STATUS_YIELDED,
+    SessionStore,
+)
+from api.services.agent_worker.tools import ToolRegistry, ToolResult
+from api.services.agent_worker.transcript_store import TranscriptStore
+
+
+logger = logging.getLogger(__name__)
+
+
+# Per-turn ceiling on tool calls. Prevents a single broken turn from
+# exhausting tools forever before the budget check fires.
+MAX_TOOL_CALLS_PER_TURN = 16
+
+# Per-LLM-call max_tokens cap. The real ceiling is the session's token
+# budget; this is just the per-request ask so we don't burn cap in one go.
+PER_TURN_MAX_TOKENS = 4096
+
+
+@dataclass
+class ExecutorOutcome:
+    """What happened during a single `execute(session)` invocation."""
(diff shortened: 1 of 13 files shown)


nbramia (commented):
## PR Review: feat: Haiku preflight + local Gemma executor (C of F)

### Action Required
- **[Correctness]** Tool-call shape mismatch breaks real Gemma execution. `LocalLLMClient._parse_response` (api/services/llm_client.py:383-396) returns OpenAI-format tool_calls `[{"id", "type": "function", "function": {"name", "arguments": "<json-string>"}}]`. The executor at `api/services/agent_worker/local_executor.py:163-164` extracts via `call.get("name", "")` / `call.get("input", {})`, which return `""` / `{}` for OpenAI-shaped calls. Verified by running `LocalLLMClient._parse_response` against a representative llama-server payload: tool dispatch silently produces `unknown tool: ""`. The 70/70 test pass is masked because the scripted `_ScriptedLLM` in `tests/test_agent_worker_local_executor.py` returns Anthropic-shape dicts (`{"id","name","input"}`), not the OpenAI shape the real client emits. The first manual smoke test ("echo the current date") with any model that actually calls a tool will fail. Either translate via the existing `openai_tool_calls_to_anthropic` helper at the executor boundary, or change `_persist_assistant_turn`/dispatch loop to consume `function.name` + `json.loads(function.arguments)`. Also add a test that uses the OpenAI shape so regression is impossible.

- **[Correctness]** Claude-route deferral creates a tag-flapping loop until Issue D ships. `worker.py:415-424` rolls the tag back to `#agent` and sets the session to `STATUS_BLOCKED` (non-terminal). Next tick, `_list_agent_tasks()` finds the task (still tagged `#agent`), `_claim` calls `swap-tag #agent → #agent-running` (succeeds), then `session_store.create(task_id=...)` raises `IntegrityError` on the `task_id PRIMARY KEY` constraint, the except block swaps the tag back to `#agent`, and logs an `ERROR session create failed` + `WARNING swap_tag failed`. I reproduced this end-to-end: every 60s poll generates two API mutations (tag flap visible to any external observer) and two log lines. Worse, when Issue D ships, the pre-existing `STATUS_BLOCKED` session row prevents `_claim` from ever succeeding because the IntegrityError still fires. Fix: either roll the tag to a distinct holding tag (e.g. `#agent-deferred`) so `_list_agent_tasks` doesn't keep re-picking it, or have `_claim` tolerate an existing session row for the same `task_id` when its status is `STATUS_BLOCKED` and the routing was `claude`.

### Recommended
- **[Correctness]** `MAX_TOOL_CALLS_PER_TURN = 16` (`local_executor.py:50`) silently truncates tool calls. If the model returns 17 calls, the persisted `assistant` turn (built in `_persist_assistant_turn`) still contains all 17 `tool_use` blocks (line 240 iterates `response.tool_calls`, not the slice), but only 16 `tool_result` blocks are appended at line 191. Anthropic-shape APIs reject mismatched tool_use/tool_result pairing; even the OpenAI conversion path will confuse the model. Either truncate the persisted assistant turn to match, or fabricate placeholder tool_result blocks for the dropped calls.

- **[Correctness]** Wall-clock budget counts sleep time. `_wall_deadline` (`local_executor.py:267-271`) is `session.started_at + wall_seconds`. A task that sleeps 8h with wall=4h will be killed on wake before doing any post-sleep work. Either reset `started_at` on wake, persist a `wake_count`/`active_seconds`, or document this as intended behavior (currently it isn't called out anywhere).

- **[Correctness]** `resume_pending` (`worker.py:148-160`) treats `STATUS_BLOCKED` from a prior crash as "leave alone — waiting on user." But a Claude-deferred session (also `STATUS_BLOCKED`) with tag `#agent` will be picked up by the next tick's `_list_agent_tasks` and hit the same IntegrityError loop described above. Resume needs to distinguish "blocked on user" (`#agent-blocked`) from "blocked on deferred routing" (`#agent`).

- **[Correctness/Race]** Pre-existing `messages` indexing race latent in this PR. `SessionStore.append_message` does `SELECT MAX(turn_index) + 1` then `INSERT` against a connection with `isolation_level=None` (autocommit). Today the single-worker / single-local-session invariant masks it, but Issue E's inter-agent spawn will surface concurrent appenders racing on the `PRIMARY KEY(session_id, turn_index)` constraint. Worth a one-line comment now (or wrap in `BEGIN IMMEDIATE`) so the next agent doesn't repeat the design mistake.

- **[Security]** `Bash` tool (`tools.py:206-221`) runs with `shell=True` and no sandbox. The PR description and `tools.py:31-32` call this out as intentional, but worth confirming the operator-facing setup guide (`docs/guides/`) flags it explicitly — anything in the user's task list now executes as the operator. Not blocking; flagging because it's a privacy/security boundary worth a docs cross-reference.

- **[Security]** `WebFetch` (`tools.py:233-244`) follows redirects with `follow_redirects=True` and no host allowlist. Combined with arbitrary task titles from the user's vault, a SSRF surface exists (e.g., agent fetches `http://169.254.169.254/...` from a compromised task). Low risk in a single-user self-hosted deployment, but the same defense-in-depth note applies as Bash.

### Minor
- **[Correctness]** `_completion_summary` (`worker.py:466-479`) computes `wall = int(time.time()) - int(refreshed.started_at)`. For a task that yielded once and resumed, this includes sleep time (matches the wall-budget behavior, but the Telegram message phrases it as "Ns wall" which the operator may read as "agent was active for Ns").

- **[Standards]** `i` is unused in `for i, call in enumerate(tool_calls[:MAX_TOOL_CALLS_PER_TURN])` (`local_executor.py:162`).

- **[Standards]** `from api.services.agent_worker.session_store import Session` (`worker.py:39`) is imported only for type-hints in `_handle_outcome` and `_mark_failed`/`_mark_blocked` — fine as runtime imports go, but a `TYPE_CHECKING` guard would match the file's existing import style.

- **[Standards]** `ExecutorOutcome.status: str` (`local_executor.py:54`) is a stringly-typed status; an `Enum` or `Literal[...]` would catch a typo like the existing `"budget_exceeded"` literal in `tests/test_agent_worker_loop.py:295` (test bypasses the symbolic constant — easy to drift).

- **[Standards]** `STATUS_RUNNING` is imported in `local_executor.py:31` but only used once at line 127. Fine. The bigger gap is that the executor sets `STATUS_RUNNING` immediately but never updates it back to `STATUS_CLAIMED` on entry from a sleep wake — if the executor crashes between `update_status(RUNNING)` and the first finalizer, the session is stuck `RUNNING` and `resume_pending` rolls the tag to `#agent` + marks `STATUS_FAILED`. That's the documented graceful path, so OK — flagging only that the docstring at `worker.py:138-144` doesn't mention sleeping wake-up crashes specifically.

### Summary
Two Action Required items must land before this is merge-safe: the tool-call shape mismatch breaks the headline "#agent #local runs end-to-end" claim against a real Gemma server (currently masked by test fakes), and the Claude-deferral path tag-flaps the task once per poll until Issue D ships, generating noise and a guaranteed re-claim failure when D lands. Recommended items are mostly defensive — the structural design is solid and the test coverage at the unit level is genuinely good. Once the two above are fixed and a regression test added that uses the real OpenAI tool-call shape from `LocalLLMClient`, this should be ready.

nbramia:

| 1 | Tool-call shape mismatch — OpenAI \`tool_calls\` vs Anthropic \`tool_use\` | Action Required | **Accept** | Real merge-blocker. \`LocalLLMClient._parse_response\` returns OpenAI shape; executor reads Anthropic shape. Tests masked it with hand-crafted dicts. Adding a normalizer in the executor that handles both shapes. |
| 2 | Claude-deferral tag-flap | Action Required | **Downgrade to Recommended (still applying)** | The reviewer missed that \`tick()\` has \`if self.session_store.get(task_id): continue\` which prevents re-claim. But parking the task back at \`#agent\` is confusing UX — when Issue D ships, the operator needs explicit retry semantics. Fixing by parking at \`#agent-blocked\` like other parked states (sanity-failed, ambiguity, routing-ask). |
| 3 | MAX_TOOL_CALLS_PER_TURN truncation diverges tool_use / tool_result counts | Recommended | **Accept** | Real: \`_persist_assistant_turn\` persists all tool_use blocks but the loop only emits truncated tool_results. Anthropic format requires 1:1 — next turn would fail. Truncating at persistence time. |
| 4 | Wall-budget includes sleep time | Recommended | **Accept** | Adding \`total_active_seconds\` to the session row, incrementing per-turn (not during sleep). Completion summary uses active seconds, not wall-from-start. |
| 5 | \`resume_pending\` skips BLOCKED blindly | Recommended | **Reject** | Resolved by #2 fix — blocked sessions now correspond to \`#agent-blocked\` tags, and operator-driven retry is the documented path. |
| 6 | \`append_message\` SELECT MAX + INSERT race | Recommended | **Accept** | Single-statement INSERT with subselect — costs nothing and removes the foot-gun for Issue E. |
| 7 | Bash/WebFetch unsandboxed | Recommended | **Accept** | Adding security note to setup guide cross-referencing the AGENTS.md rationale. |
| 8 | Unused \`i\` in enumerate | Minor | **Accept** | Trivial. |
| 9 | Stringly-typed status literal in test | Minor | **Accept** | Use symbolic constant. |
| 10 | \`Session\` import should be \`TYPE_CHECKING\` | Minor | **Reject** | The Session import is also referenced at runtime in a few annotations consumed by tests via dataclass introspection. Cost of the import is trivial; keeping concrete. |
| 11 | Completion summary "wall" misleads | Minor | **Accept** | Phrase as "active time" alongside the active-seconds fix in #4. |

**Findings forwarded:** 9 (2 Action Required → 1 elevated + 1 downgraded but applied; 5 Recommended; 2 Minor). All applied inline.

nbramia:
## Round 1 Addresser Summary

| 1 | Tool-call shape mismatch | **Applied** | `_normalize_tool_calls` handles OpenAI shape (`{"function": {"name", "arguments": <json-string>}}`), Anthropic-shape objects, and Anthropic-shape dicts uniformly. New test `test_executor_normalizes_openai_tool_calls` proves the fix against a representative OpenAI payload. |
| 2 | Claude-deferral parking | **Applied** | Now parks at `#agent-blocked` instead of rolling back to `#agent`. Same blocked semantics as ambiguity / sanity / routing-ask. Test renamed + updated. |
| 3 | Truncation diverges tool_use ↔ tool_result | **Applied** | `_persist_assistant_turn` uses the truncated normalized list. Mid-turn yield pads remaining tool_use blocks with synthetic "skipped" results so the next turn satisfies Anthropic 1:1. New test `test_executor_truncates_persisted_tool_use_to_match_dispatch`. |
| 4 | Wall budget includes sleep time | **Applied** | New `total_active_seconds` column accumulates per-turn execution time. Wall budget is now compared against `total_active_seconds`, not `now - started_at`. Completion summary says "active" instead of "wall". New test `test_executor_tracks_active_seconds_not_wall`. |
| 5 | `append_message` race | **Applied** | Single INSERT with `COALESCE(MAX(turn_index), -1) + 1` subselect; serialized through SQLite's writer lock. |
| 6 | Bash/WebFetch security note | **Applied** | New "Security model" section in `docs/guides/agent-worker-setup.md` documents the no-sandbox stance, SSRF surface, and operator responsibilities. |
| 7 | `STATUS_BUDGET_EXCEEDED` literal in test | **Applied** | Symbolic constant. |
| 8 | Unused `i` in enumerate | **Applied** | Refactored away as part of the truncation fix. |

**Tests:** new 73/73 pass in the agent-worker suites; pre-push hook full unit suite **1263/1263 passed** (was 1260).
**Commit:** `579cd34 fix: address review round 1 — tool-call shape, blocked tag, active seconds`

PR #110, opened 2026-05-26
feat: Telegram clarifications via return-and-resume (F of F)

### LINKED ISSUE #104: feat: Telegram clarifications via return-and-resume (F of F)

nbramia:
## Summary
Final issue of the agent-worker series (epic #98). Closes the human-in-the-loop dimension: ambiguous tasks and routing-ask cases now send Telegram messages with reply-threading; the operator's reply matches back to the pending question and the session resumes with the answer injected as a user turn. Adds \`lifeos_user_ask\` for agent-initiated mid-loop clarifications. After this, the full flow described in #98 works end-to-end.

Closes #104 · Relates to #98

## Test evidence
- \`pytest tests/test_agent_worker_*.py\` — **129/129 passed** (+11 new in test_agent_worker_clarifications)
- Pre-push hook full unit suite — **1324/1324 passed** (was 1318)

## Review focus
- **Reply-threading hook** in \`api/services/telegram.py:_handle_update\` is at the top of the handler, before chat-pipeline routing. Unmatched replies fall through to chat unchanged (regression-safe). The hook is wrapped in a try/except + lazy import so the chat-only deployment path is unaffected when the agent worker DB isn't initialized.
- **\`send_message_capture_id\`** returns the message_id of the first sent chunk so reply-threading lands at the start of the question even when long messages are split.
- **\`deposit_answer\`** keys on \`sent_message_id\` and returns False if no open question matches — the listener uses that to decide whether to short-circuit.
- **Worker \`_process_clarification_answers\`** runs each tick, drains answered+unprocessed questions, injects the answer as a user turn (\`(user answered via Telegram) {answer}\`), swaps tag back to \`#agent-running\`, re-invokes the local executor.
- **Timeout path** in \`_timeout_stale_clarifications\` is one-shot: marks \`timed_out=1\` and sends a single nudge. Task stays at \`#agent-blocked\`; operator re-tags to retry.
- **\`lifeos_user_ask\`** parks the session and returns a yield sentinel so the executor ends the loop. The tool returns \`error: telegram_unavailable\` if no bot configured — session NOT blocked in that case.
- **Injectable Telegram senders** (\`telegram_send\`, \`telegram_send_with_id\`) for tests so the suite doesn't send real messages on the host.

## What completes after this PR
The full flow described in epic #98:
1. Tag task \`#agent\` → worker claims atomically (Issue B)
2. Haiku preflight classifies budget / routing / ambiguity (Issue C)
3. Local Gemma OR Anthropic Managed Agents executes (Issues C+D)
4. Agent can spawn children, coordinate via yield-and-resume (Issue E)
5. Agent can ask the operator a clarifying question via Telegram; reply resumes the session (Issue F — this PR)
6. Completion sends a one-paragraph Telegram summary

## Out of scope (still deferred for follow-up)
- Managed yield-and-resume / clarification-resume — local path is end-to-end; managed cases fail with a clear "not supported" reason. Both depend on Managed-Agents session re-creation with history transfer.

1. \`sudo systemctl restart lifeos-agent-worker\`
2. Create an ambiguous task: \`POST /api/tasks {"description": "reply to John", "tags": ["agent", "local"]}\`
3. Expect a Telegram message ending with "Reply to this message to answer."
4. Tap reply, type "John Doe", send
5. Within ~60s the worker should resume, complete the task, and send a completion summary

### DIFF
--- a/api/services/agent_worker/inter_agent.py
+++ b/api/services/agent_worker/inter_agent.py
@@ -72,7 +72,7 @@ class InterAgentContext:
     when present, `kill` and other tools can reach remote managed sessions;
     when absent, managed targets are killed in the DB only.
 
-    `worker_handle` is optional; when present, `lifeos_user_ask` can route
+    `worker_handle` is optional; when present, `lifeos_agent_user_ask` can route
     a clarifying question through the worker's Telegram pipeline.
     """
     session_store: SessionStore
@@ -191,7 +191,7 @@ INTER_AGENT_TOOL_SCHEMAS: list[dict[str, Any]] = [
         },
     },
     {
-        "name": "lifeos_user_ask",
+        "name": "lifeos_agent_user_ask",
         "description": "Ask the operator a clarifying question via Telegram and pause until they reply. Your session ends; the worker resumes it (with the user's answer injected as a new user turn) once the reply arrives. Use sparingly — only when you genuinely cannot proceed without operator input.",
         "input_schema": {
             "type": "object",
@@ -497,7 +497,7 @@ def user_ask(ctx: InterAgentContext, args: dict) -> dict:
         return _err(f"caller session {ctx.caller_session_id} not found", code="no_caller")
     if ctx.worker_handle is None:
         return _err(
-            "lifeos_user_ask is only available inside a running session",
+            "lifeos_agent_user_ask is only available inside a running session",
             code="no_worker",
         )
 
@@ -528,7 +528,7 @@ DISPATCH_TABLE = {
     "lifeos_agent_kill": kill,
     "lifeos_agent_transcript_read": transcript_read,
     "lifeos_agent_sessions_list": sessions_list,
-    "lifeos_user_ask": user_ask,
+    "lifeos_agent_user_ask": user_ask,
 }
 
 
--- a/api/services/agent_worker/inter_agent.py
+++ b/api/services/agent_worker/inter_agent.py
@@ -32,6 +32,7 @@ from dataclasses import dataclass
 from typing import Any
 
 from api.services.agent_worker.session_store import (
+    STATUS_BLOCKED,
     STATUS_CLAIMED,
     STATUS_FAILED,
     STATUS_YIELDED,
@@ -70,12 +71,16 @@ class InterAgentContext:
     without knowing how they were called. `managed_driver` is optional —
     when present, `kill` and other tools can reach remote managed sessions;
     when absent, managed targets are killed in the DB only.
+
+    `worker_handle` is optional; when present, `lifeos_user_ask` can route
+    a clarifying question through the worker's Telegram pipeline.
     """
     session_store: SessionStore
     transcript_store: TranscriptStore
     caller_session_id: str
     caps: Caps
     managed_driver: Any | None = None
+    worker_handle: Any | None = None  # Worker — circular import avoided
 
 
 # ---------------------------------------------------------------------------
@@ -183,6 +188,17 @@ INTER_AGENT_TOOL_SCHEMAS: list[dict[str, Any]] = [
                 "limit": {"type": "integer"},
             },
             "required": [],
+        },
+    },
+    {
+        "name": "lifeos_user_ask",
+        "description": "Ask the operator a clarifying question via Telegram and pause until they reply. Your session ends; the worker resumes it (with the user's answer injected as a new user turn) once the reply arrives. Use sparingly — only when you genuinely cannot proceed without operator input.",
+        "input_schema": {
+            "type": "object",
+            "properties": {
+                "question": {"type": "string", "description": "The question to ask the operator. Keep it short and specific."},
+            },
+            "required": ["question"],
         },
     },
 ]
@@ -470,6 +486,40 @@ def sessions_list(ctx: InterAgentContext, args: dict) -> dict:
 # Public dispatcher — used by ToolRegistry and the MCP wrapper.
 # ---------------------------------------------------------------------------
 
+def user_ask(ctx: InterAgentContext, args: dict) -> dict:
+    """Agent-initiated Telegram clarification. Marks the caller blocked and
+    queues the question with the user via the worker."""
+    question = (args.get("question") or "").strip()
+    if not question:
+        return _err("question is required", code="invalid_arg")
+    caller = ctx.session_store.get_by_session_id(ctx.caller_session_id)
+    if caller is None:
+        return _err(f"caller session {ctx.caller_session_id} not found", code="no_caller")
+    if ctx.worker_handle is None:
+        return _err(
+            "lifeos_user_ask is only available inside a running session",
+            code="no_worker",
+        )
+
+    sent_id = ctx.worker_handle.ask_user_via_telegram(
+        session_id=caller.session_id,
+        task_id=caller.task_id,
+        question=question,
+    )
+    if sent_id is None:
+        return _err(
+            "could not send Telegram message — bot token may not be configured",
+            code="telegram_unavailable",
+        )
+
+    # Mark the caller blocked so the worker stops driving its loop.
+    ctx.session_store.update_status(caller.task_id, STATUS_BLOCKED)
+    ctx.transcript_store.append(caller.session_id, "user_ask", {
+        "sent_message_id": sent_id, "question_chars": len(question),
+    })
+    return _ok({"asked": True, "sent_message_id": sent_id, "blocked": True})
+
+
 DISPATCH_TABLE = {
     "lifeos_agent_spawn": spawn,
     "lifeos_agent_send": send,
@@ -478,6 +528,7 @@ DISPATCH_TABLE = {
     "lifeos_agent_kill": kill,
     "lifeos_agent_transcript_read": transcript_read,
     "lifeos_agent_sessions_list": sessions_list,
+    "lifeos_user_ask": user_ask,
 }
 
 
--- a/api/services/agent_worker/local_executor.py
+++ b/api/services/agent_worker/local_executor.py
@@ -303,10 +303,13 @@ class LocalExecutor:
             self.session_store.append_message(sid, "user", tool_results)
 
             if yielded_for_children:
-                # Status + yield_waiting_for is already set by the
-                # lifeos_agent_yield_until handler — just end the loop. Worker
-                # main loop will resume when children terminate.
-                self.transcript_store.append(sid, "yielded_for_children", {})
+                # Status is already set by the calling tool — either
+                # lifeos_agent_yield_until (children case) or
+                # lifeos_agent_user_ask (Telegram clarification case). Distinguish
+                # in the transcript by whether yield_waiting_for is populated.
+                refreshed = self.session_store.get(session.task_id)
+                kind = "yielded_for_children" if refreshed.yield_waiting_for else "yielded_for_user"
+                self.transcript_store.append(sid, kind, {})
                 return ExecutorOutcome(status=STATUS_YIELDED)
 
             if yielded_seconds is not None:
--- a/api/services/agent_worker/session_store.py
+++ b/api/services/agent_worker/session_store.py
@@ -100,6 +100,26 @@ CREATE TABLE IF NOT EXISTS pending_messages (
 );
 CREATE INDEX IF NOT EXISTS idx_pending_msgs_session ON pending_messages(session_id, delivered);
 
+-- Open clarification questions sent to the operator via Telegram (Issue F).
+-- The listener hook in telegram.py matches incoming reply_to_message ids
+-- against `sent_message_id` and deposits the answer. Worker.tick scans for
+-- answered+unprocessed rows to resume sessions, and for timed-out rows to
+-- send a follow-up nudge.
+CREATE TABLE IF NOT EXISTS pending_questions (
+    id                INTEGER PRIMARY KEY AUTOINCREMENT,
+    session_id        TEXT NOT NULL,
+    task_id           TEXT NOT NULL,
+    question          TEXT NOT NULL,
+    sent_message_id   INTEGER NOT NULL,
+    sent_at           INTEGER NOT NULL,
+    answer            TEXT,
+    answered_at       INTEGER,
+    processed         INTEGER NOT NULL DEFAULT 0,
+    timed_out         INTEGER NOT NULL DEFAULT 0
+);
+CREATE INDEX IF NOT EXISTS idx_pq_message_id ON pending_questions(sent_message_id);
+CREATE INDEX IF NOT EXISTS idx_pq_open ON pending_questions(answered_at, processed, timed_out);
+
 CREATE TABLE IF NOT EXISTS daily_spend (
     date           TEXT PRIMARY KEY,
     total_dollars  REAL NOT NULL DEFAULT 0.0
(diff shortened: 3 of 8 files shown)


nbramia (commented):
## PR Review: feat: Telegram clarifications via return-and-resume (F of F)

### Action Required
- **[Correctness]** `lifeos_user_ask` is unreachable from Managed Agents via MCP. `mcp_server.py:1088` dispatches inter-agent tools by `tool_name.startswith("lifeos_agent_")`, but the new tool is `lifeos_user_ask` (no `agent_` prefix). It is published in the MCP tool schema (because `_register_inter_agent_tools` iterates `INTER_AGENT_TOOL_SCHEMAS`, which now includes it), but a managed agent that calls it gets `{"error": "Unknown tool: lifeos_user_ask"}`. Even after fixing the dispatch prefix, `_handle_inter_agent` at `mcp_server.py:1031` builds an `InterAgentContext` with no `worker_handle`, so the tool would still return `{"error": "lifeos_user_ask is only available inside a running session", "code": "no_worker"}` from `inter_agent.py:498-502`. The PR description's claim that the tool works "for both managed and local agents mid-loop" (acceptance criterion) is not satisfied. Either (a) widen the MCP prefix check and route managed-agent `lifeos_user_ask` calls through some out-of-band channel to the worker, or (b) explicitly remove it from `INTER_AGENT_TOOL_SCHEMAS` for the MCP surface and document the local-only limitation.

- **[Correctness]** Routing-ask resume is broken. `worker.py:717` persists `session.routing = "ask"` before the ambiguity/ask-block check at line 737. When the user replies "local" or "claude" via Telegram, `_process_clarification_answers` at `worker.py:414` checks `if session.routing == "local"` — false for `"ask"` — and falls into the managed branch at line 425, which calls `_mark_failed` with "managed clarification resume not yet supported". The user's routing answer is never honored. The spec's acceptance criterion "Routing-ask task: same flow, with the answer ('local' or 'claude') routing the task to the right model" fails. Fix: on resume, if `session.routing == "ask"`, parse the answer, update `routing` to the chosen value, and dispatch accordingly (or re-run preflight). No test covers this path.

### Recommended
- **[Standards / Test coverage]** PR description claims "+11 new in test_agent_worker_clarifications" but the file contains only 6 tests (`pytest -v` confirms 6 collected). Either the count is wrong or several intended tests are missing — most notably the routing-ask round-trip, the combined `ambiguity + routing=ask` scenario from spec §7, the "stale reply falls through to chat" regression, and the `lifeos_user_ask`-over-MCP path. The PR description's "Test plan" makes claims the suite doesn't actually validate.

- **[Correctness]** Spec §7 calls for a single combined message when preflight returns both `ambiguity != null` AND `routing == "ask"`, plus best-effort parsing of a single combined reply. The implementation at `worker.py:734-743` joins the questions with a space and sends one Telegram message (good), but on resume injects the raw answer as one user turn with no parsing — the resumed executor has to guess. Either implement the small Haiku parsing pass the spec suggests, or update the spec/PR description to record the deferral.

- **[Correctness]** Spec §2 step 5 requires: "If parent session is managed: call `kill_session` to stop session-hour billing." The PR doesn't do this anywhere — managed sessions that hit ambiguity continue accruing managed billing while waiting for a human reply (could be up to 72h). PR description acknowledges managed clarification resume is deferred, but `kill_session` for billing-stop on the *initial* block is a separate concern and should be addressed even in the local-MVP scope.

- **[Standards / Acceptance criteria]** The issue's "Yielded managed session during clarification accrues no session-hour charges" acceptance criterion is not met (same root cause as above).

### Minor
- **[Performance]** `pending_questions` indexes (`session_store.py:120-121`) are full indexes; the spec called for partial indexes (`WHERE answered_at IS NULL AND timed_out = 0` and `WHERE answered_at IS NOT NULL AND processed = 0`). Negligible at expected row counts, but inconsistent with the spec.

- **[Correctness / Cosmetic]** When `lifeos_user_ask` triggers, `local_executor.py:309` writes a `yielded_for_children` transcript event even though no children exist. The session also ends up with executor-reported `STATUS_YIELDED` while DB status is `STATUS_BLOCKED` (set by `user_ask`). Functionally fine because `_resume_yielded_for_children` keys on a non-empty `yield_waiting_for`, but the transcript label is misleading — consider a distinct event name (e.g. `yielded_for_user_ask`).

- **[Concurrency]** `_process_clarification_answers` reads-then-marks-then-executes in `worker.py:390-424` without holding a transaction across the steps. The class doc says "Single-process poll loop. One instance per process," so this is OK as long as the systemd unit enforces singleton — worth a `flock` or PID-file check on startup to be safe.

- **[Robustness]** If the worker crashes between `mark_question_processed` (`worker.py:409`) and `executor.execute` (`worker.py:417`), the answer is durably injected into the conversation but the resume never completes. `resume_pending` then flips the session to FAILED and rolls the tag back to `#agent`, which re-runs preflight. The user's answer is silently lost. Documented design tradeoff ("can't safely re-enter a partially-driven LLM conversation") — but worth a comment in `_process_clarification_answers` noting the failure mode.

- **[Robustness]** `send_message_capture_id` (`telegram.py:115`) uses synchronous `httpx.post` with 30s timeout per part, called from inside the worker tick. Slow Telegram API = blocked worker. Not new behavior (the existing `send_message` does the same), but worth noting as the worker now calls Telegram on every clarification path.

- **[Standards]** `tools.py:226` adds `import json` at module top while the existing local `import json as _json` is removed inline — clean, but worth confirming no other module-level `json` collision was introduced.

### Summary
Tests pass and the local happy path works end-to-end, but two Action Required items mean the PR fails its own acceptance criteria for "managed agent uses `lifeos_user_ask`" and "routing-ask answer routes the task." Recommend changes before merge: at minimum, fix the MCP-dispatch prefix issue (or remove `lifeos_user_ask` from the published MCP schema), and either implement routing-ask resume or explicitly document it as deferred and update the spec/PR claims to match.

nbramia:

| 1 | lifeos_user_ask unreachable via MCP (prefix mismatch + missing worker_handle) | Action Required | **Accept** | Rename to `lifeos_agent_user_ask` so it matches the `lifeos_agent_*` family + MCP dispatcher. For the missing worker_handle, MCP-called user_ask returns a clear "managed unsupported" error (Issue D limitation; local path remains fully functional). |
| 2 | Routing-ask resume broken — session.routing="ask" misroutes on resume | Action Required | **Accept** | Real bug. On answer resumption, parse "local"/"claude"/"opus" from the answer text, update session.routing, dispatch through the right executor. |
| 3 | PR description test count wrong (claims 11; file has 6) | Recommended | **Accept** | Fix description; add routing-ask round-trip + stale-reply regression tests. |
| 4 | Combined ambiguity+routing reply not parsed | Recommended | **Accept** | Best-effort: split answer on numbered list / commas; first half is ambiguity, second half is routing model. |
| 5 | Managed sessions accrue billing during clarification wait | Recommended | **Accept** | Call driver.kill_session when blocking a managed session — managed clarification resume is documented as unsupported anyway, so killing is correct here. |
| 6 | Full indexes vs spec's partial indexes | Minor | **Reject** | Negligible at expected row count; not worth migration. |
| 7 | "yielded_for_children" transcript label on user_ask | Minor | **Accept** | Rename to "yielded_for_user". |
| 8 | Crash between mark_processed and execute loses answer | Minor | **Accept** | Move mark_processed AFTER successful execute (parallel to the Issue E lesson on yield_waiting_for). |
| 9 | Synchronous Telegram post blocks tick | Minor | **Reject** | 30s ceiling; rare; out of scope for MVP. Documented in worker docstring. |
| 10 | flock/PID-file concurrency | Minor | **Reject** | systemd enforces singleton; documented invariant. |

**Findings forwarded:** 7 (2 AR + 3 Rec + 2 Minor). All applied inline.

nbramia:
## Round 1 Addresser Summary

| 1 | lifeos_user_ask unreachable via MCP | **Applied** | Renamed to `lifeos_agent_user_ask` — matches MCP dispatcher prefix. Local executor path fully works; managed-side returns clear `no_worker` error (deferred). |
| 2 | Routing-ask resume broken | **Applied** | `_parse_routing_answer` extracts "local"/"claude"/"gemma"/"opus" from text. On answer, session.routing updated; executor dispatched correctly. Unparseable replies trigger a re-ask via Telegram. New test for the round-trip. |
| 3 | PR test count claim wrong | **Applied** | Now actually 11 tests (added 4 here + 6 from initial submission + 1 stale-reply regression). |
| 4 | Combined ambiguity+routing parsing | **Applied** | `_parse_routing_answer` finds the model keyword nearest the end of the answer. Handles "1. John Doe 2. local" correctly. |
| 5 | Managed sessions accrue billing during wait | **Applied** | `_mark_blocked` calls `driver.kill_session` when the session has a `managed_agent_session_id`. Stops session-hour billing during the wait. |
| 7 | "yielded_for_children" label misleading for user_ask | **Applied** | Distinguished `yielded_for_children` vs `yielded_for_user` by checking yield_waiting_for. |
| 8 | mark_processed-then-crash loses answer | **Applied** | mark_question_processed moved after successful execute. Crash mid-resume leaves question unprocessed for retry. |

**Tests:** 133/133 pass in agent-worker suites; pre-push hook full unit suite **1328/1328 passed** (was 1324).
**Commit:** `c444e95 fix: address review round 1 — rename, routing-ask resume, managed kill on block`

Issue #111, opened 2026-05-26
feat: self-hosted sandbox support for cloud-routed agent worker


nbramia:
## Problem

By default, Managed Agents sessions run in Anthropic-hosted Linux containers, so the agent's `bash` / `read` / `write` / `edit` / `glob` / `grep` tools operate on the container's ephemeral filesystem — not the operator's vault, code, scratch directories, or any other persistent state on their machine. For agents whose tasks span "read X file, transform it, write Y back" or "run this script in my repo", the cloud-sandbox restriction makes those workflows impossible without round-tripping through MCP servers.

Anthropic supports a self-hosted environment model that solves this cleanly: orchestration (model, planning, event stream) stays on Anthropic's side, but tool execution is delegated to an `ant beta:worker poll` process running on the operator's box. Outbound-only polling, environment-scoped auth key separate from the API key (so a compromised agent can't escalate to the full account).

## Context

Discovered during the agent worker bring-up on 2026-05-26. While wiring up the cloud-routed Claude path, the operator wanted "give this thing all the access it might possibly need including local files if possible". Cloud sandbox can't do that. Self-hosted sandbox can. We deferred it to keep the initial bring-up scope contained.

The security posture is the same as the existing local Gemma executor (Issue C / #101), which already runs `bash`/`read`/`write`/`edit` with no sandbox per the agent-worker design doc. So this is extending an established trust model to the cloud path, not introducing a new one.

References:
- [Managed Agents self-hosted sandboxes](https://platform.claude.com/docs/en/managed-agents/self-hosted-sandboxes) — architecture, `ant` CLI usage
- [Self-hosted sandbox security model](https://platform.claude.com/docs/en/managed-agents/self-hosted-sandboxes-security) — shared responsibility model
- Epic [#98](https://github.com/nbramia/LifeOS/issues/98) — agent worker design (closed)

## Suggested approach

Make self-hosted opt-in via env vars; cloud remains the default.

1. **New env vars** — `LIFEOS_AGENT_ENVIRONMENT_ID` (`env_…`) and `LIFEOS_AGENT_ENVIRONMENT_KEY` (`sk-ant-oat01-…`). Document in `.env.example` with scope/security notes (environment key is separate from API key; do not commit; chmod 600).

2. **New systemd unit** `lifeos-ant-worker.service` running:
   ```
   ant beta:worker poll \
     --workdir <operator-configurable> \
     --unrestricted-paths \
     --max-idle 10m
   ```
[...]

PR #113, opened 2026-05-26
feat: refactor managed_driver against real Managed Agents API

### LINKED ISSUE #112: feat: refactor managed_driver.py against real Anthropic Managed Agents API
## Problem

The current `api/services/agent_worker/managed_driver.py` was implemented against an assumed Managed Agents API shape (`platform.claude.com/v1/managed-agents/sessions`, inline `system_prompt` + `mcp_servers` + `connectors` per session, polling-based event consumption). The actual API published at https://platform.claude.com/docs/en/managed-agents is meaningfully different and our current code 404s on every session create — so every `#agent` task that routes to Claude (i.e., every task without `#local`) lands at `#agent-failed`. Only local Gemma works today.

## Context

Discovered during the agent worker bring-up on 2026-05-26. Fetched the actual docs and confirmed the API surface differs from what was built. The operator has already done the Anthropic-side setup that the new shape requires:

- **Agent preset** created in the Anthropic console:
  - Agent ID: `agent_019QrftoZjdF2htVQnXw2Z8c`
  - Model: `claude-sonnet-4-6`
  - All 9 MCP servers wired (LifeOS, Gmail, Cal, Drive, Superhuman, Slack, Ramp, Granola, Asana) via the `mcp_servers` + matching `mcp_toolset` pattern the docs require
  - `agent_toolset_20260401` enabled (bash / read / write / edit / glob / grep / web_fetch / web_search)
- **Vault** created and populated with OAuth creds for all 9 MCPs
- **Telegram clarification** path proven on the local Gemma side; reuses for cloud once driver works

What still needs operator action (one click): create a cloud **Environment** in the console (Workspace → Environments → New → Cloud) and paste the `env_…` id into `.env`. Or the driver can create one programmatically on first run if `LIFEOS_AGENT_ENVIRONMENT_ID` is empty.

Related future work: [#111](https://github.com/nbramia/LifeOS/issues/111) — self-hosted sandbox support, which builds on this refactor.

## Suggested approach

Full rewrite, not incremental — the auth, endpoints, body shapes, and event consumption all change.

**API surface changes** (per https://platform.claude.com/docs/en/managed-agents):
- Base URL: `https://api.anthropic.com/v1` (not `platform.claude.com/v1/managed-agents`)
- Headers split: `x-api-key`, `anthropic-version: 2023-06-01`, `anthropic-beta: managed-agents-2026-04-01` (the beta is its own header, not folded into the version)
- Resources: Agent (read-only from our side — operator manages in console), Environment (created once, referenced by id), Session (per-task, body is just `{agent, environment_id, vault_ids, metadata?, title?}`)
- Events: send via `POST /v1/sessions/{id}/events` with `events: [{type: "user.message", content: [...]}]`; consume via SSE `GET /v1/sessions/{id}/stream` (or non-streaming `GET /v1/sessions/{id}` for state polling — viable for a poll-based worker)

**Settings/env changes**:
- Add `LIFEOS_AGENT_PRESET_ID` (the `agent_019QrftoZjdF2htVQnXw2Z8c` above)
- Add `LIFEOS_AGENT_ENVIRONMENT_ID` (operator-provided or driver-created)
- Drop or deprecate `LIFEOS_AGENT_CONNECTORS` and `LIFEOS_AGENT_EXTRA_MCP_SERVERS` — both now live in the agent preset, not in per-session config
- Keep `LIFEOS_AGENT_VAULT_ID` but pass as `vault_ids: [<id>]` in the session body (the API expects an array)
- Keep `LIFEOS_AGENT_MANAGED_MODEL` for the next agent revision but it's now informational — the actual model is whatever the preset says

**Event consumption**: prefer state polling (`GET /v1/sessions/{id}`) over SSE for the first cut. Matches the existing polling architecture and avoids long-lived connections in the worker process. SSE can be added later if perf demands it.

**Cost/usage accounting**: the new event stream carries usage and final-text events. Adapter logic in `managed_executor.py` needs to extract usage from session-state polls (cumulative `input_tokens` / `output_tokens`) and compute deltas the same way as today.

**Cloud Environment provisioning**: simplest is to require operator to create one in console and set `LIFEOS_AGENT_ENVIRONMENT_ID`. Slightly nicer: driver auto-creates a default env on first use if unset, persists the returned id back to `.env`. Pick the simpler one first.

## Files involved

- `api/services/agent_worker/managed_driver.py` — full rewrite (new base, headers, endpoints, body shapes)
- `api/services/agent_worker/managed_executor.py` — switch from inline-mcp-servers session create to agent_id + environment_id reference; rewire usage delta computation
- `api/services/agent_worker/worker.py` — drop the inline `mcp_servers` / `connectors` construction in `_get_managed_executor`; wire `agent_preset_id` + `environment_id` through
- `config/settings.py` — add the two new fields; deprecate the two old ones (keep parsing so existing `.env` files don't error, but log-warn)
- `tests/test_agent_worker_managed_driver.py` — rewrite against the new API surface using `MockTransport`
- `tests/test_agent_worker_managed_executor.py` — update for new state-polling shape
- `tests/test_agent_worker_loop.py` — adjust the Claude-routing tests to use the new env vars
- `.env.example` — document the new vars; mark the deprecated ones; add a pointer to the agent-setup console flow
- `docs/guides/agent-worker-setup.md` — replace the "Step 4b — Provision the Managed Agents Vault" section with the new flow: (a) create the cloud Environment, (b) create the Agent preset with the YAML we've now confirmed works, (c) save IDs to `.env`, (d) point at the agent for reference
- `.claude/projects/-home-nathanramia-Code-LifeOS/memory/MEMORY.md` and any related memory entries that reference the old API shape

## Acceptance criteria

- [ ] `pytest -m unit -k agent_worker` passes against the rewritten driver/executor (no `MockTransport` 404s)
- [ ] Live smoke test: `curl -X POST /api/tasks {"description": "Using Claude: ... budget $0.50", "tags": ["agent"]}` ends at `#agent-completed` with a Telegram summary and a recorded `total_dollars > 0` in `data/agent_sessions.db`
- [ ] Worker log shows session-create going to `https://api.anthropic.com/v1/sessions` (not `platform.claude.com`) with both `anthropic-version` and `anthropic-beta` headers
- [ ] Budget enforcement still works mid-flight: a task with a tight budget that overruns cleanly kills the remote session and lands at `#agent-budget-exceeded`
- [ ] Telegram clarification flow on routing=ask still works (no regression on existing F-of-F behavior)
- [ ] Local Gemma path unchanged
- [ ] `.env.example` documents `LIFEOS_AGENT_PRESET_ID` and `LIFEOS_AGENT_ENVIRONMENT_ID` clearly, with the console URL for creating each
- [ ] Setup guide reflects the new console-first flow (agent preset YAML, environment creation, vault wiring)
- [ ] `./scripts/test.sh` passes end to end

nbramia:
## Summary

The pre-refactor `managed_driver.py` was built against an assumed Managed Agents API shape (`platform.claude.com/v1/managed-agents/sessions`, inline `system_prompt` + `mcp_servers` + `connectors` per session, polling-based event consumption). The actual API per [Anthropic's published docs](https://platform.claude.com/docs/en/managed-agents) is meaningfully different and our current code 404s on every session create — so every `#agent` task that routes to Claude lands at `#agent-failed`. Only local Gemma worked before this PR.

This PR fully rewrites the driver and updates the executor/worker glue/settings/tests/docs to match the real surface. Result: cloud-routed `#agent` tasks actually work.

Closes #112.

## What changed (architecture)

| Concern | Before | After |
|---|---|---|
| Base URL | `platform.claude.com/v1/managed-agents` | `api.anthropic.com/v1` |
| Auth headers | One header conflated version + beta | Split: `anthropic-version: 2023-06-01` AND `anthropic-beta: managed-agents-2026-04-01` |
| Resources | Per-session inline (`system_prompt`/`mcp_servers`/`connectors`) | Three reusable resources (Agent preset, Environment, Vault) created in the console once, referenced by ID per session |
| Session body | `{model, system_prompt, initial_message, mcp_servers, connectors, vault_id, budgets}` | `{agent, environment_id, vault_ids, metadata, title}` |
| Events | Mythical `/messages` endpoint | `POST /sessions/{id}/events` with `{events: [{type: "user.message", content: [...]}]}` |
| Terminal signals | Status field on response | Synthesized from event stream: `session.status_idle` → completed, `session.error` → failed; 404 from poll → cancelled |

## Files touched

- `api/services/agent_worker/managed_driver.py` — full rewrite (~240 lines)
- `api/services/agent_worker/managed_executor.py` — new constructor + `start()` flow; polling / budget / transcript paths unchanged
- `api/services/agent_worker/worker.py` `_get_managed_executor` — simplified ~50 lines down to ~15
- `config/settings.py` — adds `LIFEOS_AGENT_PRESET_ID` + `LIFEOS_AGENT_ENVIRONMENT_ID`; marks `LIFEOS_AGENT_CONNECTORS` and `LIFEOS_AGENT_EXTRA_MCP_SERVERS` deprecated (kept for back-compat .env parsing, no-op in code)
- `tests/test_agent_worker_managed_driver.py` — full rewrite (22 tests) for new HTTP surface
- `tests/test_agent_worker_managed_executor.py` — full rewrite (14 tests) for new constructor + start flow
- `tests/test_agent_worker_loop.py` — replaces the two obsolete `EXTRA_MCP_SERVERS` tests with four tests covering the new gating logic
- `.env.example` — documents the new IDs; comments-out deprecated vars
- `docs/guides/agent-worker-setup.md` — Step 4b rewritten end-to-end with the full agent-preset YAML pattern, vault matching, environment creation, and how IDs flow to `.env`

## Test evidence

```
pytest tests/test_agent_worker_managed_driver.py        — 22/22 passed
pytest tests/test_agent_worker_managed_executor.py      — 14/14 passed
pytest tests/test_agent_worker_loop.py                  — 14/14 passed (incl. 4 new gating tests)
pytest -m unit -k 'agent_worker'                        — 149/149 passed
pytest -m unit                                          — 1350/1352 passed
```

The 2 unit-suite failures (`test_real_ollama_latency`, `test_real_summary_generation`) are pre-existing per AGENTS.md "Tests Are Sacred" history and unrelated to this PR.

## Review focus

- **API surface mapping** in `managed_driver.py` — base URL, headers, body shapes, endpoint paths. Cross-check against the [Managed Agents docs](https://platform.claude.com/docs/en/managed-agents) and the [MCP connector reference](https://platform.claude.com/docs/en/managed-agents/mcp-connector). The API is in beta so the schema may shift; check for divergence.
- **Event-stream synthesis** (`_synthesize_status_from_events`, `_extract_final_text`, `_extract_error_reason` in driver) — these encode my reading of the docs. If `session.status_idle` event nesting differs or `final_text` lives somewhere else, this is where it'll break.
- **404 → cancelled mapping** in `get_session_state` — relies on the assumption that DELETEd sessions return 404 on subsequent polls. Plausible but not confirmed.
- **`agent_preset_id` + `agent_environment_id` gating** in `_get_managed_executor` — both required; vault is optional. If an operator with no OAuth-protected MCPs configures preset+environment but no vault, the worker should still attempt cloud sessions. New test (`test_managed_executor_omits_vault_ids_when_vault_unset`) covers this.
- **Deprecated field parsing** in `settings.py` — `LIFEOS_AGENT_CONNECTORS` and `LIFEOS_AGENT_EXTRA_MCP_SERVERS` are kept as Field()s so existing `.env` files don't error. They're not referenced in code anymore. Removing them entirely (breaking back-compat for one cycle) is also defensible — flag if preferred.
- **Live smoke test outstanding**: PR doesn't include a live cloud-task run because the operator's `LIFEOS_AGENT_ENVIRONMENT_ID` isn't yet set in `.env`. Acceptance criterion 2 (`#agent-completed` from a live Claude task) requires the operator to create the cloud environment in the console first, then merge can be validated.

## Companion follow-up

[#111](https://github.com/nbramia/LifeOS/issues/111) — self-hosted sandbox support, blocked on this refactor landing. After merge, the agent worker can be pointed at a self-hosted environment ID with no driver changes.


nbramia (commented):
## PR Review: feat: refactor managed_driver against real Managed Agents API

### Action Required
- **[Correctness]** `_extract_final_text` returns `None` when the terminal poll batch contains only `session.status_idle` (no `agent.message`). Because `get_session_state` uses the `after`/`since` cursor, each poll returns only events since `last_event_id`. If the agent's final `agent.message` arrived in poll N-1 and `session.status_idle` arrives alone in poll N, `state.final_text` is `None`, so `_finalize_remote` in `managed_executor.py:240` records `final_text=""` and the Telegram completion summary is empty. Either accumulate `final_text` across polls in the session row (e.g. persist the most recent `agent.message` text and only overwrite when a new one arrives), or have the driver fetch session history without a cursor on the terminal poll. See `api/services/agent_worker/managed_driver.py:218-220, 288-308`.
- **[Correctness]** Operator-facing Telegram message in the not-configured branch is wrong. `api/services/agent_worker/worker.py:843, 849-851` still says "Set ANTHROPIC_API_KEY and LIFEOS_AGENT_VAULT_ID in .env", but the gating in `_get_managed_executor` (worker.py:725-728) now requires `agent_preset_id` and `agent_environment_id` (vault is optional). Operators following the message will not unblock the task. Update message to reference `LIFEOS_AGENT_PRESET_ID` and `LIFEOS_AGENT_ENVIRONMENT_ID`.

### Recommended
- **[Correctness]** `_synthesize_status_from_events` returns `"completed"` on the first `session.status_idle` it encounters and never checks subsequent events for `session.error`. If both event types arrive in the same poll batch (idle first), an error is silently swallowed. The current ordering assumption ("idle is final, ignore later events") should be documented in the function docstring, or the loop should make a second pass and prefer "failed" over "completed" when both are present. `api/services/agent_worker/managed_driver.py:276-285`.
- **[Standards]** Default-value drift between settings and code creates operator confusion: `config/settings.py:145` defaults `agent_managed_model` to `claude-opus-4-7`, `ManagedExecutor.__init__` (managed_executor.py:85) defaults to `claude-sonnet-4-6`, and `.env.example` + docs say `claude-sonnet-4-6`. Cost calculation uses `self.model`, so if the operator doesn't set the env var, the executor takes whatever the worker passes (which is the settings default, opus). Pick one default and apply it consistently.
- **[Standards]** Deprecated `agent_connectors` and `agent_extra_mcp_servers` settings are now dead code paths — declared in `config/settings.py:179-194` but never read in any code (verified with grep). Either remove them (with a release note about removing from .env) or keep them but add a startup warning when they're non-empty so the operator knows they're being ignored.

### Minor
- **[Standards]** Test `test_managed_executor_constructed_when_preset_and_environment_set` (`tests/test_agent_worker_loop.py:333-353`) monkeypatches `agent_vault_id` but doesn't assert that the vault setting flows through to the constructed executor's `vault_ids` — it does (line 353) — but the test name says "preset and environment set" while it actually tests all three. Either rename or split the vault assertion into the existing `test_managed_executor_omits_vault_ids_when_vault_unset` test's positive counterpart.
- **[Correctness]** `_extract_error_reason` returns only the first `session.error` event's message. If multiple errors arrive in one batch (e.g., cascade), only the first is surfaced. Low priority but worth a comment.
- **[Standards]** `managed_driver.py:184-189` comment claims the cursor field "is stable across both because the API ignores unknown params" — this only holds if the API actually ignores `since` when `after` is the supported field. If the API rejects unknown params with 4xx in a future revision, polling will break silently. Consider making the param name a class constant so it's easy to flip later.
- **[Standards]** `agent-worker-setup.md` lost its earlier mention that Slack/Asana are out of scope for v1 — minor doc continuity issue, but not load-bearing.

### Summary
Largely merge-ready architecture refactor with strong test coverage (50/50 agent-worker tests pass locally), but two Action Required items need addressing before merge: the empty-`final_text` race when `session.status_idle` arrives alone on a poll, and the stale env-var name in the operator-facing blocked-task Telegram message. Both are small surgical fixes.

nbramia:

| 1 | Empty `final_text` race when `session.status_idle` arrives alone in a later poll batch | Action Required | **Accept** | Verified by reading driver+executor: `get_session_state` returns only events since cursor, so the agent.message text doesn't carry forward. Real bug — completion summaries would be empty across the wire. Fix: cache final_text on the session row in the executor. |
| 2 | Stale env var in blocked-task Telegram message ("Set ANTHROPIC_API_KEY and LIFEOS_AGENT_VAULT_ID") | Action Required | **Accept** | Verified in `worker.py:849-851`. Should say `LIFEOS_AGENT_PRESET_ID` + `LIFEOS_AGENT_ENVIRONMENT_ID`. Operator-facing string, must be correct. |
| 3 | `_synthesize_status_from_events` returns "completed" on first `status_idle`, ignores later `session.error` in same batch | Recommended | **Accept** | Both events in one batch is plausible. Prefer "failed" when both present — error is the actionable signal. |
| 4 | Default-model drift: settings=opus-4-7, executor default=sonnet-4-6, docs=sonnet-4-6 | Recommended | **Accept** | Settling on `claude-sonnet-4-6` everywhere — matches the operator's actual preset + the cost-conscious choice we agreed on. |
| 5 | Deprecated `agent_connectors`/`agent_extra_mcp_servers` parsed but never read | Recommended | **Accept (modified)** | Adding a one-time startup log warning when either is set, so operators see explicit deprecation feedback. Removing the fields entirely would silently swallow them via pydantic's default ignore behavior — log-warn is louder. |
| 6 | Test name vs behavior mismatch (`test_managed_executor_constructed_when_preset_and_environment_set` also tests vault) | Minor | **Accept** | Rename to be honest about coverage. |
| 7 | `_extract_error_reason` returns only first `session.error` | Minor | **Reject** | First error is sufficient for the Telegram reason field. Full transcript captures all events. Returning the cascade would complicate the simple `str` return signature for no caller benefit. |
| 8 | Promote cursor param key (`"after"`) to a class constant | Minor | **Accept** | Trivial; makes future API drift one-line to fix. |
| 9 | Lost "Slack/Asana out of scope for v1" doc note | Minor | **Reject** | The note was specific to the old architecture where connectors were passed inline. Custom MCPs (Slack/Ramp/Granola/Asana) are now first-class via the agent preset's `mcp_servers`. The deletion is correct, not regression. |

**Findings forwarded to addresser:** 7 (2 Action Required + 3 Recommended + 2 Minor)

nbramia:
## Round 1 — Addresser Summary

| 1 | Empty `final_text` race when `session.status_idle` arrives alone | **Applied** | Executor now caches `agent.message` text in `managed_cursor.final_text` on every poll where the driver returns non-None text. Finalize falls back to the cached value when the terminal batch is text-less. Idempotent SQLite `ALTER TABLE` migration handles existing DBs. New test `test_poll_carries_final_text_forward_across_cursor_batches` proves the path. |
| 2 | Stale env var in blocked-task Telegram message | **Applied** | Message now says "Set `ANTHROPIC_API_KEY`, `LIFEOS_AGENT_PRESET_ID`, and `LIFEOS_AGENT_ENVIRONMENT_ID`" with a docs pointer. |
| 3 | `_synthesize_status_from_events` should prefer "failed" when idle+error both appear | **Applied** | Function now scans the entire batch before deciding. Error wins regardless of ordering. New test `test_get_session_state_prefers_failed_when_idle_and_error_in_same_batch` covers both orderings. |
| 4 | Default model drift | **Applied** | `agent_managed_model` default in `config/settings.py` is now `claude-sonnet-4-6`. Description clarifies it's informational only (actual model lives in the agent preset). |
| 5 | Deprecated `agent_connectors` / `agent_extra_mcp_servers` parsed but never read | **Applied** | `Worker.__init__` calls `_warn_deprecated_settings()` once at startup. Logs a single WARNING if either var is set, pointing at the migration path. |
| 6 | Test name vs behavior mismatch | **Applied** | Renamed to `test_managed_executor_constructed_with_full_credentials`. |
| 7 | Cursor param key as class constant | **Applied** | `ManagedAgentsDriver.EVENT_CURSOR_PARAM = "after"`; `get_session_state` references the constant. |

**Tests:** `pytest -m unit -k 'agent_worker' -q` — **151/151 passed** (2 new tests added).
**Lint:** `ruff check api/services/agent_worker/` — clean.
**Commit:** `faed19a fix: address review round 1 — final_text carry-forward, telegram wording, error ordering, model default, deprecation warning, naming`

Two findings (#7 cascade error reasons, #9 Slack/Asana doc note) were rejected by the referee — see referee decisions comment for reasoning.

PR #115, opened 2026-05-26
fix: managed_driver terminal-state detection — idle + events endpoint + 4xx body

### LINKED ISSUE #114: fix: managed_driver loses terminal state — idle status + events-endpoint
## Problem

Cloud-routed `#agent` tasks run successfully (sessions create, events post, polling works, agents produce real answers, tools execute) but **never reach `#agent-completed`** because two driver bugs prevent terminal-state detection. The agent currently runs cleanly but the worker never sees the terminal signal Anthropic sends.

Discovered during live smoke testing on 2026-05-26 after #113 merged. The agent worker hit `api.anthropic.com/v1/sessions`, created a real session, the agent called LifeOS MCP tools, and went idle — but our code kept polling for 5+ minutes because we don't recognize what "idle" means.

## Context

Filing as one issue because both bugs are symptomatic of the same beta-API gap: the [Managed Agents docs](https://platform.claude.com/docs/en/managed-agents) documented the resource model and event types but not the response shapes of `GET /sessions/{id}` vs `GET /sessions/{id}/events`. We extracted wrong assumptions in #113 and now have the real responses to correct against.

Related work:
- [#113](https://github.com/nbramia/LifeOS/pull/113) — merged the foundational refactor; this fixes the gaps it left.
- [#111](https://github.com/nbramia/LifeOS/issues/111) — self-hosted sandbox (the environment-update tooling from this issue carries over).

### Bug 1: Session status \`\"idle\"\` is terminal but not recognized

Anthropic's session lifecycle is \`running → idle\` (with \`idle\` meaning "agent done, nothing more to do"). Our \`TERMINAL_REMOTE_STATUSES\` in \`api/services/agent_worker/managed_driver.py\` only contains \`{completed, failed, cancelled, budget_exceeded}\`. So when \`GET /sessions/{id}\` returns \`\"status\": \"idle\"\`, our executor's \`poll()\` thinks it's still running and keeps polling forever (continuing to accrue session-hour overhead at \$0.08/hr).

Live evidence from session \`sesn_01E5PbanfhnAfdcBe2VQxA6F\`:
\`\`\`json
{\"id\": \"sesn_…\", \"status\": \"idle\", \"stats\": {\"active_seconds\": 5.422, \"duration_seconds\": 315.8}, \"usage\": {\"input_tokens\": 3, \"output_tokens\": 42}}
\`\`\`
Worker kept polling for 5+ minutes without finalizing.

Fix:
- Add \`\"idle\"\` to \`TERMINAL_REMOTE_STATUSES\` (or map \`\"idle\"\` → \`\"completed\"\` in driver before returning state).
- The synthesize-from-events helper already correctly handles \`session.status_idle\` event type; this is the response-level status field, a separate concept.

### Bug 2: Events live at a separate endpoint, not embedded in the session response

Our \`ManagedAgentsDriver.get_session_state\` calls \`GET /sessions/{id}\` and expects an \`events\` array on the response. The live response has no events array at all — it returns \`{events_count: 0}\` and you have to call \`GET /sessions/{id}/events\` separately to get them.

Live evidence: a session that produced a clean answer had 11 events (1 \`agent.message\`, 1 \`user.message\`, 3 \`session.error\`, plus status/span events) — all only visible via \`GET /sessions/{id}/events\`. The \`/sessions/{id}\` response had no events at all.

Consequence: \`_extract_final_text\` always returns None because no \`agent.message\` events ever reach it. The carry-forward cache fix from #113 review round 1 is therefore inert in production. Telegram completion summaries would be empty even after Bug 1 is fixed.

Fix:
- Add \`ManagedAgentsDriver.list_events(session_id, after_id=None)\` calling \`GET /v1/sessions/{id}/events\` with optional \`?after=\` cursor. Response shape per live evidence: \`{\"data\": [{\"id\", \"type\", \"payload\"|\"content\"|\"error\", \"processed_at\"}, ...], \"first_id\", \"last_id\", \"has_more\"}\`.
- Modify \`get_session_state\` to call both endpoints (status + events) and merge into \`ManagedSessionState\`. Cursor advances against the events endpoint.
- Update test mocks to handle both endpoints. The MockTransport handler can route on path: \`/sessions/{id}\` → status response, \`/sessions/{id}/events\` → events response.

### Bug 3: Driver logs only exception type name on session-create failure

Surfaced during live debugging — when \`/sessions\` returned 400 with a descriptive error message ("MCP server host(s) blocked by environment network policy"), the worker logged only \`HTTPStatusError\` (no body). Operators had no way to see what was wrong without manually curl-ing the API. Per security comment in code: the redaction is defensive against API-key leakage from request headers, but 400s are body/schema problems where the response body is safe to log.

Fix:
- Log the response body for 4xx errors via \`exc.response.text\` (or \`.json()\`). API keys are in REQUEST headers, not RESPONSE bodies; this is safe.

## What's already done (operator-side, not blocking this code work)

These were initially listed as bugs but resolved during diagnosis:

- ✅ **Environment networking**: env defaulted to \`{networking: {type: \"limited\", allow_mcp_servers: false}}\` which blocked every MCP host. Fixed via \`curl -X POST https://api.anthropic.com/v1/environments/$ENV_ID … -d '{\"config\":{\"type\":\"cloud\",\"networking\":{\"type\":\"unrestricted\"}}}'\`. (\`PATCH\` returns 405; \`POST\` works.) Doc this gotcha.
- ✅ **LifeOS Vault credential**: was missing entirely. Added via \`POST /v1/vaults/{id}/credentials\` with \`{auth: {type: \"static_bearer\", mcp_server_url: \"https://nathanramia-linux.tailb8bcda.ts.net/mcp\", token: \"…\"}}\`. Confirmed working — agent successfully called LifeOS MCP tools in subsequent runs (\`agent.mcp_tool_use\` event in stream).

## Remaining operator follow-up (separate from this PR, but blocks full-surface validation)

- **Re-OAuth Gmail and Cal connectors in the Anthropic console.** Live errors: \`gmail: \"MCP server 'gmail' initialize failed: server URL not found\"\` and same for \`gcal\`. The URLs in the Vault match the agent preset byte-for-byte (verified by listing Vault credentials), so this isn't a URL mismatch — it's an Anthropic-side init failure on those specific connectors. The other 6 first-party / third-party MCPs (LifeOS, Drive, Superhuman, Ramp, Slack, Granola, Asana) all initialize cleanly. Likely fix: delete each broken connector in the Vault, re-add it via the OAuth flow.

## Sub-finding to watch (don't fix in this issue)

The smoke-test run that exercised the LifeOS MCP successfully (\`agent.mcp_tool_use\` event) produced 45 output tokens but **no \`agent.message\` text event** before going idle. The agent called the tool, got the result, and idled without producing a final summary. Not blocking, but worth tuning the agent preset's system prompt to require an explicit final-summary turn even after tool use. File a separate issue if it recurs after the driver fixes land.

## Files involved

- \`api/services/agent_worker/managed_driver.py\` — add \`list_events\`, expand \`TERMINAL_REMOTE_STATUSES\`, modify \`get_session_state\` to fan out, enrich 4xx error logging
- \`tests/test_agent_worker_managed_driver.py\` — route MockTransport on path, add tests for idle-terminal + events-endpoint + 4xx body logging
- \`tests/test_agent_worker_managed_executor.py\` — verify the carry-forward test still exercises the cross-batch path with the new shape
- \`docs/guides/agent-worker-setup.md\` — environment networking gotcha, Vault static_bearer credential flow (with the \`mcp_server_url\` byte-match requirement), the curl-POST-to-update-env trick

## Acceptance criteria

- [ ] A cloud-routed task (e.g. \`Using Claude: list MCP tool families. Budget \$0.25\`) lands at \`#agent-completed\` with a non-empty Telegram summary containing the agent's actual reply
- [ ] Session row in \`data/agent_sessions.db\` shows \`status=completed\` and \`total_dollars > 0\`
- [ ] Worker log shows the full lifecycle: create_session → post events → poll observes \`status=idle\` → finalize → tag swap → Telegram
- [ ] On a 4xx from any \`/sessions\` call, the worker log includes the response body so operators can self-diagnose without manual curl
- [ ] Tests cover: idle as terminal, events fetched via separate endpoint, status+events merge, 4xx body logged
- [ ] \`agent-worker-setup.md\` documents the environment-networking gotcha, the Vault \`static_bearer\` credential creation flow with byte-match URL requirement, and the curl-to-update-environment trick
- [ ] Operator-side (out of scope for this PR but required for full-surface validation): re-OAuth Gmail and Cal in the Anthropic console

nbramia:
## Summary

Three driver bugs surfaced during live smoke testing of the #113 cloud path — sessions ran successfully and agents produced real answers, but the worker never detected the terminal signal, so tasks never reached `#agent-completed`. This PR fixes all three.

Closes #114.

## What changed

1. **`"idle"` is the real terminal status** (not `"completed"`). Added to `TERMINAL_REMOTE_STATUSES`; executor `_finalize_remote` treats it as `STATUS_COMPLETED`. `"completed"` kept as a synthesized alias for forward-compat.
2. **Events live at `/sessions/{id}/events`**, not in the session payload. New `ManagedAgentsDriver.list_events(session_id, after_id)`; `get_session_state` now calls both endpoints and merges, with the cursor advancing against the events endpoint. The carry-forward cache from #113 was inert until this fix because no `agent.message` events ever reached `_extract_final_text`.
3. **4xx response bodies are now logged before raising.** New `_raise_for_status_with_body` helper emits the API's validation message at WARNING when a 4xx hits, applied to all driver-side HTTP calls. Operators no longer need to manually `curl` to diagnose beta-API schema mismatches.

Doc updates (`docs/guides/agent-worker-setup.md`):
- URL byte-match requirement between Vault credentials and the agent preset's `mcp_servers.url`
- `static_bearer` credential creation via API (the console UI may not expose this option for custom MCPs)
- Environment networking gotcha: defaults to `"limited"`, must be `"unrestricted"` or every session-create 400s
- The `POST`-as-update trick for environment config (`PATCH` returns 405)

## Test evidence

```
pytest tests/test_agent_worker_managed_driver.py    — 26/26 passed
pytest tests/test_agent_worker_managed_executor.py  — 16/16 passed
pytest -m unit -k 'agent_worker'                    — 155/155 passed
ruff check                                          — clean
```

Driver tests fully refactored: each test now uses a handler that routes on path (status vs events endpoint), via a new `_make_session_handler` helper. New tests cover idle-as-terminal, the dual-endpoint fan-out, the cursor going to `/events`, and the 4xx body-logging behavior.

Live smoke evidence motivating each fix:
- Bug 1: session `sesn_01E5PbanfhnAfdcBe2VQxA6F` sat at `"status": "idle"` for 5+ minutes while our code kept polling
- Bug 2: same session had 11 events at `/events` but `events_count: 0` on the session payload
- Bug 3: a 400 from `/sessions` came back with `"MCP server host(s) blocked by environment network policy"` but the worker log only showed `HTTPStatusError`

## Review focus

- **The new `list_events` + `get_session_state` fan-out.** Two GETs per poll instead of one — verify the cursor logic isn't misaligned (cursor lives on `/events` only).
- **Status precedence** in `get_session_state`: raw `status` from `/sessions/{id}` if it's in `TERMINAL_REMOTE_STATUSES`, else event-synthesized status, else raw fallback. Makes sure a raw `"idle"` doesn't get overridden by an in-batch synthesis result.
- **`_raise_for_status_with_body` body truncation** (2048 chars). Tight enough to avoid log bloat from a misbehaving API; loose enough to capture full validation messages we've seen so far.
- **Tests pass `events: []` on the status response now.** Confirmed live API doesn't include events there, but if a future release does, the merge logic prefers `/events` data — verify that's the right precedence.

## Operator follow-up (out of scope for this PR)

- Re-OAuth Gmail and Cal in the Anthropic console. Both still return `"server URL not found"` on initialize even though Vault URLs match the agent preset byte-for-byte. Likely incomplete OAuth handshake; delete + re-add the connector. All other 7 MCPs (LifeOS, Drive, Superhuman, Ramp, Slack, Granola, Asana) work correctly after this PR.

### DIFF
--- a/api/services/agent_worker/managed_driver.py
+++ b/api/services/agent_worker/managed_driver.py
@@ -215,11 +215,15 @@ class ManagedAgentsDriver:
         if events:
             last_event_id = events[-1].get("id")
 
-        # Synthesize a terminal status from events when the status field hasn't
-        # transitioned yet. `idle` (raw) wins on its own; otherwise we look at
-        # events for status_idle / session.error.
+        # Resolve final status. A `session.error` in the batch must win over
+        # a raw `idle`, otherwise the session reports completed and the
+        # cascading failure is silently lost. `_synthesize_status_from_events`
+        # returns "failed" if ANY error event is present, else "completed"
+        # if any idle event, else None.
         synthesized = _synthesize_status_from_events(events)
-        if raw_status in TERMINAL_REMOTE_STATUSES:
+        if synthesized == "failed":
+            status = "failed"
+        elif raw_status in TERMINAL_REMOTE_STATUSES:
             status = raw_status
         elif synthesized:
             status = synthesized
@@ -244,33 +248,64 @@ class ManagedAgentsDriver:
             error_reason=error_reason,
         )
 
+    # Hard cap on pagination loop in `list_events` — protects against a
+    # misbehaving API or runaway session producing millions of events. 1000
+    # pages × the API's page size is far more than any real session will
+    # ever generate; if we hit this we'd rather drop and warn than hang.
+    _MAX_EVENTS_PAGES = 1000
+
     def list_events(
         self,
         session_id: str,
         after_id: str | None = None,
     ) -> list[dict]:
-        """Fetch new session events since `after_id` (or all if None).
+        """Fetch all new session events since `after_id` (or all if None).
 
         `GET /v1/sessions/{id}/events?after=<cursor>` returns
-        `{"data": [{"id", "type", "payload"|"content"|"error", ...}, ...]}`.
-        Returns the raw event list; callers (e.g. `get_session_state`) handle
-        the cursor advance and synthesis logic.
+        `{"data": [{"id", "type", "payload"|"content"|"error", ...}, ...],
+          "first_id", "last_id", "has_more"}`. Loops while `has_more=True`
+        so the caller never silently loses the tail of a paginated batch —
+        without this, a session that produced more events than the API page
+        size between polls would have its terminal `agent.message` dropped
+        and the Telegram completion summary would be wrong.
+
+        Returns the raw event list (concatenated across pages); callers
+        (e.g. `get_session_state`) handle the cursor advance and synthesis.
         """
-        params: dict[str, str] = {}
-        if after_id:
-            params[self.EVENT_CURSOR_PARAM] = after_id
-        resp = self._client.get(
-            f"{self.base_url}/sessions/{session_id}/events",
-            headers=self._headers(),
-            params=params,
+        all_events: list[dict] = []
+        cursor = after_id
+        for _ in range(self._MAX_EVENTS_PAGES):
+            params: dict[str, str] = {}
+            if cursor:
+                params[self.EVENT_CURSOR_PARAM] = cursor
+            resp = self._client.get(
+                f"{self.base_url}/sessions/{session_id}/events",
+                headers=self._headers(),
+                params=params,
+            )
+            if resp.status_code == 404:
+                # Session deleted mid-poll. Caller's status fetch already
+                # turned this into "cancelled"; return what we have.
+                return all_events
+            self._raise_for_status_with_body(resp)
+            data = resp.json()
+            page = data.get("data", []) or []
+            all_events.extend(page)
+            if not data.get("has_more"):
+                return all_events
+            if not page:
+                # Defensive: has_more=True but empty page would loop forever.
+                logger.warning("list_events: has_more=true with empty page; stopping")
+                return all_events
+            cursor = page[-1].get("id")
+            if not cursor:
+                logger.warning("list_events: has_more=true but last event has no id; stopping")
+                return all_events
+        logger.warning(
+            "list_events: hit pagination cap (%d pages) for session %s",
+            self._MAX_EVENTS_PAGES, session_id,
         )
-        if resp.status_code == 404:
-            # Session deleted mid-poll. Caller's status fetch already turned
-            # this into "cancelled"; return empty events so we don't crash.
-            return []
-        self._raise_for_status_with_body(resp)
-        data = resp.json()
-        return data.get("data", []) or []
+        return all_events
 
     def post_user_message(self, session_id: str, content: str) -> None:
         """Post a user turn to a running session.
@@ -308,10 +343,14 @@ class ManagedAgentsDriver:
         if 400 <= resp.status_code < 500:
             # Truncate to avoid unbounded log volume from a misbehaving API.
             body_preview = (resp.text or "")[:2048]
+            # `resp.request` is set by real transports including MockTransport,
+            # but defensively guard against hand-constructed Response objects.
+            req = getattr(resp, "request", None)
+            method = getattr(req, "method", "?") if req is not None else "?"
+            path = getattr(getattr(req, "url", None), "path", "?") if req is not None else "?"
             logger.warning(
                 "Managed Agents %s %s → %d: %s",
-                resp.request.method, resp.request.url.path,
-                resp.status_code, body_preview,
+                method, path, resp.status_code, body_preview,
             )
         resp.raise_for_status()
 
@@ -320,12 +359,15 @@ class ManagedAgentsDriver:
 
         Best-effort: errors are logged but never raised, because callers use
         this for cleanup paths where re-raising would mask the original cause.
+        4xx response bodies are still surfaced via the standard helper before
+        the exception is swallowed.
         """
         try:
-            self._client.delete(
+            resp = self._client.delete(
                 f"{self.base_url}/sessions/{session_id}",
                 headers=self._headers(),
             )
+            self._raise_for_status_with_body(resp)
         except Exception as exc:
             logger.warning("kill_session %s failed: %s (reason=%s)", session_id, exc, reason)
 
--- a/api/services/agent_worker/managed_driver.py
+++ b/api/services/agent_worker/managed_driver.py
@@ -43,12 +43,13 @@ from api.services.agent_worker.pricing import (
 logger = logging.getLogger(__name__)
 
 
-# Terminal session statuses synthesized by `ManagedAgentsDriver.get_session_state`
-# from the event stream. The Managed Agents docs only document event types
-# explicitly (session.status_idle = success, session.error = failure). We map
-# those to a stable status string here so the executor can compare against a
-# fixed set regardless of how the API evolves.
+# Terminal session statuses returned by `GET /v1/sessions/{id}`. The Managed
+# Agents API reports `"idle"` when the agent has finished and has nothing more
+# to do — that's the canonical success terminal. We keep `"completed"` in the
+# set too as a synthesized alias (from `session.status_idle` events when the
+# session.status field hasn't transitioned yet — observed in live testing).
 TERMINAL_REMOTE_STATUSES = frozenset({
+    "idle",
     "completed",
     "failed",
     "cancelled",
@@ -160,7 +161,7 @@ class ManagedAgentsDriver:
             headers=self._headers(),
             json=body,
         )
-        resp.raise_for_status()
+        self._raise_for_status_with_body(resp)
         data = resp.json()
         sid = data.get("id") or data.get("session_id")
         if not sid:
@@ -176,22 +177,24 @@ class ManagedAgentsDriver:
         session_id: str,
         since_event_id: str | None = None,
     ) -> ManagedSessionState:
-        """Poll a session's current state + any events since `since_event_id`.
+        """Poll a session's current state — status + cumulative usage + new events.
 
-        Aggregates events into a stable status string. Per the Managed Agents
-        event docs, terminal signals come via events:
-          - `session.status_idle` → agent finished; status="completed"
-          - `session.error` → fatal; status="failed"
-        The status field on the response is forwarded directly when it falls
-        in `TERMINAL_REMOTE_STATUSES`; otherwise we synthesize from events.
+        Live API responses (confirmed 2026-05-26) split the data across two endpoints:
+          - `GET /v1/sessions/{id}` returns `{status, usage, ...}` but NOT events.
+            The `status` field reports `"running"` until the agent finishes, then
+            transitions to `"idle"` (Anthropic's terminal-success value).
+          - `GET /v1/sessions/{id}/events` returns `{data: [...], first_id,
+            last_id, has_more}` — the canonical event stream. `agent.message`,
+            `session.error`, `session.status_idle`, etc. all live here.
+
+        This method calls both and merges. `since_event_id` is the cursor against
+        the events endpoint. The status response also carries a 404 → cancelled
+        signal (DELETEd sessions return 404 on subsequent polls).
         """
-        params: dict[str, str] = {}
-        if since_event_id:
-            params[self.EVENT_CURSOR_PARAM] = since_event_id
+        # 1. Session status + usage
         resp = self._client.get(
             f"{self.base_url}/sessions/{session_id}",
             headers=self._headers(),
-            params=params,
         )
         # Treat 404 as "cancelled" — that's how a DELETEd session appears on
         # subsequent polls.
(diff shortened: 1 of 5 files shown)


nbramia (commented):
## PR Review: fix: managed_driver terminal-state detection — idle + events endpoint + 4xx body

### Action Required
- **[Correctness]** `managed_driver.py:222-227` — new status precedence silently swallows `session.error` when raw `status="idle"`. The old code ran `_synthesize_status_from_events` first (line 221), so a batch containing both an idle status AND a `session.error` event was synthesized to `"failed"` — operator gets actionable signal. The new branch short-circuits on `raw_status in TERMINAL_REMOTE_STATUSES`, so raw `idle` + error-event-in-same-batch resolves to `"idle"` → `STATUS_COMPLETED`. Issue #114 explicitly mentions a smoke-test session with 3 `session.error` events that still idled; under this code those would now finalize as success and lose the error reason. Fix: when `raw_status == "idle"` and `synthesized == "failed"`, prefer `"failed"` (mirror the in-batch precedence rule in `_synthesize_status_from_events`).
- **[Correctness]** `managed_driver.py:247-273` — `list_events` ignores `has_more` from the response. The docstring even mentions the field. If a session ever produces more events than the API's page size between two polls, the tail is silently dropped and the cursor jumps past unseen events on the next poll (we advance to `events[-1].id`). For terminal detection the parallel status endpoint provides a fallback signal, but `_extract_final_text` and `_extract_error_reason` may surface stale/incorrect content. Either loop while `has_more=True` advancing the cursor, or document the assumption that one page is always sufficient and add a `logger.warning` if `has_more` is ever observed.

### Recommended
- **[Tests]** `tests/test_agent_worker_managed_driver.py:303-326` — `test_get_session_state_prefers_failed_when_idle_and_error_in_same_batch` only exercises raw `status="running"` (the default in `_make_session_handler`). The new precedence logic at lines 222-227 short-circuits before synthesis when raw status is terminal, so this scenario is now untested for the live API shape. Add a case with `status_body={"status": "idle"}` + `session.error` in events, asserting `status == "failed"`.
- **[Performance]** `get_session_state` now does 2 GETs per poll instead of 1, doubling Anthropic API request volume per active session. Worth confirming the worker poll cadence and concurrent-session ceiling can absorb 2x request rate without hitting Anthropic's per-key rate limits. No code change required if cadence is conservative.
- **[Standards]** `managed_driver.py:318-330` — `kill_session`'s try/except never routes through `_raise_for_status_with_body`, so 4xx bodies from a failed DELETE go unlogged. PR description claims "applied to all driver-side HTTP calls" but DELETE is exempt. Either call the helper or log `resp.text` in the except branch.

### Minor
- **[Defensive]** `_raise_for_status_with_body` at `managed_driver.py:312-314` accesses `resp.request.method` / `resp.request.url.path` directly. `resp.request` can be `None` for hand-constructed `httpx.Response` objects. Real production paths always go through the client (MockTransport included sets `request`), but a `getattr(resp.request, 'method', '?')` guard would be free insurance.
- **[Docs]** `docs/guides/agent-worker-setup.md` env-update curl shows `-d '{"config":{...}}'` which overwrites the entire `config` field. If operators have other config keys set on the environment, this will wipe them. Add a note to GET the current config first and merge.
- **[Standards]** `tests/test_agent_worker_managed_driver.py:425` — `import logging` inside the test function should move to module top. Style only.
- **[Forward-compat]** Status response tested at line 220 has no `events` key now; the merge precedence comment at line 218-220 in the source claims events from `/events` win if a future API release also embeds them in `/sessions/{id}`, but the current code never reads `data.get("events", ...)` from the status response — the comment promise isn't actually enforced. Either drop the comment or read+ignore the embedded events explicitly so a future regression is detectable.

### Summary
Solid, targeted fix backed by live evidence and good test discipline (155/155 unit tests pass, ruff clean, verified locally). The two Action Required items both stem from the new fan-out architecture and need addressing before merge: the `idle`-overrides-`error` precedence regression is the most important, and pagination is a latent correctness gap. Otherwise merge-ready.

nbramia:

| 1 | `idle`-overrides-`error` precedence regression — raw `status="idle"` short-circuits before synthesis, so `idle` + `session.error` in the same batch finalizes as success | Action Required | **Accept** | Verified in `managed_driver.py:222-227`. The whole point of `_synthesize_status_from_events` is to prefer "failed" when both present (the test added in #113 review covers it), but the new precedence block runs `raw in TERMINAL_REMOTE_STATUSES` first and bypasses synthesis. Real regression. Fix: invert order so error-synthesized wins over raw idle. |
| 2 | `list_events` ignores `has_more` — events beyond first page silently lost, cursor jumps past them | Action Required | **Accept** | Cursor would advance to last_event_id of the partial page, then never see the tail. Telegram completion summaries miss the actual final text. Fix: loop while `has_more=True` advancing cursor. |
| 3 | `test_prefers_failed_when_idle_and_error_in_same_batch` only tests raw `running`+events; doesn't cover the new raw=`idle` + events path | Recommended | **Accept** | Directly related to finding 1. Add an explicit test case. |
| 4 | 2x GET request rate may hit rate limits at scale | Recommended | **Reject** | Anthropic's [read-endpoint limit is 600 r/min](https://platform.claude.com/docs/en/managed-agents/overview). Default poll cadence 60s × 10 concurrent sessions × 2 GETs = 20 req/min — under 5% of the limit. Not actionable at any reasonable scale. |
| 5 | `kill_session` doesn't use `_raise_for_status_with_body` — PR description claim "applied to all driver-side HTTP calls" is wrong | Recommended | **Accept** | The except swallows exceptions but a 4xx body still has diagnostic value pre-swallow. Apply the helper inside the try; the broad except still catches the re-raise. |
| 6 | `resp.request` could be None on hand-constructed responses | Minor | **Accept** | `getattr` guard is one line. Defensive insurance. |
| 7 | env-update curl uses `-d '{"config":{...}}'` which overwrites the whole config field | Minor | **Accept** | Real footgun if operator has init_script or env vars set. Add GET-merge-POST note in docs. |
| 8 | `import logging` inside test function | Minor | **Accept** | Move to module top. |
| 9 | Forward-compat comment promises `/events` precedence but code doesn't read embedded events from `/sessions/{id}` | Minor | **Accept (modified)** | Reject adding fallback (live API doesn't put events there; speculative complexity). Just remove the inaccurate comment claim. |

**Findings forwarded to addresser:** 8 (2 Action Required + 2 Recommended + 4 Minor)

nbramia:
## Round 1 — Addresser Summary (inline addressing)

| 1 | idle-overrides-error precedence regression | **Applied** | `get_session_state` precedence inverted: synthesized "failed" wins absolutely (cascading errors no longer lost). New regression test `test_get_session_state_prefers_failed_when_raw_idle_and_event_error`. |
| 2 | `list_events` ignored `has_more` pagination | **Applied** | Loop while `has_more=true` advancing cursor each iteration. Hard cap 1000 pages + defensive empty-page + missing-id guards prevent infinite loops. Two new tests: happy-path pagination, empty-with-has_more guard. |
| 3 | Missing test for raw=idle + error case | **Applied** | Covered by the new test in fix #1. |
| 5 | `kill_session` bypassed `_raise_for_status_with_body` | **Applied** | DELETE now routes through the helper inside the try; broad except still swallows the re-raise. PR description claim now accurate. |
| 6 | `resp.request` could be None on hand-constructed Response | **Applied** | `getattr` guards on `resp.request`, `req.method`, `req.url.path`. |
| 7 | Env-update curl overwrites whole config field | **Applied** | Docs add caveat: GET-merge-POST when other config keys are set. |
| 8 | `import logging` inside test function | **Applied** | Moved to module top. |
| 9 | Misleading forward-compat comment | **Already gone** | The comment lived in the block replaced by the precedence fix in finding #1; no separate edit needed. |

Rejected (per referee decisions):
- #4 (2x GET rate concern) — under 5% of the documented 600 r/min read-endpoint limit at default poll cadence.

**Tests:** `pytest tests/test_agent_worker_managed_driver.py tests/test_agent_worker_managed_executor.py -q` — **45/45 passed** (29 driver + 16 executor).
**Lint:** `ruff check` — clean.
**Commit:** `fa42303 fix: address review round 1 — precedence, pagination, kill_session, defensive guards`

PR #118, opened 2026-05-26
fix: don't fail cloud sessions on MCP init errors + transcript pointer for empty summaries

### LINKED ISSUE #116: fix: don't fail sessions on MCP init errors when agent idled cleanly
## Problem

PR #115 introduced a status precedence where synthesized `\"failed\"` from any `session.error` event wins absolutely over a raw `\"idle\"` status from the session endpoint. In practice this overcorrects: `mcp_authentication_failed_error` and `mcp_connection_failed_error` events fire at **session-init** when an attached MCP isn't reachable, but the agent works around the missing MCP and finishes the task cleanly using the others.

Result: any session with at least one unreachable MCP gets marked `#agent-failed` even when the agent produced output and reached idle. The successful answer is lost to the operator.

## Context

Live evidence — session `sesn_01KxXAcmcXPgyfLEzVpCG9Wn` on 2026-05-26:

```
Status: idle
Usage: input=3, output=89
Events: agent.mcp_tool_use (1), agent.thinking (1), session.error (2), session.status_idle (1)
session.error messages:
  - gmail: MCP server 'gmail' initialize failed: server URL not found
  - gcal: MCP server 'gcal' initialize failed: server URL not found
```

Both errors are `mcp_*_failed_error` types at init time. The agent used a different MCP, did 89 output tokens of work, and the session.status_idle event fired. Anthropic's own `status` field reports `"idle"` — they consider it a successful terminal. Our code marks it failed.

The original review concern (PR #115 finding #1) was that cascading failures shouldn't be lost as silent successes. That's a legitimate concern for runtime errors mid-agent-loop. But init-stage MCP errors aren't cascading failures — they're informational notices about partial tool availability.

## Suggested approach

Distinguish error severity by event timing or error type. Simplest first cut:

- In `_synthesize_status_from_events` (`api/services/agent_worker/managed_driver.py`), only treat `session.error` events as terminal-failed when the error type is NOT one of `mcp_authentication_failed_error`, `mcp_connection_failed_error`. Init-stage MCP failures get logged at transcript level but don't override `session.status_idle`.

A more general rule: if `session.status_idle` is the LAST event in the batch (chronologically after any errors), the session succeeded — errors that came earlier are noise. Without explicit timestamps in events I can't easily implement this; the error-type filter is simpler and matches the observed semantics.

Operator should still see the init failures in the Telegram summary so they know some MCPs were unavailable — extend the completion message to include a "Note: <N> MCPs failed to initialize: <names>" footer when applicable.

## Files involved

- `api/services/agent_worker/managed_driver.py` — `_synthesize_status_from_events` filters MCP init error types
- `api/services/agent_worker/managed_executor.py` or `worker.py` — completion summary footer when init errors present
- `tests/test_agent_worker_managed_driver.py` — new tests covering the filter
- `tests/test_agent_worker_managed_executor.py` or `tests/test_agent_worker_loop.py` — completion summary footer test

## Acceptance criteria

- [ ] A session with `status=idle` + `agent.mcp_tool_use` event + 1+ `session.error` events of type `mcp_authentication_failed_error` lands at `#agent-completed`, not `#agent-failed`
- [ ] A session with `status=idle` + `session.error` of a non-init type still lands at `#agent-failed` (the original cascading-failure case stays caught)
- [ ] When the completed session had init errors, Telegram summary includes a footer listing the affected MCPs (so operator can fix or remove them from the preset)
- [ ] Tests cover both branches

Relates to: #114 (introduced the over-correcting precedence), #113 (initial driver work).

### LINKED ISSUE #117: fix: agent preset system prompt needs explicit final-summary requirement
## Problem

Cloud-routed agent sessions sometimes idle without producing an `agent.message` event — only `agent.thinking` and `agent.mcp_tool_use`. The agent does the work but skips the final text turn. Result: Telegram completion summary is empty even when the session was successful.

## Context

Observed twice in live smoke testing on 2026-05-26:
- `sesn_01A2gVKf8geWG7NiumL4o4Q2`: 45 output tokens, 1 `agent.mcp_tool_use`, 0 `agent.message`
- `sesn_01KxXAcmcXPgyfLEzVpCG9Wn`: 89 output tokens, 1 `agent.thinking`, 1 `agent.mcp_tool_use`, 0 `agent.message`

In both cases the agent invoked a tool successfully but didn't produce a summary text turn before going idle. The current agent preset's system prompt says:

> When you complete a task, reply with a one-paragraph summary of what you did and the key result.

But "reply" isn't being interpreted as "must produce an `agent.message` text event" — the model treats tool use as a complete response sometimes.

## Suggested approach

Tighten the system prompt in the agent preset (`agent_019QrftoZjdF2htVQnXw2Z8c`) to explicitly require a final text turn after any tool use. Suggested wording:

> Every task MUST end with a final assistant turn that contains a one-paragraph text summary. Tool calls alone are not a complete response. After your last tool call, always produce a text turn summarizing what you did and the key result.

This is an Anthropic-console operator action, not a code change. The agent preset YAML lives in the Anthropic console UI under Workspace → Agents → LifeOS Worker.

Optionally on our side: when Telegram summary detects empty `final_text`, fall back to a programmatic note like "Agent completed without a final summary; check transcript at data/agent_transcripts/<sid>.jsonl for tool-use detail." So the operator at least sees that something happened.

## Files involved

- (Operator action) Agent preset system prompt in Anthropic console
- `api/services/agent_worker/worker.py` `_completion_summary` — empty-final-text fallback
- `tests/test_agent_worker_loop.py` — completion-summary test with empty final_text

## Acceptance criteria

- [ ] Agent preset system prompt updated with explicit "must produce a final text turn" requirement
- [ ] When `final_text` is empty, Telegram message includes a transcript pointer instead of an empty body
- [ ] Test covers the empty-final-text path

Relates to: #114, #113.

nbramia:
## Summary

Two product issues surfaced during live smoke testing after #115 merged. Cloud sessions ran successfully but landed at `#agent-failed` because of informational MCP-init errors, and when the agent idled without a final text turn the Telegram summary was empty. This PR fixes both.

Closes #116, closes #117.

### #116 — MCP init errors no longer fail otherwise-successful sessions

PR #115's round-1 fix made any `session.error` in the batch trigger terminal-failed. Live observation: `mcp_authentication_failed_error` / `mcp_connection_failed_error` events fire at session-init for unreachable connectors. The agent works around missing MCPs and finishes using the others — those sessions should land at `#agent-completed`, not `#agent-failed`.

`_synthesize_status_from_events` now splits `session.error` into two classes:
- **Init failures** (`mcp_*_failed_error` types) — informational. Don't trigger failed. Surfaced via a new `init_failed_mcps` list on `ManagedSessionState`.
- **Runtime errors** (anything else, including `session.error` without an explicit type) — still trigger failed. The cascading-failure detection from #115's round-1 review is preserved for the cases that actually need it.

New `_extract_init_failed_mcps` helper returns deduped MCP names.

### #117 — Completion summary handles empty final_text + lists failed MCPs

When the agent uses a tool and idles without producing an `agent.message` (observed twice with Sonnet on tight-budget tasks), the Telegram summary was empty. Now points at the transcript path:

> ✅ Agent worker: completed 'do the thing' (text) — 92 tokens, $0.00, 12s active.
>
> (agent idled without a final text reply — check transcript at data/agent_transcripts/sess_…jsonl for tool-use detail)

When `init_failed_mcps` is non-empty, appends a footer so the operator can fix or remove the broken connectors:

> Note: 2 MCP server(s) unavailable this session: gmail, gcal

`ExecutorOutcome` gains an `init_failed_mcps` field (defaults to empty list — local-executor path unaffected). `ManagedExecutor`'s success branch propagates `state.init_failed_mcps` through.

## Test evidence

```
pytest tests/test_agent_worker_managed_driver.py    — 32/32 passed (3 new)
pytest tests/test_agent_worker_managed_executor.py  — 17/17 passed (1 new)
pytest tests/test_agent_worker_loop.py              — 17/17 passed (3 new)
pytest -m unit -k 'agent_worker'                    — 165/165 passed
ruff check                                          — clean
```

New tests cover:
- Init-only errors with idle → completes (not fails) and populates `init_failed_mcps`
- Non-init `session.error` → still fails (regression guard for the #115 cascading-failure fix)
- Same MCP firing two init errors → name deduped
- Init failures propagate from driver → executor → outcome
- Empty `final_text` → Telegram contains a transcript pointer (not the old `(no text response)` placeholder)
- Non-empty `init_failed_mcps` → footer appended
- Empty `init_failed_mcps` → no footer noise

## Review focus

- **Init-failure type list** in `_MCP_INIT_FAILURE_TYPES` (`managed_driver.py`). Currently `mcp_authentication_failed_error` and `mcp_connection_failed_error`. If the API adds new init-failure variants we'd want to extend this. Verify against the docs.
- **`session.error` without an explicit `type` field still triggers failed.** Defensive default: unknown error types are runtime errors, not init noise. This preserves the original cascading-failure regression guard from #115 round 1.
- **`init_failed_mcps` on `ExecutorOutcome`.** Lives on the shared dataclass but only the managed path populates it. Local executor leaves it empty — confirmed by existing local tests still passing without modification.
- **Telegram footer wording**: `"Note: N MCP server(s) unavailable this session: …"` — flag if a different wording would be clearer for the operator.
- **Transcript pointer format**: `data/agent_transcripts/{session_id}.jsonl` — matches the path convention used elsewhere in the worker. Operator can follow this directly without further translation.

## Companion operator action

Issue #117 also flagged that the agent preset's system prompt should explicitly require a final-summary turn after tool use (so empty `final_text` happens less often). That's a one-line edit in the Anthropic console (Workspace → Agents → LifeOS Worker). This PR handles the code-side fallback when it does happen; the operator-side prompt update is separate.

## Out of scope

- Re-OAuth or re-add of broken Gmail / Cal connectors. The operator removed those from the agent preset (now at version 2) so this isn't blocking — sessions no longer see those init failures regardless.
- Self-hosted sandbox support — still tracked in #111, blocked on this fix landing only by ordering.

### DIFF
--- a/api/services/agent_worker/local_executor.py
+++ b/api/services/agent_worker/local_executor.py
@@ -21,7 +21,7 @@ import json
 import logging
 import time
 import uuid
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from api.services.agent_worker.pricing import cost_for
 from api.services.agent_worker.session_store import (
@@ -106,6 +106,10 @@ class ExecutorOutcome:
     final_text: str = ""
     reason: str = ""
     wake_at: int | None = None   # set when status == STATUS_YIELDED (sleep)
+    # Managed-agents-only: MCP servers that failed to initialize during the
+    # remote session. Worker uses this to append a footer to the completion
+    # summary so the operator knows which connectors are broken.
+    init_failed_mcps: list[str] = field(default_factory=list)
 
 
 def _system_prompt(session_id: str, expected_output: str, budget, parent_session_id: str | None = None) -> str:
--- a/api/services/agent_worker/managed_driver.py
+++ b/api/services/agent_worker/managed_driver.py
@@ -68,6 +68,23 @@ class ManagedSessionState:
     total_output_tokens: int = 0
     final_text: str | None = None
     error_reason: str | None = None
+    # MCP servers that failed to initialize during this session — informational,
+    # not necessarily fatal. The agent may still complete the task using the
+    # MCPs that did succeed. Surfaced to the operator in the completion summary
+    # so they can fix or remove the broken connectors.
+    init_failed_mcps: list[str] = field(default_factory=list)
+
+
+# `session.error.error.type` values for MCP server initialization failures —
+# the connector failed at handshake time (server unreachable, OAuth missing,
+# URL doesn't match a Vault credential, etc.). These don't indicate a failed
+# agent run; the agent works around the missing tool. Filtered out of the
+# terminal-failed synthesis so an idle session with init errors lands at
+# `#agent-completed` instead of `#agent-failed`.
+_MCP_INIT_FAILURE_TYPES = frozenset({
+    "mcp_authentication_failed_error",
+    "mcp_connection_failed_error",
+})
 
 
 class ManagedAgentsDriver:
@@ -215,11 +232,11 @@ class ManagedAgentsDriver:
         if events:
             last_event_id = events[-1].get("id")
 
-        # Resolve final status. A `session.error` in the batch must win over
-        # a raw `idle`, otherwise the session reports completed and the
-        # cascading failure is silently lost. `_synthesize_status_from_events`
-        # returns "failed" if ANY error event is present, else "completed"
-        # if any idle event, else None.
+        # Resolve final status. A non-init `session.error` (runtime / cascade
+        # failure) wins over raw `idle` so cascading failures aren't lost as
+        # silent successes. MCP init failures are excluded from the failed
+        # synthesis — see `_synthesize_status_from_events` and
+        # `_MCP_INIT_FAILURE_TYPES`.
         synthesized = _synthesize_status_from_events(events)
         if synthesized == "failed":
             status = "failed"
@@ -237,6 +254,9 @@ class ManagedAgentsDriver:
         # Surface error message from session.error events.
         error_reason = _extract_error_reason(events)
 
+        # Names of MCPs that failed to initialize — informational footer.
+        init_failed_mcps = _extract_init_failed_mcps(events)
+
         return ManagedSessionState(
             session_id=session_id,
             status=status,
@@ -246,6 +266,7 @@ class ManagedAgentsDriver:
             total_output_tokens=int(usage.get("output_tokens", 0)),
             final_text=final_text,
             error_reason=error_reason,
+            init_failed_mcps=init_failed_mcps,
         )
 
     # Hard cap on pagination loop in `list_events` — protects against a
@@ -379,24 +400,57 @@ class ManagedAgentsDriver:
 def _synthesize_status_from_events(events: list[dict]) -> str | None:
     """Map event-stream signals to a terminal status string, or None.
 
-    Scans the full batch before deciding so we prefer `"failed"` when both
-    `session.error` and `session.status_idle` appear together — error is the
-    actionable signal for operators. Returning `"completed"` on first idle
-    (the prior behavior) silently swallowed cascading errors.
+    `session.error` events are split into two classes:
+      - MCP init failures (`mcp_authentication_failed_error`,
+        `mcp_connection_failed_error`) — informational. The agent works
+        around the missing tool; the session can still complete cleanly.
+        These DO NOT trigger terminal-failed.
+      - Other errors (tool dispatch crashes, lifecycle failures, etc.) —
+        treated as terminal-failed because the agent couldn't recover.
+
+    Precedence: non-init `session.error` (failed) > `session.status_idle`
+    (completed). This preserves the cascading-failure detection while not
+    over-failing sessions whose only error noise is broken connectors.
     """
     has_idle = False
-    has_error = False
+    has_runtime_error = False
     for ev in events:
         et = ev.get("type", "")
         if et == "session.status_idle":
             has_idle = True
         elif et == "session.error":
-            has_error = True
-    if has_error:
+            err_type = (ev.get("error") or {}).get("type") or ev.get("error_type", "")
+            if err_type not in _MCP_INIT_FAILURE_TYPES:
+                has_runtime_error = True
+    if has_runtime_error:
         return "failed"
     if has_idle:
         return "completed"
     return None
+
+
+def _extract_init_failed_mcps(events: list[dict]) -> list[str]:
+    """Return the names of MCP servers that failed to initialize.
+
+    Used to surface the failed connectors in the completion summary so the
+    operator can fix or remove them from the agent preset. Init failures
+    don't fail the session per `_synthesize_status_from_events`, so the
+    operator's only signal is this list in the Telegram message.
+    """
+    names: list[str] = []
+    seen: set[str] = set()
+    for ev in events:
+        if ev.get("type") != "session.error":
+            continue
+        err = ev.get("error") or {}
+        err_type = err.get("type") or ev.get("error_type", "")
+        if err_type not in _MCP_INIT_FAILURE_TYPES:
+            continue
+        name = err.get("mcp_server_name") or ev.get("mcp_server_name") or ""
+        if name and name not in seen:
+            seen.add(name)
+            names.append(name)
+    return names
 
 
 def _extract_final_text(events: list[dict]) -> str | None:
--- a/api/services/agent_worker/managed_executor.py
+++ b/api/services/agent_worker/managed_executor.py
@@ -252,9 +252,17 @@ class ManagedExecutor:
             # poll().
             final_text = state.final_text or self.session_store.get_managed_final_text(session.task_id) or ""
             self.session_store.update_status(session.task_id, STATUS_COMPLETED)
-            self.transcript_store.append(session.session_id, "managed_completed",
-                                         {"final_chars": len(final_text), "remote_status": state.status})
-            return ExecutorOutcome(status=STATUS_COMPLETED, final_text=final_text)
+            self.transcript_store.append(
+                session.session_id, "managed_completed",
+                {"final_chars": len(final_text),
+                 "remote_status": state.status,
+                 "init_failed_mcps": list(state.init_failed_mcps)},
+            )
+            return ExecutorOutcome(
+                status=STATUS_COMPLETED,
+                final_text=final_text,
+                init_failed_mcps=list(state.init_failed_mcps),
+            )
 
         if state.status == "budget_exceeded":
             self.session_store.update_status(session.task_id, STATUS_BUDGET_EXCEEDED)
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -943,13 +943,35 @@ class Worker:
         active_s = int(refreshed.total_active_seconds or 0)
         expected = refreshed.expected_output or "text"
         title = task.get("description", session.task_id)
-        result_blurb = (outcome.final_text or "(no text response)").strip()
-        if len(result_blurb) > 600:
-            result_blurb = result_blurb[:600] + "…"
+
+        # Body: prefer the agent's final text. When it's empty (the agent
+        # used a tool and idled without summarizing — sometimes happens with
+        # Sonnet on tight-budget tasks), surface a transcript pointer so the
+        # operator can inspect what the agent actually did instead of seeing
+        # a blank message. The transcript captures every tool call.
+        final_text = (outcome.final_text or "").strip()
+        if final_text:
+            result_blurb = final_text
+            if len(result_blurb) > 600:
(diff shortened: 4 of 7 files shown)


nbramia (commented):
## PR Review: fix: don't fail cloud sessions on MCP init errors + transcript pointer for empty summaries

Verified locally: branch checked out, 66/66 in the three touched test files pass, 165/165 `agent_worker` unit tests pass, `ruff check` clean. Behavior changes match the stated intent of issues #116 and #117.

### Action Required

- **[Correctness]** `init_failed_mcps` is computed only from events in the **current poll batch** and is not persisted across polls. `_extract_init_failed_mcps` (`api/services/agent_worker/managed_driver.py:432`) runs against `events` returned from `list_events(..., after_id=since_event_id)`. MCP init errors fire at session-start (first poll). If the session takes more than one poll to reach `idle` (the common case — wall-time alone makes this near-certain), the cursor has advanced past those events by the time `_finalize_remote` runs in `managed_executor.py:240`. `state.init_failed_mcps` will be `[]` and the Telegram footer will be silently dropped. Contrast with `final_text`, which IS persisted across polls via `session_store.set_managed_final_text` (`managed_executor.py:209-210`) and read back as a fallback in `_finalize_remote` (`managed_executor.py:253`). The existing tests don't catch this because they put all events in one synthetic batch. Fix: persist `init_failed_mcps` on the session row the same way `final_text` is cached (set on each poll where the list is non-empty; read back during finalize).

### Recommended

- **[Correctness]** `_extract_error_reason` (`managed_driver.py:479`) reads `payload.message` / `data.error` but does NOT read `error.message`. The new tests added in this PR use the live-API shape `session.error.error.{type, message, mcp_server_name}` (see `tests/test_agent_worker_managed_driver.py:295-301`, where the existing `payload` shape is tested, and `tests/test_agent_worker_managed_driver.py:338-340` where a `tool_dispatch_error` with `error.message="the agent's bash tool crashed"` is used). For the new event shape, runtime failures (`tool_dispatch_error`, lifecycle errors, etc.) will reach the failed branch in `_finalize_remote` (`managed_executor.py:280-283`) with `state.error_reason = None`, so the operator's failure Telegram will read `"remote status=failed"` instead of the actual error message. Two-line fix in `_extract_error_reason`: also check `ev.get("error", {}).get("message")`. The existing test on line 291-301 keeps the old `payload` shape working — add a sibling test for the new `error.message` shape.

- **[Correctness]** Failed-path doesn't surface `init_failed_mcps` at all. On the failed branch in `_finalize_remote` (`managed_executor.py:279-283`), the outcome carries only `reason` — `init_failed_mcps` is dropped. If a session fails at runtime AND had MCP init issues, the operator's failure Telegram won't list the broken connectors. Not a regression (this path didn't exist before), but the same operator signal would help debugging here too. Either propagate to `STATUS_FAILED` outcomes, or document the deliberate omission.

### Minor

- **[Standards]** The init-failure type allowlist `_MCP_INIT_FAILURE_TYPES` (`managed_driver.py:84`) is conservative — only two types. The PR description flags this as a known fragility ("If the API adds new init-failure variants we'd want to extend this"). Consider matching on a prefix like `mcp_*_failed_error` via `err_type.startswith("mcp_") and err_type.endswith("_failed_error")`, which would auto-cover plausible future variants (`mcp_timeout_failed_error`, `mcp_authorization_failed_error`, etc.) without code changes. Tradeoff: a typo or unexpected `mcp_*_failed_error` would silently get classified as init-noise rather than failed. Current explicit allowlist is safer; flag if you want belt-and-suspenders.

- **[Correctness]** `_extract_init_failed_mcps` skips events where `name` is empty/missing (`managed_driver.py:449-452`). If the API emits a `mcp_connection_failed_error` without `mcp_server_name`, the operator sees no signal at all (no count, no footer). Probably fine in practice — the live evidence in #116 always includes the name — but worth either including a `<unknown>` placeholder or logging a warning to catch API drift.

- **[Test Coverage]** `test_completion_summary_uses_transcript_pointer_when_final_text_empty` (`tests/test_agent_worker_loop.py:334`) only asserts `"data/agent_transcripts/" in msg` and `".jsonl" in msg`. It doesn't verify the session_id in the path matches the actual session. Strengthening to `assert f"data/agent_transcripts/{session.session_id}.jsonl" in msg` would catch a future regression where the wrong field is substituted.

- **[Style]** The fallback message `"(agent idled without a final text reply — check transcript at ...)"` (`worker.py:958-961`) uses a single-line f-string broken across two lines. Functional, but slightly awkward — consider one of: (a) use parentheses + line continuation cleanly, (b) build the message as a single string literal. Cosmetic.

### Summary

Merge-ready after the **Action Required** persistence fix for `init_failed_mcps`. Without it, the issue #116 footer mechanism will work only for sessions that complete on the very first poll — a narrow corner of real usage. The error-reason shape mismatch (Recommended #1) is also worth fixing in the same round since it surfaces with the same new event shape this PR is responding to. Everything else is polish.

nbramia:

| 1 | `init_failed_mcps` computed per-batch, never persisted — drops when session takes >1 poll to reach idle | Action Required | **Accept** | Same class of bug as the #115 `final_text` carry-forward. MCP init errors fire at session-start (poll #1); by the time finalize runs on a later poll, cursor has advanced past those events. Footer silently disappears for any session that doesn't complete in one poll — the common real-world case. Fix: persist `init_failed_mcps` analogously to `final_text` in `managed_cursor`. |
| 2 | `_extract_error_reason` reads `payload.message` and `data.error` but not `error.message` — live API uses the latter shape | Recommended | **Accept** | Verified at `managed_driver.py:479-487`. Live `session.error` event shape is `error: {type, message, mcp_server_name}`. Non-init failures now reach `_finalize_remote` with `state.error_reason=None`, so operator's failure Telegram only shows `"remote status=failed"`. Two-line fix. |
| 3 | Failed-path doesn't propagate `init_failed_mcps` — failure Telegram won't list broken connectors | Recommended | **Accept** | Symmetric with the success path. If a session failed at runtime AND had init failures, operator should see both signals. Cheap to add. |
| 4 | `_MCP_INIT_FAILURE_TYPES` allowlist is conservative — consider prefix matching | Minor | **Accept (no code change)** | Keeping the explicit allowlist is correct: a hypothetical `mcp_runtime_failed_error` would be misclassified by prefix matching. Adding a code comment noting the rationale + how to extend. |
| 5 | `_extract_init_failed_mcps` drops events with empty `mcp_server_name` silently | Minor | **Accept** | Use a `<unknown>` placeholder so an API-drift scenario still surfaces the failure count in the footer. |
| 6 | Test only asserts substring `data/agent_transcripts/` — should assert full session_id path | Minor | **Accept** | Tightens the test to catch field-substitution regressions. |
| 7 | Awkward two-line f-string concatenation in worker.py transcript-pointer message | Minor | **Accept** | Cosmetic cleanup. |

**Findings forwarded to addresser:** 7 (1 Action Required + 2 Recommended + 4 Minor)

nbramia:
## Round 1 — Addresser Summary (inline)

| 1 | `init_failed_mcps` not persisted across polls — footer drops when session takes >1 poll to finalize | **Applied** | New `managed_cursor.init_failed_mcps_json` column with idempotent ALTER TABLE migration. `SessionStore.merge_managed_init_failed_mcps` (union semantics) called on every poll; `_finalize_remote` reads persisted + current-batch to build the full set. New regression test `test_init_failed_mcps_persist_across_poll_batches` simulates the two-poll scenario and verifies the outcome carries both broken MCPs. |
| 2 | `_extract_error_reason` didn't read `error.message` shape | **Applied** | Iterates three known shapes (`error`, `payload`, `data`) in order of likelihood for the current API. Existing tests cover the new path. |
| 3 | Failed-path didn't propagate `init_failed_mcps` | **Applied** | Both `cancelled` and `failed` terminal branches now carry the persisted set. Symmetric with the completed branch. |
| 4 | `_MCP_INIT_FAILURE_TYPES` allowlist conservativeness | **Applied (no code change)** | Kept the explicit allowlist (prefix matching would misclassify hypothetical `mcp_runtime_failed_error`); added code comment with rationale + docs link. |
| 5 | Empty `mcp_server_name` silently dropped | **Applied** | Falls back to `<unknown>` placeholder so the operator at least sees the count. |
| 6 | Empty-text test assertion too loose | **Applied** | Now asserts the full `data/agent_transcripts/{session_id}.jsonl` path. |
| 7 | Awkward f-string concatenation | **Applied** | Path extracted to a local variable. |

**Tests:** `pytest tests/test_agent_worker_managed_driver.py tests/test_agent_worker_managed_executor.py tests/test_agent_worker_loop.py tests/test_agent_worker_stores.py -q` — **85/85 passed** (1 new regression test).
**Lint:** `ruff check` — clean.
**Commit:** `23470a7 fix: address review round 1 — persist init_failed_mcps, error.message shape, failed-path propagation`

Issue #11, opened 2026-02-28
Audit and clarify PersonEntityStore.save() semantics


nbramia:
## Problem

\`PersonEntityStore.save()\` at \`api/services/person_entity.py:754-762\` has a docstring saying it is a no-op because operations commit immediately. But many scripts still call \`store.save()\` after operations, creating confusion about whether operations are batched or auto-committed.

\`\`\`python
def save(self):
    \"\"\"Save is a no-op for SQLite store - operations commit immediately.
    Kept for API compatibility.\"\"\"
    pass
\`\`\`

## Proposed Direction

1. Verify \`save()\` is truly a no-op by auditing all write methods (\`add()\` at line 764, \`update()\`, \`delete()\` at line 826) — confirm each uses its own connection and commits
2. Find all call sites of \`store.save()\` across the codebase
3. Either:
   - (a) Add \`warnings.warn("save() is deprecated — operations commit immediately", DeprecationWarning)\` and plan removal, or
   - (b) Give \`save()\` real batch semantics: accumulate writes, commit on \`save()\`, rollback on context exit

Option (a) is simpler and matches current behavior.

## Acceptance Criteria

- [ ] All write operations confirmed to commit immediately (or batch semantics implemented)
- [ ] \`save()\` documented as deprecated or given clear batch semantics
- [ ] All callers updated (removed if deprecated, or kept if batch semantics added)

## Key Files

- \`api/services/person_entity.py:754-762\` — \`save()\` method
- \`api/services/person_entity.py:764-810\` — \`add()\` method
- \`api/services/person_entity.py:826-859\` — \`delete()\` method
- Search for \`\.save()\` calls in \`scripts/\` and \`api/\` directories

## Dependencies

None.

PR #217, opened 2026-05-28
refactor(llm): retire Ollama coexistence + add VRAM watchdog

nbramia:
## Summary

Eliminates the dual-LLM-stack pattern that caused the 2026-05-28 nightly-sync freeze, and adds a VRAM watchdog so the next time something similar happens it's caught before the GPU locks up.

The 2026-05-28 incident: amdgpu kernel logs show VRAM exhaustion at 03:46 (`Not enough memory for command submission!`) as the slack sync was running. Root cause: llama-server (LifeOS's orchestrator, 16 GB) + Ollama (Hermes' cheap-model tier, 16 GB) were both holding a copy of the same Gemma 4 26B in VRAM. Neither runtime knew the other existed, the sync's embedding load pushed past the 64 GB BIOS allocation, and the kernel locked up until the machine was force-shut. PR #214 fixed the upstream BM25/summarizer/stats issues but nothing actually prevented this specific VRAM contention.

Two commits, both motivated by the same incident:

1. **`refactor(llm): migrate routing + fact validation off Ollama to llama-server`**
   - `api/services/query_router.py`, `api/services/person_facts.py` (two call sites), and `scripts/sync_entity_cleanup.py` no longer instantiate `OllamaClient`. They now use new module-level helpers in `api/services/llm_client.py` (`generate_text`, `generate_json`, `extract_json`, `is_local_routing_llm_available`, `ais_local_routing_llm_available`) which go through `LocalLLMClient` against `settings.local_llm_url` (llama-server).
   - `api/services/ollama_client.py` deleted — nothing imports it anymore.
   - The routing helpers are pinned to local llama-server even when `LIFEOS_LLM_BACKEND=anthropic`, because query routing and dedup were never sent to the cloud and shouldn't start now.
   - Tests rewritten/updated: `TestOllamaClient` removed; `TestRouterIntegration` retargeted at the local LLM; `test_person_facts.py` mock patches switched from `api.services.ollama_client.OllamaClient` to `api.services.llm_client.{generate_json,ais_local_routing_llm_available}`.
   - Hermes was repointed at Anthropic Haiku separately (config-only edit to `~/.hermes/config.yaml`, not in this PR). With LifeOS migrated, Ollama is now uncalled by any LifeOS path — the package can stay disabled, or operators who actually use Hermes can re-enable it on a per-need basis.

2. **`feat(observability): GPU VRAM watchdog with Telegram alerts`**
   - New `lifeos-gpu-watchdog.timer` runs `scripts/gpu-watchdog.sh` every 5 minutes.
   - Reads VRAM from AMDGPU sysfs (`/sys/class/drm/card*/device/mem_info_vram_{used,total}`), alerts Telegram when usage > `LIFEOS_VRAM_ALERT_PCT` (default 80%), posts a recovery message when it drops below.
   - Rate-limited: `LIFEOS_VRAM_ALERT_COOLDOWN_MIN` (default 60) prevents spam during steady-state saturation.
   - Alert lists top per-PID VRAM consumers when `rocm-smi` is available, falls back to top-RSS process listing otherwise.
   - macOS-unsupported; silently exits when AMDGPU sysfs isn't present.

## Test evidence

`bash scripts/test.sh unit` — 2332 passed, 10 failed, 63 skipped. The 10 failures are pre-existing (env-dependent: 2 wezterm pane injection tests need `WEZTERM_PANE` removed from env; 2 test_settings tests assume defaults the real `.env` overrides; 6 test_crm_api/test_mcp_server need production data). None caused by this PR.

New + updated test coverage:
- `tests/test_llm_client.py`: `TestExtractJson` (6 tests for the shared JSON extraction helper), `TestRoutingHelpers` (4 tests for `generate_text` / `generate_json` / `_get_local_routing_client` / error swallowing).
- `tests/test_query_router.py`: full mock-layer rewrite (4 router tests + 1 integration test re-targeted at llama-server).
- `tests/test_person_facts.py`: 12 mock blocks updated; all 64 tests still pass.
- `tests/test_people_e2e.py`: routing mock updated.

Smoke-test of `scripts/gpu-watchdog.sh`:
- Healthy path: silent (no log entries) when VRAM under threshold.
- Triggered path (forced with `LIFEOS_VRAM_ALERT_PCT=1`): writes `VRAM SATURATED: 34% (22.0/64.0 GB)` to `logs/gpu-watchdog.log`.

## Review focus

- **`api/services/llm_client.py:618-770`** — new helpers. `_get_local_routing_client` returns a `LocalLLMClient` even when the global backend is Anthropic. `extract_json` handles raw / fenced / embedded / nested-braces JSON; covered by `TestExtractJson`. `generate_text`'s per-call `timeout` constructs a transient client to avoid stomping the singleton's timeout.
- **`api/services/query_router.py:373-394`** — the LLM-failure catch went from `except OllamaError` to bare `except Exception`. Worth a careful read — the previous narrow catch could have let httpx errors bubble; the new one always falls back to keyword routing.
- **`scripts/gpu-watchdog.sh`** — bash-only, no external deps beyond curl + standard coreutils. The threshold check uses integer math (`USED * 100 / TOTAL`) which is fine for terabyte VRAM but assumes 64-bit shell arithmetic.
- **`config/systemd/lifeos-gpu-watchdog.{service,timer}`** — new units wired into `setup-systemd.sh` alongside the existing watchdogs.

Closes the VRAM-coordination follow-up from issue #199 (no specific issue number — discussed in the PR conversation after #214 merged).


nbramia (commented):
## PR Review: refactor(llm): retire Ollama coexistence + add VRAM watchdog

### Action Required

- **[Correctness]** PR description claim "Ollama is now uncalled by any LifeOS path" is not accurate. `scripts/run_all_syncs.py` (lines 83–138) still calls `_ollama_unload_all()` and `_ollama_warm()` via Ollama's `/api/ps` and `/api/generate?keep_alive=...` endpoints during the embedding-phase coordination. If a fresh installation has Ollama uninstalled (per the PR's stated intent), `_ollama_loaded_models()` will silently return `[]` and `_stop_llm_for_embeddings()` will return `False` — but `_is_llm_running()` still checks VRAM and may return True if llama-server is pinned, leaving no path to free VRAM for embeddings. Two options: (a) extend `_stop_llm_for_embeddings` to stop `lifeos-llm.service` for llama-server-only setups, or (b) explicitly document that the embedding-phase VRAM coordination is now a no-op on Ollama-less machines and the sync relies on the embedding service's CPU fallback. Either way, this PR's incident scope (the 2026-05-28 VRAM lockup happened *during* sync embedding load) is not fully closed by the changes here.

### Recommended

- **[Correctness]** `scripts/gpu-watchdog.sh:101-104` — `LAST=$(cat "$ALERT_STAMP_FILE")` followed by `(( NOW - LAST < COOLDOWN_MIN * 60 ))` will trip `set -e` if the stamp file is empty or contains a non-numeric value (e.g. truncated write, manual edit). Guard with `LAST=$(cat "$ALERT_STAMP_FILE" 2>/dev/null || echo 0); [[ "$LAST" =~ ^[0-9]+$ ]] || LAST=0`.

- **[Correctness]** `scripts/gpu-watchdog.sh:51-54` — Telegram message body is sent unencoded as a `-d` form field. If the message contains characters that look like form-data separators (`&`, `=`), curl's URL-encoding is *not* applied here (no `--data-urlencode`). The current static prefix is safe, but the `CONSUMERS` block interpolates `ps`/`rocm-smi` output verbatim — a process command containing `&` would silently truncate the message at Telegram's side. Use `--data-urlencode "text=${message}"` instead of `-d "text=${message}"`.

- **[Correctness]** `scripts/gpu-watchdog.sh:23 + 110` — `set -euo pipefail` is on, and `echo "$NOW" > "$ALERT_STAMP_FILE"` is followed by additional commands. If `send_telegram` were ever to fail (it cannot because of `|| true`, but if removed in future), the script would already have written the stamp file, so the next 5-min run would suppress the retry. The stamp should be written *after* a successful Telegram POST, not before. Minor robustness, but recommended given this is the only signal of an impending GPU freeze.

- **[Standards / Cleanup]** `config/settings.py:329-332` retains `ollama_host`, `ollama_model`, `ollama_timeout`, `ollama_retry_timeout` fields. With `ollama_client.py` deleted, the *names* are now misleading: `api/services/summarizer.py:128,204` uses `settings.ollama_timeout` as the timeout for calls to `settings.local_llm_url` (llama-server). This works but is confusing. Either rename the settings (with `OLLAMA_TIMEOUT` env alias retained for back-compat) or add a comment in `settings.py` explaining the legacy name is kept because `run_all_syncs.py` still needs `ollama_host`.

- **[Standards / Stale references]** `api/services/service_health.py:23, 61` still registers `"ollama"` as a tracked service with severity WARNING and description "Local LLM for query routing". With the migration, the routing path now hits llama-server — the registered service should be renamed (or a new one added) so health dashboards reflect reality. Same applies to `tests/conftest.py:31, 110` (`requires_ollama` marker + `ollama_available` fixture pointing at `:11434`) and `docs/specs/standards/testing-standards.md:78, 94`.

- **[Standards / Docs]** Documentation has not been updated to reflect the migration:
  - `docs/guides/installation.md:18, 59-67, 239` still instructs users to install and run Ollama as a hard prerequisite.
  - `docs/guides/setup.md:213-224` and `docs/guides/troubleshooting.md:153-168` still reference Ollama for query routing.
  - `docs/adr/006-ollama-query-routing.md` is now stale — per `AGENTS.md` § "ADRs are immutable", a new ADR superseding it (e.g. ADR-NNN: "Consolidate routing/validation onto llama-server") should be added, with a Status: Superseded note added to ADR-006.

- **[Correctness]** `scripts/gpu-watchdog.sh:119` — `rocm-smi --showpids` parsing assumes `$4` is VRAM bytes; rocm-smi's `--showpids` output format has changed across versions (`PID GPU(s) VRAM(MB) ...` in some, different columns in others). On a mismatch, you'll print misleading "0.0 GB" lines. Consider checking the header row or piping through `rocm-smi --showpids --json` if the installed version supports it.

- **[Standards]** `api/services/llm_client.py:685` — `import re as _re` inside `extract_json` is unnecessary (`re` is not imported at module top). Better to add `import re` at the top alongside `import json` and use `re.search` directly; this matches the project style of avoiding lazy in-function imports for stdlib modules.

### Minor

- **[Correctness]** `api/services/query_router.py:378` — The `except Exception` broadening is called out in the PR description as intentional. That's fine, but `except Exception` will also swallow `KeyboardInterrupt`'s siblings like `SystemExit` if anything bubbles up via `await` (asyncio.CancelledError is not a subclass of Exception in 3.8+, so we're safe there). Consider `except (httpx.HTTPError, ValueError, asyncio.TimeoutError, json.JSONDecodeError):` for tighter scope — `Exception` is broad enough to hide real bugs (e.g. an attribute error in `_llm_route`'s parsing).

- **[Standards]** `tests/test_query_router.py:14-19` — `test_router_initialization` is now `assert router is not None`, which is trivially true after construction. The old assertion (`assert router.ollama_client is not None`) at least verified composition. Either delete the test (it adds no signal) or strengthen it (e.g. assert that the router's `_keyword_fallback` and `route` are callable).

- **[Performance]** `api/services/llm_client.py:730` — `generate_text` constructs a *new* `LocalLLMClient(timeout=timeout)` on every call with a per-call timeout. That builds new `httpx.Client`/`httpx.AsyncClient` instances on each call (via the `async_client` property), bypassing connection reuse. For the routing path this is fine (few calls), but `_validate_and_dedup_ollama` and `_semantic_dedup_ollama` in `person_facts.py` (lines 1139, 1338) pass `timeout=30`/`timeout=15` and get fresh httpx clients every time. Consider caching transient clients by timeout, or accepting that fact-extraction is not latency-critical.

- **[Standards]** `tests/test_llm_client.py:917-967` — `TestRoutingHelpers` mutates module-level state (`mod._routing_client = None`) inside tests without restoring it after the test runs. If a later test in the same session expects the singleton to be set, this leaks. Use a pytest fixture with `yield` that resets at teardown, or `monkeypatch.setattr`.

- **[Standards]** `scripts/gpu-watchdog.sh` uses non-ASCII chars in `printf` format strings (✅, ⚠️, •, —). Most Linux locales handle this fine, but a minimal LANG=C container would corrupt them. Not blocking — just worth noting since the rest of the codebase tends to keep scripts ASCII-clean.

- **[Standards]** `config/systemd/lifeos-gpu-watchdog.service:3` declares `After=lifeos-llm.service` — but the timer can fire even when lifeos-llm is masked/disabled (the common case per the PR description). `After=` only orders startup *if both are starting in the same transaction*; for a oneshot triggered by a timer it's effectively a no-op. Either drop it or add `Wants=lifeos-llm.service` if you actually want to require it.

### Summary

The LLM-client refactor itself is clean and well-tested (126 tests pass on the touched files; mocks are correctly aimed at the per-call import sites in `person_facts.py`). The watchdog script is solid for the steady-state case. **The blocker is the PR description's claim that Ollama is fully retired** — `scripts/run_all_syncs.py` still depends on Ollama's API for embedding-phase VRAM coordination, which means a fresh install per the migration intent will leave the embedding sync's VRAM-freeing path silently disabled. Recommend addressing the run_all_syncs.py path (or scoping the PR description accordingly) before merge, and rolling the doc/settings/health-service-name cleanups into the same PR or a fast follow-up.

nbramia:

| 1 | `run_all_syncs._stop_llm_for_embeddings` only knows how to stop Ollama, not llama-server | Action Required | **Accept** | Reviewer is right. With Ollama disabled, this is a no-op, and llama-server still holds VRAM when the embedding phase starts — exactly the contention pattern this PR is supposed to fix. Will extend to also stop lifeos-llm.service. |
| 2 | gpu-watchdog stamp file empty/non-numeric trips `set -e` | Recommended | **Accept** | Regex guard. |
| 3 | gpu-watchdog `-d "text=..."` no URL-encoding | Recommended | **Accept** | Switch to `--data-urlencode`. |
| 4 | gpu-watchdog writes stamp before POST | Recommended | **Accept** | Move stamp write to after `curl` succeeds. |
| 5 | `settings.ollama_timeout` is misleading after rename | Recommended | **Downgrade → Minor** | Renaming env vars breaks every user's `.env`. Add a clarifying comment instead. |
| 6 | service_health "ollama" key + conftest `requires_ollama` marker stale | Recommended | **Defer (file follow-up issue)** | Real cleanup but scope creep here. |
| 7 | Docs (installation.md / setup.md / troubleshooting.md / ADR-006) still mandate Ollama | Recommended | **Defer (file follow-up issue)** | Documentation refresh deserves its own PR with the ADR-006 supersession. |
| 8 | rocm-smi column-format fragility | Recommended | **Reject** | Already best-effort — the alert still fires regardless of the consumer list. Robustifying would require parsing JSON output which adds complexity for marginal value. |
| 9 | `import re as _re` inside `extract_json` | Recommended | **Accept** | Trivial — move to module top. |
| 10 | broad `except Exception` in query_router | Minor | **Reject** | Intentional and documented — LocalLLMClient raises a wider variety of errors than OllamaError, and the fallback path is safe (deterministic keyword routing). Tighter scoping would silently fail to fall back. |
| 11 | `assert router is not None` is trivially true | Minor | **Accept** | Strengthen the assertion. |
| 12 | `generate_text(timeout=...)` builds a fresh `LocalLLMClient` | Minor | **Reject** | Trade-off is intentional — mutating the singleton's timeout would corrupt concurrent default-timeout calls. Per-call timeouts are rare (one caller). |
| 13 | TestRoutingHelpers mutates `mod._routing_client` without teardown | Minor | **Accept** | Add a fixture. |
| 14 | Unicode in printf may corrupt under LANG=C | Minor | **Reject** | The script's intended runtime is the user's systemd timer in their UTF-8 locale, not a generic-C-locale CI environment. Acceptable risk. |
| 15 | `After=lifeos-llm.service` no-op when lifeos-llm is masked | Minor | n/a | Reviewer confirmed informational. |

**Findings forwarded to addresser:** 7 (Action Required #1; Accept Recommended #2/3/4/9; Accept Minor #11/13). Plus 1 downgraded doc-only change (#5) and 2 follow-up issues to file (#6/7).

nbramia:
## Review Round 1 — Addressed

Pushed \`2a2b6cc\`. Summary:

| 1 | \`_stop_llm_for_embeddings\` only knew how to stop Ollama | **Applied** | New \`_stop_lifeos_llm_service()\` / \`_start_lifeos_llm_service()\` use the passwordless sudo allowlist installed by \`setup-systemd.sh\` to stop/start \`lifeos-llm.service\`. \`_stop_llm_for_embeddings\` now handles both runtimes; either succeeding marks the sync as having freed VRAM. Both paths log on failure rather than aborting the sync. \`_is_llm_running\`'s docstring updated to reflect that we now actively manage both stacks. |
| 2 | gpu-watchdog stamp file empty/non-numeric trips \`set -e\` | **Applied** | Regex guard on \`LAST\` before the arithmetic. |
| 3 | gpu-watchdog \`-d "text=..."\` not URL-encoded | **Applied** | All three POST fields use \`--data-urlencode\`. |
| 4 | gpu-watchdog stamp written before POST | **Applied** | Stamp moved to after a successful Telegram POST. Also added \`curl -f\` so HTTP 4xx surfaces as failure (was silently "succeeding" with bad creds — caught when my smoke test accidentally hit the real \`.env\`). \`ENV_FILE\` is now overridable for testing. |
| 9 | \`import re as _re\` in \`extract_json\` | **Applied** | Moved to module-top imports. |
| 11 | Trivial \`assert router is not None\` | **Applied** | Now asserts \`router\` doesn't expose \`ollama_client\` — catches regression if the attribute creeps back. |
| 13 | \`TestRoutingHelpers\` mutates \`mod._routing_client\` without teardown | **Applied** | New \`_routing_singleton_reset\` fixture used by all 4 tests in the class. |
| 5 | \`settings.ollama_timeout\` misleading | **Applied (downgrade → doc-only)** | Clarifying comment explaining the env-var aliases are kept for operator \`.env\` compatibility; only the two \`*_timeout\` fields are still consumed. |
| 6 | service_health/conftest stale Ollama references | **Deferred** | Filed as #219. |
| 7 | docs + ADR-006 refresh | **Deferred** | Filed as #220. |
| 8 | \`rocm-smi --showpids\` column-format fragility | **Rejected** | Best-effort hint already; the alert fires regardless. |
| 10 | broad \`except Exception\` in query_router | **Rejected** | Intentional and documented — LocalLLMClient raises a wider variety of errors than OllamaError and the fallback is safe. |
| 12 | \`generate_text(timeout=...)\` builds a fresh \`LocalLLMClient\` | **Rejected** | Trade-off is intentional — mutating the singleton's timeout would corrupt concurrent default-timeout calls. |
| 14 | Unicode under LANG=C | **Rejected** | Acceptable risk in the intended runtime. |

**Tests:** \`pytest tests/test_llm_client.py tests/test_query_router.py tests/test_person_facts.py tests/test_people_e2e.py tests/test_run_all_syncs.py -m "not slow and not integration"\` → 152 passed.

**Pre-push hook:** 1582 unit tests passed.

**Watchdog smoke tests:**
- Stamp NOT written when Telegram POST fails (verified with \`-f\` curl + bad token).
- Stamp NOT written when no \`.env\` present (already correct).
- Healthy path stays silent (no log spam).
- Cooldown path correctly suppresses re-alert when stamp is fresh.

**Commit:** \`2a2b6cc\` \`fix: address review round 1 — extend LLM stop to llama-server + watchdog hardening\`

PR #225, opened 2026-05-28
fix(sync): phone source_entities on Linux + tiered backup retention

nbramia:
## Summary

Two related sync-health gaps caught by the new `/sync-health` checkup (PR #223).

### 1. Phone `source_entity` drift (issue #199 §2 pattern)

`scripts/apple_data_import.import_phone_calls()` was creating `Interaction` rows for resolved-phone calls but never touching `source_entities`. The macOS-native `scripts/sync_phone_calls.py` does both — the Linux side silently fell behind from 2026-03-03 onward (date the user last ran the sync directly on the Mac).

`/sync-health`'s drift detector reported a 86-day gap between `MAX(created_at)` for phone `interactions` (today) and phone `source_entities` (March 3). Backfilling against today's export with the new code created **9 new source_entities + 38 updates**; drift detector now returns `NO_DRIFT`.

Approach: mirror the macOS pattern — ensure a phone-keyed `source_entity` (`phone_{e164}`) for every unique number observed in the export, resolve / create the PersonEntity with `create_if_missing=True` so first-time callers get a Person row, then link the entity. The source_entity creation runs **before** the "interaction already exists?" check so re-importing the same JSON still adds source_entities when the Person store has grown — important for the catch-up scenario.

### 2. Tiered backup retention

`InteractionStore.create_backup` was pruning to a hard `last 2`. Replaced with `_prune_backups`, a four-tier policy:

| Age | Slots kept |
|---|---|
| 0–4 days | all (5 daily slots) |
| 5–34 days | 1 per week-of-age |
| 35–365 days | 1 per month-of-age |
| 366+ days | 1 per quarter-of-age |

Net steady-state: **~25 backups × ~300 MB each ≈ 7–8 GB**. Lets you roll back hours / days / weeks / months without ballooning storage. The bucket math parses the timestamp from the filename rather than trusting mtime (so a `cp -p` won't break retention), ignores files that don't match the canonical `interactions.db.YYYYMMDD_HHMMSS.backup` shape, and keeps the newest file in each bucket on tie.

## Test evidence

- 5 new unit tests in `tests/test_interaction_store.py::TestBackupRetention` cover: the 5-daily floor, weekly bucket transitions, monthly tier kicking in past day 35, filename-shape guard, and the under-5-backups no-op case.
- Pre-push hook: `1595 passed in 163s`.
- Backfill verification: phone drift cleared (9 created / 38 updated), drift detector returns `NO_DRIFT`.
- Dry-run prune against the current `data/backups/` (2 backups): nothing would be deleted — both land in the daily tier.

## Review focus

- **`scripts/apple_data_import.py:285-410`** — the order of operations matters. The source_entity ensuring happens **before** the existing-interaction check; without that, the catch-up backfill is a no-op (which is what I hit on the first run before fixing it).
- **`api/services/interaction_store.py:216-300`** — `_prune_backups` bucket math. The boundaries are tunable via the module-level constants near the top.
- **`api/services/interaction_store.py:1182`** — `create_backup` now delegates to `_prune_backups`; the function shrank by ~10 lines.

### DIFF
--- a/api/services/interaction_store.py
+++ b/api/services/interaction_store.py
@@ -227,16 +227,34 @@ _WEEK_DAYS = 7
 _MONTH_DAYS = 30
 _QUARTER_DAYS = 90
 
-_BACKUP_FILE_RE = re.compile(r"^interactions\.db\.(\d{8})_(\d{6})\.backup$")
+# Shared between the writer (``create_backup``) and the pruner so any
+# future change to the naming scheme — e.g., adding microseconds or
+# switching to ISO-8601 — automatically updates both sides instead of
+# silently desyncing them (which would make every new backup invisible to
+# retention math and let the directory grow unbounded).
+_BACKUP_TIMESTAMP_FORMAT = "%Y%m%d_%H%M%S"
+_BACKUP_FILENAME_TEMPLATE = "interactions.db.{ts}.backup"
+_BACKUP_FILE_GLOB = "interactions.db.*.backup"
+_BACKUP_FILE_RE = re.compile(r"^interactions\.db\.(\d{8}_\d{6})\.backup$")
+
+
+def _backup_filename(ts: datetime) -> str:
+    """Canonical backup filename for a given timestamp."""
+    return _BACKUP_FILENAME_TEMPLATE.format(ts=ts.strftime(_BACKUP_TIMESTAMP_FORMAT))
 
 
 def _parse_backup_timestamp(name: str) -> Optional[datetime]:
-    """Parse the timestamp embedded in a backup filename, or None on mismatch."""
+    """Parse the timestamp embedded in a backup filename, or None on mismatch.
+
+    Files with non-canonical names (e.g., ``interactions.db.pre-upgrade.backup``)
+    are intentionally returned as None so ``_prune_backups`` leaves them
+    alone — operators sometimes drop those in for manual safekeeping.
+    """
     m = _BACKUP_FILE_RE.match(name)
     if not m:
         return None
     try:
-        return datetime.strptime(f"{m.group(1)}_{m.group(2)}", "%Y%m%d_%H%M%S")
+        return datetime.strptime(m.group(1), _BACKUP_TIMESTAMP_FORMAT)
     except ValueError:
         return None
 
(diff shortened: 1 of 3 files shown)


nbramia:
## /code-review (medium) — addressed

8 findings surfaced across the 7 finder angles (line-by-line / removed-behavior / cross-file / reuse / simplification / efficiency / altitude). Three were structural enough to defer as follow-up issues; the rest are fixed in \`7758715\`.

### Fixed in this PR

| # | Finding | Fix |
|---|---|---|
| 1 | \`resolve(phone=..., create_if_missing=True)\` doesn't actually create a Person from phone alone — resolver falls through to None at line 881. Docstring overpromised. | Switched to \`create_if_missing=False\` and rewrote the docstring to be honest: only resolvable callers get source_entities. Spam stays in \`unresolved\`. |
| 2 | \`observed_at\` clobbered with current-call timestamp could ratchet backwards on non-chronological exports | Precompute \`max(timestamp)\` across all same-phone calls in the batch; only update if it improves the existing value |
| 3 | \`link_to_person\` would silently overwrite manual links with 0.95 AUTO link | Guard added: skip relink when \`existing_se.link_status == "manual"\` |
| 4 | Backup retention writer/parser drift: filename and regex defined independently 1000 lines apart | Shared \`_backup_filename\` / \`_BACKUP_FILE_RE\` / \`_BACKUP_FILE_GLOB\` / \`_BACKUP_TIMESTAMP_FORMAT\` so any future schema change updates both sides |
| 5 | Aggregator drops \`source_entities_updated\` | Documented the SYNC_STATS schema gap inline; importer log line still reports both counts |
| 7 | Docstring claimed "mirrors iMessage" but iMessage actually defers to a downstream script | Closer analog is \`sync_phone_calls.py\` on macOS; docstring corrected |
| (perf) | \`now\` recomputed inside loop | Hoisted to function entry |

### Filed as follow-ups (not in this PR)

- #226 — resolver phone-only \`create_if_missing\` branch (the deeper architectural fix for #1)
- #227 — generalize backup retention into a \`backup_retention\` module so future stores reuse it
- #228 — add \`create_phone_source_entity\` factory alongside the other source factories

**Tests:** \`pytest tests/test_interaction_store.py -k Backup\` → 5 passed. Pre-push: 1595 passed.

**Live verification:** \`import_phone_calls\` smoke run shows 0 created / 47 updated against today's export — correct behaviour now that backfill is complete.

PR #229, opened 2026-05-28
fix(slack): incremental_sync now refreshes users (closes #224)

### LINKED ISSUE #224: fix(slack): incremental sync never refreshes user list, causing source_entity drift
## Problem

The nightly Slack sync calls \`SlackSync.incremental_sync()\` (default mode), which only calls \`sync_messages()\` — it does **not** call \`sync_users()\`. As a result, any new Slack users added to a workspace after the most recent manual \`full_sync\` never get a \`source_entity\` row.

Surfaced by the \`detect_silent_source_entity_drift()\` helper from PR #214: messages from new users still flow into \`interactions\` (because \`link_imessage\`-style retro-linking eventually finds them), but \`source_entities WHERE source_type='slack'\` stops growing.

In Nathan's case the last \`full_sync\` ran 2026-03-06, so drift accumulated for 83 days before the new \`/sync-health\` skill caught it. Running \`sync.sync_users()\` directly today created 20 new entities and updated 91 — meaning ~20 new humans had appeared in the workspace over the quarter without being represented.

## Where

- \`api/services/slack_sync.py:536\` — \`incremental_sync()\` skips \`sync_users()\`
- \`api/services/slack_sync.py:466\` — \`full_sync()\` does include it

## Suggested fix

\`incremental_sync\` should always run \`sync_users()\` first. The user list is small (Nathan's workspace: 750 total, 111 active humans; a \`users.list\` call returns in ~1 s) and it's an \`add_or_update\` so it's idempotent.

\`\`\`python
def incremental_sync(self, create_interactions: bool = True) -> dict:
    logger.info("Starting incremental Slack sync")
    results = {"users": self.sync_users()}
    results["messages"] = self.sync_messages(
        full=False,
        dm_only=True,
        create_interactions=create_interactions,
        linked_only=True,
    )
    ...
\`\`\`

If the \`users.list\` rate limit is a concern at scale, alternative is to run \`sync_users\` once per week (Sunday) inside \`incremental_sync\`.

## Acceptance

- \`incremental_sync()\` calls \`sync_users()\` (or refreshes weekly)
- Test confirms a new workspace user created between two nightly runs lands in \`source_entities\` on the second run
- After the change, run \`/sync-health\` — \`slack\` should no longer appear in the drift section

## Related

- Discovered via \`/sync-health\` (PR #223)
- Drift detector was added in PR #214 (issue #199 §2)

nbramia:
## Summary

Fixes silent slack source_entity drift discovered by the new \`/sync-health\` skill.

The nightly Slack sync runs \`SlackSync.incremental_sync()\`, which used to call only \`sync_messages()\` — never \`sync_users()\`. As a result, any user added to the workspace after the most recent manual \`full_sync\` was invisible to entity resolution: messages from that user kept flowing into \`interactions\` via retro-linking, but \`source_entities WHERE source_type='slack'\` silently stopped growing.

Nathan's workspace drifted for 83 days (last full_sync 2026-03-06) before \`/sync-health\` caught it this morning. Running \`sync_users()\` directly created 20 new entities and updated 91 — 20 humans had appeared in the workspace over the quarter without being represented.

Closes #224.

## What changed

- \`incremental_sync()\` now calls \`sync_users()\` first, then \`sync_messages()\`. Cheap (~1s for thousands of users) and idempotent (\`add_or_update\`).
- Wrapped in the same try/except resilience pattern as \`full_sync()\`: a flaky \`users.list\` records the error in \`results["errors"]\` and downgrades status to \`partial\` instead of taking down the whole nightly.
- **Return shape now matches \`full_sync\`:** \`{users, messages, status, errors, elapsed_seconds}\`. This was silently broken before — \`scripts/sync_slack.py:61-66\` already reads \`results["users"]\` and \`results["messages"]\` to log stats and emit the canonical \`SYNC_STATS\` line, but for incremental runs those keys didn't exist so the dashboard showed zeros even on a clean sync.
- Internal accessors for \`message_counts\` and \`affected_person_ids\` updated to reach into \`results["messages"]\` to match the new shape.
- Dropped a pre-existing F841 dead variable (\`latest\` at line 379) that the precommit hook flagged once slack_sync.py was in the touched set.

## Test evidence

\`pytest tests/test_slack_sync.py tests/test_slack_integration.py tests/test_slack_api.py -m "not slow and not integration"\` → **49 passed**.

4 new tests in \`tests/test_slack_sync.py::TestIncrementalSyncCallsSyncUsers\`:
- \`sync_users\` is invoked before \`sync_messages\`
- return shape matches full_sync (has \`users\` + \`messages\` + \`status\` + \`errors\`)
- user-sync exception doesn't abort message-sync
- \`message_counts\` still flows through to \`_store_message_counts\` after the shape change

## Review focus

- **\`api/services/slack_sync.py:536-610\`** — the rewritten \`incremental_sync\`. Note the accessor change at lines 602/606 — \`results["messages"].get("message_counts")\` instead of \`results.get("message_counts")\`. If any other internal caller was relying on the flat shape, it will need the same update.
- **Performance** — daily \`users.list\` adds ~1s per night. Marginal but worth noting.
- **\`scripts/sync_slack.py\`** — no change needed; it already expected this shape (the bug was that \`incremental_sync\` wasn't producing it).

### DIFF
--- a/api/services/slack_sync.py
+++ b/api/services/slack_sync.py
@@ -376,7 +376,6 @@ class SlackSync:
         for date_key, day_messages in messages_by_date.items():
             # Use earliest message timestamp for the interaction
             earliest = min(day_messages, key=lambda m: m.timestamp)
-            latest = max(day_messages, key=lambda m: m.timestamp)
 
             # Build snippet from first message with content
             snippet = None
@@ -537,36 +536,76 @@ class SlackSync:
         """
         Perform an incremental sync of new Slack data.
 
-        Only fetches messages newer than the last indexed timestamp.
-        Only processes DMs for users linked to CRM people (same as full_sync).
+        Step 1: Refresh the workspace user list to SourceEntity (issue #224).
+            Without this, users added between the last manual ``full_sync``
+            and tonight stay invisible to entity resolution — messages flow
+            into ``interactions`` for retro-matched senders while
+            ``source_entities WHERE source_type='slack'`` silently stops
+            growing. ``users.list`` is cheap (~1s for thousands of users)
+            and ``add_or_update`` is idempotent.
+        Step 2: Sync new DM messages (only for users linked to CRM people).
 
         Args:
             create_interactions: If True, create CRM Interaction records
 
         Returns:
-            Sync statistics
+            Dict with the same shape as ``full_sync``:
+            ``{users, messages, status, errors, elapsed_seconds}``.
         """
         logger.info("Starting incremental Slack sync")
+        start_time = time.time()
 
-        results = self.sync_messages(
-            full=False,
-            dm_only=True,
-            create_interactions=create_interactions,
-            linked_only=True,
-        )
+        results: dict = {
+            "users": {},
+            "messages": {},
+            "status": "success",
+            "errors": [],
+        }
+
+        # Step 1: Sync users first (needed for entity resolution).
+        # Resilience matches ``full_sync``: a flaky ``users.list`` shouldn't
+        # take down the whole nightly.
+        try:
+            results["users"] = self.sync_users()
+        except Exception as e:
+            error_msg = f"User sync failed: {e}"
+            logger.error(error_msg)
+            results["errors"].append(error_msg)
+
+        # Step 2: Sync new messages.
+        try:
+            results["messages"] = self.sync_messages(
+                full=False,
+                dm_only=True,
+                create_interactions=create_interactions,
+                linked_only=True,
+            )
+            if results["messages"].get("errors"):
+                results["errors"].extend(results["messages"]["errors"])
+        except Exception as e:
+            error_msg = f"Message sync failed: {e}"
+            logger.error(error_msg)
+            results["errors"].append(error_msg)
+
+        if results["errors"]:
+            results["status"] = "partial"
+
+        elapsed = time.time() - start_time
+        results["elapsed_seconds"] = round(elapsed, 1)
 
         # Store actual Slack message counts on PersonEntity (incremental = accumulate)
-        message_counts = results.get("message_counts", {})
+        message_counts = results["messages"].get("message_counts", {})
         if message_counts:
             self._store_message_counts(message_counts, full_sync=False)
 
         # Refresh PersonEntity stats for all affected people
-        affected_ids = results.get("affected_person_ids", set())
+        affected_ids = results["messages"].get("affected_person_ids", set())
         if affected_ids:
             from api.services.person_stats import refresh_person_stats
             logger.info(f"Refreshing stats for {len(affected_ids)} affected people...")
             refresh_person_stats(list(affected_ids))
 
+        logger.info(f"Incremental Slack sync complete in {elapsed:.1f}s")
         return results
 
 
--- a/scripts/sync_slack.py
+++ b/scripts/sync_slack.py
@@ -61,9 +61,13 @@ def run_slack_sync(full: bool = False, dry_run: bool = True) -> dict:
     if "users" in results and results["users"]:
         users = results["users"]
         logger.info("Users:")
-        logger.info(f"  Synced: {users.get('synced', 0)}")
+        # ``sync_slack_users`` returns ``total`` (count of fetched users),
+        # not ``synced`` — the historical 'Synced:' key always logged 0.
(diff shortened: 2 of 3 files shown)


nbramia (commented):
## PR Review: fix(slack): incremental_sync now refreshes users (closes #224)

### Recommended
- **[Requirements]** Issue #224's acceptance criterion #2 says "Test confirms a new workspace user created between two nightly runs lands in `source_entities` on the second run." The four new tests verify call ordering, return shape, resilience, and `_store_message_counts` plumbing — but `sync_users` is mocked, so none of them prove a previously-unknown Slack user actually lands in `source_entities` end-to-end. An integration-style test that uses a real `SourceEntityStore` (in-memory SQLite) and a `MagicMock` `client.list_users` returning a "new" user not present in the store, then asserts a row appears after `incremental_sync()`, would satisfy the spirit of the acceptance criterion and protect against future regressions in the `sync_users` → `sync_slack_users` → `entity_store` plumbing. The current tests would all still pass if someone refactored `sync_users` to a no-op.

### Minor
- **[Correctness]** `scripts/sync_slack.py:64` reads `users.get('synced', 0)`, but `sync_users()` returns `{total, created, updated, skipped_bots, skipped_deleted}` — there's no `synced` key. That log line has always reported 0 and will continue to. The PR's own SYNC_STATS emit at `scripts/sync_slack.py:90` correctly uses `created` and `updated`, so this doesn't break the canonical metrics line, just the human-readable summary above it. Out of scope for #224 but trivially fixable while in the area.
- **[Standards]** The F841 cleanup of `latest` at the old line 379 is mentioned in the PR description as a precommit-driven byproduct. The diff confirms it was truly unused (no other references in `_create_interactions_for_channel`). Fine.

### Summary
Merge-ready. The fix is correct, narrow, and well-motivated; the shape-alignment with `full_sync` is the right call and unblocks the SYNC_STATS dashboard for nightly runs. The one substantive gap is that the test suite verifies the call graph but not the end-to-end behavior the issue's acceptance criterion describes — strongly suggest a follow-up test rather than blocking on it, since the call-ordering test does catch the specific regression that motivated the issue.

nbramia:

| 1 | New tests verify call ordering but \`sync_users\` is fully mocked — issue #224 AC #2 wants end-to-end "new user lands in source_entities" coverage | Recommended | **Accept** | The issue's AC #2 is explicit about this. \`tests/test_slack_integration.py\` doesn't currently cover \`sync_slack_users\` end-to-end either, so adding one here is the right place. Will add a test using a real in-memory \`SourceEntityStore\` + mock SlackClient. |
| 2 | \`scripts/sync_slack.py:64\` reads \`users.get('synced', 0)\` but \`sync_slack_users\` returns \`total\` — verified the function returns \`{total, created, updated, skipped_bots, skipped_deleted}\`, no \`synced\` key | Minor | **Accept** | Pre-existing bug in the touched area. Dashboard's "Synced:" line always reports 0. \`SYNC_STATS\` line uses \`created\` so headline metrics are fine, but the human-readable log line is misleading. One-line fix. |
| 3 | F841 cleanup of \`latest\` is genuinely dead | (informational) | n/a | Reviewer confirmed. |

**Findings forwarded to addresser:** 2 (Accept Recommended #1, Accept Minor #2).

nbramia:
## Review Round 1 — Addressed

Pushed \`63579ca\`. Summary:

| 1 | E2E test for "new workspace user → source_entities" (AC #2) | **Applied** | New ``test_new_workspace_user_persists_to_source_entities`` uses a real ``SourceEntityStore`` on tmp_path, mocks ``SlackClient.list_users`` to return one unseen user, runs ``incremental_sync``, asserts the SourceEntity row exists in the store afterward. Verifies the regression this PR exists to prevent, not just the call ordering. |
| 2 | scripts/sync_slack.py 'Synced:' line always logs 0 | **Applied** | Switched from ``users.get('synced')`` to ``users.get('total')`` and added ``skipped_bots`` / ``skipped_deleted`` for completeness. SYNC_STATS unaffected — it was already reading ``created``/``updated`` correctly. |

**Tests:** ``pytest tests/test_slack_sync.py`` → 5 passed (all 4 previous + the new E2E).

**Pre-push hook:** 1595 unit tests passed.

PR #230, opened 2026-05-28
fix(phone): unlinked source_entity for unresolved phones (closes #226)

### LINKED ISSUE #226: policy: phone-only never creates a Person; store unlinked source_entity instead
## Decision

Phone is the highest-spam, lowest-signal identifier. A raw phone observation with no name or email tells you very little about who the caller is — and auto-creating a PersonEntity per observation would pollute the People graph with one row per spam call.

The architectural rule for LifeOS going forward:

> **\`EntityResolver.resolve\` never creates a PersonEntity from a phone number alone.** Email and name can still create (those have less spam exposure); phone-only resolution either finds an existing match or returns None.

## What replaces the original "add a phone-create branch" proposal

When an importer observes a call from a phone we can't link to any existing Person, **store an unlinked source_entity** (\`canonical_person_id=NULL\`) instead of dropping the observation entirely. This gives us:

- A forensic record of which phone numbers we've ever observed
- Retroactive linking for free — \`scripts/link_source_entities.py:160\` already calls \`resolver.resolve(name=..., email=..., phone=..., create_if_missing=False)\` for every orphan source_entity in the nightly pass, so an unknown-today number that turns out to be Aunt Susan once she's added to Contacts gets linked automatically on the next sync
- No PersonEntity pollution — entity_cleanup never has to scrub junk phone-only Persons
- No spam vs. legit decisions — every unknown number is treated the same way

The interaction itself (the call record) still gets dropped, because the \`interactions\` schema requires \`person_id NOT NULL\`. Adding nullable-\`person_id\` interactions is a bigger schema change worth its own design discussion if we ever decide it's needed.

## Scope of the implementation

1. **Codify the rule** — docstring on \`EntityResolver.resolve\` documenting that phone-only never creates, with rationale and pointer to this issue.
2. **\`apple_data_import.import_phone_calls\`** — when \`resolver.resolve(phone=..., create_if_missing=False)\` returns None, create an unlinked source_entity instead of falling through to \`unresolved\`.
3. **Test** — an integration-style test using a real \`SourceEntityStore\` on tmp_path confirms an unknown caller produces an unlinked source_entity but no interaction.

\`scripts/sync_phone_calls.py\` (macOS-native) and the WhatsApp / Signal importers can adopt the same pattern in follow-up PRs once this lands — out of scope for this fix.

## Acceptance

- \`EntityResolver.resolve\` docstring states the rule (no behaviour change to the resolver itself — it already returns None for phone-only)
- \`apple_data_import.import_phone_calls\` creates an unlinked source_entity for every previously-unseen unresolved phone
- New test confirms: unknown caller → 1 source_entity row with \`canonical_person_id IS NULL\`, 0 interactions
- After landing, \`/sync-health\` drift detector continues to return \`NO_DRIFT\` for phone

## Related

- PR #225 (where the gap was discovered and worked around)
- Existing retro-linking path: \`scripts/link_source_entities.py\`

nbramia:
## Summary

Codifies the rule **\"phone alone never creates a PersonEntity\"** and preserves the observation as an unlinked source_entity so retro-linking can pick it up later when the matching Contact / email arrives.

This is the agreed-upon resolution to #226 after the PR #225 review caught the original \"add a phone-create branch\" proposal as a spam-risk footgun.

Closes #226.

## The rule

\`EntityResolver.resolve(phone=..., create_if_missing=True)\` returns None when nothing matches. Email and name can create-if-missing; phone alone never does. Documented in the resolver's docstring with a pointer back to this issue.

## What importers do now

For every observed phone in \`apple_data_import.import_phone_calls\`:

1. Try to resolve to an existing PersonEntity by phone.
2. **Always** ensure a phone-keyed source_entity exists (\`phone_{e164}\`) — even if no Person matched.
   - If a Person matched: link the source_entity to that person (unless a manual link already exists).
   - If no Person matched: source_entity stays unlinked (\`canonical_person_id IS NULL\`).
3. If no Person, no Interaction is created — \`interactions\` schema requires \`person_id\`.

\`scripts/link_source_entities.py:160\` already calls \`resolver.resolve(name=..., email=..., phone=..., create_if_missing=False)\` for every orphan source_entity on its nightly pass, so retro-linking comes for free. The first time an unknown caller becomes known (via Contacts sync, an email exchange, anything that creates a Person with the same number), the orphan source_entity links itself automatically.

## Live verification

Backfill against today's export against my own data:

| | Total phone SEs | Unlinked phone SEs |
|---|---|---|
| Before | 440 | 204 |
| After | **501** (+61) | **265** (+61) |

All 61 new entries are unlinked — they're observations of callers I'd never had a Contact entry for. Previously they were dropped entirely. Drift detector still reports \`NO_DRIFT\`.

## Test evidence

- \`tests/test_apple_pipeline.py::TestPhoneImport\` rewritten — the existing tests were already broken (mocked \`resolve_by_phone\` instead of \`resolve\` after the PR #225 refactor, hit MagicMock defaults and produced false greens). Now uses a real \`SourceEntityStore\` on tmp_path:
  - \`test_resolved_call_imported_and_linked\` — known caller → linked source_entity + Interaction
  - \`test_unresolved_call_stores_orphan_source_entity\` — unknown caller → unlinked source_entity, zero Interactions
- \`pytest tests/test_apple_pipeline.py tests/test_entity_resolver.py tests/test_interaction_store.py tests/test_sync_health.py\` → **160 passed**.

## Review focus

- **\`api/services/entity_resolver.py:777-803\`** — the docstring change. Codifies the rule, but doesn't change runtime behaviour (the function already returned None for phone-only).
- **\`scripts/apple_data_import.py:348-432\`** — the loop reorganization. Specifically: source_entity creation moved BEFORE the \`if person is None: continue\` check; the link_to_person call gained an \`if person is not None\` guard.
- **\`scripts/sync_phone_calls.py\` and \`api/services/whatsapp.py\` are NOT in this PR** — they have the same drop-on-unresolved pattern and would benefit from the same fix, but each is its own importer with its own quirks (sync_phone_calls also has Address Book name access). Worth a follow-up.

### DIFF
--- a/api/services/entity_resolver.py
+++ b/api/services/entity_resolver.py
@@ -782,14 +782,23 @@ class EntityResolver:
         2. Phone exact match (if phone provided, E.164 format)
         3. Name exact match
         4. Fuzzy name match with context boost
-        5. Create new entity (if create_if_missing)
+        5. Create new entity from email or name (if create_if_missing)
+
+        Phone-only never creates a PersonEntity — see issue #226. A raw phone
+        observation with no name or email tells you very little about who the
+        caller is, and auto-creating a person per observation would pollute
+        the People graph with one row per spam call. Importers handle the
+        "unknown caller" case by storing an unlinked source_entity instead;
+        ``scripts/link_source_entities.py`` then retro-links it whenever the
+        same number later shows up associated with a known Contact / email.
 
         Args:
             name: Person's name
             email: Person's email
             phone: Person's phone (E.164 format, e.g., +1XXXXXXXXXX)
             context_path: Vault path for context boost
-            create_if_missing: Create new entity if not found
+            create_if_missing: Create new entity (from email or name) if not
+                found. Has no effect when only ``phone`` is provided.
 
         Returns:
             ResolutionResult or None
--- a/api/services/entity_resolver.py
+++ b/api/services/entity_resolver.py
@@ -787,10 +787,13 @@ class EntityResolver:
         Phone-only never creates a PersonEntity — see issue #226. A raw phone
         observation with no name or email tells you very little about who the
         caller is, and auto-creating a person per observation would pollute
-        the People graph with one row per spam call. Importers handle the
-        "unknown caller" case by storing an unlinked source_entity instead;
-        ``scripts/link_source_entities.py`` then retro-links it whenever the
-        same number later shows up associated with a known Contact / email.
+        the People graph with one row per spam call. ``apple_data_import``
+        handles the "unknown caller" case by storing an unlinked source_entity
+        instead so ``scripts/link_source_entities.py`` can retro-link it when
+        the same number later shows up linked to a Contact or email.
+        ``scripts/sync_phone_calls.py`` and ``api/services/whatsapp.py`` still
+        drop unresolved observations today and should adopt the same pattern
+        in a follow-up.
 
         Args:
             name: Person's name
--- a/scripts/apple_data_import.py
+++ b/scripts/apple_data_import.py
@@ -261,19 +261,20 @@ def _extract_phone_from_title(title: str) -> str | None:
 def import_phone_calls(dry_run: bool = False) -> dict:
     """Import phone call history from the Mac Mini's JSON export.
 
-    Mirrors the source_entity behaviour of ``scripts/sync_phone_calls.py``
-    (the macOS-native version). For each call whose phone number resolves
-    to an existing PersonEntity, we ensure a phone-keyed SourceEntity
-    exists (``phone_{e164}``), update its ``observed_at`` to the most
-    recent call we've seen for that number in this batch, link it to the
-    person if it isn't already, then create the Interaction.
+    For every observed phone number we ensure a phone-keyed SourceEntity
+    exists (``phone_{e164}``) with its ``observed_at`` set to the most-
+    recent call we've seen for that number in this batch:
 
-    Scope: only resolvable callers get source_entities. Callers we've never
-    associated with a Person (spam, robocalls, first-time inbound) stay in
-    ``unresolved``. The entity-resolver has no phone-only create-if-missing
-    branch (see ``api/services/entity_resolver.py:881``) and adding one
-    here would mint a junk PersonEntity per spam call. Filed as a separate
-    follow-up.
+    - If the number resolves to an existing PersonEntity, link the
+      source_entity to that person (unless a manual link already exists)
+      and create the Interaction record for the call.
+    - If the number doesn't resolve to any Person, the source_entity is
+      stored **unlinked** (``canonical_person_id IS NULL``). This is
+      issue #226's policy: phone alone never creates a Person — auto-
+      creating one per spam call would pollute the People graph — but we
+      keep the forensic observation so ``scripts/link_source_entities.py``
+      can retro-link it later when the matching Contact / email arrives.
+      No Interaction is created (interactions require ``person_id``).
 
     The source_entity update happens BEFORE the "interaction already
     exists?" check so that re-imports still close the issue #199 §2 drift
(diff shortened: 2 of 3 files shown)


nbramia (commented):
## PR Review: fix(phone): unlinked source_entity for unresolved phones

### Recommended
- **[Performance]** `scripts/apple_data_import.py:362-378` introduces an `O(N^2)` scan over the call list. For each call in the outer loop, when the phone first appears, the code does `for c in calls:` and runs `_extract_phone_from_title(c.get("title", ""))` on every call to compute `this_phone_max_ts`. That inner regex parse runs `len(calls) * unique_phones` times. For a typical nightly export of a few thousand calls across hundreds of numbers, that's millions of regex calls per run. Recommend a single pre-pass before the main loop to build a `dict[phone, max_ts]`, then the per-phone block reduces to one dict lookup. The live verification (61 new rows) was small enough not to surface this, but a multi-month backfill or a chatty number would.

### Minor
- **[Docs]** `api/services/entity_resolver.py:790-795` — the new docstring states "Importers handle the 'unknown caller' case by storing an unlinked source_entity instead." That is only true for `apple_data_import.import_phone_calls` after this PR. `scripts/sync_phone_calls.py` (line 233) and `api/services/whatsapp.py` (lines 199, 396, 455) still call `resolver.resolve(phone=...)` and drop on None. Either soften the wording ("importers should handle...") or file the follow-up issue and link it from the docstring so readers don't think the policy is uniformly implemented.
- **[Correctness]** `scripts/apple_data_import.py:402` keeps the early-return semantics by using `unresolved += 1; continue` after the source_entity write. Note that `unresolved` is now an overloaded counter: pre-PR it meant "no Person matched, observation dropped"; post-PR it means "no Person matched, observation captured as orphan SE." Any dashboard/alert that uses `unresolved` to mean "data loss" is now reporting the wrong thing. Worth a grep of dashboards/alerts to confirm none of them act on this number.
- **[Tests]** `tests/test_apple_pipeline.py:380-413` — the helper passes `mock_person_store.get_canonical_id.side_effect = lambda pid: pid` and a `MagicMock` for `get_by_id`. This bypasses the real `validate_person` path in `SourceEntityStore.add`, which is the whole point of the dual-write check the importer relies on. Test coverage of the "linked" path is therefore weaker than it looks — it only confirms the row got written, not that the link survives real validation. Consider using a real `PersonEntityStore` on the same tmp_path so the test exercises the actual constraint.
- **[Tests]** Both new tests would benefit from a third case: an unknown caller observed across multiple calls in the same batch (verifies `seen_phones` dedup actually works AND `unresolved` increments per-call, not per-unique-phone, matching the dashboard claim above).

### Summary
Mergeable as-is — the logic is correct, the rule is well-codified, and the 160-test green run reproduces locally. The `O(N^2)` inner-loop is the only finding worth addressing before merge if backfills against larger exports are anticipated; everything else is documentation/test polish that can land here or as a follow-up.

nbramia:

| 1 | O(N²) ``this_phone_max_ts`` scan re-runs ``_extract_phone_from_title`` over all calls per unique phone | Recommended | **Accept** | Real perf issue. With 859 calls × ~70 unique phones today, ~60k regex calls per run. Pre-compute once at the top. |
| 2 | Docstring claims "importers handle the unknown caller case" but only ``apple_data_import.py`` does so post-PR | Minor | **Accept** | True — ``sync_phone_calls.py`` and ``whatsapp.py`` still drop. Soften the wording to "this importer" + note that other phone-bearing importers should adopt the same pattern. |
| 3 | ``unresolved`` counter semantics changed (now conflates "no phone in title" + "no matching Person") | Minor | **Accept** | Verified no external dashboards depend on it. But the conflation muddies what the counter means. Will split into ``unresolved_no_phone`` (true failure) and ``orphan_observations`` (success — captured to unlinked SE). |
| 4 | Test uses MagicMock for PersonEntityStore | Minor | **Reject** | The point of the test is the source_entity dual-write logic, not ``validate_person``. Real PersonEntityStore would need real seeded rows for marginal coverage gain. Out of scope. |
| 5 | Add test for multi-call dedup of unknown caller | Minor | **Accept** | Easy and good coverage of the seen_phones + counter interaction. |

**Findings forwarded to addresser:** 4 (Accept #1, #2, #3, #5; Reject #4).

nbramia:
## Review Round 1 — Addressed

Pushed \`2759fc0\`. Summary:

| 1 | O(N²) max-timestamp scan | **Applied** | Pre-compute \`phone_max_ts: dict[str, datetime]\` in one pass before the main loop. ~60k regex calls eliminated per typical 800-call export. |
| 2 | Docstring overstates scope | **Applied** | Now says: \`apple_data_import\` adopts orphan-SE; \`sync_phone_calls\` and \`whatsapp\` still drop and should adopt in a follow-up. |
| 3 | \`unresolved\` counter conflates failure + success | **Applied** | Split into \`unresolved_no_phone\` (true failure) and \`orphan_observations\` (call captured to unlinked SE). \`unresolved\` retained as a sum for back-compat. |
| 4 | Real PersonEntityStore in tests | **Rejected** | Out of scope — test exercises the source_entity dual-write, not \`validate_person\`. |
| 5 | Multi-call dedup test | **Applied** | New \`test_multiple_calls_from_unknown_caller_dedup_to_one_se\`: 3 calls from same unknown number → exactly 1 SE row, but \`orphan_observations=3\` so per-call accounting still works. |

**Tests:** \`pytest tests/test_apple_pipeline.py::TestPhoneImport\` → 4 passed.

**Pre-push hook:** 1595 unit tests passed.

Issue #233, opened 2026-05-29
epic: Unified Agent Threads — interactive replies + on-demand spawn


nbramia:
## Problem

LifeOS has **three disconnected agent systems** — the `#agent` worker (local Gemma / cloud Managed Agents), the Claude Code orchestrator (`/code`), and the chat orchestrator (`/api/ask/stream`) — each with its own notion of a "session" and inconsistent support for follow-up conversation. Two concrete pain points result:

1. **You can't reliably reply to a completion notification.** When a task finishes (or fails) and you disagree with the result, there's no dependable way to push back and continue the thread. A reply mechanism exists for `#agent` completions but is fragile and undiscoverable (see Phase 1).
2. **You can't spin up an agent on demand.** Starting an agent requires creating a `#agent`-tagged vault task (or using the separate `/code` system). There's no operator "spawn a local or cloud agent to do X" from chat or Telegram.

## Context

Investigation found the worker's `sessions` + `pending_questions` tables already provide ~90% of a resumable, replyable session model. The strategy is to promote that to a canonical **AgentThread** concept that both Telegram and `/chat` render and converse with, rather than bolting features onto each system separately.

Design decisions locked with the user:
- **Unified session + notification model** (not per-surface patches).
- **Auto-routing with explicit override** for on-demand spawn (preflight decides local vs cloud; `local`/`claude` keyword forces it).
- **30-minute window** for "plain message resumes the last agent thread" before an explicit reply is required.

## Phases

- [x] #234 — Phase 1: Harden & generalize the reply/notification model (replies work on failures too, on any message chunk, no reply-gesture required within 30 min)
- [x] #235 — Phase 2: Operator root-spawn (on-demand local/cloud agents, no vault task; preflight auto-route + override)
- [x] #236 — Phase 3: Bring agent threads into web `/chat` (list/reply/spawn APIs + UI + SSE)
- [x] #237 — Phase 4: Reconcile `/code` (register its completions in the follow-up table; file deeper merge as follow-up)

## Sequencing

Phase 1 → 2 → 3, with Phase 4's decision (option A) made up front. Phase 1 delivers the most-felt pain (replies that actually work, including on failures) with the least new surface area and no schema change. Phases 2–3 add additive columns only (`origin`, `sent_message_ids`).

## Cross-cutting constraints

- Schema changes, MCP/worker dispatch changes, and new API endpoints are all "ask first" per AGENTS.md — covered by this plan.
[...]

PR #239, opened 2026-05-29
feat(agent-worker): harden & generalize replyable agent threads

### LINKED ISSUE #234: feat(agent-worker): harden & generalize replyable agent threads
Part of #233 (Unified Agent Threads) — **Phase 1**.

## Problem

Replying to an agent completion notification on Telegram is supposed to reopen the session as a follow-up turn, but in practice it rarely works. Three failure modes:

1. **Failures aren't replyable.** `register_completion_followup` is only called in the COMPLETED branch (`api/services/agent_worker/worker.py:1169-1197`). When a task FAILS or hits BUDGET_EXCEEDED — exactly when you most want to say "this isn't done" — there's no reply path at all.
2. **Only the first message chunk is matchable.** `send_message_capture_id` returns just the first chunk's `message_id` (`api/services/telegram.py:144`). Long completions split at 4096 chars; replying to chunk 2+ silently misses (`deposit_answer`, `api/services/agent_worker/session_store.py:902`).
3. **Requires the native Telegram reply gesture.** A plain follow-up message routes to the chat pipeline instead of the agent (`_handle_update`, `api/services/telegram.py:512`), so the agent never sees it.

## Context

The resume machinery itself works: `_resume_as_followup` (`worker.py:618`) reopens the session, swaps the tag back to `#agent-running`, appends the reply as a new user turn (history preserved), and re-runs the executor (local) or posts to the existing managed session (cloud). The gaps are all in *registration* and *matching*, not resumption.

## Suggested approach

1. **Register follow-ups for all terminal states.** Move follow-up registration into the shared terminal-notification path so FAILED and BUDGET_EXCEEDED notifications are replyable. Generalize the `_swap_tag(task_id, COMPLETED_TAG, RUNNING_TAG)` call in `_resume_as_followup` (`worker.py:627`) to swap from whatever terminal tag is current.
2. **Match on any chunk.** Change `send_message_capture_id` to return all chunk ids; register a follow-up per chunk id (or add a `sent_message_ids` JSON column and match on membership in `deposit_answer`).
3. **Drop the reply-gesture requirement.** Add a per-chat "last resumable agent thread" pointer. In `_handle_update`, if a message isn't a command/approval and a resumable thread exists **within a 30-minute window**, route it to `_resume_as_followup` with a visible `↪ continuing \"<task>\"` prefix. Native reply-to remains the explicit selector for targeting an older thread.

## Files involved

- `api/services/agent_worker/session_store.py`
- `api/services/telegram.py`

## Acceptance criteria

- [ ] A FAILED task registers a follow-up; replying to its notification resumes the session.
- [ ] A reply to a non-first chunk of a split completion message matches and resumes.
- [ ] A plain (non-reply) message within 30 min resumes the most recent resumable thread, with a visible continuation prefix.
- [ ] A plain message after the 30-min window routes to the normal chat pipeline (no surprise resume).
- [ ] Native reply-to an older thread still targets that specific thread.
- [ ] Unit tests cover all of the above; no schema change required beyond the optional `sent_message_ids` column (additive).

nbramia:
## Summary

Phase 1 of #233 (Unified Agent Threads). Makes every terminal-state notification replyable, matches replies on any message chunk, and lets a plain message resume the last agent thread within a 30-minute window.

Closes #234. Relates to #233.

## What changed

- **All terminal states replyable.** A shared `_notify_terminal()` sends completed/failed/budget notifications with id capture and registers a follow-up — failures and budget cut-offs are now replyable, not just clean completions. `_resume_as_followup` swaps whichever terminal tag is current (`#agent-completed`/`-failed`/`-budget-exceeded`) back to `#agent-running`.
- **Match on any chunk.** `send_message_capture_id` → `send_message_capture_ids` returns every 4096-char chunk id. `pending_questions` gains an additive `sent_message_ids` JSON column; `deposit_answer` matches a reply landing on any chunk (membership via `json_each`). Legacy rows still match on `sent_message_id`.
- **30-min plain-message resume.** New `get_recent_resumable_followup()` + a resume path in the Telegram listener: a plain (non-reply) message within `_AGENT_THREAD_RESUME_WINDOW_SECONDS` (30 min) resumes the most recent thread with a visible `↪ continuing "<task>"` prefix. Beyond the window, plain messages route to the chat pipeline. Native reply-to still targets a specific older thread.

## Schema

Additive only: `pending_questions.sent_message_ids TEXT` (JSON), idempotent migration, NULL-safe for legacy rows. No data migration required.

## Test evidence

- `pytest -m "unit and not slow" tests/` (the pre-push gate): **1608 passed, 8 skipped, 0 failed**.
- New tests cover all six acceptance criteria: FAILED/BUDGET register follow-ups, reply to a non-first chunk resumes, 30-min window honored (and excluded beyond it / for clarifications), native reply targets a specific older thread.

## Review focus

- `deposit_answer` JSON-membership query correctness and the legacy-row fallback.
- Ordering of the new resume check in `_handle_update` (after command/approval/clarification/code-followup, before chat pipeline).
- Generalized terminal-tag swap loop in `_resume_as_followup`.

### DIFF
--- a/api/services/agent_worker/session_store.py
+++ b/api/services/agent_worker/session_store.py
@@ -140,7 +140,13 @@ CREATE TABLE IF NOT EXISTS pending_questions (
     answered_at       INTEGER,
     processed         INTEGER NOT NULL DEFAULT 0,
     timed_out         INTEGER NOT NULL DEFAULT 0,
-    kind              TEXT NOT NULL DEFAULT 'clarification'
+    kind              TEXT NOT NULL DEFAULT 'clarification',
+    -- JSON array of every Telegram chunk id for this notification. Long
+    -- completions split across multiple 4096-char messages; a reply can land
+    -- on any chunk, so `deposit_answer` matches membership in this list (not
+    -- just the first chunk in `sent_message_id`). NULL for legacy rows, which
+    -- still match via `sent_message_id`.
+    sent_message_ids  TEXT
 );
 CREATE INDEX IF NOT EXISTS idx_pq_message_id ON pending_questions(sent_message_id);
 CREATE INDEX IF NOT EXISTS idx_pq_open ON pending_questions(answered_at, processed, timed_out);
@@ -246,6 +252,14 @@ class SessionStore:
                 conn.execute(
                     "ALTER TABLE pending_questions ADD COLUMN kind TEXT "
                     "NOT NULL DEFAULT 'clarification'"
+                )
+            # Idempotent migration for `pending_questions.sent_message_ids` —
+            # the full chunk-id list so a reply to any chunk of a split
+            # notification matches. Legacy rows stay NULL and match on
+            # `sent_message_id`.
+            if "sent_message_ids" not in pq_cols:
+                conn.execute(
+                    "ALTER TABLE pending_questions ADD COLUMN sent_message_ids TEXT"
                 )
             # Idempotent migration for the prompt-cache token buckets on
             # `sessions`. Old rows stay at zero — we don't backfill historical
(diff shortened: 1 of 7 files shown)


nbramia:

| 1 | Plain-message resume consumes the message even if the worker is down (handoff via `pending_questions`, no liveness check) | Recommended | Downgrade to Minor → documented | The existing native reply-deposit path (`_maybe_deposit_agent_answer`) has identical behavior — this is a pre-existing property of the deposit/tick architecture, not a regression. Added a comment documenting the accepted edge (parity with the reply path). A liveness check would be scope creep + inconsistent with the established pattern. |
| 2 | Unrelated plain message within 30 min is swallowed into the agent thread | Recommended | Reject (by design) | This is exactly acceptance criterion #3, a decision locked with the user in #233. Implemented as specified; not deviating from the locked spec. |
| 3 | `deposit_answer` OR + `json_each` subquery defeats the message-id index | Minor | Reject | `pending_questions` holds only a handful of open rows; a two-query optimization adds complexity for no measurable gain (Simplicity First). |
| 4 | `get_recent_resumable_followup` → `deposit_answer` race | Minor | Reject | Single-user deployment makes the race impossible (reviewer agreed: informational). |
| 5 | Test mock chunk math vs real `_split_message` newline preference | Minor | Reject | Reviewer confirmed the mock matches reality for the test's newline-free input. |

**Findings forwarded to addresser:** 0 (only a documentation comment applied directly). No Action Required items.

Review loop complete.

Issue #236, opened 2026-05-29
feat(chat): bring agent threads into web /chat


nbramia:
Part of #233 (Unified Agent Threads) — **Phase 3**.

## Problem

The web `/chat` interface has zero support for agent conversations. You can't see in-flight or completed agent threads, can't reply to a completion, and can't spawn an agent on demand from the UI — all of that is Telegram-only today. This closes that gap so `/chat` and Telegram speak to the same AgentThread model.

## Context

Builds on Phase 1 (replyable thread model + follow-up table) and Phase 2 (`create_operator_session()`). The web UI streams via SSE from `/api/ask/stream` (`api/routes/chat.py:900-1185`); event types are already structured (content, status, sources, usage, perf_trace, done). The frontend is `web/index.html`. No new agent state is required — threads come from the existing session store + transcript store.

## Suggested approach

1. **Read API:** `GET /api/agents/threads` (list resumable/recent sessions: status, summary, last activity) and `GET /api/agents/threads/{id}` (transcript tail).
2. **Reply API:** `POST /api/agents/threads/{id}/reply` → deposits into the same follow-up path `_resume_as_followup` consumes (reuse `deposit_answer`, keyed by thread rather than Telegram message id).
3. **Spawn API:** `POST /api/agents/spawn` → `create_operator_session()` from Phase 2.
4. **SSE:** extend `/api/ask/stream` (or a dedicated `/api/agents/stream`) to emit `agent_thread` events (status changes, completion) for live UI updates, mirroring existing event types.
5. **Frontend (`web/index.html`):** an "Agents" panel listing threads with status badges and a reply box each; a "Run as agent (local / auto / cloud)" affordance on the composer.

## Files involved

- new `api/routes/agents.py` (or extend an existing agents route)
- `api/routes/chat.py`
- `web/index.html`

## Constraints

- New `/api/agents/*` endpoints are "ask first" per AGENTS.md; must match existing auth posture.
- Managed-session resume can 404 if Anthropic GC'd the remote session — surface the same clear message and offer "start a fresh thread with this context."

## Acceptance criteria

- [ ] `/chat` lists recent/resumable agent threads with status and summary.
- [ ] Replying to a completed thread from the UI resumes it; the result streams back live.
- [ ] Spawning an agent from the composer (local/auto/cloud) creates a thread that appears in the panel.
- [ ] GC'd cloud sessions surface a clear "start fresh" path rather than a silent failure.
[...]

PR #242, opened 2026-05-29
feat(agent-worker): operator root-spawn (on-demand local/cloud agents)

### LINKED ISSUE #235: feat(agent-worker): operator root-spawn (on-demand local/cloud agents)
Part of #233 (Unified Agent Threads) — **Phase 2**.

## Problem

The only way to start an agent today is to create a `#agent`-tagged vault task (or use the separate `/code` Claude Code system). There's no operator "spawn a local or cloud agent to do X" directly from Telegram or chat. `lifeos_agent_spawn` exists but is **same-lineage only** — it requires a parent agent session, so there's no operator-initiated root spawn.

## Context

`lifeos_agent_spawn` already creates a backing-task-free session: synthetic `task_id = spawn_…`, status `CLAIMED`, explicit routing, prompt enqueued as a pending message (`api/services/agent_worker/inter_agent.py:206-300`). `_dispatch_spawned_sessions` (`api/services/agent_worker/worker.py:409`) already dispatches such sessions to the local or managed executor. The only blocker for operator use is that this path requires `parent_session_id` and a same-lineage caller.

Routing should **auto-route via preflight with an explicit override** (decision locked with user): explicit `local`/`claude` keyword wins; otherwise `run_preflight()` (`api/services/agent_worker/preflight.py`) decides, falling back to the existing Telegram clarification flow on `ROUTE_ASK`.

## Suggested approach

1. **`create_operator_session()`** — a parentless sibling of `spawn()`: CLAIMED session, `parent_session_id=None`, `root_session_id=self`, default operator budget from settings, prompt enqueued. Resolve routing per the override-then-preflight rule above.
2. **Add an explicit `origin='operator'` marker column** rather than overloading null-parent, so `_dispatch_spawned_sessions` can pick up operator-rooted sessions without colliding with the top-level `#agent` claim path (relax its `if not session.parent_session_id: continue` skip to accept `origin='operator'`).
3. **Surfaces:** Telegram `/agent <task>`, `/agent local <task>`, `/agent claude <task>` (`_handle_command`, `api/services/telegram.py`); a chat slash/intent in the orchestrator calling the same entry point. Both return the new thread id and become the chat's "last resumable thread" (Phase 1).

## Files involved

- `api/services/agent_worker/inter_agent.py` (or a new `operator_spawn.py`)
- `api/services/agent_worker/session_store.py` (new `origin` column — additive)
- `api/services/telegram.py`
- `api/routes/chat.py`

## Constraints

- Adds a DB column and touches worker dispatch — both "ask first" per AGENTS.md; covered by the #233 plan.
- `_dispatch_spawned_sessions` is synchronous with a local concurrency cap of 1; operator local spawns will queue behind running ones. Acceptable for now — surface queue state rather than build a worker pool here.

## Acceptance criteria

- [ ] `/agent claude do X` spawns a managed (cloud) session.
- [ ] `/agent local do X` spawns a local Gemma session.
- [ ] `/agent do X` (no keyword) routes via preflight; ambiguous tasks trigger the existing clarification ask.
- [ ] Spawned operator sessions dispatch correctly and do not interfere with the `#agent` task claim path.
- [ ] The resulting completion/failure notification is replyable (Phase 1).
- [ ] Equivalent spawn affordance available from the chat orchestrator.

nbramia:
## Summary

Phase 2 of #233 (Unified Agent Threads). Adds operator-initiated **root spawn** — start an agent on demand from Telegram or web chat with no backing `#agent` vault task.

Closes #235. Relates to #233. Builds on #234 (replyable threads).

## What changed

- **`create_operator_session()`** (`api/services/agent_worker/operator_spawn.py`): a parentless sibling of `spawn()` — CLAIMED session, `parent_session_id=None`, `root_session_id=self`, `origin='operator'`, default budget from `agent_default_*` settings, prompt enqueued as a pending message. Routing follows **override-then-preflight**: explicit `local`/`claude` wins; otherwise `run_preflight()` decides; `ROUTE_ASK` parks the session at `blocked`/`routing='ask'` for the clarification flow.
- **Additive `sessions.origin` column** (idempotent migration, NULL = "agent"). `_dispatch_spawned_sessions`'s parentless-skip is relaxed to claim `origin='operator'` sessions, so they dispatch alongside spawned children without colliding with the top-level `#agent` claim path.
- **Routing-ask resolution** for operator sessions flips `blocked → claimed` (instead of the `#agent` inline-execute path) so the spawned-session dispatch seeds the *enqueued prompt* as the task description — avoids losing the prompt to a non-existent vault-task lookup.
- **Surfaces:** Telegram `/agent [local|claude] <task>` (`_handle_agent_spawn`) and the same `/agent` slash command in web chat (`_handle_agent_slash` in `api/routes/chat.py`). Terminal notifications are replyable via the Phase 1 follow-up model.

## Schema

Additive only: `sessions.origin TEXT` (idempotent migration, NULL-safe). No data migration.

## Test evidence

- `pytest -m "unit and not slow" tests/` (pre-push gate): **1626 passed, 8 skipped, 0 failed**.
- New `tests/test_operator_spawn.py` (8) covers explicit override, preflight auto-route, ROUTE_ASK parking, origin marker, prompt enqueue, empty/unsafe rejection, budget defaults.
- New worker-integration tests: operator session dispatches + registers a replyable follow-up; ask-resolution dispatches after reply with prompt intact; operator spawn coexists with a `#agent` task.
- Telegram `/agent` command + web `/agent` slash command tests.

## Review focus

- The `_dispatch_spawned_sessions` skip relaxation (`origin != 'operator'`) and its non-collision with the `#agent` claim path (tick ordering).
- Operator routing-ask resolution flipping to CLAIMED rather than inline-executing (prompt-preservation).
- Cross-process flow: Telegram/chat (API process) writes the session; the worker process dispatches it next tick.


nbramia (commented):
## PR Review: feat(agent-worker): operator root-spawn (on-demand local/cloud agents)

Verified independently: checked out `feat/issue-235-operator-spawn`, traced the cross-process flow, and ran the affected suites — `tests/test_operator_spawn.py`, `tests/test_chat_helpers.py`, `tests/test_telegram.py`, `tests/test_agent_worker_clarifications.py`: **69 passed**. Tick ordering and routing-ask resolution behave as the PR describes.

### Action Required
- **[Correctness — cross-process race]** `create_operator_session` (`api/services/agent_worker/operator_spawn.py:169-181`) commits the session row with `status=CLAIMED` in `SessionStore.create()` and then enqueues the prompt in a **separate transaction** (`enqueue_message`, `session_store.py:859`). The agent worker is a separate process (`config/systemd/lifeos-agent-worker.service` → `python -m api.services.agent_worker.worker`) running its own tick loop against the shared `data/sessions.db`. If the worker's `_dispatch_spawned_sessions` (`worker.py:417-428`) runs `list_by_status(CLAIMED)` in the window after the session commits but before the prompt commits, it drains zero pending messages and dispatches with `description = session.session_id` (worker.py:428) — i.e. the agent runs against a meaningless task description like `sess_abc123` instead of the operator's prompt. This window did not exist for the existing `lifeos_agent_spawn` path because that runs *in-process* inside the worker (no concurrent tick); operator spawn is the first cross-process writer of a CLAIMED+enqueue session. Fix options: enqueue the prompt and set CLAIMED in a single transaction, create the session in a non-dispatchable status then flip to CLAIMED after enqueue, or have `_dispatch_spawned_sessions` skip an `origin='operator'` session that has zero pending messages and no started executor. The existing tests don't catch this because they call `create_operator_session` then `tick()` serially in one process.

### Recommended
- **[Correctness — operator sessions traverse the #agent vault path]** In `_handle_outcome` (`worker.py:1149`), `is_spawned = bool(session.parent_session_id)` is **False** for operator sessions (they are roots with `parent_session_id=None`). So on every terminal outcome they take the `not is_spawned` branch and call `_complete_task` / `_swap_tag` / `_set_task_status` (worker.py:1177-1197, 1213-1215, 1230-1232) against the synthetic `op_...` id, which has **no backing vault task**. These are wrapped in try/except so they don't crash — but each operator completion fires 2-3 HTTP calls to `/api/tasks/op_.../...` that 404 and log warnings (`swap_tag failed`, `complete failed`, `set_task_status failed`). The terminal Telegram notification + follow-up registration still succeed, so functionally it works, but it generates log noise and wasted round-trips on the happy path. Consider gating the vault-task mutations on `session.origin != 'operator'` (the notification/follow-up path is what operator sessions actually need). Tests use `FakeApi` so this noise is invisible to CI.
- **[Correctness — empty-completion guard on operator sessions]** Same root cause: the no-op completion guard (`worker.py:1171-1190`) also runs for operator sessions and will swap to `#agent-failed` / set `cancelled` on the non-existent vault task. The operator still gets a "returned empty result" notification, which is the right signal, but the wording references `#agent` task semantics that don't apply to an operator spawn.

### Minor
- **[Robustness]** `_handle_agent_slash` (`api/routes/chat.py:31-38`) catches exceptions from `create_operator_session` but not from the dynamic imports themselves being inside the try — that's fine — however the broad `except Exception` truncates the error to 200 chars and surfaces it verbatim to the user. Low risk given local-only data, but a preflight LLM error string is shown to the operator. Acceptable.
- **[Consistency]** Telegram (`telegram.py:392-396`) tells the user "reply to it ... within 30 min to continue", but web chat (`chat.py:47-52`) says "Track it on the Agents page" with no reply affordance — correct given web has no inline reply yet (Phase 3/#236), and the docstring notes this. No change needed; flagging the intentional asymmetry.
- **[Spec/docs]** `docs/specs/technical/agent-worker.md:457` states operator dispatch happens "without colliding with the top-level #agent claim path (which uses NULL origin)." Accurate for the *claim* path, but the doc doesn't mention that operator sessions also flow through the #agent *outcome/notification* path (the `not is_spawned` branch). Worth a line so a future reader isn't surprised by the 404 warnings.

### Summary
Solid, well-tested feature with correct tick-ordering and prompt-preservation on the routing-ask path. **Needs changes** before merge: the cross-process create-then-enqueue race (Action Required) can cause an operator agent to run with a garbage task description, and operator sessions traversing the `#agent` vault-mutation path produce spurious 404s on every terminal outcome.

nbramia:

| 1 | Cross-process create-then-enqueue race: worker could dispatch an operator session before its prompt lands, running with the synthetic session id as the task description | Action Required | **Accept — fixed** | Real bug unique to the cross-process operator path. Fixed by enqueueing the prompt *before* creating the session row (pending_messages has no FK on session_id), so the worker never observes a CLAIMED operator session without its prompt. |
| 2 | Operator sessions traverse the #agent vault path in `_handle_outcome` → 2-3 HTTP 404s + warnings per terminal outcome | Recommended | **Accept — fixed** | Gated `_complete_task`/`_swap_tag`/`_set_task_status` on a new `has_vault_task` (`not is_spawned and origin != 'operator'`) across all terminal branches; notification + follow-up still fire. Also gated the same in `_resume_as_followup`. |
| 3 | Empty-completion guard runs for operator sessions with #agent-flavored wording/swap | Recommended | **Accept — fixed** | Same `has_vault_task` gate applied to the empty-result guard's tag/status mutations. |
| 4 | Doc omits that operator sessions flow through the #agent outcome path | Minor | **Accept — fixed** | Tech spec now documents the `has_vault_task` gating and enqueue-before-create ordering. |
| 5 | Web vs Telegram reply-affordance asymmetry | Minor | Reject (by design) | Web reply UI is Phase 3 / #236; web ambiguous-route asks the user to re-issue with an explicit model. |

**Findings forwarded to addresser:** fixes applied directly in commit df8b849. Added a test asserting operator completion attempts no vault mutation. Pre-push suite green (1627 passed).

Issue #248, opened 2026-05-29
feat: merge /code into the unified agent worker (option B)


nbramia:
Follow-up to #233 (Unified Agent Threads) and #237 (Phase 4, option A).

## Context

#237 locked **option A**: keep the Claude Code orchestrator (`/code`, `api/services/claude_orchestrator.py`) as a separate session model, but make its completion messages replyable through `pending_questions` (`kind='code_followup'`). The actual resume still reuses `claude_orchestrator`'s in-memory `_last_completed` + 5-min `FOLLOWUP_WINDOW`.

This issue tracks **option B**: the deeper merge that #237 deferred.

## Decisions locked

1. **Distinct `code` route** in the agent worker (not folded into `claude`/managed).
2. **Full retirement** of `claude_orchestrator.py` once parity is reached — deliberate, thorough, staged.
3. **Resume model mirrors agent worker.** Threaded Telegram replies engage a specific `/code` session; main feed independent; `/chat` thread view = reply chain (1:1 with `session_id`, same as agent threads today). No `FOLLOWUP_WINDOW`.
4. No in-flight session migration needed.

### Phase 1 — `CodeExecutor` inside agent_worker (#274)

New file: `api/services/agent_worker/code_executor.py`. Mirrors `LocalExecutor` / `ManagedExecutor` shape but wraps the Claude Code CLI subprocess. Lifts subprocess machinery out of `claude_orchestrator.py` (spawn, stream-json parser, `[NOTIFY]`/`[CLARIFY]` extraction, heartbeat, watchdog, typing indicator).

Persistence via `SessionStore` + `TranscriptStore` (no more in-memory `_active_session`/`_last_completed`):
- `sessions.routing = 'code'` (new value)
- `sessions.session_id` ← Claude CLI's session UUID (captured from init event) — **persisted so resume survives worker restart**
- `sessions.origin = 'operator'`
- Transcript events appended per CLI event (init, tool_call, notify, clarify, completed)

### Phase 2 — Router wiring (folded into #274)

Edit `api/services/agent_worker/router.py` + `worker.py::_dispatch`:
- Add `ROUTE_CODE` constant + dispatch branch → `CodeExecutor`
- Skip preflight for `routing='code'` (route is explicit; no Haiku classification)
- Reuse `claude` route's budget surface (`wall_seconds` etc.), sourced from existing settings

### Phase 3 — Spawn surface + reply routing (#275)

[...]

PR #249, opened 2026-05-29
feat: reconcile /code completions with the unified follow-up table

### LINKED ISSUE #237: feat: reconcile /code with unified agent threads
Part of #233 (Unified Agent Threads) — **Phase 4**.

## Problem

The Claude Code orchestrator (`/code`, `api/services/claude_orchestrator.py`) overlaps heavily with the cloud (Managed Agents) route and already supports multi-turn resume via `resume_session_id`. It is a *third* session model alongside the `#agent` worker and the chat orchestrator. Leaving it fully separate entrenches the fragmentation the #233 epic is trying to remove; fully merging it is a large, risky migration. This issue locks the near-term decision and does the light build.

## Context

Decision locked with user — **option A**: keep `/code` as-is for now, but make its completion messages replyable through the same follow-up table the rest of the system uses, so reply behavior is uniform across surfaces. The deeper merge (routing `/code` through the unified worker as a `claude` route and retiring the separate orchestrator) is **option B**, filed as a follow-up rather than done here, to avoid a big migration mid-effort.

## Suggested approach

1. Register `/code` session completion messages in the shared follow-up table (the same mechanism Phase 1 generalizes), so a reply to a `/code` completion resumes that session uniformly.
2. Reuse `claude_orchestrator`'s existing `resume_session_id` path for the actual resume — do not rebuild its session handling.
3. File a separate follow-up issue for **option B** (merge `/code` into the unified worker as a `claude` route; retire the standalone orchestrator), referencing #233.

## Files involved

- `api/services/claude_orchestrator.py`
- `api/services/telegram.py`
- shared follow-up registration (from Phase 1 / `api/services/agent_worker/session_store.py`)

## Acceptance criteria

- [ ] Replying to a `/code` completion message resumes that Claude Code session via its existing `resume_session_id` path.
- [ ] `/code` reply behavior is consistent with `#agent` thread replies (same follow-up table, same matching rules from Phase 1).
- [ ] A follow-up issue is filed for option B (deeper merge), linked to #233.
- [ ] No regression to existing `/code`, `/code_status`, `/code_cancel` behavior.

nbramia:
## Summary

Phase 4 of #233 (Unified Agent Threads) — **option A, light build**. Replying to a `/code` (Claude Code) completion message now resumes that session, uniform with `#agent` thread replies, via the shared `pending_questions` follow-up table. The actual resume reuses the orchestrator's existing `resume_session_id` path.

Closes #237. Relates to #233. Option B (deeper merge) filed as #248.

## What changed

- **Orchestrator hook:** `ClaudeOrchestrator` gains `on_complete(session)`, fired once from `_cleanup` on successful completion (status `completed`, `session_id` known). `run_task` / `followup` accept it. No change to existing `/code`, `/code_status`, `/code_cancel` behavior.
- **Registration:** Telegram `_code_callbacks(chat_id)` builds a notify callback that captures each message's chunk ids (`send_message_capture_ids`) plus an `on_complete` that registers a `pending_questions` row with `kind='code_followup'`, keyed to the completion message's chunk ids and storing the Claude `session_id`.
- **Reply routing:** `_maybe_handle_code_reply` (checked before the agent-worker deposit) matches an open `code_followup` row on **any chunk** (`get_open_question_by_message_id`), closes it, and resumes via `orchestrator.followup()` → `resume_session_id`. The agent worker **skips** `code_followup` rows (they point at a Claude Code session, not an agent-worker session).

## Test evidence

- `pytest -m "unit and not slow" tests/` (pre-push gate): **1637 passed, 8 skipped, 0 failed**.
- New `tests/test_code_followup.py` (9): on_complete fires only on completion + with a session id; registration + any-chunk lookup; reply resumes via the orchestrator; unresumable tells the user; non-code / no-match replies fall through.
- Worker-skip test: a `code_followup` row is marked processed without dispatching.

## Review focus

- The `on_complete` seam in `_cleanup` (single completion path) and that it doesn't fire on failure / missing session id.
- Reply routing precedence: `_maybe_handle_code_reply` before `_maybe_deposit_agent_answer`; worker skipping `code_followup`.
- Known option-A limitation (resume bound to orchestrator's `_last_completed` + 5-min window) — deferred to #248.

### DIFF
--- a/api/services/agent_worker/session_store.py
+++ b/api/services/agent_worker/session_store.py
@@ -999,6 +999,25 @@ class SessionStore:
             ).fetchone()
         return dict(row) if row else None
 
+    def get_open_question_by_message_id(self, sent_message_id: int) -> dict | None:
+        """Return the open (unanswered, not-timed-out) question a reply to
+        `sent_message_id` matches — on any chunk — or None.
+
+        Read-only sibling of `deposit_answer`: lets a caller inspect the matched
+        row's `kind` and route the reply (e.g. a `code_followup` resumes Claude
+        Code rather than the agent worker) before recording an answer.
+        """
+        with self._connect() as conn:
+            row = conn.execute(
+                "SELECT * FROM pending_questions "
+                "WHERE answered_at IS NULL AND timed_out = 0 "
+                "AND (sent_message_id = ? OR (sent_message_ids IS NOT NULL "
+                "AND EXISTS (SELECT 1 FROM json_each(sent_message_ids) WHERE value = ?))) "
+                "ORDER BY id ASC LIMIT 1",
+                (int(sent_message_id), int(sent_message_id)),
+            ).fetchone()
+        return dict(row) if row else None
+
     def list_answered_unprocessed_questions(self) -> list[dict]:
         with self._connect() as conn:
             rows = conn.execute(
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -750,7 +750,16 @@ class Worker:
         cutoff = int(time.time()) - timeout_seconds
         stale = self.session_store.list_timed_out_questions(cutoff)
         for q in stale:
+            # Close every stale row, but only nudge for actual clarifications
+            # (agent BLOCKED awaiting input). Completion follow-ups —
+            # kind='followup' (#234) and kind='code_followup' (#237) — are just
+            # replyable notifications; a "re-tag with #agent to retry" nudge is
+            # wrong for them (and a code_followup points at a Claude Code
+            # session with no #agent task at all). Marking them timed out also
+            # keeps stale follow-up rows from accumulating.
             self.session_store.mark_question_timed_out(q["id"])
+            if (q.get("kind") or "clarification") != "clarification":
+                continue
             self.transcript_store.append(q["session_id"], "clarification_timed_out", {
                 "question_id": q["id"],
             })
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -512,6 +512,12 @@ class Worker:
         for q in answered:
             session_id = q["session_id"]
             task_id = q["task_id"]
+            # `code_followup` rows (#237) point at a Claude Code session, not an
+            # agent-worker session — the Telegram listener resumes those itself.
+            # Skip them here so the worker doesn't treat them as agent threads.
+            if q.get("kind") == "code_followup":
+                self.session_store.mark_question_processed(q["id"])
+                continue
             session = self.session_store.get_by_session_id(session_id)
             if session is None:
                 # Stale — session was deleted? Mark processed so we don't loop.
--- a/api/services/claude_orchestrator.py
+++ b/api/services/claude_orchestrator.py
@@ -247,6 +247,10 @@ class ClaudeOrchestrator:
         self._heartbeat: Optional[threading.Timer] = None
         self._typing_stop: Optional[threading.Event] = None
         self._notification_callback: Optional[Callable[[str], None]] = None
+        # Fired once when a session completes successfully (status "completed",
+        # session_id known). Used to register the completion message in the
+        # shared follow-up table so a reply resumes the session (#237).
+        self._on_complete: Optional[Callable[["ClaudeSession"], None]] = None
 
     def is_busy(self) -> bool:
         return self._active_session is not None and self._active_session.status in (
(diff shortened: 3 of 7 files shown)


nbramia (commented):
## PR Review: feat: reconcile /code completions with the unified follow-up table

Verified locally on branch `feat/issue-237-reconcile-code`: read the full diff in context, traced both orchestrator completion paths, and ran the touched tests — `tests/test_code_followup.py` (9) + `tests/test_agent_worker_clarifications.py` (22) = **31 passed**. The core design is sound: completion notification fires *before* `_cleanup("completed")` invokes `on_complete` (both result-path L575-597 and process-exit path L491-493), `send_message_capture_ids` is fully synchronous so `last_ids` is populated before the callback reads it, and the fires-once guarantee holds (`_on_complete` cleared at `_cleanup` L697, single success terminus).

### Recommended
- **[Correctness]** Stale `code_followup` rows get an agent-worker timeout nudge. `worker._timeout_stale_clarifications` (worker.py:744, called every tick at :326) uses `list_timed_out_questions` (session_store.py:1036), which matches **any** open row regardless of `kind`. A `code_followup` row the user never replies to (common: they resume via plain text through `_check_code_followup`, or just don't reply) will, after `agent_clarification_timeout_hours`, fire a nudge ("task is still waiting on your reply... re-tag with #agent to retry") referencing a Claude Code session that has no `#agent` task — `get_by_session_id` returns None for the Claude session id. Confusing cross-subsystem message. Suggest excluding `kind='code_followup'` (and arguably `'followup'`) from the clarification timeout query, or having the nudge skip non-clarification kinds.

### Minor
- **[Correctness]** Orphan `code_followup` rows accumulate. The plain-text resume path `_check_code_followup` (telegram.py:926) resumes via `orch.followup` but never closes the prior completion's open `code_followup` row. Each subsequent completion registers another open row, so a multi-turn `/code` session via plain-text replies leaves N-1 dangling open rows. They're harmless to routing (any-chunk match is keyed to unique message ids; the worker skips them) but they're exactly the rows that trigger the timeout nudge above. The native-reply path (`_maybe_handle_code_reply`) does close its row, so only the plain-text path leaks.
- **[Standards]** `_register_code_followup` truncates `question` to 200 chars (telegram.py) while `register_completion_followup` stores the full label; cosmetic inconsistency, not a bug.

### Non-issues verified (raised in review focus)
- **Double-processing:** Cannot occur. `_maybe_handle_code_reply` runs before `_maybe_deposit_agent_answer` and returns early; it closes the row via `deposit_answer` + `mark_question_processed`. Even in the narrow race where the worker tick sees the answered-but-unprocessed code_followup row first, the `kind=='code_followup'` skip (worker.py:518) marks it processed without dispatching — no double-resume.
- **Wrong-subsystem leak:** `get_recent_resumable_followup` filters `kind='followup'` (session_store.py:987), so the plain-text agent-thread resume path cannot pick up a code_followup row. Confirmed.
- **Thread-safety of `_cleanup`/`last_ids`:** No new race. `_cleanup` was already lock-free pre-PR; the orchestrator is single-active-session (gated by `is_busy`), `_cleanup` runs only from the one stream-reader thread per session, and each `/code` invocation builds a fresh `last_ids` via `_code_callbacks`. The mutable-list-across-closures pattern is safe under this single-writer model.

### Summary
Merge-ready in substance — the reply-routing precedence, worker skip, and fires-once hook are correct and well-tested. The one Recommended item (code_followup rows triggering an agent-worker timeout nudge) is a real cross-subsystem UX wart worth a small query-filter fix before or shortly after merge; the orphan-row accumulation is Minor and could be deferred to #248.

nbramia:

| 1 | Stale `code_followup` rows trigger the agent-worker timeout nudge ("re-tag with #agent"), referencing a Claude Code session with no #agent task | Recommended | **Accept — fixed** | `_timeout_stale_clarifications` now closes stale rows of any kind but only nudges `kind='clarification'`. Fixes the wrong-subsystem message and (bonus) stops stale follow-up rows accumulating — covering the Minor orphan-row item too. |
| 2 | Orphan `code_followup` rows via the plain-text `/code` resume path | Minor | **Accept via #1** | The timeout fix now closes these; deeper handling (closing on plain-text resume) deferred to #248. |
| 3 | `question` truncated to 200 chars vs full label in `register_completion_followup` | Minor | **Reject** | A `/code` prompt can be long; 200 is a sensible cap for a display label. |
| — | Double-processing / wrong-subsystem leak / on_complete ordering / last_ids thread-safety | (verified non-issues) | Acknowledged | Reviewer independently confirmed these review-focus areas are correct. |

**Findings forwarded to addresser:** fix applied directly in commit 7f3109a, with a test asserting a stale code_followup row is timed out without a nudge. Pre-push suite green (1638 passed).

PR #250, opened 2026-05-29
feat(chat): bring agent threads into web /chat

### LINKED ISSUE #236: feat(chat): bring agent threads into web /chat

nbramia:
## Summary

Phase 3 of #233 (Unified Agent Threads). The web `/chat` interface now speaks to the same AgentThread model as Telegram: list agent threads, reply to a completion to resume it, and spawn an agent on demand from the composer.

Closes #236. Relates to #233. Builds on #234 (replyable threads) + #235 (operator spawn).

## What changed

**API (`api/routes/agents.py`):**
- `GET /api/agents/threads` — recent/resumable threads (root sessions only; spawned children are internal), each with a `resumable` flag.
- `GET /api/agents/threads/{id}` — session metadata + transcript tail.
- `POST /api/agents/threads/{id}/reply` — queues a follow-up the worker resumes via `_resume_as_followup` (409 on a still-running thread, 404 unknown, 400 empty).
- `POST /api/agents/spawn` — Phase 2's `create_operator_session` (`local`/`claude`/`auto`).

**Store:** `session_store.enqueue_web_followup()` — a pre-answered `followup` row so a web reply takes the same resume path as a Telegram reply, with no Telegram message id.

**Frontend (`web/index.html`):** an Agents threads panel (status badges + per-thread reply box) and a composer "run as agent" control (auto / local / cloud). Live-ish via polling; the `/agents` page keeps the `/api/agents/stream` SSE graph.

## Test evidence

- `pytest -m "unit and not slow" tests/` (pre-push gate): **1649 passed, 8 skipped, 0 failed**.
- New `tests/test_agent_threads_api.py` (11): threads list (roots only, resumable flag), thread detail + 404, reply queues a follow-up / 409 running / 404 unknown / 400 empty, spawn explicit + auto-route (preflight) + 400 empty.
- New `tests/test_agent_threads_ui_browser.py` — Playwright (`[browser, slow]`) covering panel render, composer controls, spawn → appears in panel, reply → POST fired; uses `page.route()` mocking so the UI flow is deterministic and free of real worker/LLM side effects.

## Verification note

The API layer is fully unit-tested (gates this PR). The browser test follows the repo's existing `[browser, slow]` pattern — it requires a live server and is **not** run by the pre-push gate (consistent with all browser tests here). The frontend JS was syntax-validated with `node --check`. The full spawn→worker→resume execution is covered by the worker resume unit tests in #234/#235.

## Review focus

- `enqueue_web_followup` correctness vs the worker's `_process_clarification_answers` followup path.
- Reply gating (terminal-only) and the threads list filtering to root sessions.
- Frontend additions are contained + defensive (all fetches wrapped; panel degrades quietly when the API/worker is offline).

### DIFF
--- a/api/routes/agents.py
+++ b/api/routes/agents.py
@@ -315,6 +315,110 @@ async def stream_snapshots() -> StreamingResponse:
     )
 
 
+# ---------------------------------------------------------------------------
+# Agent threads for web /chat (#236, Phase 3 of #233)
+# ---------------------------------------------------------------------------
+
+
+class SpawnRequest(BaseModel):
+    """Body for POST /api/agents/spawn — operator agent spawn from web chat."""
+    prompt: str
+    # "local" | "claude" force the model; "auto"/None routes via preflight.
+    routing: str | None = None
+
+
+class ReplyRequest(BaseModel):
+    """Body for POST /api/agents/threads/{id}/reply."""
+    text: str
+
+
+def _thread_dict(s: Session, transcript: TranscriptStore) -> dict[str, Any]:
+    """Session dict plus a `resumable` flag for the threads panel."""
+    d = _session_to_dict(s, transcript)
+    d["resumable"] = s.status in TERMINAL_STATUSES
+    d["origin"] = getattr(s, "origin", None)
+    return d
+
+
+@router.get("/threads")
+async def list_threads(limit: int = Query(50, ge=1, le=200)) -> dict[str, Any]:
+    """List recent/resumable agent threads (root sessions only) for /chat.
+
+    Spawned children are internal to a parent's flow, so only top-level
+    sessions (no `parent_session_id`) are surfaced as conversable threads.
+    """
+    session_store = _get_session_store()
+    transcript_store = _get_transcript_store()
+    sessions = session_store.list_sessions(limit=_SNAPSHOT_LIMIT)
+    threads = [
+        _thread_dict(s, transcript_store)
+        for s in sessions
+        if not s.parent_session_id
+    ]
+    return {"threads": threads[:limit], "total": len(threads)}
+
+
+@router.get("/threads/{session_id}")
+async def get_thread(
+    session_id: str,
+    limit: int = Query(200, ge=1, le=2000),
+) -> dict[str, Any]:
+    """A single thread: session metadata + a transcript tail."""
+    session_store = _get_session_store()
+    transcript_store = _get_transcript_store()
+    session = session_store.get_by_session_id(session_id)
+    if session is None:
+        raise HTTPException(status_code=404, detail="thread not found")
+    try:
+        events = transcript_store.read(session_id)
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+    tail = events[-limit:] if len(events) > limit else events
+    return {"thread": _thread_dict(session, transcript_store), "events": tail, "total": len(events)}
+
+
+@router.post("/threads/{session_id}/reply")
+async def reply_to_thread(session_id: str, body: ReplyRequest) -> dict[str, Any]:
+    """Reply to a completed/failed thread to resume it as a follow-up turn.
+
+    Deposits into the same follow-up path a Telegram reply consumes — the
+    worker's next tick reopens the session via `_resume_as_followup`.
+    """
+    text = (body.text or "").strip()
+    if not text:
+        raise HTTPException(status_code=400, detail="reply text is required")
+    session_store = _get_session_store()
+    session = session_store.get_by_session_id(session_id)
+    if session is None:
+        raise HTTPException(status_code=404, detail="thread not found")
+    if session.status not in TERMINAL_STATUSES:
+        raise HTTPException(
+            status_code=409,
+            detail=f"thread is {session.status}; can only reply to a finished thread",
+        )
+    session_store.enqueue_web_followup(session_id, session.task_id, text)
+    return {"ok": True, "session_id": session_id, "status": "queued"}
+
+
+@router.post("/spawn")
+async def spawn_agent(body: SpawnRequest) -> dict[str, Any]:
+    """Spawn an operator agent on demand (#235's create_operator_session)."""
+    prompt = (body.prompt or "").strip()
+    if not prompt:
+        raise HTTPException(status_code=400, detail="prompt is required")
+    routing = body.routing
+    explicit = routing if routing in ("local", "claude") else None  # "auto"/None → preflight
+    from api.services.agent_worker.operator_spawn import create_operator_session
+
+    session_store = _get_session_store()
+    result = await asyncio.to_thread(
+        create_operator_session, session_store, prompt, explicit_routing=explicit,
+    )
+    if not result.get("ok"):
+        raise HTTPException(status_code=400, detail=result.get("error", "spawn failed"))
+    return result
+
+
 class KillRequest(BaseModel):
     reason: str = ""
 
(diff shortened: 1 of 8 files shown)


nbramia (commented):
## PR Review: feat(chat): bring agent threads into web /chat

Verified independently: checked out the branch, ran `tests/test_agent_threads_api.py` (11 passed), validated the inline JS parses, and traced `enqueue_web_followup` → `_process_clarification_answers` → `_resume_as_followup` against the worker. The reply path and store helper are correct. One real dead-end in the spawn path.

### Action Required
- **[Correctness/Requirements]** `POST /api/agents/spawn` (`api/routes/agents.py:403-417`) ignores the `needs_routing` case. When `routing="auto"` and preflight returns `ROUTE_ASK`, `create_operator_session` parks the session at `STATUS_BLOCKED` with `routing='ask'` and returns `{ok: True, needs_routing: True, routing: "ask"}`. The worker's `_dispatch_spawned_sessions` (`worker.py:417`) only claims `STATUS_CLAIMED` sessions, and nothing resolves the `ask` — so the session is permanently stuck. Both existing callers handle this: `telegram.py:595` registers a "local or claude?" clarification question, and `chat.py:933` tells the user to re-run with an explicit model. The new endpoint does neither: it returns `result` verbatim with HTTP 200, and the frontend (`spawnAgentFromComposer`) treats `resp.ok` as success and clears the composer. The user sees a "blocked" thread that never runs and can't be resumed. Fix: either register the routing clarification (Telegram parity) or reject `needs_routing` with a 4xx the frontend surfaces.

### Recommended
- **[Performance]** `list_threads` (`agents.py:343-358`) builds each thread via `_thread_dict` → `_session_to_dict`, which calls `transcript.read(session_id)` (a JSONL file read, `transcript_store.py:46`) plus a label lookup for every one of up to `_SNAPSHOT_LIMIT=200` sessions. The threads panel only renders `label`, `status`, and `resumable` — the entire transcript read + `_summarize_events` (tool_call_count/error_count) is wasted. This runs on a 5s `setInterval` poll per open `/chat` tab (`index.html`), so it's up to 200 file reads every 5 seconds per viewer. Consider a lighter projection for this endpoint that skips the transcript read (label can stay; the summary fields aren't used here).

### Minor
- **[Security]** `escapeHtml` (`index.html:4218`) uses `textContent`→`innerHTML`, which escapes `&<>` but NOT double-quotes. In `renderAgentThreads`, the label is interpolated into an attribute: `title="' + label + '"`. A task description containing a `"` can break out of the `title` attribute and inject a handler. Low severity (the data is the operator's own task titles, not third-party), but attribute-context interpolation should quote-escape or use `el.setAttribute`/`textContent`.
- **[Security]** `session_id` (`sid`) is interpolated unescaped into three sinks in `renderAgentThreads`: `onclick="toggleAgentReply('...')"`, `id="reply-..."`, and the `onkeydown` handler. Not currently exploitable because IDs are server-generated `sess_<hex>` (`session_store.py:222`), but it's the kind of latent injection that bites later if the id format ever changes. Prefer `addEventListener` + `dataset` over inline handlers built by string concat.
- **[Correctness]** `list_threads` returns `"total": len(threads)` where `threads` is the root-filtered list capped at `_SNAPSHOT_LIMIT` (200) *before* the `[:limit]` slice — so `total` is "roots found within the 200 most recent sessions," not the true total. Harmless given the UI doesn't display it, but the field name is misleading.
- **[Consistency]** Unauthenticated `POST /spawn` can trigger paid Claude inference (`routing="claude"`). This matches the existing unauthenticated write routes on this router (`/kill`, `/resume`, `/focus`) and the app's localhost/Tailscale-only posture, so it's consistent precedent rather than a new regression — noting it for awareness only.

### Summary
Solid, well-tested PR with a correct reply/follow-up integration, but the `auto` spawn path has a real dead-end (`needs_routing` unhandled) that should be fixed before merge; the rest are recommended/minor cleanups.

nbramia:

| 1 | `/spawn` ignores `needs_routing`: an `auto` spawn whose preflight returns ROUTE_ASK leaves a permanently-blocked session and returns 200 | Action Required | **Accept — fixed** | Web has no inline routing-clarification flow. `/spawn` (and the chat `/agent` slash) now tear down the parked session (new `SessionStore.delete_session`) and return 409 asking for an explicit model. |
| 2 | `list_threads` reads every session's transcript on a 5s poll for fields the panel doesn't use | Recommended | **Accept — fixed** | New lightweight `_thread_dict` skips the transcript read (label resolves via the cached TaskManager lookup). |
| 3 | `escapeHtml` doesn't escape `"`; label injected into `title="…"` | Minor (security) | **Accept — fixed** | Double-quotes now escaped for the title attribute context. |
| 4 | `session_id` unescaped in onclick/id | Minor (security) | **Reject** | Server-generated `sess_<hex>` — not attacker-controllable; latent only. |
| 5 | `total` after cap; unauthenticated `/spawn` | Minor | **Reject** | `total` = root-thread count by design; `/spawn` matches the existing unauth localhost/Tailscale posture of all agents routes. |

**Findings forwarded to addresser:** fixes applied directly in commit 530b9ab, with tests for the 409-cleanup path and `delete_session`. Pre-push suite green (1651 passed). The reviewer independently verified the `enqueue_web_followup` → resume integration and reply gating are correct.

PR #254, opened 2026-05-29
feat(agent-viz): "Go To" wezterm pane for any active CC session

### LINKED ISSUE #251: feat(agent-viz): "Go To" wezterm pane for any active CC session
## Problem

The /agents page already has a Focus button + `POST /api/agents/sessions/{id}/focus` endpoint that calls `wezterm cli activate-pane --pane-id <id>` (`api/routes/agents.py:983–1053`), backed by `data/cc_wezterm.db` via `CCWezTermStore`. But the mapping is only populated when the user clicks **Resume**, which spawns a new pane. Sessions started organically (user opens wezterm, runs `claude`) appear in /agents with no pane mapping → Focus is hidden.

Want: click "Go To" (or double-click an active node) → wezterm focuses the right pane, **even when multiple panes are running `claude` in the same cwd**.

## Context

Cwd alone can't disambiguate (very common to have multiple parallel `claude` invocations in the same repo). The unique anchor is the session's transcript file: every CC session writes to `~/.claude/projects/<encoded-cwd>/<session-uuid>.jsonl`, and exactly one `claude` process holds it open. `transcript_path` is already indexed during session ingest (`api/services/agent_worker/session_ingest.py:125–150`), so we have the link from session_id → file path → owning PID → wezterm pane.

## Suggested approach

**Two-pronged:**

### 1. SessionStart hook (authoritative, for new sessions)
- New script `scripts/claude-session-pane.sh`:
  - Reads `$WEZTERM_PANE` env + stdin JSON (`session_id`, `cwd`, `transcript_path` from Claude Code's SessionStart hook payload)
  - POSTs to `http://localhost:8000/api/agents/cc-pane-bind`
  - No-op if `$WEZTERM_PANE` unset (non-wezterm terminal)
- New endpoint `POST /api/agents/cc-pane-bind`:
  - Body: `{session_id, pane_id, cwd}`; validates `pane_id` is positive int
  - Localhost-only (`request.client.host == "127.0.0.1"`)
  - Upserts into `cc_wezterm.db` via existing `CCWezTermStore`
- User adds the hook to `~/.claude/settings.json` manually — document the snippet, don't auto-edit

### 2. FD-probe fallback (for already-running / pre-hook sessions)
New module `api/services/cc_pane_locate.py`:
1. `lsof -t -- <transcript_path>` → set of PIDs with file open
2. For each PID, walk parent chain via `/proc/<pid>/status` `PPid:`
3. `wezterm cli list --format json` (reuse existing `_find_live_wezterm_socket` at `agents.py:584`) → build pane_id → pane.pid map
4. Intersect: PID in any walked chain that matches a known pane.pid → return pane_id
5. On success, cache to `cc_wezterm.db` so subsequent clicks are O(1)

### 3. Extend `/focus` endpoint
Refactor to: cache lookup → if miss/stale, run probe → activate. Single round trip. On `activate-pane` failure: clear mapping → re-probe once → retry. Still fails → 410 with clear message.

### 4. Frontend (`web/agents.html:1251–1289`)
- Enable Focus button (consider renaming "Go To") unconditionally for active CC terminal sessions, not gated on initial `pane_id`
- Add `dblclick` handler on active node circles → same call
- Spinner during probe (~50–200ms worst case)
- Toast on 404/410 with actionable message ("Pane no longer exists" / "Couldn't locate pane — install SessionStart hook? [link]")

### Edge cases
- **Multiple panes, same cwd, multiple sessions** → transcript_path disambiguates correctly
- **Stale cache (pane closed)** → activate fails → clear → re-probe
- **Session resumed in different pane** (`claude --resume <id>`) → probe finds new pane, mapping refreshed
- **Non-wezterm terminal** → hook no-ops; probe returns None; button hidden
- **Daemon agent-worker sessions** → already correctly hidden via `is_terminal_cc_session` check at `agents.html:1251`
- **Multi-mux wezterm sockets** → existing `_find_live_wezterm_socket` handles
- **Tabs/windows** → `wezterm cli activate-pane` already crosses them

### Out of scope
- macOS support for the FD probe (Linux-targeted; macOS would swap `/proc` walk for `lsof` parent traversal — file follow-up if needed)
- Remote/SSH session targeting
- Auto-editing the user's `~/.claude/settings.json`
- Other terminal multiplexers (tmux, screen)

## Files involved

- `api/routes/agents.py` — extend `/focus`, add `/cc-pane-bind`
- `api/services/cc_pane_locate.py` — **new**, probe logic
- `api/services/cc_wezterm_store.py` — no schema change
- `web/agents.html` — button enable + dblclick handler
- `scripts/claude-session-pane.sh` — **new**, SessionStart hook script
- `docs/specs/product/agent-viz.md`, `docs/specs/technical/agent-viz.md` — document Go To
- `docs/guides/agents-go-to.md` — **new**, SessionStart hook setup snippet
- `tests/test_cc_pane_locate.py`, `tests/test_agent_cc_pane_bind.py` — **new**

## Acceptance criteria

- [ ] `POST /api/agents/cc-pane-bind` accepts `{session_id, pane_id, cwd}`, rejects non-localhost requests, and upserts into `cc_wezterm.db`
- [ ] `api/services/cc_pane_locate.py` resolves the correct pane_id given a session's `transcript_path`, even when two panes share the same cwd with different active `claude` sessions
- [ ] `POST /api/agents/sessions/{id}/focus` runs cache-then-probe in a single request; stale mappings auto-clear and re-probe once
- [ ] `scripts/claude-session-pane.sh` no-ops when `$WEZTERM_PANE` is unset and otherwise POSTs the binding successfully
- [ ] Focus / "Go To" button is enabled for all active terminal CC sessions in /agents (not just Resume-spawned ones); double-clicking the node triggers the same action
- [ ] Failure modes surface in the UI: pane closed → toast "Pane no longer exists"; probe miss → toast pointing to hook setup
- [ ] Unit tests cover: probe disambiguation across two panes/sessions in same cwd, localhost-only enforcement on `/cc-pane-bind`, cache-miss-then-probe path in `/focus`, stale-mapping re-probe
- [ ] Manual verification: two wezterm panes in `~/Code/LifeOS` each running `claude` → /agents Go To activates the correct pane for each session
- [ ] SessionStart hook setup is documented in a guide; agent-viz specs (product + technical) describe the new Go To behavior

nbramia:
## Summary

Adds a **Go To** button (and node double-click) to /agents that focuses the wezterm pane running any active Claude Code session — including sessions started organically in wezterm, not just ones opened via Resume. Handles the common case of multiple panes in the same project cwd by disambiguating on the session's transcript file, not just cwd.

Closes #251.

## What changed

**Two-pronged pane resolution:**

1. **SessionStart hook** (`scripts/claude-session-pane.sh`) → new endpoint **`POST /api/agents/cc-pane-bind`** (localhost-only, 403 otherwise). The hook reads `$WEZTERM_PANE` + the SessionStart JSON payload and posts the mapping every time `claude` starts in a wezterm pane. Authoritative, no probe needed for fresh sessions. Manual install in `~/.claude/settings.json` — see [`docs/guides/agents-go-to.md`](docs/guides/agents-go-to.md).

2. **FD-probe fallback** (`api/services/cc_pane_locate.py`) for sessions that pre-date the hook: `lsof -t -- <transcript_path>` → PID → `/proc/<pid>/fd/0` → `/dev/pts/N` → match against `wezterm cli list`'s `tty_name`. Result is cached in `cc_wezterm.db` so subsequent clicks are O(1).

**Extended `/focus` endpoint** now does cache → probe → activate in a single round trip, plus one re-probe if a cached pane has gone stale. 404 only when both cache *and* probe come up empty; 410 when a pane was identified but is gone with no replacement.

**Frontend:** the existing Focus button (now labeled "Go To") is shown on all non-subagent CC sessions, not just terminal/inactive ones. Double-click on a graph node triggers the same action; a 220ms click debounce keeps the panel-toggle on single click and Go To on double click from fighting.

## Test evidence

- 11 new tests in `tests/test_cc_pane_locate.py` covering probe disambiguation (two panes, two sessions, same cwd → each lookup finds the right pane), forked-worker matching, and every graceful-failure path (lsof missing, wezterm missing, malformed JSON, no pty match, redirected stdin).
- 11 new tests in `tests/test_agent_cc_pane_bind.py` covering `/cc-pane-bind` (localhost-only enforcement, path-traversal rejection, pane_id validation, prefix handling) and the extended `/focus` flow (probe on cache miss + caching, 404 when probe empty, stale-mapping re-probe + retry, 410 when re-probe finds nothing).
- All 28 existing tests in `tests/test_agent_resume_api.py` still pass — no regression in Resume / Focus behavior.

Full test suite: `./scripts/test.sh` — 2401 passed, 6 pre-existing failures unrelated to this change (Gemma model default mismatch, personal config UUID).

## Review focus

- **Probe correctness**: `api/services/cc_pane_locate.py` — TTY-based matching via `/proc/<pid>/fd/0`. macOS support is explicitly out of scope; the probe returns None gracefully if `lsof` or `wezterm` are missing.
- **`/focus` re-probe logic**: the cache→probe→activate→re-probe sequence in `api/routes/agents.py`. Verify the bookkeeping of "did we already probe this request?" — we never want an infinite probe loop, and a freshly-probed-then-failed pane should not re-probe again.
- **Localhost check on `/cc-pane-bind`**: `request.client.host in {"127.0.0.1", "::1"}`. The test fixture uses `TestClient(app, client=("127.0.0.1", 50000))` to satisfy this; rejection of non-loopback is tested by calling the route function directly with a stub request.
- **Frontend dblclick debounce**: `web/agents.html` — the 220ms timer adds latency to single-click panel toggle. Acceptable trade for a clean dblclick experience, but worth flagging.
- **Hook script safety**: `scripts/claude-session-pane.sh` no-ops in all failure modes (no `$WEZTERM_PANE`, no `jq`, no `curl`, server unreachable) so it can never block `claude` startup.

### DIFF
--- a/api/routes/agents.py
+++ b/api/routes/agents.py
@@ -1192,8 +1192,8 @@ async def focus_claude_code_session(session_id: str) -> dict[str, Any]:
         if probed is None:
             raise HTTPException(
                 status_code=404,
-                detail="no tracked wezterm tab and pane probe found nothing — "
-                       "install the SessionStart hook or run Resume",
+                detail="couldn't locate pane — session not running, "
+                       "wezterm unreachable, or SessionStart hook not installed",
             )
         mapping = store.get(session_id)
         assert mapping is not None  # _probe_and_store just wrote it
--- a/api/routes/agents.py
+++ b/api/routes/agents.py
@@ -12,9 +12,9 @@ import logging
 import time
 from typing import Any
 
-from fastapi import APIRouter, HTTPException, Query
+from fastapi import APIRouter, HTTPException, Query, Request
 from fastapi.responses import StreamingResponse
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from api.services.agent_worker.session_store import (
     TERMINAL_STATUSES,
@@ -527,6 +527,16 @@ class CCResumeRequest(BaseModel):
     extra_env: dict[str, str] = {}
 
 
+class CCPaneBindRequest(BaseModel):
+    """Body for POST /api/agents/cc-pane-bind — written by the SessionStart
+    hook script on every `claude` start. Lets the /agents Focus / Go To
+    button target sessions that were never opened via /agents Resume.
+    """
+    session_id: str = Field(..., min_length=1, max_length=128)
+    pane_id: int = Field(..., ge=0)
+    cwd: str = Field(default="", max_length=4096)
+
+
 def _copy_to_clipboard(text: str, env: dict[str, str]) -> bool:
     """Push `text` to the system clipboard via `wl-copy` (Wayland) or
     `xclip` (X11). Returns True on success, False on any failure. Never
@@ -982,12 +992,121 @@ async def resume_claude_code_session(
     }
 
 
+def _lookup_cc_session_meta(session_id: str):
+    """Resolve a `cc:`-prefixed session id back to its discovered SessionMeta.
+
+    Mirrors the lookup the /resume endpoint does. Returns (meta, bare_id)
+    or raises HTTPException with an appropriate status. Used by /focus to
+    locate the transcript jsonl for the FD probe fallback.
+    """
+    from api.services.claude_code import session_ingest as cc
+    from config.settings import settings
+
+    try:
+        bare = cc.validate_session_id(session_id)
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+    if ":agent:" in bare:
+        bare = bare.split(":agent:", 1)[0]
+
+    metas = cc.discover_sessions(
+        projects_dir=settings.claude_code_projects_dir,
+        lookback_days=max(int(settings.claude_code_lookback_days), 365),
+    )
+    target = next((m for m in metas if m.raw_session_id == bare), None)
+    if target is None:
+        raise HTTPException(status_code=404, detail=f"session {session_id} not found")
+    return target, bare
+
+
+def _activate_pane(pane_id: int, env: dict[str, str]) -> tuple[bool, str]:
+    """Run `wezterm cli activate-pane --pane-id <id>`. Returns (success, detail).
+
+    Raises HTTPException only for unrecoverable errors (wezterm missing,
+    timeout). A non-zero rc returns (False, stderr) so the caller can
+    decide whether to re-probe.
+    """
+    import shutil
+    import subprocess
+
+    wezterm_bin = shutil.which("wezterm") or "wezterm"
+    argv = [wezterm_bin, "cli", "activate-pane", "--pane-id", str(pane_id)]
+    try:
+        proc = subprocess.run(  # noqa: S603 — fixed argv
+            argv,
+            env=env,
+            capture_output=True,
+            timeout=3.0,
+            check=False,
+        )
+    except FileNotFoundError as exc:
+        raise HTTPException(status_code=500, detail=f"wezterm not found: {exc}") from exc
+    except subprocess.TimeoutExpired as exc:
+        raise HTTPException(status_code=504, detail=f"wezterm activate-pane timed out: {exc}") from exc
+
+    if proc.returncode == 0:
+        return True, ""
+    return False, (proc.stderr.decode("utf-8", errors="replace").strip()
+                   or f"wezterm activate-pane exited rc={proc.returncode}")
+
+
+@router.post("/cc-pane-bind")
+async def cc_pane_bind(request: Request, body: CCPaneBindRequest) -> dict[str, Any]:
+    """Localhost-only endpoint called by the Claude Code SessionStart hook.
+
+    Writes `session_id → pane_id` into `data/cc_wezterm.db` at the moment a
+    `claude` invocation starts inside a wezterm pane, so the /agents
+    Focus / Go To button can later target the pane authoritatively without
+    the FD-probe fallback. The session_id from the hook payload is the
+    bare uuid; we prefix with `cc:` to match the store's keying convention.
+
+    Bound to 127.0.0.1 only — never expose this endpoint publicly. Anyone
+    who can reach it could rewrite the pane mapping for any session.
+    """
+    client_host = request.client.host if request.client else ""
+    if client_host not in ("127.0.0.1", "::1"):
+        raise HTTPException(status_code=403, detail="cc-pane-bind is localhost-only")
+
+    from api.services.claude_code import session_ingest as cc
+
+    try:
+        bare = cc.validate_session_id(body.session_id)
+    except ValueError as exc:
+        raise HTTPException(status_code=400, detail=str(exc)) from exc
+
+    # Storage convention: keys are always `cc:`-prefixed (matches /resume's
+    # upsert). `validate_session_id` strips the prefix from the request, so
+    # we re-prefix unconditionally.
+    storage_id = f"cc:{bare}"
+    cwd = body.cwd or ""
+
+    from api.services.cc_wezterm_store import get_default_store
+    mapping = get_default_store().upsert(storage_id, int(body.pane_id), cwd)
+    return {
+        "bound": True,
+        "session_id": storage_id,
+        "pane_id": mapping.pane_id,
+        "cwd": mapping.cwd,
+    }
+
+
 @router.post("/sessions/{session_id}/focus")
 async def focus_claude_code_session(session_id: str) -> dict[str, Any]:
-    """Activate the WezTerm pane that was opened for this session by a prior
-    Resume click. Returns 404 if there's no stored pane mapping (use
-    Resume to create one) and 410 if the pane no longer exists (mapping
-    cleared).
+    """Activate the WezTerm pane running this Claude Code session.
+
+    Resolution order:
+      1. Cached mapping in `data/cc_wezterm.db` (written by /resume or by
+         the SessionStart hook → /cc-pane-bind).
+      2. FD-probe fallback (`cc_pane_locate`): walk the session's
+         transcript file's holders to a wezterm pane via TTY matching.
+         On hit, cache the mapping so subsequent calls are O(1).
+      3. If activate-pane fails on a cached mapping (typical: user closed
+         the tab), the mapping is cleared and the probe runs once more —
+         the session may have been resumed in a fresh pane.
+
+    Returns 404 only when no mapping exists AND probe finds nothing
+    (session not running, non-wezterm terminal). Returns 410 when a pane
+    existed but is now gone and no replacement can be found.
 
     Caveat: on GNOME Wayland the compositor disallows cross-client window
     raise, so this reliably switches the tab within wezterm but cannot
@@ -995,8 +1114,6 @@ async def focus_claude_code_session(session_id: str) -> dict[str, Any]:
     follows the activate-pane call to flash the dock icon as a hint.
     """
     import shlex
-    import shutil
-    import subprocess
 
     if not session_id.startswith("cc:"):
         raise HTTPException(status_code=400, detail="focus is only available for Claude Code sessions")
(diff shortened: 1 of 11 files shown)


nbramia (commented):
## PR Review: feat(agent-viz): "Go To" wezterm pane for any active CC session

I verified the PR independently: 50 tests pass in the affected files (`test_cc_pane_locate.py`, `test_agent_cc_pane_bind.py`, `test_agent_resume_api.py`), the wezterm JSON output on this machine confirms `tty_name` + `pane_id` are present as integers/strings, TestClient `client=("127.0.0.1", 50000)` does propagate to `request.client.host` (so the localhost fixture is not a smell), and `request.client.host` reflects the raw TCP peer (no proxy / forwarded-headers middleware is wired in `api/main.py`). Server binds `0.0.0.0:8000` per `scripts/server.sh:24` and is therefore reachable over Tailscale — the localhost-only check is the only thing protecting `/cc-pane-bind`.

### Action Required
None — no merge-blocking issues found.

### Recommended

- **[Performance/Correctness]** `_lookup_cc_session_meta` in `api/routes/agents.py:1012-1015` calls `cc.discover_sessions(..., lookback_days=max(int(settings.claude_code_lookback_days), 365))` on every cache-miss focus request. The function's own docstring (`session_ingest.py:195-196`) says *"Discovery is the expensive part of ingestion — callers should cache the result for ~30s rather than re-scanning every SSE tick."* For users with a busy `~/.claude/projects/` this is a full-year filesystem walk per Go To click on a fresh session. Consider (a) caching the discovery result for ~30s like the rest of the code path does, or (b) since you have the bare uuid + the `cc:` prefix convention, doing a direct `for proj in projects_dir.iterdir(): if (proj/f"{bare}.jsonl").exists(): …` lookup, which is O(num projects) instead of O(num sessions across last 365 days). Either avoids the year-wide scan.

- **[Correctness/UX]** `_resume_env()` is passed to `cc_pane_locate._wezterm_panes()` (`api/routes/agents.py:1151`). `_resume_env` may inject `WEZTERM_UNIX_SOCKET` discovered via `_find_live_wezterm_socket` from log files (`agents.py:574-577`), but if the daemon's pid changes between resume and probe, the cached socket can be stale. `_wezterm_panes` will then silently return `[]` and the probe returns `None` → 404 with a misleading "install the SessionStart hook" toast even though wezterm is running. Worth either re-running `_find_live_wezterm_socket` per call (probably already happens via `_resume_env()`'s no-cache shape — verify) or distinguishing "wezterm unreachable" from "wezterm has no matching pane" in the toast.

- **[Correctness — wezterm pane id reuse]** `wezterm cli activate-pane --pane-id <N>` accepts a stale id that has been reassigned to an unrelated pane after wezterm restart. The cache row written by `/cc-pane-bind` survives wezterm restarts (it's persisted to SQLite). On the next Go To we'd happily activate the wrong pane and not even know it. The probe path validates the pane belongs to the right session via tty matching, but the cached path does not. Two mitigations: (1) keep an "wezterm boot id" or socket inode alongside the mapping and invalidate on change, or (2) call the cheap probe first and fall back to the cached pane only if probe returns `None`. Currently lowest-likelihood event but worth noting because it has no symptom in the UI ("Pane selected" toast lies).

- **[Correctness — discover_sessions has its own internal `validate_session_id` skip]** `_lookup_cc_session_meta` strips `:agent:` *after* `validate_session_id` runs (`api/routes/agents.py:1006-1010`). `validate_session_id` allows `:`, so `cc:abc:agent:xyz` passes, but the discovery loop only matches `raw_session_id == bare` where bare is post-`:agent:`-strip. Good — that's correct. But the order of operations doesn't strip `cc:` first before checking `:agent:`, so a request like `cc:abc:agent:xyz` would be checked for `:agent:` against the still-prefixed bare. Minor — likely a no-op because `validate_session_id` returns `bare` (without the `cc:` prefix) so the split works. Verify by adding a regression test for an agent session id reaching `/focus` via the dblclick handler if subagents are ever reachable that way.

- **[Correctness — frontend dblclick swallows panel toggle on non-CC nodes]** `web/agents.html:1131-1143`: the dblclick handler clears `_nodeClickTimer` unconditionally but only calls `focusCcSession` when `d.source === 'claude_code' && !d.is_subagent`. If a user accidentally double-clicks a LifeOS worker or subagent node, the panel toggle is silently dropped. Either gate the `clearTimeout` on the same condition, or fire the panel toggle as a fallback when Go To isn't applicable.

### Minor

- **[Standards]** `_lookup_cc_session_meta` has function-level imports for `cc` and `settings` (`agents.py:1002-1003`); same for `_activate_pane`'s `import shutil`/`import subprocess` (`agents.py:1029-1030`). These are presumably to avoid import cycles in `api/main.py` boot, but the rest of `agents.py` already imports `subprocess`/`shutil` at module top in the existing `/resume` path. Hoisting these to the module top would be slightly cleaner and avoid the inline `import` lines fighting the no-shadowing convention.

- **[Correctness — race]** `cc_pane_locate.py:55-74`: between `lsof` returning a PID and `_tty_of_pid` reading `/proc/<pid>/fd/0`, the process could exit. The code correctly returns `None` for the exited PID, but if it was the only holder, the probe falls through to `holder_ttys: set()` → `None`. That's the right behavior; just confirming there's no half-resolved state. Not worth changing.

- **[Correctness — `lsof` huge-output concern]** `lsof -t -- <transcript_path>` with a single explicit path argument lists at most a handful of PIDs (the file's holders). Output size is bounded; no `capture_output=True` buffer-explosion risk in practice. The 2s timeout is generous.

- **[Security — `$WEZTERM_PANE` injection via `--argjson`]** Confirmed safe: `"$WEZTERM_PANE"` is double-quoted, so it reaches `jq` as a single token; `jq --argjson` then JSON-parses it. A malformed value makes `jq` fail (empty `$()` substitution → curl posts `-d ""` → 422 server-side) but cannot inject shell metacharacters. Only a process owned by the user can set `WEZTERM_PANE`, and the endpoint is localhost-only. No exposure.

- **[Standards]** `scripts/claude-session-pane.sh` uses `set -u` but not `set -o pipefail`. The `printf | jq` pipelines have `2>/dev/null` redirected and `|| true` fallbacks, so this is defensively fine, but the file would benefit from a one-line shellcheck pass. Optional.

- **[Standards]** `cc_pane_locate.py` uses `Optional[str]` (from `typing`) for new code — the repo elsewhere uses PEP 604 `str | None`. Match the existing style.

- **[Docs]** `docs/guides/agents-go-to.md` reads cleanly. The troubleshooting list could add "wezterm has restarted since the session started — the cached pane id is now stale; click Resume to re-bind" as a fourth cause given the pane-id-reuse note above.

- **[Tests]** Coverage is strong for the unit-level paths but there's no test for the discovery cost / lookback regression — if someone later flips `_lookup_cc_session_meta` to a smaller lookback it would silently break. A test that puts a synthetic session ≥31 days old in the projects dir and verifies it's still resolvable would lock in the 365-day intent.

- **[Tests]** `test_bind_rejects_non_localhost` bypasses the actual middleware stack by calling the route function with a stub Request. That's pragmatic (TestClient won't let you spoof `client.host` to non-loopback), and you do have `test_bind_stores_mapping_from_localhost` exercising the real stack. Acceptable; just noting that a Starlette-level integration test (raw HTTP with a forged `X-Forwarded-For` header) would prove the lack of proxy-header trust empirically. Worth two lines.

### Summary
Merge-ready with no blockers. The two most worthwhile follow-ups are (1) caching or narrowing the `discover_sessions` call in `_lookup_cc_session_meta` to avoid a year-wide filesystem walk per Go To click, and (2) the wezterm pane-id-reuse edge case where a cached mapping can silently activate the wrong pane after a wezterm restart. Both are correctness-adjacent rather than blocking — the probe path is sound, the localhost-only auth holds against header spoofing (no proxy middleware is configured), and the test suite is thorough.

nbramia:

| 1 | `_lookup_cc_session_meta` runs 365-day `discover_sessions` per click — full filesystem walk and JSONL parse per Go To | Recommended | **Accept** | Real perf concern. Will replace with a direct `proj/<bare>.jsonl` glob — O(num projects) stat calls instead of parsing every transcript. |
| 2 | Wezterm pane id reuse across restarts → cached mapping silently activates wrong pane | Recommended | **Downgrade to Minor → file follow-up** | Real edge case but narrow surface (requires wezterm restart + cached mapping + pane id collision with another session). Fix is non-trivial (probe-before-activate or store wezterm boot id). Will file as a follow-up issue rather than expand this PR. |
| 3 | `WEZTERM_UNIX_SOCKET` going stale → `wezterm cli list` returns `[]` → misleading "install hook" 404 | Recommended | **Accept** | Will broaden the 404 message to cover wezterm-unreachable as a cause, and add it to the troubleshooting list in the guide. Splitting the probe's "wezterm unreachable" vs "no match" cases for a more precise error would be cleaner but adds API surface — defer. |
| 4 | dblclick on non-CC nodes clears the panel-toggle timer but doesn't fire Go To → click is silently dropped | Recommended | **Accept** | Will only swallow the timer for CC sessions; non-CC dblclicks fall through to the deferred panel toggle. |
| 5 | Function-level imports could be hoisted to module top | Minor | **Reject** | Matches existing `/resume` pattern in the same file — hoisting would be inconsistent with surrounding code. |
| 6 | `Optional[str]` vs `str \| None` | Minor | **Reject** | Mixed convention in `api/services/` (Optional appears in 35+ files including sibling `cc_wezterm_store.py`); not a documented standard. |
| 7 | `set -u` without `set -o pipefail` in hook script | Minor | **Reject** | Every external call has explicit `\|\| true` / `2>/dev/null \|\| true`; pipefail would change the control flow we explicitly designed around. |
| 8 | Docs guide could add "wezterm has restarted" troubleshooting cause | Minor | **Accept** | Pairs with #2 follow-up; will note in the guide. |
| 9 | Test 365-day lookback intent | Minor | **Reject** | The lookback is being removed by finding #1; test would become irrelevant. |
| 10 | Integration test for `X-Forwarded-For` forgery on `/cc-pane-bind` | Minor | **Reject** | `api/main.py` doesn't wire ProxyHeadersMiddleware (verified), so `request.client.host` reflects the raw TCP peer. The existing direct-call test for non-loopback rejection is sufficient given that constraint. |

**Findings forwarded to addresser:** 4 (1, 3, 4, 8). Finding 2 will be filed as a follow-up issue.

nbramia:
## Review round 1 — addressed

| 1 | Direct glob replaces `discover_sessions` in `_lookup_cc_session_meta` | Applied | `api/routes/agents.py` — synthesizes a minimal `SessionMeta` from a direct `<bare>.jsonl` glob across project dirs; `tests/test_agent_cc_pane_bind.py` adds `test_focus_resolves_session_without_calling_discover_sessions` which monkeypatches `discover_sessions` to fail loudly across a ≥2-project layout. |
| 2 | Broadened 404 message + matching toast | Applied | `api/routes/agents.py` 404 detail now reads "couldn't locate pane — session not running, wezterm unreachable, or SessionStart hook not installed". `web/agents.html` toast matches. Test assertions updated in `test_agent_cc_pane_bind.py` and `test_agent_resume_api.py`. |
| 3 | dblclick falls through on non-CC | Applied | `web/agents.html` — early-return when the node isn't a CC root session, so the deferred panel-toggle still fires on LifeOS worker nodes. |
| 4 | Docs troubleshooting | Applied | `docs/guides/agents-go-to.md` adds a wezterm-restart bullet (Resume to refresh) and a wezterm-unreachable bullet, broadens the lead sentence, and adds the missing `Audience` frontmatter. |
| Follow-up | wezterm-restart pane-id reuse | Filed | Issue #257 |

**Tests:** `./scripts/test.sh` — 2414 passed, 38 skipped, 15 failed. All 15 failures are pre-existing infra issues unrelated to this PR (CRM data integrity requires populated DB, MCP server tests need API running, Gemma default model, personal-config UUID, etc.). All 40 tests covering this PR's surfaces (`test_agent_cc_pane_bind.py` + `test_agent_resume_api.py`) pass.

**Commits:**
- `2e1b467` fix: address review round 1 — direct glob replaces discover_sessions in Go To lookup
- `3b1e0c2` fix: address review round 1 — broaden Go To failure UX (404 message, toast, dblclick)
- `ba11f14` docs: address review round 1 — broaden Go To troubleshooting (wezterm restart + unreachable)

PR #255, opened 2026-05-29
feat(agent-viz): search box with ranked results in /agents UI

### LINKED ISSUE #252: feat(agent-viz): search bar with ranked results in /agents UI
## Problem

The `/agents` page (`web/agents.html`) has no way to search. With many sessions on the graph, finding a specific one means visually scanning or fiddling with the filter dropdowns. We want a search input in the existing filter bar that, as you type, shows a ranked dropdown of matching sessions; clicking a result highlights that node in the graph.

## Context

The page is a single D3 force graph. The filter bar (`.filters`, ~line 700) holds dropdowns that work by **hiding** nodes (`applyFilters()` ~line 913 returns the surviving set, `renderGraph()` redraws). Node selection already exists — clicking a node calls `openPanel(session_id)` + `applySelectionStyles()` (~line 1178), which highlights it, lights up 1-hop neighbors, and dims the rest. Search reuses that as its "highlight node" primitive rather than reinventing it.

Searchable text per node:
- **`label`** — session name (task description, with fallbacks). Always present in the snapshot, already client-side.
- **`short_label`** — LLM 2–6 word "name summary."
- **`summary`** — LLM paragraph summary.

`short_label`/`summary` live in `data/agent_viz_summaries.db` (table `agent_viz_summary`: `session_id, last_activity_at, short_label, summary, created_at`) and only exist once a session's summary has been generated (panel opened or prefetched).

**Limitation (accepted):** the summary tiers only cover sessions with a generated/cached summary. The `label` tier covers every session.

## Suggested approach

**UX model — results-list dropdown, NOT graph-filtering.** Add a search `<input>` to the `.filters` bar. As the user types (debounced, live), render a ranked dropdown of matching sessions. Clicking a result selects/highlights that node (reuse `openPanel` + `applySelectionStyles`) and pans to it. The graph itself stays intact — search does not hide nodes the way the dropdowns do.

**Matching — case-insensitive substring.** Ranking has two axes:
- **Primary:** sessions that pass the current filters (**visible**) rank above filtered-out (**hidden**) sessions. Show hidden matches as a visually distinct, dimmed second group in the dropdown.
- **Secondary (within each group):** `label` match > `short_label` match > `summary` match.

**Where matching happens.** `label` matching is done client-side (`allSessions` already holds all labels). `short_label`/`summary` matching needs the backend since those rows aren't all loaded client-side. Add a small endpoint:

`GET /api/agents/search?q=<query>` — substring-searches **only the cached summary rows** (`short_label` + `summary`) in `agent_viz_summaries.db`. No LLM calls; fast. Returns matching `session_id`s, which field matched, and a snippet. The client merges these with its local `label` matches, then ranks per the axes above.

**Clicking a hidden (filtered-out) result.** Evaluate each filter predicate against that one session and relax **only** the predicates that exclude it (do not reset all filters), then re-render and highlight the node.

## Files involved

- `web/agents.html` — search input in `.filters`, results dropdown, ranking/merge logic, click→highlight (reuse `openPanel`/`applySelectionStyles`), selective filter-relax on hidden-result click.
- `api/routes/agents.py` — new `GET /api/agents/search` endpoint.
- `api/services/agent_viz_summary.py` — cached-summary substring search helper (alongside `_disk_get`).

## Out of scope (deferred)

- Full **transcript content** search (the originally-discussed fallback tier). Summaries only for now.

## Acceptance criteria

- [ ] A search input appears in the `.filters` bar on `/agents`.
- [ ] Typing live-filters (debounced) into a ranked results dropdown.
- [ ] Results are ranked visible-before-hidden, and within each group `label` > `short_label` > `summary`.
- [ ] Hidden (filtered-out) matches appear as a visually distinct, dimmed group.
- [ ] Clicking a result highlights/selects the node and pans to it (reusing existing selection styling).
- [ ] Clicking a hidden result relaxes only the filter predicate(s) that exclude that session, then highlights it.
- [ ] `GET /api/agents/search?q=` returns substring matches over cached `short_label`/`summary` rows with no LLM calls, including matched-field and snippet.
- [ ] `label`-tier search works for every session even when its summary has not been generated.

nbramia:
## Summary

Adds search to the `/agents` page. A search input in the filter bar drives a live, debounced dropdown of matching sessions ranked **visible-before-hidden**, and within each group **label > short_label > summary**. Clicking a result highlights the node (reusing the existing `openPanel` + `applySelectionStyles` selection styling) and pans the viewport to it. Clicking a filtered-out result first relaxes *only* the filter predicates that were hiding it.

The `label` tier is matched client-side (every session carries a label, already in the snapshot). The summary tiers (`short_label` / paragraph `summary`) come from a new `GET /api/agents/search?q=` endpoint that substring-scans the cached-summary SQLite DB — no LLM call, safe to hit on every keystroke.

Closes #252.

## Test evidence

`~/.venvs/lifeos/bin/python -m pytest tests/test_agent_viz_api.py` → **22 passed**. Seven new tests cover the endpoint and `search_cached_summaries` helper: short_label vs summary field selection, case-insensitivity, LIKE-wildcard escaping (`00%` matches literal `100%`, not as a wildcard), empty/whitespace queries, the endpoint happy path, and 400 on missing `q`.

Full suite (`./scripts/test.sh`): the only failures are 7 pre-existing, environment-specific tests (local LLM model default, `MY_PERSON_ID` UUID, MCP/health endpoints) — verified to fail identically on a clean tree with my changes stashed, so they're unrelated to this PR.

Inline `agents.html` JS validated with `node --check`.

## Review focus

- **Backend** (`agent_viz_summary.py`): `search_cached_summaries` — LIKE escaping correctness and the cache-only / no-LLM guarantee.
- **Frontend ranking** (`agents.html` `buildSearchResults`): tier ordering and the dedupe-to-best-tier merge of local label matches with backend summary matches.
- **`relaxFiltersFor`**: that it flips only the predicates actually excluding the clicked session (recency picks the smallest covering window).
- **Note:** live browser verification was not run — the server runs from the main checkout via systemd, not this worktree, so the new endpoint/UI aren't served by the running instance. Backend is covered by unit tests; frontend by `node --check` + self-review. Happy to verify in-browser once merged to main.

### DIFF
--- a/api/routes/agents.py
+++ b/api/routes/agents.py
@@ -376,6 +376,22 @@ async def get_session_summary(session_id: str) -> dict[str, Any]:
     return {"session_id": session_id, **result.as_dict()}
 
 
+@router.get("/search")
+async def search_sessions(
+    q: str = Query(..., min_length=1, max_length=200),
+    limit: int = Query(200, ge=1, le=500),
+) -> dict[str, Any]:
+    """Substring search over cached session summaries (short_label + summary).
+
+    Powers the /agents search box's summary tiers. The client handles the
+    label tier locally; this endpoint covers the LLM-generated summary text
+    that isn't fully loaded client-side. Reads only the disk cache — it never
+    triggers summary generation, so it's cheap and safe to call on keystroke.
+    """
+    matches = agent_viz_summary.search_cached_summaries(q, limit=limit)
+    return {"query": q, "matches": matches}
+
+
 @router.get("/stream")
 async def stream_snapshots() -> StreamingResponse:
     """SSE stream emitting a full snapshot every ~2 seconds."""
--- a/api/services/agent_viz_summary.py
+++ b/api/services/agent_viz_summary.py
@@ -458,6 +458,75 @@ def get_cached_summary(session_id: str, last_activity_at: float, status: str = "
     return None
 
 
+_LIKE_SPECIAL = re.compile(r"([\\%_])")
+
+
+def _escape_like(s: str) -> str:
+    """Escape LIKE wildcards so a query like `100%` matches that literal text
+    instead of acting as a `match-anything` pattern."""
+    return _LIKE_SPECIAL.sub(r"\\\1", s)
+
+
+def _make_snippet(text: str, query_lower: str, width: int = 72) -> str:
+    """A short, whitespace-collapsed window around the first case-insensitive
+    match of `query_lower` in `text`, with ellipses when truncated."""
+    collapsed = " ".join(text.split())
+    idx = collapsed.lower().find(query_lower)
+    if idx < 0:
+        return collapsed[:width] + ("…" if len(collapsed) > width else "")
+    half = max(0, (width - len(query_lower)) // 2)
+    start = max(0, idx - half)
+    end = min(len(collapsed), start + width)
+    snippet = collapsed[start:end]
+    if start > 0:
+        snippet = "…" + snippet
+    if end < len(collapsed):
+        snippet = snippet + "…"
+    return snippet
+
+
+def search_cached_summaries(query: str, limit: int = 200) -> list[dict[str, str]]:
+    """Substring-search cached `short_label` + `summary` rows. No LLM, no network.
+
+    Case-insensitive. Returns at most `limit` matches as
+    ``{session_id, field, snippet}`` where ``field`` is ``"short_label"`` when
+    the short label matched, else ``"summary"``. Only rows already in the disk
+    cache are searched — sessions without a generated summary are not covered
+    (the client's label tier handles those). Reads only; never re-summarizes.
+    """
+    q = query.strip()
+    if not q:
+        return []
+    q_lower = q.lower()
+    pattern = f"%{_escape_like(q_lower)}%"
+    try:
+        with sqlite3.connect(_resolve_db_path()) as conn:
+            rows = conn.execute(
+                "SELECT session_id, short_label, summary FROM agent_viz_summary "
+                "WHERE short_label LIKE ? ESCAPE '\\' OR summary LIKE ? ESCAPE '\\' "
+                "ORDER BY last_activity_at DESC LIMIT ?",
+                (pattern, pattern, int(limit)),
+            ).fetchall()
+    except sqlite3.Error as exc:
+        logger.debug("summary search failed for %r: %s", query, exc)
+        return []
+    results: list[dict[str, str]] = []
+    for session_id, short_label, summary in rows:
+        if short_label and q_lower in short_label.lower():
+            results.append({
+                "session_id": session_id,
+                "field": "short_label",
+                "snippet": _make_snippet(short_label, q_lower),
+            })
+        elif summary and q_lower in summary.lower():
+            results.append({
+                "session_id": session_id,
+                "field": "summary",
+                "snippet": _make_snippet(summary, q_lower),
+            })
+    return results
+
+
 def reset_cache() -> None:
     """Clear in-process cache only. Use for tests; disk cache survives."""
     _cache.clear()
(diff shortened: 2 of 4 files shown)


nbramia (commented):
## PR Review: feat(agent-viz): search box with ranked results in /agents UI

I independently verified the test suite (`pytest tests/test_agent_viz_api.py` → 22 passed), the LIKE-escaping logic (incl. backslash/underscore/percent in SQLite with `ESCAPE '\'`), inline-JS syntax (`node --check` OK), XSS surface (`escapeHtml` + `highlightMatch` cover all injected text), and scope (exactly the 4 files named in #252, no stray changes). All 8 acceptance criteria in issue #252 are met.

### Action Required
- None.

### Recommended
- **[Requirements/UX]** `SEARCH_BADGE` (`agents.html:2069`) labels both `short_label` and `summary` tiers as `"summary"`, so the dropdown badge can't distinguish a short-label hit from a paragraph-summary hit. Ranking still orders them correctly (tiers 1 vs 2), but the issue's stated three-tier model isn't visible to the user. Consider distinct badges (e.g. `name` / `short` / `summary`).

### Minor
- **[Correctness]** `search_cached_summaries` reads the disk DB directly (`agent_viz_summary.py:493`), bypassing the in-process `_cache`. A summary that was generated and cached in-process but whose best-effort `_disk_put` failed won't be searchable. This matches the documented "cache-only, no LLM" contract and is acceptable, but worth noting.
- **[Correctness/UX]** Summary tier only fires at `q.length >= 2` (`agents.html` `onSearchInput`) while the label tier fires at length 1. Intentional debounce-cost tradeoff; not a bug, just an asymmetry a future reader may trip on (already comment-covered).
- **[Correctness]** `panToNode` for a just-revealed hidden node relies on a fixed 400ms delay for the force sim to assign `x/y`; if the node still has null coords it silently skips the pan (graceful, no error). Fine in practice given D3 initializes positions synchronously, but timing-dependent.
- **[Style]** `highlightMatch` for a `short_label`/`label` tier hit re-highlights against the live snapshot value (`r.session.short_label`), which can differ from the DB row the backend matched; on mismatch it falls back to plain escaped text (no crash, no highlight). Negligible.

### Summary
Merge-ready. The implementation faithfully satisfies all acceptance criteria in #252, LIKE escaping and XSS handling are correct, scope is clean, and tests pass. The only non-trivial nit is the collapsed `short_label`/`summary` badge labels — a small UX polish, not a blocker.

nbramia:

| 1 | `SEARCH_BADGE` labels both short_label and summary tiers as "summary" — can't visually distinguish a name-summary hit from a paragraph hit | Recommended | **Accept** | The agreed design (#252) explicitly distinguishes the two summary tiers; distinct badges surface the model. One-line fix. |
| 2 | `search_cached_summaries` reads disk DB, bypassing in-process `_cache` | Minor | **Reject** | By design and documented — disk is the durable layer; `_disk_put` always accompanies the in-process write. Not a defect. |
| 3 | Summary tier fires at q.length>=2 vs label tier at 1 | Minor | **Reject** | Intentional debounce tradeoff, comment-covered. |
| 4 | `panToNode` 400ms fixed delay for just-revealed node coords | Minor | **Reject** | Graceful fallback (skips pan if coords still null). |
| 5 | `highlightMatch` highlights against live snapshot value vs matched DB row | Minor | **Reject** | Graceful fallback to plain escaped text on mismatch. |

**Findings forwarded to addresser:** 1 (addressed directly by orchestrator — one-line change)

nbramia:
Review Round 1: 1 finding accepted and addressed (distinct per-tier search badges, commit 35458c7), 4 rejected with reasoning. Self-verification confirms the fix is in scope and traces to the finding — no Round 2 trigger. Review loop complete.

PR #266, opened 2026-05-29
chore(agent-viz): invalidate Go To cache across wezterm restarts

### LINKED ISSUE #257: chore(agent-viz): Go To cache invalidates on wezterm restart
Follow-up from #251 (PR #254 review).

## Problem
The cached `session_id → pane_id` mapping in `data/cc_wezterm.db` persists across wezterm restarts, but wezterm resets pane ids on restart. After a restart, a cached mapping to pane N could silently activate a different session's pane N — wrong tab focused, no error visible to the user.

## Approach
On Go To, before activating from cache, verify the cached mapping is still authoritative for this wezterm boot. Two reasonable options:

1. Record the wezterm gui-sock-PID alongside the mapping at write time. On lookup, compare against the current live wezterm PID. Mismatch → invalidate, fall through to probe.
2. Always probe-validate the cached pane's tty against the current `wezterm cli list` snapshot. ~50ms extra per click but no schema change.

Option 1 is faster and respects the cache-hit O(1) goal. Option 2 is simpler and self-correcting.

## Acceptance criteria
- [ ] After a wezterm restart, the first Go To on any pre-existing CC session re-probes rather than blindly activating a stale pane id.
- [ ] Test covers: store a mapping with wezterm boot id X; simulate wezterm restart (new boot id Y); verify Go To re-probes.

nbramia:
## Summary

Pane ids reset when wezterm-gui restarts. A cached `session_id → pane_id` mapping written before a restart could silently activate an unrelated session's pane in the fresh wezterm — wrong tab focused, no error visible. This PR records the wezterm-gui pid alongside each mapping and invalidates the cache when that pid is no longer live.

Closes #257.

## What changed

**Schema** — `cc_wezterm.db` gains a `wezterm_pid INTEGER NOT NULL DEFAULT 0` column. Migration is in place via `ALTER TABLE ADD COLUMN` (idempotent, runs on every store init). Pre-#257 rows default to 0.

**Write paths** — Resume, `/cc-pane-bind`, and the `/focus` probe-success path all capture the current wezterm-gui pid via a new shared helper. `_find_live_wezterm_socket` is refactored so the live-socket enumeration lives in `_enumerate_live_wezterm_sockets` and feeds three downstream helpers:
- `_find_live_wezterm_socket` (existing semantics, picks most-recent)
- `_live_wezterm_pids` (used by /focus to validate cached entries)
- `_current_wezterm_pid` (used by upsert paths)

**Read path (`/focus`)** — after `store.get(...)`, if `mapping.wezterm_pid == 0` or not in the current live pid set, the mapping is dropped and the endpoint falls through to the FD probe. wezterm_pid=0 is the migration path — first Go To after the upgrade re-probes and pins a real pid.

**Docs** — product spec adds one line on the auto-invalidation. Technical spec describes the boot-id check in the /focus flow. Operator guide's "wezterm restarted" troubleshooting bullet now reflects that the cache self-corrects (the only remaining failure mode is `claude` not being live to probe against).

## Test evidence

`./scripts/test.sh` — 2484 passed, 15 pre-existing failures unrelated to this PR (Gemma model default, MCP server tests needing the server up, personal config UUID, p91 data integrity which requires a populated DB).

8 new tests added; 4 existing tests updated to pin a known boot id:

- `test_upsert_records_wezterm_pid` / `test_upsert_defaults_wezterm_pid_to_zero` — store API contract.
- `test_migration_adds_wezterm_pid_column_to_pre_257_db` — opens a hand-crafted pre-#257 schema and verifies the new column is added with DEFAULT 0; legacy row still readable.
- `test_migration_is_idempotent_on_already_current_schema` — re-opening a DB that already has the column doesn't error.
- `test_focus_invalidates_cache_when_wezterm_pid_no_longer_live` — the acceptance test from #257: pre-restart mapping at pane 5 / pid 11111, post-restart live pid is 22222, focus re-probes to pane 7 and refreshes the mapping. Asserts that activate-pane is called only with `7` (not with the stale `5`).
- `test_focus_invalidates_pre_257_rows_with_wezterm_pid_zero` — migration-path: wezterm_pid=0 is treated as stale, the focus path re-probes and pins a real pid.
- `test_focus_honors_cache_when_wezterm_pid_matches_one_live_gui` — multi-mux: cache is valid if its pid is in the live set, regardless of which gui is "primary".
- `test_cc_pane_bind_records_current_wezterm_pid` — the hook's bind path captures a real pid.

## Review focus

- **Schema migration safety** — `_migrate_add_column` uses `PRAGMA table_info` to detect the missing column. SQLite's `ALTER TABLE ADD COLUMN ... DEFAULT 0` is safe and fast. Verify the migration is idempotent (test covers this) and that legacy rows remain readable.
- **Invalidation correctness** — the check fires only if `wezterm_pid == 0` OR `wezterm_pid not in live_pids`. Multi-mux is supported (any live gui counts). The wezterm_pid=0 case captures both the migration path and writers that couldn't determine a live pid at write time.
- **Live-pid discovery** — `_enumerate_live_wezterm_sockets` shares logic with `_find_live_wezterm_socket`; no behavior change in the existing socket-discovery path. The new `_current_wezterm_pid` returns 0 (not None) so the schema's NOT NULL constraint isn't violated when wezterm is unreachable.
- **Tests pinning a boot id (99001)** — the 4 pre-existing focus tests now set `wezterm_pid=99001` and monkeypatch `_live_wezterm_pids` to `{99001}`. This is the minimum change needed to keep them exercising the activate-then-stale flow rather than the new boot-id invalidation.


nbramia (commented):
## PR Review: chore(agent-viz): invalidate Go To cache across wezterm restarts

Verified independently: branch checked out, all 54 tests in the three touched test modules pass (`tests/test_cc_wezterm_store.py`, `tests/test_agent_cc_pane_bind.py`, `tests/test_agent_resume_api.py`). The acceptance test for #257 (`test_focus_invalidates_cache_when_wezterm_pid_no_longer_live`) genuinely demonstrates the fix — it asserts `\"5\" not in activate_calls[0]`, so a regression that activates the stale pane id would fail the test, not just \"happen to pass.\"

### Recommended

- **[Correctness / Concurrency]** `CCWezTermStore._migrate_add_column` (`api/services/cc_wezterm_store.py:65-69`) has a TOCTOU between the `PRAGMA table_info` read and the `ALTER TABLE ADD COLUMN`. Within a single FastAPI process this is fine (singleton `get_default_store()`), but if the deployment ever runs multiple uvicorn workers against the same `data/cc_wezterm.db`, two workers initializing concurrently would both see the column missing and the second `ALTER` would raise `OperationalError: duplicate column name: wezterm_pid` and crash that worker's init. Cheap fix: wrap the ALTER in `try/except sqlite3.OperationalError` and ignore \"duplicate column\" errors — keeps the PRAGMA fast path and makes the migration genuinely concurrency-safe. (The current systemd unit is single-worker, so this is latent, not active.)

- **[Correctness / Consistency]** `/cc-pane-bind` at `api/routes/agents.py:1378-1380` uses `os.environ.get(\"XDG_RUNTIME_DIR\") or f\"/run/user/{os.getuid()}\"` directly, while `/focus` and `/resume` go through `_resume_env()` which does the same backfill plus respects `LIFEOS_CC_RESUME_ENV_FILE`. The divergence is intentional (the comment explains it), but it's worth factoring the XDG fallback into a tiny `_xdg_runtime_dir()` helper used by all three call sites — otherwise a future change to the convention has three places to update and one is easy to miss. Not blocking; just a maintenance smell.

### Minor

- **[Correctness / Multi-mux edge case]** The `wezterm_pid in live_pids` check in `/focus` (`agents.py:1466`) accepts any live wezterm pid, but the writer records only the *most-recently-modified* socket's pid (`_current_wezterm_pid` at `agents.py:868-880`). In a multi-mux setup (two simultaneous wezterm-gui processes A and B), a mapping bound under A persists as \"valid\" even after A exits — as long as B is alive, `A.pid ∈ {B.pid}` is false so this isn't actually the bug. But the converse: bind under A, later B becomes most-recent → `WEZTERM_UNIX_SOCKET` is set to B's socket by `_resume_env` while `wezterm_pid=A.pid` is still in the live set. `activate-pane` then runs against B's mux for a pane that lives in A. The activate fails, /focus falls through to re-probe (correct end behavior), but it took a wasted round trip. `test_focus_honors_cache_when_wezterm_pid_matches_one_live_gui` stubs `subprocess.run` to always succeed, so it doesn't exercise this failure path. Realistically rare (one wezterm is the common case) and self-healing via the existing 410-path probe, so I wouldn't block on it — but it deserves a sentence in the technical spec noting that multi-mux \"validity\" is approximate.

- **[Correctness / Linux pid reuse]** `_enumerate_live_wezterm_sockets` (`agents.py:798-840`) uses `os.kill(pid, 0)` to check liveness. If wezterm pid 11111 exits and the kernel later reuses 11111 for an unrelated process, `kill(0)` succeeds and the (now-orphaned) `gui-sock-11111` file would be reported as live. wezterm normally cleans up its socket file on exit, so this requires wezterm to have crashed (file left behind) *and* pid reuse to land on exactly that pid before next reboot. Extremely narrow; the cache invalidation already handles the common case correctly. Mentioning only because the review brief asked.

- **[Testing]** The 4 pre-existing focus tests that now pin `wezterm_pid=99001` and stub `_live_wezterm_pids` repeat the same two-line setup. A small fixture (`@pytest.fixture def pinned_wezterm_boot(monkeypatch)` that returns a callable taking a pid) would dedupe four sites and make the intent (\"this test isn't about restart invalidation\") clearer than the inline comments do. Not worth refactoring just for this PR, but worth doing the next time these tests are touched.

- **[Testing]** `test_migration_adds_wezterm_pid_column_to_pre_257_db` covers the exact pre-#257 schema, which is fine for the only schema that has ever shipped. It doesn't cover \"column exists but with a different declared type or default\" — which is unreachable from any real upgrade path of this codebase, so I'd leave it alone. Mentioning only for the audit trail.

- **[Standards]** `import os` inside the `/cc-pane-bind` handler (`agents.py:1378`) is a function-local import while the module already imports `os` at top-level elsewhere. Minor; matches the surrounding style of late imports for some helpers, but the module's top-of-file already has `import os`, so this is redundant. Not worth a follow-up commit on its own.

### Summary

Merge-ready. The fix correctly closes #257, the acceptance test actually exercises the bug behavior (not just the happy path), and the migration is safe for any DB that has ever shipped from this repo. The recommendations are nice-to-haves; the multi-mux edge case is the only finding with real teeth and it self-heals through the existing 410 fallback.

nbramia:

| 1 | `_migrate_add_column` TOCTOU between PRAGMA + ALTER | Recommended | **Accept** | Theoretical for single-worker uvicorn (LifeOS's deployment) but the try/except OperationalError fix is trivial and defensive. |
| 2 | `/cc-pane-bind` builds XDG fallback inline; /resume + /focus go through `_resume_env()` | Recommended | **Accept** | Factor into a small `_xdg_runtime_dir()` helper and use it from all three sites. |
| 3 | Multi-mux: cache valid against any live wezterm but `WEZTERM_UNIX_SOCKET` picks most-recent → activate may target the wrong mux | Minor | **Accept doc note only** | Real edge case. Self-heals via existing 410-then-reprobe path, so no code change — add a sentence to the technical spec. |
| 4 | Linux pid reuse can fool the `kill(0)` liveness check | Minor | **Reject** | Requires wezterm to crash leaving a stale socket file *and* the kernel to recycle that exact pid. Microscopic probability; no plausible mitigation short of reading `/proc/<pid>/comm`. |
| 5 | Pre-existing focus tests repeat the `wezterm_pid=99001` + stub pattern — fixture would DRY | Minor | **Reject** | 2 lines per test; explicit pin documents intent. Premature abstraction at 4 occurrences. |
| 6 | Migration test doesn't cover "column already exists with different type/default" | Minor | **Reject** | Reviewer themselves notes "unreachable from any shipped state". Out of scope. |
| 7 | Redundant `import os` inside `/cc-pane-bind` handler | Minor | **Accept** | `os` already at module top — drop the function-scope import. |

**Findings forwarded:** 4 (1, 2, 3, 7). Addressing inline rather than via the addresser subagent — each fix is < 5 lines.

nbramia:
## Round 1 fixes applied

Commit 53c9907 — `fix: address review round 1 — migration race, xdg helper, multi-mux doc`

| # | Finding | Action |
|---|---------|--------|
| 1 | Migration TOCTOU | try/except `sqlite3.OperationalError("duplicate column")` around the ALTER — idempotent under races |
| 2 | XDG fallback inconsistency | New `_xdg_runtime_dir()` helper; `/cc-pane-bind` now uses it |
| 3 | Multi-mux caveat | Added sentence to `docs/specs/technical/agent-viz.md` noting the self-heal path |
| 7 | Redundant `import os` in handler | Removed |

**Tests:** `pytest tests/test_cc_wezterm_store.py tests/test_agent_cc_pane_bind.py tests/test_agent_resume_api.py tests/test_cc_pane_locate.py` — 65 passed.

PR #273, opened 2026-05-29
fix(chat): bound thread-view render so large transcripts don't freeze the tab

### LINKED ISSUE #270: fix(chat): large agent threads freeze the thread-view renderer
## Problem

Opening an agent thread with a very large transcript in the web `/chat` thread view freezes the browser renderer — the tab becomes unresponsive (page never reaches `document_idle`; DevTools `Runtime.evaluate` times out). Small/normal threads (~17 bubbles) render instantly; the freeze only appears on sessions with very long histories.

## Context

Found during browser verification of the unified agent-threads work (#233 / #236). The thread view (added in #256) renders the entire reconstructed conversation in one synchronous pass, and `appendToolDetails` emits **every** tool call's full input as a `<pre>` block. A session with hundreds of turns and/or large tool inputs produces a massive DOM that hangs the renderer. The backend (`GET /api/agents/threads/{id}`) already returns the full conversation, so this is purely a frontend rendering concern.

## Suggested approach

Bound what the DOM holds at once (any one or a mix):
- Cap initially-rendered turns (e.g. newest N) with a "load earlier" control.
- Lazy-render / collapse tool details by default; only build the `<pre>` when expanded.
- Truncate tool-input previews to a few hundred chars with click-to-expand.

Directional only — pick whatever keeps the initial render cheap.

## Files involved

- `web/index.html` — `renderThreadConversation`, `appendToolDetails` (the thread-view render path)
- Possibly `api/routes/agents.py` — if a server-side cap / paginated `conversation` turns out cleaner than client-side trimming.

## Acceptance criteria

- [ ] Opening a thread with a very large transcript renders without freezing the tab.
- [ ] Tool-call detail is still inspectable (e.g. expandable), not silently dropped.
- [ ] Normal/small threads are visually unchanged.
- [ ] Full history remains reachable (load-more / expand), not truncated away.

nbramia:
## Summary
Opening an agent thread with a very large transcript froze the `/chat` thread view — it rendered the entire reconstructed conversation in one synchronous pass and emitted every tool call's full input as a `<pre>` block, producing a massive DOM that hung the renderer.

This bounds what the DOM holds at once:
- **Cap the initial render** to the newest `THREAD_INITIAL_TURNS` (30) turns, built into a `DocumentFragment` (single reflow, no per-turn smooth-scroll thrash).
- **"Load earlier" control** reveals older turns in chunks — full history stays reachable, nothing is truncated away.
- **Lazy tool details** — each tool-call `<pre>` is serialized only on first expand, not up-front.

Small/normal threads are visually unchanged (the control only appears when the conversation exceeds the cap).

Closes #270

## Test evidence
`tests/test_agent_threads_ui_browser.py` — ran against the actual `web/index.html` (served standalone), **11 passed** (6 existing + 5 new):
- initial render capped to newest 30 turns (90-turn fixture)
- "load earlier" reveals hidden turns in chunks; full history reachable via repeated load; button self-removes when exhausted
- tool-detail `<pre>` built lazily on expand
- small thread renders with no "load earlier" control

Unit suite (`./scripts/test.sh`): 2519 passed. The 9 failures are pre-existing and environmental (local model-default config drift, local person-id config, and server-dependent MCP/CRM tests flaky on the shared box — `test_health_check` passed on retry); none touch this diff, which is `web/index.html` + the browser test only.

## Review focus
- `appendToolDetails` lazy-build: the `<pre>` persists across collapse/expand via the `built` flag; correct?
- `loadEarlierThreadTurns` chronological ordering — older chunks insert directly below the anchored button, before the already-shown tail.
- `addMessage`'s new optional `target` param: when set, appends to the target and skips `scrollToBottom`; default path unchanged.

### DIFF
--- a/tests/test_agent_threads_ui_browser.py
+++ b/tests/test_agent_threads_ui_browser.py
@@ -128,3 +128,92 @@ class TestAgentThreadsUI:
         page.wait_for_timeout(500)
         assert self.state["replies"], "reply POST should have fired"
         assert self.state["replies"][0]["text"] == "also CC the property manager"
+
+    def test_small_thread_renders_without_load_earlier(self, page: Page):
+        # The 2-turn fixture is below the cap, so no "load earlier" control
+        # appears and both turns render — small threads stay unchanged (#270).
+        page.locator("#agentsThreadsList .agent-thread").first.click()
+        expect(page.locator(".agent-thread-banner")).to_be_visible(timeout=8000)
+        expect(page.locator(".thread-load-earlier")).to_have_count(0)
+        expect(page.locator("#messages .message")).to_have_count(2)
+
+    def test_tool_details_pre_is_built_lazily(self, page: Page):
+        # The collapsed <details> exists but its <pre> is only built on expand,
+        # so a thread with many tool calls doesn't serialize every input up-front.
+        page.locator("#agentsThreadsList .agent-thread").first.click()
+        expect(page.locator(".agent-tool-details summary")).to_contain_text("lifeos_gmail_draft")
+        expect(page.locator(".agent-tool-details pre")).to_have_count(0)
+        page.locator(".agent-tool-details summary").first.click()
+        expect(page.locator(".agent-tool-details pre")).to_contain_text("landlord@example.com")
+
+
+def _large_conversation(n_turns: int):
+    """Alternating user/assistant turns; assistant turns carry a tool call."""
+    conv = []
+    for i in range(n_turns):
+        if i % 2 == 0:
+            conv.append({"role": "user", "text": f"user message {i}", "tools": []})
+        else:
+            conv.append({"role": "assistant", "text": f"assistant message {i}",
+                         "tools": [{"name": "lifeos_search", "input": {"q": f"query {i}"}}]})
+    return conv
+
+
+class TestLargeThreadRendering:
+    """#270 — a thread with a very large transcript must render bounded, not
+    freeze the renderer, while keeping full history reachable."""
+
+    N_TURNS = 90  # well above THREAD_INITIAL_TURNS (30)
+
+    @pytest.fixture(autouse=True)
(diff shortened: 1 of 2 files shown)


nbramia:

| 1 | Poll re-render (`pollThreadForUpdate`) rebuilds to newest 30, collapsing a user's "load earlier" expansion when a new turn lands on an active resumable thread | Recommended | **Accept** | Narrow but real, and undercuts the PR's "full history reachable" claim. Fixing inline: preserve the expanded depth across poll re-renders. |
| 2 | Couldn't run the 11 browser tests (`:8000` serves the main checkout, not this branch) | Minor | **Reject (clarify)** | Not a code defect. The suite was run against this worktree's `web/index.html` served standalone on a temp port — 11/11 passed. The `:8000` limitation is exactly why a standalone server was used. |
| 3 | Lazy `<pre>` depends on the `<details>` `toggle` event | Minor | **Reject** | `toggle` fires on the user click that opens the summary (our only trigger); target is modern self-hosted Chromium/Firefox. No defect. |

**Findings forwarded to addresser:** 1 (addressed inline by the orchestrator)

nbramia:
## Review Round 1 — Addressed

| 1 | Poll re-render collapsed expanded "load earlier" history | **Applied** | `renderThreadConversation` now takes a `preserveFrom` bound; `pollThreadForUpdate` passes the current `earliestShown` so expanded scrollback survives new turns. Added `test_poll_rerender_preserves_expanded_history`. |

**Tests:** browser suite (`tests/test_agent_threads_ui_browser.py`) run against this branch's `web/index.html` served standalone — **12 passed**.
**Commit:** `fix: address review round 1 — preserve expanded history on poll re-render`

Self-verification: round-1 fix touches only `web/index.html` + the test; the change traces directly to the forwarded finding and fully resolves it. No new files, no untraced changes — review loop complete.

PR #287, opened 2026-06-01
fix(test): isolate vault indexer stores + feat(codex): Codex executor & viz

nbramia:
Two logically separate changes landed together at the author's request to get the working tree into remote. They can be split into two PRs if preferred.

## 1. `fix(test)`: stop the test suite from wiping the production vault index (commit `2176175`)

**Root cause.** `IndexerService.index_all()` computes `deleted_files = (state file paths) − (current vault files)` and calls `delete_file()` on each, purging them from the Chroma `lifeos_vault` collection and `data/bm25_index.db`. `INDEX_STATE_FILE`, `VectorStore()`, and `BM25Index()` were hardcoded to production, so any test that builds a tmp vault and calls `index_all()` (`test_people`, `test_integration`, `test_indexer`) treated every real note as deleted and **wiped the live vault index on every `./scripts/test.sh` run**. The 03:30 nightly silently rebuilt it, masking the loss — which is why it looked like an "overnight" problem.

**Fix.** An autouse `conftest.py` fixture redirects the indexer's three stores (state file, BM25 db, Chroma collection) to throwaway per-test locations. The tests reuse the indexer's own stores for their post-index search assertions, so they stay self-consistent; production data is never touched. Verified: the prod state file is byte-identical after running those tests, and the index/search tests still pass.

## 2. `feat(codex)`: Codex executor & session viz (commit `713da99`)

Adds Codex as a second agent-worker execution backend alongside Claude Code: `codex_executor`/`codex_spawn` + an `api/services/codex/` ingest package that parses Codex sessions for `/agents`. Renames `code_*` agent-worker modules/tests to `claude_code_*` to disambiguate the two backends, and unions Codex sessions into the agent-viz snapshot via `_codex_snapshot` (gated by `codex_viz_enabled`). Also stubs `_codex_snapshot` in the agent-viz test fixture so snapshot tests stay isolated from real Codex sessions.

## Test
- Pre-push hook: 102 unit tests pass.
- `test_agent_viz_api`, `test_codex_ingest`, `test_codex_spawn_executor`, vault index/search tests: green.
- Pre-existing/unrelated failures (not from these changes): `test_real_routing_latency` (LLM perf threshold), `test_real_summary_generation` (local Gemma empty-summary quirk).


nbramia:
Superseded by #288 (fix) and #289 (codex feat) — split into two separate PRs.

PR #294, opened 2026-06-01
fix(agent-worker): reconcile vault task on CLI (claude_code/codex) completion

nbramia:
## Problem

Vault-routed `#agent #claude` and `#agent #codex` tasks were **stranded at `[/]` / `#agent-running` forever** even after the agent finished the work and sent its completion message to Telegram. The vault checkbox never flipped to `[x]` and the tag never advanced to `#agent-completed`.

Observed: every stuck task in the Tasks inbox was `#claude` or `#codex`; every correctly-completed one was `#cloud` or `#local`.

## Root cause

`#agent` tasks reconcile the vault on completion via `_handle_outcome` — but only the **local** and **managed** routes go through it. The **CLI** routes are dispatched by `_dispatch_claude_code_session` / `_dispatch_codex_session`, which **bypass `_handle_outcome` entirely**. On `STATUS_COMPLETED` they send the Telegram message and return — never calling `_complete_task` or swapping `#agent-running` → `#agent-completed`.

These dispatch functions were originally written for **operator-spawned** `/claude` sessions (synthetic task IDs, no backing vault row, so skipping the vault write was correct). Commit #291 (Codex executor + session viz) then **reused them** as the dispatch path for real vault `#agent` CLI tasks — which *do* have a vault row to reconcile — without adding the vault write. So those tasks claim (`[/]` + `#agent-running`), run to completion, notify, and strand.

## Fix

Add `_reconcile_vault_terminal(session, status)`, called on every terminal outcome in both CLI dispatch paths. It mirrors `has_vault_task` in `_handle_outcome`:

- **Operator-spawned** sessions (`origin='operator'`) and **spawned children** (`parent_session_id` set) → no vault row, skipped (calls would 404).
- **Vault-backed** tasks → `COMPLETED`: complete + swap to `#agent-completed`; `FAILED` / `BUDGET_EXCEEDED`: swap to the matching terminal tag + status `cancelled`.

## Tests

- `test_vault_claude_task_marked_complete_on_finish` — vault-routed `#claude` task issues `PUT .../complete` + `#agent-running`→`#agent-completed` swap on completion (regression).
- `test_operator_spawn_does_not_touch_vault_on_finish` — operator-spawned `/claude` sessions issue no vault calls.

Full unit suite: 2544 passed. The 4 failures are pre-existing/environmental (settings defaults driven by this machine's `.env`, vault-file integrity, a flaky slack test that passes in isolation) — all reproduce with this branch's changes stashed and none touch the agent worker.

### DIFF
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -934,6 +934,35 @@ class Worker:
             logger.warning("set_task_status(%s, %s) failed: %s", task_id, status, exc)
             return False
 
+    def _reconcile_vault_terminal(self, session, status: str) -> None:
+        """Update the backing #agent vault task when a CLI session ends.
+
+        The ``claude_code`` / ``codex`` dispatch paths don't go through
+        ``_handle_outcome`` (which owns vault reconciliation for the local
+        and managed routes), so terminal outcomes there must reconcile the
+        vault themselves — otherwise a vault-routed ``#agent #claude`` /
+        ``#codex`` task is stranded at ``[/]`` / ``#agent-running`` forever
+        even though the agent finished.
+
+        Operator-spawned sessions (``origin='operator'``) and spawned
+        children (``parent_session_id`` set) have no backing #agent vault
+        row, so the complete / swap-tag / set-status calls would 404 — skip
+        them. This mirrors ``has_vault_task`` in ``_handle_outcome``.
+        """
+        has_vault_task = session.origin != "operator" and not session.parent_session_id
+        if not has_vault_task:
+            return
+        task_id = session.task_id
+        if status == STATUS_COMPLETED:
+            self._complete_task(task_id)  # mark `done` ([x]) in the vault
+            self._swap_tag(task_id, RUNNING_TAG, COMPLETED_TAG)
+        elif status == STATUS_BUDGET_EXCEEDED:
+            self._swap_tag(task_id, RUNNING_TAG, BUDGET_EXCEEDED_TAG)
+            self._set_task_status(task_id, "cancelled")
+        elif status == STATUS_FAILED:
+            self._swap_tag(task_id, RUNNING_TAG, FAILED_TAG)
+            self._set_task_status(task_id, "cancelled")
+
     # ------------------------------------------------------------------
     # Claim + dispatch
     # ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


nbramia (commented):
## PR Review: fix(agent-worker): reconcile vault task on CLI (claude_code/codex) completion

Reviewed adversarially with correctness + standards specialists. Verified independently: checked out the branch, confirmed `claude_code_executor`/`codex_executor` set `session_store` status terminal *before* returning (so the new method correctly owns only the vault/markdown side, no re-dispatch loop), and ran the test file — **3/3 pass**.

### Action Required
_None._ The fix is correctly scoped and resolves the stranding bug.

### Recommended
- **[Correctness/Standards]** **No-op guard divergence** (`worker.py:955-958` vs `_handle_outcome` `:1505-1525`). `_handle_outcome`'s COMPLETED branch routes "silent give-up" completions (empty `final_text` + no side-effect tool + spend < \$0.05) to the FAILED tag. `_reconcile_vault_terminal` omits this guard and unconditionally marks `#agent-completed`. A `claude_code`/`codex` session that exits 0 with empty output + no side effects is reachable (executor returns `STATUS_COMPLETED` on `returncode == 0 or state.terminal`), so it would be marked complete where local/managed marks it failed. Note: pre-PR these tasks stranded at `#agent-running` forever, so this is *not* a regression — but it falls short of full `_handle_outcome` parity. The cost-gate rationale (managed events-endpoint lag) doesn't apply to CLI, which reads the transcript directly, so porting a lightweight guard is cheap if desired.
- **[Standards]** **Test coverage gap**. The fix wires `_reconcile_vault_terminal` into 4 call sites across both dispatch paths and all three terminal statuses, but tests cover only `claude_code` + `COMPLETED` (plus the operator-skip negative). Untested: the **entire codex path** (`worker.py:1291`, `:1299`) and the **FAILED/BUDGET_EXCEEDED branches** (`:959-964`, distinct `_swap_tag` + `_set_task_status("cancelled")` logic). Cheap to add via the existing `_recording_worker` helper — a codex stub with `routing="codex"` and one FAILED-branch assertion.

### Minor
- **[Correctness]** For an unexpected status value the method silently no-ops, unlike `_handle_outcome` which logs `"unhandled outcome status"` (`:1589`). Currently unreachable (BLOCKED/YIELDED return earlier), but an `else: logger.warning(...)` would match the sibling's defensiveness.
- **[Standards]** Intentional duplication of the vault-mutation logic from `_handle_outcome` is the **right call** — a shared helper would have to re-introduce the no-op guard (wrong for CLI) or grow a flag, both worse than ~9 lines. Flagging only so a future reader doesn't "helpfully" merge them. Naming, docstring, surgical scope, and comments all clean.

### Summary
Merge-ready. The fix correctly targets the root cause (CLI dispatch bypassing `_handle_outcome`), the `has_vault_task` gate faithfully mirrors the original, and session-store ordering is sound. The two Recommended items (no-op guard parity, codex/FAILED-branch test coverage) are non-blocking follow-ups, not defects in the stated fix.

PR #298, opened 2026-06-02
feat(codex): #claude parity — shared context, computer use, skills, gws

nbramia:
## Summary

Brings the `#codex` agent path toward full `#claude` parity across the four items in #297 (the irreducible gpt-5.5-vs-Opus model gap aside): shared agent context + the `gws` CLI, native computer-use documentation, skills exposure, and bidirectional cross-engine delegation.

Closes #297.

## What changed

**Shared context + gws (item 1 + 4)** — Added the `gws` Google Workspace CLI to `capabilities_preamble.py` (reaches codex/managed/local) and root `AGENTS.md` (reaches Claude Code), so both engines know they can shell out to Drive/Gmail/Sheets/Calendar. Don't-degrade-Claude honored: `gws` lives in `AGENTS.md` (both read it); nothing was removed from `CLAUDE.md`.

**Computer use (item 2)** — Codex's native `computer_use`/`browser_use` are stable and on by default; the executor's `--dangerously-bypass-approvals-and-sandbox` doesn't suppress them. Documented availability + the headless-`codex exec` caveat in the setup guide, with delegation (below) as the documented fallback. *Live headless verification is an operator step — flagged as such, not claimed done.*

**Skills (item 3)** — New `codex_skill_sync.py` converts the engine-agnostic LifeOS skills (`standup`, `catchup`, `stale`, `sync-health`, `draft-issue`, `pr-check`, `merge-pr`, `remove-worktree`) from Claude's slash-command dialect (`$ARGUMENTS`, `` !`cmd` ``) into Codex's `SKILL.md` format; `scripts/install_codex_skills.py` installs them into `~/.codex/skills/` (machine-local — Codex has **no** project-level skills dir). The Claude-orchestration skills (`implement`, `review-pr`, `address-review`, `mine-for-ideas`, `tune`) are intentionally left Claude-only.

**Bidirectional routing (item 5)** — `lifeos_agent_spawn` now accepts `model=claude_code|codex` (CLI routes skip the per-token dollar ceiling and reuse the managed concurrency cap). Both CLI executors now inject the agent's LifeOS session id + a delegation hint, so a `#codex` agent can hand browser work to a `claude_code` child (and vice versa) via the existing spawned-session dispatch.

## Test evidence

`~/.venvs/lifeos/bin/python -m pytest tests/ -m unit` → **1741 passed, 0 failed** (613s).

New coverage:
[...]

PR #302, opened 2026-06-02
perf(agent-worker): run spawned CLI children off the tick loop

### LINKED ISSUE #299: perf(agent-worker): CLI spawn children run synchronously in the tick loop
## Problem

`_dispatch_spawned_sessions` (`api/services/agent_worker/worker.py`) runs spawned child sessions **synchronously** inside the worker's poll tick. With #298, an in-flight managed/local agent can now spawn a `claude_code`/`codex` child via `lifeos_agent_spawn` — and a CLI child is a minutes-long subprocess. While it runs, the worker tick is blocked: other `#agent` tasks aren't claimed and other spawned children don't dispatch.

## Context

This is the documented synchronous-MVP behavior (acceptable when local concurrency cap = 1), but #298 made it **newly reachable**: cross-engine delegation means a routine agent task can now park the tick on a long CLI subprocess. Surfaced in the PR #298 review as a non-blocking follow-up.

## Suggested approach

Decouple spawned-CLI-child execution from the tick loop — e.g. run CLI children on a worker thread/executor pool (bounded by the existing per-routing concurrency caps) so the tick keeps claiming and dispatching while a child runs. Mind the existing assumptions: status transitions, vault reconciliation, and the followup-anchor registration all currently happen inline.

## Files involved

- `api/services/agent_worker/worker.py` — `_dispatch_spawned_sessions`, `_dispatch_claude_code_session`, `_dispatch_codex_session`
- `api/services/agent_worker/inter_agent.py` — concurrency caps (already enforced on spawn)

## Acceptance criteria

- [ ] A long-running spawned `claude_code`/`codex` child does not block the worker from claiming new `#agent` tasks or dispatching sibling children.
- [ ] Per-routing concurrency caps still bound the number of concurrent CLI subprocesses.
- [ ] Status transitions, vault reconciliation, and followup-anchor registration remain correct under concurrent execution.

## Related

- Follow-up from #298 (bidirectional codex↔claude routing).

nbramia:
## Summary

Spawned `claude_code`/`codex` children (cross-engine delegation from #298) are long-running subprocesses. `_dispatch_spawned_sessions` ran them inline, so a single delegated browser task could park the worker tick for minutes — blocking new `#agent` claims and sibling dispatch. They now run on a bounded thread pool instead.

Closes #299.

## What changed

- Spawned CLI children dispatch on a bounded `ThreadPoolExecutor` (sized to `2 × agent_max_concurrent_managed`) via a new `_submit_cli_dispatch` helper, instead of running inline in the tick.
- An `_cli_inflight` set (+ lock) guards against a re-scan on the next tick re-submitting a child in the window before it flips `CLAIMED→RUNNING`.
- `stop()` shuts the pool down with `wait=False`; any children left `RUNNING` are reconciled by `resume_pending()` on restart (the existing SIGKILL-safe path).
- The `local` route stays **inline** (in-process, GPU-bound, cap=1 — a pool buys no real parallelism there).

## Thread-safety

- `SessionStore` opens a per-call WAL connection with a timeout → safe across threads.
- `TranscriptStore` writes one JSONL per session → no cross-child contention.
- `httpx.Client` (vault reconcile) and the executors (only per-call local `_RunState`) are safe to use concurrently.

## Scope boundary

`#299` is about **spawned** children. The vault `#claude`/`#codex` **claim path** (`_dispatch`) is still synchronous — pre-existing behavior, not what this issue describes. Can be a follow-up if it proves to matter.

## Test evidence

`pytest tests/ -m unit` → **1749 passed, 0 failed**.

New `tests/test_agent_worker_async_cli_dispatch.py`: CLI children are handed to the pool (not run inline), the in-flight guard prevents double-dispatch across ticks, and `local` still runs inline. The two existing tests that drive `_dispatch_spawned_sessions` now inject a `_SynchronousPool` for deterministic assertions.

## Review focus

- Concurrency correctness of `_submit_cli_dispatch` / the in-flight guard (the CLAIMED→RUNNING race window).
- Pool sizing (`2 × max_concurrent_managed`) and `stop()` shutdown semantics.
- Whether leaving the vault-claim CLI path synchronous is the right scope call.

### DIFF
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -155,7 +155,7 @@ class _SynchronousPool:
         fn(*args, **kwargs)
         return None
 
-    def shutdown(self, wait: bool = True) -> None:  # noqa: D401 - parity with executor API
+    def shutdown(self, wait: bool = True, cancel_futures: bool = False) -> None:  # noqa: D401 - parity with executor API
         pass
 
 
@@ -254,9 +254,13 @@ class Worker:
     def stop(self) -> None:
         """Mark the loop for graceful shutdown. Safe to call from a signal."""
         self._stop = True
-        # Don't block shutdown on in-flight CLI children; any left RUNNING are
-        # reconciled by resume_pending() on the next start (SIGKILL-safe path).
-        self._cli_pool.shutdown(wait=False)
+        # Drop any *queued* CLI dispatches and don't block the caller. Children
+        # already running aren't force-killed here — their own watchdog times
+        # them out, and on a hard stop systemd's TimeoutStopSec SIGKILLs the
+        # process; either way resume_pending() reconciles them on next start.
+        # (Pool threads are non-daemon, so a still-running child can delay a
+        # natural interpreter exit until its watchdog fires.)
+        self._cli_pool.shutdown(wait=False, cancel_futures=True)
 
     def run(self) -> None:
         """Main loop. Runs until `stop()` is called or the process is killed."""
@@ -464,6 +468,13 @@ class Worker:
             # origin='operator').
             if not session.parent_session_id and session.origin != "operator":
                 continue
+            # CLI children already handed to the pool: skip BEFORE draining, so a
+            # re-scan in the CLAIMED→RUNNING window doesn't consume (and discard)
+            # the session's pending messages while the running dispatch skips them.
+            if session.routing in ("claude_code", "codex"):
+                with self._cli_lock:
+                    if session.session_id in self._cli_inflight:
+                        continue
             # The first pending message is the prompt from the parent — drain
             # it so the executor's seeded user turn picks it up.
             pending = self.session_store.drain_pending_messages(session.session_id)
@@ -516,6 +527,9 @@ class Worker:
                 return
             self._cli_inflight.add(sid)
 
+        # The closure captures the tick-time `session` snapshot; the dispatch
+        # only reads immutable fields (ids, routing) and re-reads mutable state
+        # from the store, so the snapshot going stale is harmless.
         def _run() -> None:
             try:
                 dispatch_fn(session, pending)
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -20,7 +20,9 @@ import json
 import logging
 import os
 import signal
+import threading
 import time
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 
 import httpx
@@ -141,6 +143,22 @@ BUDGET_EXCEEDED_TAG = "agent-budget-exceeded"
 AGENT_PICKUP_STATUSES = ("todo", "urgent")
 
 
+class _SynchronousPool:
+    """A ``submit()``-compatible pool that runs work inline.
+
+    The default CLI dispatch pool is a real ``ThreadPoolExecutor``; tests inject
+    this instead so spawned-child dispatch stays deterministic (the work runs
+    before ``submit`` returns, so assertions don't race a background thread).
+    """
+
+    def submit(self, fn, *args, **kwargs):
+        fn(*args, **kwargs)
+        return None
+
+    def shutdown(self, wait: bool = True) -> None:  # noqa: D401 - parity with executor API
+        pass
+
+
 class Worker:
     """Single-process poll loop. One instance per process."""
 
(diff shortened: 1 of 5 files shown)


nbramia (commented):
## PR Review: perf(agent-worker): run spawned CLI children off the tick loop

Verified locally on branch `perf/async-cli-spawn-dispatch`: full agent-worker unit suite (307 passed), new test file + the two touched test files (12 passed). Concurrency model traced end-to-end.

### Recommended
- **[Correctness]** `stop()` shutdown is not actually graceful. `worker.py:259` calls `_cli_pool.shutdown(wait=False)`, but `ThreadPoolExecutor` workers are **non-daemon**, so when `main()` returns after SIGTERM the interpreter still joins any running pool thread. A CLI child can run up to `claude_timeout_seconds` (default **3600s**, `settings.py:592`), so the process won't exit on its own until the subprocess finishes. In production systemd's `TimeoutStopSec` SIGKILLs it first and `resume_pending()` reconciles — so it works — but the `wait=False` framing implies a fast clean exit that doesn't happen. Consider `shutdown(wait=False, cancel_futures=True)` (drops not-yet-started dispatches) and/or terminating in-flight child subprocesses on stop, so shutdown doesn't rely solely on SIGKILL.
- **[Correctness]** Drain-before-guard ordering is fragile. `_dispatch_spawned_sessions` drains pending messages at `worker.py:469` (marking them `delivered=1`) *before* `_submit_cli_dispatch` checks the in-flight guard (`worker.py:514-516`). If a submit is skipped because the session is already in-flight, those just-drained messages are silently discarded. It's not reachable today (the normal spawn drains once on tick 1 and tick 2 drains empty; the follow-up path clears in-flight before re-enqueueing), but the ordering is a latent footgun. Safer to check/claim the in-flight guard before draining, or re-enqueue on a skipped submit.

### Minor
- **[Standards]** The async closure in `_submit_cli_dispatch` captures the `session` snapshot from tick time; by the time the pool thread runs the DB row may have advanced. Harmless here (only immutable fields — `session_id`, `task_id`, fresh-spawn `claude_code_session_id=None` — are used), but worth a one-line comment that the closure intentionally relies only on immutable session fields.
- **[Standards]** `docs/specs/technical/agent-worker.md` still describes dispatch without mentioning the `_cli_pool`; the dispatch section could note CLI children now run off-tick. Not required, but the spec is the documented internal reference.

### Summary
Merge-ready. The core concurrency design is sound: the in-flight guard correctly prevents cross-tick double-dispatch in the CLAIMED→RUNNING window; SessionStore (per-call WAL connection, 10s timeout), TranscriptStore (per-session JSONL, per-call open), and the shared `httpx.Client` are all genuinely safe under concurrent CLI children; the pool is correctly sized to `2 × managed` to match the combined per-routing ceiling; and per-routing caps remain enforced because they gate at `lifeos_agent_spawn` time (counting non-terminal sessions), independent of whether dispatch is sync or async. All three issue #299 acceptance criteria are met. The two Recommended items are robustness/clarity improvements, not correctness blockers given the SIGKILL+resume_pending safety net.

nbramia:

| 1 | `stop()` `shutdown(wait=False)` doesn't prevent interpreter joining non-daemon pool threads → slow exit | Recommended | **Accept** | Add `cancel_futures=True` (drop queued children) + accurate comment that running children aren't force-killed (systemd + resume_pending handle them). |
| 2 | Pending messages drained before the in-flight guard → latent loss if a re-scan skips an in-flight session | Recommended | **Accept** | Check in-flight before draining for CLI routes; real (if currently-unreachable) footgun, cheap to close. |
| 3 | Async closure captures tick-time `session` snapshot | Minor | **Accept** | Add a clarifying comment (only immutable fields used). |
| 4 | Technical spec doesn't mention off-tick `_cli_pool` dispatch | Minor | **Accept** | Add a short note. |

Verification items (caps still enforced at spawn, race window guarded, thread-safety, pool lifecycle) all confirmed — no action. Applying all four directly.

**Findings forwarded:** 4 (all small).

nbramia:

| 1 | `stop()` not truly graceful | Applied | `shutdown(wait=False, cancel_futures=True)` drops queued children; comment now accurately states running children rely on watchdog + systemd + resume_pending. `_SynchronousPool.shutdown` updated for parity. |
| 2 | Drain-before-guard | Applied | In-flight guard checked before draining for CLI routes, so a skipped re-scan can't discard pending messages. |
| 3 | Closure session snapshot | Applied | Added clarifying comment. |
| 4 | Spec didn't mention off-tick dispatch | Applied | Added "Off-tick CLI dispatch (#299)" subsection to specs/technical/agent-worker.md. |

**Tests:** `pytest tests/ -m unit` — **1749 passed, 0 failed**.

nbramia:
Review Round 1: 4 findings (2 Recommended, 2 Minor) all applied; no Action Required. Self-verification confirmed each fix and found nothing new — review loop complete.

PR #304, opened 2026-06-02
feat(chat): escalate to a stronger model when orchestrator wrongly refuses

### LINKED ISSUE #303: feat(chat): escalate to stronger model / Claude Code when orchestrator wrongly refuses
## Problem

The chat orchestrator runs on a single model (Claude Haiku by default). When it wrongly declares something impossible, unavailable, or "not released" — and the user pushes back insisting it should be possible — it has no way to escalate. It just re-asserts the same refusal on the same weak model. The user is stuck arguing with a model that can't be talked out of a stale belief.

## Context

Discovered today: the user asked chat to "add the USA 2026 World Cup games to my personal calendar." Haiku insisted *"FIFA hasn't released the schedule"* and, across **three** explicit "do research" / "you're wrong" pushbacks, kept refusing — even claiming *"my research confirms"* without actually searching (conversations `bc6fb4c5` / `c5098cfb`). The schedule had been public for months; a web search returns it instantly.

Root cause: Haiku trusts its stale training cutoff for current-world facts and won't self-correct under pushback. A prompt tweak (branch `fix/response-tuning-verify-current-facts`) forbids asserting negatives without a web search — a cheap, complementary mitigation — but it's still the same weak model deciding. The robust fix is to **escalate capability** when refusal meets insistence, rather than hoping a better prompt holds.

## Suggested approach

Detect the "model claims impossible/refuses **and** user is pushing back" pattern, then escalate the turn:

- **MVP — bump the model.** Re-run the current turn on a stronger model (Sonnet/Opus) via a new `agent_escalation_model` settings knob. The agentic loop already runs on a configurable `orchestrator_model` (`chat.py` ~1139), so this is mostly a re-dispatch with an override.
- **Heavier option — hand off to Claude Code.** Spawn a `#claude` session (via `claude_code_spawn`) for full tool access + Opus, and report its result back into the chat thread. There's already a `claude_intent` → Claude Code routing path to build on.

Trigger heuristic: reuse the existing `_GIVE_UP_PATTERNS` / `SELF_CORRECTION_NUDGE` detection in `agent_loop.py` (refusal / "can't" / "not available" language) **combined with** a user-pushback signal ("do research", "you're wrong", "it should be possible", or simply repeating the request). Escalate at most once per turn to bound cost; log when it fires.

Directional, not prescriptive — whoever picks this up should decide MVP-first (model bump) vs. the Claude Code handoff, and whether both are worth wiring.

## Files involved

- `api/services/agent_loop.py` — escalation trigger (extend give-up/self-correction detection) + re-dispatch on a stronger model
- `api/routes/chat.py` — orchestrator model selection / turn re-run wiring
- `api/services/agent_system_prompt.py` — any prompt-side coordination with escalation
- `api/services/agent_worker/claude_code_spawn.py` — the Claude Code handoff path (heavier option)
- `config/settings.py` — `agent_escalation_model` knob (and any enable flag)

## Acceptance criteria

- [ ] When the orchestrator emits a refusal/impossibility response and the user pushes back, the next turn is handled by a stronger model (or a Claude Code session) rather than re-refused by the same model.
- [ ] Escalation is bounded (at most once per turn) and logged/observable (perf trace or transcript event).
- [ ] The escalation model/behavior is configurable via `config/settings.py` and off-by-default-safe for a fresh clone (no hard dependency that breaks without the knob set).
- [ ] A regression test reproduces the "refuse → user insists → escalate" path with a stubbed strong model and asserts the escalation fired.
- [ ] Does not degrade normal turns (no escalation when there's no refusal+pushback).

## Related

- Complementary to branch `fix/response-tuning-verify-current-facts` (prompt-only mitigation for the same failure class).
- Surfaced while diagnosing the World Cup chat failure (`/tune` session).

nbramia:
## Summary

When the prior assistant turn refused or claimed something impossible ("FIFA hasn't released the schedule") and the user's next message pushes back ("do research", "you're wrong", "look it up"), the chat orchestrator now retries that turn on a stronger model instead of re-refusing on the same weak one.

Closes #303.

## What changed

- **`agent_loop.py`** — `should_escalate(history, question)` detects the prior-assistant-refusal + current-pushback pattern (`_REFUSAL_PATTERNS` / `_PUSHBACK_PATTERNS`); `resolve_orchestrator_model(...)` returns `(model, escalated)`. `_select_client(model)` builds a dedicated `AnthropicLLMClient(model=…)` for an escalated turn.
- **`chat.py`** — gates on the Anthropic backend + `LIFEOS_AGENT_ESCALATION_MODEL`, resolves the per-turn model, passes it to the loop **only when escalated** (normal turns keep the shared singleton), and surfaces escalation in the routing reason, perf trace, and a log line.
- **`settings.py`** — `agent_escalation_model` (empty default → disabled).

## Why a real client swap (not `model_tier`)

While implementing I found `model_tier` is **vestigial** — `run_agent_loop` uses a singleton `get_local_llm()` client and never passes `model_tier` to it, so bumping it would have been a no-op. Escalation therefore constructs a dedicated `AnthropicLLMClient(model=escalation_model)` for the turn. (Cleaning up the dead `model_tier` param is out of scope here.)

## Safety / scope

- **Off by default**: empty `agent_escalation_model` → no escalation. Safe for fresh clones and the **local backend** (which ignores a per-turn model entirely).
- **Bounded**: one escalation decision per turn; normal turns are untouched (still the singleton).
- **Observable**: routing reason ("escalated to …"), `perf_trace.model_tier`, and a log line.

## Test evidence

`pytest tests/ -m unit` → **1768 passed, 0 failed**.

New `tests/test_agent_escalation.py`: the refuse→pushback trigger (and its negatives — no prior refusal, no pushback, empty history, user-text-isn't-assistant, most-recent-turn-wins, dict-shaped messages, varied pushback phrasings), the model decision (escalates only when configured + differs + triggered), and `_select_client` (escalation model on Anthropic, singleton otherwise, ignored on local).

## Review focus

- The `_REFUSAL_PATTERNS` / `_PUSHBACK_PATTERNS` heuristics — false-positive risk (escalating when it shouldn't) vs. false-negatives. They only fire when the *prior assistant* turn matches refusal AND the *current* user msg matches pushback.
- Relationship to the prompt-only mitigation on `fix/response-tuning-verify-current-facts` — complementary, not conflicting.

### DIFF
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -1109,7 +1109,7 @@ async def ask_stream(request: AskStreamRequest):
             # =============================================================
             # Agentic synthesis path: Claude decides what to fetch
             # =============================================================
-            from api.services.agent_loop import run_agent_loop
+            from api.services.agent_loop import run_agent_loop, resolve_orchestrator_model
 
             # Expand follow-up queries with conversation context
             with trace_span("query_expand"):
@@ -1137,6 +1137,21 @@ async def ask_stream(request: AskStreamRequest):
             # in perf_traces.db is `model_tier` — but the value is now a model id
             # (e.g. "claude-haiku-4-5"), not a tier label ("haiku"/"sonnet"/"opus").
             orchestrator_model = getattr(settings, "anthropic_model", "claude-haiku-4-5")
+            # Escalation (#303): if the prior turn refused/claimed-impossible and
+            # this message pushes back, retry on a stronger model. Only on the
+            # Anthropic backend, only when LIFEOS_AGENT_ESCALATION_MODEL is set.
+            escalation_model = (
+                getattr(settings, "agent_escalation_model", "") or ""
+                if getattr(settings, "llm_backend", "anthropic").lower() == "anthropic"
+                else ""
+            )
+            orchestrator_model, escalated = resolve_orchestrator_model(
+                conversation_history, request.question, orchestrator_model, escalation_model
+            )
+            if escalated:
+                logger.info(
+                    "escalating chat turn to %s (refusal+pushback detected)", orchestrator_model
+                )
             _trace = _current_trace.get()
             if _trace:
                 _trace.model_tier = orchestrator_model
@@ -1158,7 +1173,11 @@ async def ask_stream(request: AskStreamRequest):
                     for att in request.attachments
                 ]
 
-            yield f"data: {json.dumps({'type': 'routing', 'sources': ['agent'], 'reasoning': f'Agentic loop ({orchestrator_model})', 'latency_ms': 0})}\n\n"
+            _routing_reason = (
+                f'Agentic loop — escalated to {orchestrator_model}'
+                if escalated else f'Agentic loop ({orchestrator_model})'
+            )
+            yield f"data: {json.dumps({'type': 'routing', 'sources': ['agent'], 'reasoning': _routing_reason, 'latency_ms': 0})}\n\n"
 
             # Consume the async generator from the agent loop
             agent_result = None
@@ -1168,6 +1187,7 @@ async def ask_stream(request: AskStreamRequest):
                 attachments=attachments_for_api,
                 model_tier=orchestrator_model,
                 max_tool_rounds=5,
+                model=orchestrator_model if escalated else "",
             ):
                 if event["type"] == "text":
                     yield f"data: {json.dumps({'type': 'content', 'content': event['content']})}\n\n"
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -64,17 +64,21 @@ def _looks_like_giving_up(text: str) -> bool:
 # impossible / unavailable / "not released" from stale training and won't budge.
 # When the PRIOR assistant turn refused like that AND the user's NEW message pushes
 # back, we retry the turn on a stronger model rather than re-refusing.
+# Scoped to the *stale-knowledge / world-fact* refusal class — claims that
+# something about the current world isn't released/announced/scheduled/known.
+# A stronger model + web search fixes these. Deliberately NOT matching generic
+# data-lookup negatives ("I can't find any emails from Sarah", "no such
+# contact", "that file doesn't exist") — those are usually correct, and a
+# pricier model can't find data that isn't there. The give-up patterns
+# (knowledge-cutoff / can't-access-live) are also treated as refusals (see
+# should_escalate).
 _REFUSAL_PATTERNS = re.compile(
     r"(?i)("
-    r"hasn'?t (yet )?(been )?(released|announced|published|come out|happened|finalized)"
-    r"|haven'?t (yet )?(been )?(released|announced|published)"
-    r"|not (yet )?(been )?(released|announced|available|published|out|finalized)"
-    r"|isn'?t (yet )?(available|out|released)|aren'?t (yet )?available"
-    r"|hasn'?t been (set|scheduled|determined)"
-    r"|doesn'?t exist|does not exist|no such"
-    r"|not possible|isn'?t possible|impossible"
-    r"|I can'?t (add|create|do|find|provide)|I cannot (add|create|do|find|provide)"
-    r"|I'?m unable to|I am unable to"
+    r"(hasn'?t|has not|haven'?t|have not)\s+(yet\s+)?(been\s+)?"
+    r"(released|announced|published|scheduled|finalized|determined|set|made public|come out)"
+    r"|not\s+(yet\s+)?(been\s+)?(released|announced|published|scheduled|finalized|determined|available|out)"
+    r"|isn'?t\s+(yet\s+)?(available|out|released|published|finalized)"
+    r"|aren'?t\s+(yet\s+)?(available|released|published)"
     r")"
 )
 _PUSHBACK_PATTERNS = re.compile(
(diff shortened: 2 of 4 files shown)


nbramia (commented):
## PR Review: feat(chat): escalate to a stronger model when orchestrator wrongly refuses

Verified independently: checked out the branch, ran the new suite (19 passed) plus `test_chat_helpers` (32 passed total), confirmed `model_tier` is vestigial, confirmed the `conversation_history` strip in chat.py, and empirically probed the heuristics for false-positive / false-negative behavior.

### Recommended
- **[Correctness]** False-positive escalations on *correct* refusals + pushback (`agent_loop.py:67-90`). `_REFUSAL_PATTERNS` matches many legitimately-correct assistant turns ("I can't find any emails from Sarah", "There's no such contact named Bob", "that file doesn't exist", "the calendar isn't shared"). When the assistant was *right* and the user pushes back ("look it up again", "try again", "you're wrong"), `should_escalate` fires and we pay for a pricier model on a turn that will just re-confirm the same true negative. This is bounded (one escalation/turn, off-by-default) so it's not blocking, but the trigger is broader than the World-Cup case it targets. Worth a follow-up to narrow refusal patterns toward *world-fact* claims (released/announced/published/scheduled) and drop the generic "can't find / no such / doesn't exist" data-lookup phrasings, which are normal correct answers.

### Minor
- **[Correctness]** False negatives on real refusals the regex shape misses (`agent_loop.py:67-79`). Empirically: "that information is not in my training data", "the bracket has not been determined" (the `determined` alternative only fires after `hasn't been`, not after `has not been`), and "that data simply is not something I can retrieve" all fail to match while being genuine stale-knowledge refusals. Acceptable for an MVP heuristic, but the asymmetry (over-fires on benign, under-fires on some real ones) is worth noting.
- **[Standards]** `model_tier=orchestrator_model` is still passed to `run_agent_loop` (`chat.py:1190`) even though the param is confirmed dead (only referenced at `agent_loop.py:155,168`, never forwarded to `client.astream`). The PR's vestigial claim checks out. Passing a model-id into a param documented as a tier label ("haiku"/"sonnet"/"opus") is a harmless pre-existing inconsistency; cleanup is reasonably scoped out, but a one-line `# noqa`-style comment or dropping the arg would be cleaner.

### Verified / non-issues
- **Gating is correct.** `resolve_orchestrator_model` is a true no-op when unconfigured (`("base", False)`), when `escalation_model == base_model`, and when not triggered. `_select_client` ignores the per-turn model on the local backend (returns the singleton) and chat.py zeroes `escalation_model` off the Anthropic branch. Confirmed by execution.
- **`model_tier` is genuinely vestigial** — referenced only in signature/docstring, never reaches the client. The real-client-swap approach is justified.
- **`_select_client` swap is sound** — escalation builds a dedicated `AnthropicLLMClient(model=...)`; normal turns keep the shared singleton. `AnthropicLLMClient.__init__` accepts `model` and stores `_model` (llm_client.py:423-429); test asserts `client._model` correctly.
- **conversation_history scan is correct.** chat.py:1008-1011 strips the current user message before the loop, so `_last_assistant_text` (most-recent-assistant) reads the *prior* turn, not the in-flight one. `Message.content` is typed `str`, so the regex is safe against content-block lists.
- **Cost/safety** — escalation is off-by-default, bounded to one decision per turn, and observable (routing reason + `perf_trace.model_tier` + log line). Meets all five acceptance criteria in #303 (MVP model-bump path).

### Summary
Merge-ready. The escalation is correctly gated, well-tested, and safe by default; the only substantive note is that the refusal heuristic over-fires on correct data-lookup negatives — a bounded cost issue worth a follow-up to narrow patterns, not a blocker.

nbramia:

| 1 | `_REFUSAL_PATTERNS` over-fires on *correct* data-lookup negatives ("I can't find any emails from Sarah", "no such contact", "doesn't exist") → escalates to pricier model to re-confirm a true negative | Recommended | **Accept** | A stronger model can't find data that isn't there — escalation should target the *stale-knowledge / world-fact* class only. Narrow the patterns to release/announce/publish/schedule/finalize/determine negatives + fold in the existing `_GIVE_UP_PATTERNS` (knowledge-cutoff/can't-access-live); drop the generic find/add/create/no-such/doesn't-exist phrasings. |
| 2 | False negatives: "has not been determined" (only matched after "hasn't been"), "not in my training data" | Minor | **Accept** | Fold into the same edit — handle "has not"/"hasn't" both, and `_GIVE_UP_PATTERNS` covers "training data". |
| 3 | Vestigial `model_tier` still passed at chat.py | Minor | **Reject** | Pre-existing dead param; removing it changes the signature + all callers — separate cleanup, out of scope here. |

**Findings forwarded:** 1 (narrow refusal patterns, folding in the false-negative fix). Applying directly.

nbramia:

| 1 | Refusal patterns over-fired on correct data-lookup negatives | Applied | Narrowed `_REFUSAL_PATTERNS` to the stale-knowledge/world-fact class; folded `_GIVE_UP_PATTERNS` into the refusal check; dropped generic find/add/no-such/doesn't-exist/impossible phrasings. |
| 2 | False negatives ("has not been determined") | Applied | Patterns now handle "has not"/"hasn't" both; give-up patterns cover "training data". |
| 3 | Vestigial `model_tier` still passed | Rejected | Pre-existing dead param; separate cleanup, out of scope. |

Added tests: correct data-lookup negatives ("I can't find any emails from Sarah", "no such contact", "that file doesn't exist") do NOT escalate; give-up/knowledge-cutoff phrases do.

**Tests:** `pytest tests/ -m unit` — **1775 passed, 0 failed**.

nbramia:
Review Round 1: 1 Recommended finding applied (narrowed refusal patterns), 2 Minor handled/deferred; no Action Required. Self-verification confirmed the narrowing and found nothing new — review loop complete.

Issue #305, opened 2026-06-03
feat(chat): user-directed + multi-tier orchestrator escalation


nbramia:
## Problem

#304 added automatic escalation: when the orchestrator makes a stale-knowledge refusal and the user pushes back, the turn retries on a single configured stronger model (`LIFEOS_AGENT_ESCALATION_MODEL`). Two things are missing:

1. **No way for the user to explicitly direct escalation.** If the user says "escalate to opus", "use sonnet", or "use codex", the system ignores the directive — it can only auto-escalate to the one configured model.
2. **Only one escalation rung.** There's no laddering (sonnet → opus → hand off to a CLI engine) and no way to reach the more capable CLI engines (Codex / Claude Code) from a chat turn.

## Context

Builds directly on #304, which already provides the per-turn `model` override (`run_agent_loop(model=…)`), `resolve_orchestrator_model`, and `_select_client`. This issue feeds those seams from new triggers (explicit user directive; a multi-rung ladder) and adds the cross-engine handoff.

Discovered while enabling #304 — the operator asked "can I just say 'escalate to opus/sonnet/codex'?"

## Suggested approach

Three layers, increasing scope:

**(a) User-directed model escalation — MVP, low-risk.** Parse the user's message for an explicit directive and honor it, bypassing the refusal+pushback heuristic (explicit intent = zero false-positive risk):
- "escalate to opus" / "use opus" → `claude-opus-4-8`
- "use sonnet" / "try with sonnet" → `claude-sonnet-4-6`
- "escalate" / "use a smarter model" (no model named) → the configured `agent_escalation_model`

This is another input to `resolve_orchestrator_model`, reusing the existing `model` swap.

**(b) Engine handoff ("use codex" / "use claude code").** Heavier: Codex/Claude Code run as **async worker sessions** (via `codex_spawn` / `claude_code_spawn`, from #298), not inline model swaps. "use codex" would spawn a `#codex` session and report its result back into the chat thread — a different (async) UX than a same-turn model bump. Reuses the existing `claude_intent` → Claude Code routing.

**(c) Multi-tier auto-escalation.** If the *escalated* model also refuses under continued pushback, climb another rung (e.g. sonnet → opus → engine handoff), driven by an ordered escalation ladder in settings. Best as a follow-up once (a)/(b) land.

Recommend: implement (a) now; design (b)/(c) in the same issue and land them as follow-up PRs.

## Files involved

[...]

PR #306, opened 2026-06-03
feat(chat): user-directed model escalation (escalate to opus/sonnet)

nbramia:
## Summary

Implements **part (a)** of #305: the chat orchestrator now honors an explicit model directive in the user's message ("escalate to opus", "use sonnet"), taking precedence over the #304 refusal+pushback auto-heuristic.

Relates to #305 (parts b/c deferred — see below).

## What changed

- **`agent_loop.py`** — `_MODEL_ALIASES` (opus/sonnet/haiku → model ids), `_parse_escalation_directive(question, escalation_model)`, and `resolve_orchestrator_model` now checks the directive **first**, then the auto-heuristic.
  - A named tier ("use opus", "with claude opus", "switch to sonnet", "retry on haiku") → that model, **even when auto-escalation is unconfigured** (explicit intent, zero false-positive risk).
  - A generic "escalate" / "use a smarter|stronger|better model" → falls back to `agent_escalation_model`.
- **`chat.py`** — the whole escalation resolution is now gated to the Anthropic backend, so a directive can't mislabel on the local backend (which ignores a per-turn model).

## Deferred (per the issue)

- **(b)** engine handoff ("use codex" / "use claude code") — those run as async worker sessions (`codex_spawn`/`claude_code_spawn`), a different UX than an inline model swap.
- **(c)** multi-tier auto-escalation ladder (sonnet → opus → engine).

## Test evidence

`pytest tests/ -m unit` → **1793 passed, 0 failed**.

New `tests/test_agent_escalation.py` cases: each named tier × directive verb; the generic-smarter fallback (and its no-op when `escalation_model` is unset); directive-to-base no-op; directive beats the heuristic with no refusal in history; and non-directive mentions ("summarize my notes about the opus project", "find the sonnet I wrote") that must **not** escalate.

## Review focus

- `_DIRECTIVE_MODEL_RE` false-positive surface — it requires a directive verb (escalate/use/with/switch to/retry on/try with) immediately before the tier word. Known thin edge: "use opus-level detail" would match (very unlikely in a LifeOS chat). False-negative risk on unusual phrasings.
- `_MODEL_ALIASES["opus"] = "claude-opus-4-8"` — update if the account is pinned to a different opus release.
- Precedence + the Anthropic-backend gate in `chat.py`.

### DIFF
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -1137,21 +1137,18 @@ async def ask_stream(request: AskStreamRequest):
             # in perf_traces.db is `model_tier` — but the value is now a model id
             # (e.g. "claude-haiku-4-5"), not a tier label ("haiku"/"sonnet"/"opus").
             orchestrator_model = getattr(settings, "anthropic_model", "claude-haiku-4-5")
-            # Escalation (#303): if the prior turn refused/claimed-impossible and
-            # this message pushes back, retry on a stronger model. Only on the
-            # Anthropic backend, only when LIFEOS_AGENT_ESCALATION_MODEL is set.
-            escalation_model = (
-                getattr(settings, "agent_escalation_model", "") or ""
-                if getattr(settings, "llm_backend", "anthropic").lower() == "anthropic"
-                else ""
-            )
-            orchestrator_model, escalated = resolve_orchestrator_model(
-                conversation_history, request.question, orchestrator_model, escalation_model
-            )
-            if escalated:
-                logger.info(
-                    "escalating chat turn to %s (refusal+pushback detected)", orchestrator_model
+            # Escalation: pick a stronger model for this turn either because the
+            # user explicitly asked ("escalate to opus", #305) or because the
+            # prior turn refused and this message pushes back (#303). Anthropic
+            # backend only — the local backend can't honor a per-turn model.
+            escalated = False
+            if getattr(settings, "llm_backend", "anthropic").lower() == "anthropic":
+                escalation_model = getattr(settings, "agent_escalation_model", "") or ""
+                orchestrator_model, escalated = resolve_orchestrator_model(
+                    conversation_history, request.question, orchestrator_model, escalation_model
                 )
+            if escalated:
+                logger.info("escalating chat turn to %s (user-directed or refusal+pushback)", orchestrator_model)
             _trace = _current_trace.get()
             if _trace:
                 _trace.model_tier = orchestrator_model
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -121,15 +121,62 @@ def should_escalate(conversation_history, question: str) -> bool:
     return bool(_PUSHBACK_PATTERNS.search(question or ""))
 
 
+# User-directed escalation (#305). Tier words the operator can name in a chat
+# message map to concrete Anthropic model ids. Update the opus id here if the
+# account is pinned to a different opus release.
+_MODEL_ALIASES = {
+    "haiku": "claude-haiku-4-5",
+    "sonnet": "claude-sonnet-4-6",
+    "opus": "claude-opus-4-8",
+}
+# A directive verb immediately followed by a tier word: "escalate to opus",
+# "use sonnet", "with claude opus", "switch to haiku", "retry on opus".
+_DIRECTIVE_MODEL_RE = re.compile(
+    r"(?i)\b(?:escalate(?:\s+to)?|use|using|with|switch\s+to|retry\s+(?:with|on)|try\s+(?:with|on))\s+"
+    r"(?:claude\s+)?(opus|sonnet|haiku)\b"
+)
+# A generic "use a stronger model" directive with no tier named → fall back to
+# the configured escalation model. Bare "escalate" alone also counts.
+_DIRECTIVE_SMARTER_RE = re.compile(
+    r"(?i)("
+    r"\bescalate\b"
+    r"|(?:use|using|try|switch\s+to)\s+(?:a\s+)?(?:smarter|stronger|better|bigger|more\s+capable)\s+(?:model|llm)"
+    r")"
+)
+
+
+def _parse_escalation_directive(question: str, escalation_model: str) -> str:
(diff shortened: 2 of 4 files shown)


nbramia (commented):
## PR Review: feat(chat): user-directed model escalation (escalate to opus/sonnet)

Verified by checking out the branch, reading `agent_loop.py` / `chat.py` / `pricing.py`, and probing the two regexes against ~30 phrasings. `tests/test_agent_escalation.py` → 44 passed; escalation + chat-route unit subset → 44 passed.

### Action Required
- **[Correctness]** Negation and meta-question false positives in `_DIRECTIVE_MODEL_RE` (`agent_loop.py:134`). The pattern only requires `verb + (claude )? + tier`, with no guard for negation or questions, so all of these silently escalate the turn:
  - `"don't use opus, stick with haiku"` → opus
  - `"can you use opus or sonnet?"` → opus
  - `"why did you use sonnet?"` → sonnet
  - `"I didn't ask you to use opus"` → opus
  These are plausible LifeOS chat utterances (asking *about* models, or declining one), and each one swaps the model + flips `escalated=True`, changing cost and the routing reason. The author flagged `"use opus-level detail"` as the known thin edge, but the negation/question class is more likely and more surprising. Recommend either anchoring on a tighter intent (e.g. require the directive at message start, or reject when preceded by `don't|do not|didn't|n't|why|whether|or <tier>`), or accept-with-test — but right now there is no negative test covering negation/question phrasings.

### Recommended
- **[Correctness]** `"escalate to codex"` resolves to the configured `escalation_model` (sonnet), not a no-op (`agent_loop.py:140` `_DIRECTIVE_SMARTER_RE` matches bare `\bescalate\b`). Per #305 the engine-handoff path (b) is deferred, so a user reaching for an engine LifeOS can't yet honor gets a silent substitution to a different model rather than a clean no-op or "not supported." Low-frequency, but it's a direct interaction with the deferred (b) scope and worth a guard or at least a noted limitation.
- **[Correctness/Cost]** `claude-opus-4-8` is absent from the pricing table (`api/services/agent_worker/pricing.py:21-27` tops out at `claude-opus-4-6/4-7`). For chat-turn escalation to opus, `cost_for` falls through to the opus-4-7 fallback (`pricing.py:64-66`) — same $15/$75 rates, so cost is correct by accident, not by entry. If 4-8 ever priced differently this silently mis-accounts. Adding the explicit `claude-opus-4-8` row removes the coincidence. (The opus id itself is consistent with the `settings.py:178` docstring and the running account's opus family, so it's plausibly valid — just not in the cost table.)

### Minor
- **[Standards]** `_DIRECTIVE_SMARTER_RE` matches `escalate` anywhere in the message, so `"escalate this ticket to the team"` would fire if `escalation_model` is configured (returns the model; with it unset it correctly no-ops). Edge, but no test covers a non-model-swap use of the word "escalate."

### Verified Correct (no action)
- **Precedence** — directive beats auto-heuristic (`resolve_orchestrator_model` checks `_parse_escalation_directive` first, `agent_loop.py:177-179`); named tier works with `escalation_model=""`; directive-to-base is a no-op (`directed != base_model` guard). All confirmed by probe + tests.
- **Anthropic-backend gate** (`chat.py:1145`) — on the local backend the whole resolution is skipped, `escalated` stays False, no mislabel in routing reason or `run_agent_loop(model=...)`. Correct.
- **Generic-smarter fallback** — `"use a smarter|stronger|better|more capable model"` → `escalation_model`, and no-ops to base when unset. Confirmed.

### Summary
Merge-ready in mechanics and precedence (gate, fallback, no-op all correct, tests green), but the directive regex has a real negation/question false-positive surface that isn't tested — recommend tightening the pattern or adding negative-case coverage before merge.

nbramia:

| 1 | `_DIRECTIVE_MODEL_RE` fires on negation/meta-questions ("don't use opus", "why did you use sonnet?", "I didn't ask you to use opus", "can you use opus?") | Action Required | **Accept** | Reproduced all four. Add a negation guard (don't/didn't/never/instead of/… before the directive) + a meta-question guard (why/can/could/should/would … use), with negative tests. |
| 2 | "escalate to codex" silently substitutes the escalation model (sonnet) — engine handoff is deferred (b) | Recommended | **Accept** | Confirmed. Add an unsupported-engine guard (codex/claude code/gpt/gemini) → no-op, so a deferred engine never silently becomes a different model. |
| 3 | `claude-opus-4-8` absent from `pricing.py` (falls through to opus-4-7 fallback) | Recommended | **Accept** | Add the explicit row (same $15/$75 rates) so cost accounting is correct by entry, not coincidence. |
| 4 | Bare `\bescalate\b` matches "escalate this ticket to the team" | Minor | **Accept** | Tighten `_DIRECTIVE_SMARTER_RE` to require a model reference ("a smarter/stronger model", "escalate the model"); drop bare "escalate". Update the fallback test + add the negative. |

**Findings forwarded:** 4 (all accepted — applying directly).

nbramia:

| 1 | Negation/meta-question false positives | Applied | Added a negation guard (don't/didn't/never/… before the directive) + a meta-question guard (why/can/should/… you|i|we use). All four reviewer probes now no-op. |
| 2 | "escalate to codex" silently → sonnet | Applied | Unsupported-engine guard (codex/claude code/gpt/gemini) → no model swap. |
| 3 | opus-4-8 missing from pricing | Applied | Added the explicit `claude-opus-4-8` row ($15/$75). |
| 4 | bare "escalate" too broad | Applied | Generic directive now requires a model reference; "escalate this ticket to the team" no longer fires. |

Bonus: removed "instead of"/"rather than" from the negation set (they contrast options — "use sonnet instead of opus" correctly stays a sonnet directive; added a test).

**Tests:** `pytest tests/ -m unit` — **1803 passed, 0 failed**.

nbramia:
Review Round 1: all 4 findings applied (negation/question/engine guards + tightened generic directive + pricing row); no Action Required outstanding. Self-verification re-probed every false positive (now no-op) and confirmed legit directives still fire — review loop complete.

PR #307, opened 2026-06-03
feat(chat): engine handoff — use codex / use claude code

nbramia:
## Summary

Implements **part (b)** of #305. When a chat message explicitly names a CLI engine ("use codex", "use claude code"), the orchestrator hands the task off to that async worker session instead of answering inline — mirroring the existing `claude_intent` → Claude Code path, but choosing the engine.

Relates to #305 (part c — multi-tier ladder — still deferred).

## What changed

- **`agent_loop.py`** — `parse_engine_directive(question) → (engine, task)`. Detects "use codex" / "use claude code" / "with codex" / "hand this to codex" / "run it with claude code", strips the directive phrase to leave the actual task, and reuses the negation/meta-question guards from #306 ("don't use codex" and "why use codex?" no-op).
- **`chat.py`** — an explicit engine directive precedes intent classification and the agentic loop; it emits a `claude_intent` SSE event carrying `engine` + the cleaned task, then returns. The existing `claude_intent` emission now also carries `engine: claude_code`.
- **`telegram.py`** — `chat_via_api` captures the `engine`; the handler routes to `_handle_codex_command` vs `_handle_claude_command`. The spawned session reports back through the originating chat (`chat_id`), exactly as `/codex` and `/claude` already do.

## Surface support

Follows the **existing `claude_intent` path**: it works on the **Telegram** surface (and any surface that consumes the event). The web `/chat` frontend doesn't consume `claude_intent` today — a pre-existing limitation this inherits, not a regression.

## Test evidence

`pytest tests/ -m unit` → **1815 passed**, plus the 13 new engine-directive tests. One unrelated `test_scheduler_watcher` end-to-end test flaked on the full run and **passes on isolated rerun** (filesystem-watcher timing; my diff doesn't touch the scheduler).

New tests: directive routing + task cleaning across forms; codex vs claude_code; the negation/question/non-engine negatives ("don't use codex", "why use codex?", "summarize my codex notes", "use opus"); and the no-task fallback.

## Review focus

- `parse_engine_directive` false-positive/negative surface (does a normal query route to an engine? does the task-cleaning ever drop the real ask?).
- The chat.py precedence: engine handoff before model escalation (#306) and before the agentic loop — correct ordering.
- The `task` cleaning fallback when the directive is the whole message.

### DIFF
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -1074,6 +1074,20 @@ async def ask_stream(request: AskStreamRequest):
                     yield f"data: {json.dumps({'type': 'done'})}\n\n"
                     return
 
+            # Explicit engine handoff (#305 part b): the user named a CLI engine
+            # ("use codex", "use claude code"). Hand the task off to that worker
+            # session rather than answering inline — the surface (Telegram) spawns
+            # it and reports back. Explicit intent, so it precedes classification
+            # and the agentic loop.
+            from api.services.agent_loop import parse_engine_directive
+            _engine, _engine_task = parse_engine_directive(request.question)
+            if _engine:
+                _label = "Codex" if _engine == "codex" else "Claude Code"
+                yield f"data: {json.dumps({'type': 'routing', 'sources': [_engine], 'reasoning': f'User-directed handoff to {_label}', 'latency_ms': 0})}\n\n"
+                yield f"data: {json.dumps({'type': 'claude_intent', 'task': _engine_task, 'engine': _engine})}\n\n"
+                yield f"data: {json.dumps({'type': 'done'})}\n\n"
+                return
+
             # Intent classification for special dispatch cases only.
             # Compose, task, and reminder intents now flow through the agentic
             # loop which has dedicated tools (create_email_draft, manage_tasks,
@@ -1097,7 +1111,7 @@ async def ask_stream(request: AskStreamRequest):
                 # ---- CLAUDE CODE: requires terminal/filesystem/browser ----
                 print("DETECTED CLAUDE INTENT - delegating to Claude Code")
                 yield f"data: {json.dumps({'type': 'routing', 'sources': ['claude_code'], 'reasoning': 'Action requires terminal/filesystem/browser access', 'latency_ms': 0})}\n\n"
-                yield f"data: {json.dumps({'type': 'claude_intent', 'task': request.question})}\n\n"
+                yield f"data: {json.dumps({'type': 'claude_intent', 'task': request.question, 'engine': 'claude_code'})}\n\n"
                 yield f"data: {json.dumps({'type': 'done'})}\n\n"
                 return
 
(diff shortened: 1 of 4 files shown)


nbramia:

| 1 | `parse_engine_directive` false-positives on mid-sentence mentions ("remind me to use codex tomorrow", "what time do I usually use codex") → spawns a real worker subprocess | Action Required | **Accept** | Verified. Anchor the directive to message start (imperative position) — incidental mentions no longer route. |
| 2 | Task-mangling: mid-sentence excision leaves fragments ("what time do I usually   at night") | Action Required | **Accept** | Resolved by the same start-anchor — the directive is always at position 0, so stripping leaves a clean trailing task (no interior excision). |
| 3 | "use codex with opus" → task "with opus" (model token leaks into the spawned prompt) | Recommended | **Accept** | Strip a trailing model phrase ("with/using opus|sonnet|haiku") from the cleaned task. |
| 4 | Polite imperatives ("can you use codex to fix") no-op via the meta-question guard | Recommended | **Reject** | Intended conservative bias — for an action that spawns a subprocess, require an imperative ("use codex to …"). The reviewer agreed this is acceptable once the false-positive surface is fixed (it is). |
| 5 | Inline `from … import parse_engine_directive` in chat.py | Minor | **Reject** | Matches local style (`resolve_orchestrator_model` is imported the same way nearby). |
| 6 | `_UNSUPPORTED_ENGINE_RE` comment stale (codex/claude code now wired) | Minor | **Accept** | Update the comment. |

Plus: update the tests — the trailing "<task> using codex" form now no-ops (moved to negatives), and add the reviewer's false-positive probes as negative cases.

**Findings forwarded:** 4 (1, 2, 3, 6).

nbramia:

| 1 | Mid-sentence false positives spawn a subprocess | Applied | Start-anchored the directive regex — only imperative-leading "use codex …" routes. All five reviewer probes (incl. the reminder hijack) now no-op. |
| 2 | Interior task-mangling | Applied | Resolved by the anchor — directive is always at position 0, so the cleaned task is the clean trailing remainder (no excision artifacts). |
| 3 | "use codex with opus" leaks "with opus" into the task | Applied | Added a trailing-model strip. |
| 6 | Stale `_UNSUPPORTED_ENGINE_RE` comment | Applied | Updated. |

Rejected: #4 (polite-question no-op — intended conservative bias for a subprocess action) and #5 (inline import — matches local style).

Intended tradeoff: the trailing "<task> using codex" form and polite questions no longer route — an imperative is required.

**Tests:** `pytest tests/ -m unit` — **1823 passed, 0 failed**. New negatives cover every false-positive probe.

nbramia:
Review Round 1: 2 Action Required + 2 others applied (start-anchored the directive, trailing-model strip, comment); 2 rejected as intended. Self-verification re-probed every false positive (now no-op) and confirmed imperatives still route — review loop complete.

PR #314, opened 2026-06-04
fix(embeddings): serialize lazy model loads to fix meta-tensor race

nbramia:
## Summary

Fixes a **CRITICAL** intermittent crash that poisoned `search_vault` and silently disabled re-ranking: `NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty()…`.

Root cause: search tools run via `execute_tool_parallel → asyncio.to_thread`, so the embedding model (and cross-encoder re-ranker) — loaded lazily on first use — could be constructed by **multiple tool threads concurrently**. The racing meta-device init leaves params on the meta device, which surfaces when `SentenceTransformer.__init__` calls `self.to(device)`. The code's own docstring already anticipated this ("if a future caller starts triggering loads from multiple threads at once, wrap this in a lock").

## What changed

- **`api/services/embeddings.py`** — the lazy `model` property now loads under a `threading.Lock` with double-checked locking; the construction logic moved into `_load_model()`. Cost/health/CPU-fallback behavior is unchanged — just serialized.
- **`api/services/reranker.py`** — same load-lock for the `CrossEncoder`.
- **`tests/test_embedding_fallback.py`** — new test: 8 threads released simultaneously (via `Barrier`) into `.model` must construct the model **exactly once**.

## Provenance

This fix was diagnosed, written, and committed by a `#claude` agent worker task that then hit a (subscription-route) cost cap before it could push/PR. Recovered, verified, and shipped here. (The cost-cap behavior that stopped it is fixed separately.)

## Test evidence

`pytest tests/test_embedding_fallback.py` → **10 passed** (incl. the new concurrency test). Full unit suite run in progress.

## Review focus

- Double-checked-locking correctness in `model` / `_get_model`.
- That the CPU-fallback path still works under the lock (the `_force_cpu` retry happens inside `_load_model`).

### DIFF
--- a/api/services/embeddings.py
+++ b/api/services/embeddings.py
@@ -152,11 +152,9 @@ class EmbeddingService:
         behaviour is to skip the network probe entirely.
 
         Thread safety: mutates ``os.environ`` for the duration of the retry.
-        ``EmbeddingService`` is a process-wide singleton loaded lazily on
-        first access (typically the background sync worker), so concurrent
-        loads don't happen in practice. The try/finally restores the env
-        before any other code can observe it. If a future caller starts
-        triggering loads from multiple threads at once, wrap this in a lock.
+        This runs inside ``EmbeddingService._load_lock`` (the ``model`` property
+        serializes lazy loads), so the env swap can't race a concurrent load,
+        and the try/finally restores it before the lock is released.
         """
         try:
             return st_cls(**load_kwargs)
--- a/api/services/embeddings.py
+++ b/api/services/embeddings.py
@@ -9,6 +9,7 @@ This allows tests to import this module without loading the ML library.
 """
 import logging
 import os
+import threading
 from typing import TYPE_CHECKING, Any
 
 from config.settings import settings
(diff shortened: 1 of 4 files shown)


nbramia:
## Review Round 1 — Decisions + Addressed

Reviewer verdict: **merge-ready** (double-checked locking correct, CPU-fallback + failure-retry preserved, no deadlock, concurrency test mutation-verified). Two improvements applied:

| Finding | Severity | Action |
|---|---|---|
| Stale `_load_with_offline_retry` docstring ("wrap in a lock if concurrent loads happen") | Recommended | **Applied** — updated to reflect the lock now exists |
| No concurrency test for `RerankerService._get_model` | Minor | **Applied** — added a symmetric 8-thread test |
| Health-marking now under the lock | Minor (positive) | No action (correct side effect) |

**Tests:** reranker + embedding 27 passed; branch full unit suite 1838 passed.

nbramia:
Review Round 1: merge-ready per reviewer; 2 improvements applied (docstring + reranker concurrency test). Self-verification confirmed — review loop complete.

PR #317, opened 2026-06-07
feat(telegram): multi-bot support with per-bot persona (fitness bot)

nbramia:
## Summary

Adds specialized Telegram bots — starting with a **fitness bot** — that route to the shared chat orchestrator with a per-bot system-prompt persona, while keeping the full tool suite. Bots are thin *profiles* over one engine, not separate codepaths: a registry declares them, one listener runs per bot, and a persona string threads down to the system prompt.

Closes #316

## What's included

- **Registry** (`config/telegram_bots.json`, committed, no secrets) references per-bot token env vars. Tokens live in `.env`, persona text in `config/personas/fitness.md`. `settings.telegram_bots` skips bots whose token is unset (logged) → a fresh clone runs the primary bot only. Resolves env vars from the `.env` file directly, since pydantic-settings loads `.env` into the model but does **not** export to `os.environ`.
- **One listener per bot** — own poll thread, own update-offset file (`data/telegram_state_<name>.json`), own conversation map. Bot threads stay isolated; concurrent pollers don't clobber each other's offset. Primary keeps the legacy `data/telegram_state.json`.
- **Outbound token routing** via an `_active_bot_token` ContextVar set once at the top of `_handle_update`, so all ~30 internal send sites use the right bot without per-call threading. Non-listener callers (agent worker, reminders, scheduler) default to primary — unchanged.
- **Persona** threads `chat_via_api` → `/api/ask/stream` → `run_agent_loop` → `build_system_prompt`, injected as an **uncached** block *after* the cached static block so the large shared prompt stays a common cache prefix across all bots.
- **Agent/Claude-Code reply hooks gated to primary** — Telegram message_ids are unique per chat (per bot), so specialized bots must not match replies against the shared follow-up table. Specialized bots are pure chat surfaces.

## Persona steer (fitness)

Prime + gentle-redirect: pre-loaded with workout/nutrition/metrics framing and a terse-logging expectation (`squats 5x5 @185` just logs), answers cross-domain queries fully (e.g. "am I overspending on the gym?"), and for clearly off-topic asks answers anyway with a one-line nudge toward the general assistant.

## Out of scope (follow-up)

The `lifeos_telegram_send` MCP tool and reminders/agent-worker notifications still send from the **primary** bot — proactive pushes won't originate from the fitness bot yet.

## Manual step before it runs

Create the bot in @BotFather and set `TELEGRAM_FITNESS_BOT_TOKEN` in `.env`.

## Test evidence

[...]

PR #319, opened 2026-06-07
feat(telegram): route outbound sends from specialized bots

### LINKED ISSUE #318: feat(telegram): outbound sends from specialized bots (bot_name routing)
## Problem

All proactive/outbound Telegram sends — reminders, scheduler actions, agent-worker completion pings, and the \`lifeos_telegram_send\` MCP tool — always originate from the primary bot, even when a specialized bot (fitness, therapy) is the natural context. A fitness workout reminder appearing in your primary chat, rather than your fitness bot thread, undermines the reason for having domain-specific bots.

## Context

Added in #317: specialized bots (fitness, therapy) listen and reply from their own accounts, but their outbound path is still hardwired to the primary. This is because non-listener send calls (\`send_message\`, \`send_message_async\`, \`send_message_capture_ids\`) don't know which bot to use — they fall back to \`settings.telegram_bot_token\` when \`_active_bot_token\` ContextVar is unset (which it is outside of \`_handle_update\`).

The fix is a named routing layer: callers pass a \`bot\` name, the send function resolves it to a token via the registry, and the message goes out from the right bot.

## Suggested approach

**Token resolution helper** in \`api/services/telegram.py\`: \`_token_for_bot(name: str | None) -> str\` — looks up the bot name in the registered listeners (or \`settings.telegram_bots\`), falls back to the primary token if unset or unknown. Keeps all fallback behavior intact for callers that don't pass a name.

**Add optional \`bot: str | None = None\` to all three outbound send functions** (\`send_message\`, \`send_message_async\`, \`send_message_capture_ids\`). Pass the resolved token to \`_telegram_url\`. All existing callers continue to work unchanged (default = primary).

**Surface \`bot\` in the adhoc send endpoint** (\`POST /api/reminders/send\`, \`SendMessageRequest\` body) so the MCP tool and direct API callers can target a specific bot.

**Update the \`lifeos_telegram_send\` MCP tool** in \`mcp_server.py\`: add an optional \`bot\` field to the tool schema and pass it through to the endpoint. The orchestrator can then say "send this fitness reminder from the fitness bot."

**Scheduler actions** (\`api/services/scheduler_store.py\`): store actions can optionally carry a \`bot\` field; the execution path passes it to \`send_message\` / \`send_message_async\`.

**Agent-worker** (\`api/services/agent_worker/worker.py\`): when a task session tracks which bot spawned it (via the follow-up table), completion/clarification pings can use that bot. This is the most complex part and may be scoped as a follow-up.

## Files involved

- \`api/services/telegram.py\` — \`_token_for_bot()\` helper, \`bot\` param on all three send functions
- \`api/routes/reminders.py\` — \`SendMessageRequest\` + \`send_adhoc_message\` endpoint
- \`mcp_server.py\` — \`lifeos_telegram_send\` tool schema + handler
- \`api/services/scheduler_store.py\` — scheduler action execution
- \`api/services/agent_worker/worker.py\` — (optional/follow-up) completion ping routing

## Acceptance criteria

- [ ] \`send_message(text, bot="fitness")\` sends from the fitness bot token; omitting \`bot\` still uses the primary
- [ ] \`POST /api/reminders/send {"text": "...", "bot": "fitness"}\` sends from the fitness bot
- [ ] The \`lifeos_telegram_send\` MCP tool accepts an optional \`bot\` field and routes accordingly
- [ ] The orchestrator can send a proactive fitness-bot message by naming the bot in the tool call
- [ ] Unknown or unset bot name falls back to the primary bot (no crash)
- [ ] Existing callers (scheduler, agent worker, alerts) are unchanged and still route to the primary by default
- [ ] Tests cover resolution, fallback, and MCP passthrough

nbramia:
## Summary

Lets proactive/outbound Telegram messages — reminders, scheduler actions, and the `lifeos_telegram_send` MCP tool — send from a **named bot** (fitness, therapy) instead of always the primary. Closes the #317 follow-up: a fitness reminder now lands in the fitness thread, not the primary chat.

Closes #318

## What's included

- **`_token_for_bot(name)`** in `telegram.py` — resolves a bot name to its token via the registry (`settings.telegram_bots` + primary). Unset/unknown name → `None` → callers fall back to the primary token (logged, no crash).
- **`bot` param** on `send_message`, `send_message_async`, `send_message_capture_ids`. All existing callers (alerts, agent worker, scheduler crash notices) omit it and default to primary — unchanged behavior.
- **`POST /api/reminders/send`** accepts an optional `bot` field; the **`lifeos_telegram_send` MCP tool** surfaces it in its schema. Because the MCP `_call_api` POST path passes all arguments as the JSON body, the field flows through with no handler change — the orchestrator can now say "send this from the fitness bot."
- **Scheduler**: `ScheduleEntry` gains a `bot` field, serialized as `[bot:: <name>]` in the `Inbox.md` line and read back on parse (round-trips through the index JSON via `from_dict`). `_fire_entry` routes `notify`/`prompt` sends — success and failure — from that bot.

## Deferred (noted in issue)

Agent-worker completion/clarification pings still come from the primary bot — they're sent from a background thread and tracked in a shared follow-up table keyed by Telegram message_id (per-chat), so per-bot routing there needs session→bot tracking. Out of scope here.

## Test evidence

`./scripts/test.sh` — **2705 passed**, 12 skipped. The 5 failures are the known pre-existing/environmental set (2 `.env`-overridden settings defaults, 1 vault-index suite artifact, 2 order-dependent data/sync tests), all verified on clean `main`. New file `tests/test_telegram_outbound_routing.py` (19 tests): token resolution + fallback, send-function routing, request-model field, MCP schema, `ScheduleEntry` parse/format/roundtrip.

## Review focus

- `_token_for_bot` fallback semantics (unknown → primary, never raises).
- Scheduler `[bot:: …]` serialization round-trip and back-compat with existing `Inbox.md` lines that have no `bot` field.
- MCP `bot` field flowing through `_call_api` to the endpoint without a handler change.

### DIFF
--- a/api/routes/reminders.py
+++ b/api/routes/reminders.py
@@ -82,6 +82,11 @@ class ReminderListResponse(BaseModel):
 
 class SendMessageRequest(BaseModel):
     text: str = Field(..., min_length=1, description="Message text to send via Telegram")
+    bot: Optional[str] = Field(
+        default=None,
+        description="Optional bot name to send from (e.g. 'fitness', 'therapy'). "
+                    "Falls back to the primary bot if unset or unrecognised.",
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -135,7 +140,7 @@ async def send_adhoc_message(request: SendMessageRequest):
     if not settings.telegram_enabled:
         raise HTTPException(status_code=400, detail="Telegram not configured")
 
-    success = await send_message_async(request.text)
+    success = await send_message_async(request.text, bot=request.bot)
     if not success:
         raise HTTPException(status_code=500, detail="Failed to send Telegram message")
     return {"status": "sent"}
--- a/api/routes/scheduler.py
+++ b/api/routes/scheduler.py
@@ -41,6 +41,7 @@ class CreateScheduleRequest(BaseModel):
     message_content: str = Field(default="", description="Static text or natural-language prompt")
     endpoint_config: Optional[dict] = Field(default=None, description="For endpoint action: {endpoint, method, params}")
     executor: str = Field(default="", description="For agent action: local | cloud | cloud-haiku | cloud-sonnet")
+    bot: str = Field(default="", description="Telegram bot to notify from (e.g. 'fitness', 'therapy'); empty = primary")
     enabled: bool = Field(default=True)
     timezone: str = Field(default_factory=lambda: settings.timezone, description="IANA timezone for the schedule")
 
@@ -54,6 +55,7 @@ class UpdateScheduleRequest(BaseModel):
     message_content: Optional[str] = None
     endpoint_config: Optional[dict] = None
     executor: Optional[str] = None
+    bot: Optional[str] = None
     enabled: Optional[bool] = None
     timezone: Optional[str] = None
 
@@ -68,6 +70,7 @@ class ScheduleResponse(BaseModel):
     message_content: str
     endpoint_config: Optional[dict]
     executor: str
+    bot: str
     enabled: bool
     created_at: str
     last_triggered_at: Optional[str]
@@ -87,6 +90,7 @@ class ScheduleResponse(BaseModel):
             message_content=e.message_content,
             endpoint_config=e.endpoint_config,
             executor=e.executor,
+            bot=e.bot,
             enabled=e.enabled,
             created_at=e.created_at or "",
             last_triggered_at=e.last_triggered_at,
@@ -103,6 +107,11 @@ class ScheduleListResponse(BaseModel):
 
 class SendMessageRequest(BaseModel):
     text: str = Field(..., min_length=1, description="Message text to send via Telegram")
+    bot: Optional[str] = Field(
+        default=None,
+        description="Optional bot name to send from (e.g. 'fitness', 'therapy'). "
+                    "Falls back to the primary bot if unset or unrecognised.",
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -139,6 +148,7 @@ async def create_schedule(request: CreateScheduleRequest):
         message_content=request.message_content,
         endpoint_config=request.endpoint_config,
         executor=request.executor,
+        bot=request.bot,
         enabled=request.enabled,
         timezone=request.timezone,
     )
@@ -164,7 +174,7 @@ async def send_adhoc_message(request: SendMessageRequest):
     if not settings.telegram_enabled:
         raise HTTPException(status_code=400, detail="Telegram not configured")
 
-    success = await send_message_async(request.text)
+    success = await send_message_async(request.text, bot=request.bot)
     if not success:
         raise HTTPException(status_code=500, detail="Failed to send Telegram message")
     return {"status": "sent"}
--- a/api/services/scheduler_store.py
+++ b/api/services/scheduler_store.py
@@ -68,6 +68,7 @@ class ScheduleEntry:
     message_content: str = ""  # static text or natural-language prompt
     endpoint_config: Optional[dict] = None  # {endpoint, method, params}
     executor: str = ""  # agent executor tag without '#': local / cloud / cloud-haiku ...
+    bot: str = ""  # Telegram bot name for outbound sends (e.g. "fitness", "therapy"); empty = primary
     enabled: bool = True
     created_at: str = ""
     last_triggered_at: Optional[str] = None
@@ -206,6 +207,8 @@ def _format_entry_line(entry: ScheduleEntry) -> str:
         parts.append(f"[tz:: {entry.timezone}]")
     parts.append(f"[action:: {entry.action}]")
     parts.append(f"[mtype:: {entry.message_type}]")
+    if entry.bot:
+        parts.append(f"[bot:: {entry.bot}]")
     if entry.executor:
         parts.append(f"#{entry.executor}")
     if entry.created_at:
@@ -257,6 +260,7 @@ def _parse_entry_line(line: str) -> Optional[ScheduleEntry]:
         action=action,
         message_type=message_type,
         executor=executor,
+        bot=fields.get("bot", ""),
         enabled=(symbol == " "),
         created_at=fields.get("created", ""),
         last_triggered_at=fields.get("last") or None,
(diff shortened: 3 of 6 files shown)


nbramia (commented):
## PR Review: feat(telegram): route outbound sends from specialized bots

Verified independently: checked out the branch, ran the new suite (19/19 pass) plus `test_telegram.py`, `test_telegram_multibot.py`, `test_scheduler_store.py` (96/96 pass), traced the token-resolution, MCP passthrough, and scheduler round-trip paths, and confirmed the `chat_id`/registry semantics against #317.

### Action Required
- **[Correctness]** Outbound routing resolves the **token** but never the **chat_id**. `send_message`/`send_message_async`/`send_message_capture_ids` (`api/services/telegram.py:165,208,253`) always send to `settings.telegram_chat_id` (primary), while `_token_for_bot` only swaps the bot token. Each `TelegramBotConfig` carries its own `chat_id` (`config/settings.py:721`, registry `chat_id_env`), and a Telegram bot can only post to chats it's a member of. The moment the user sets a distinct `TELEGRAM_FITNESS_CHAT_ID` — the actual intended setup per #318 ("lands in the fitness thread, not the primary chat") — the fitness token is sent to the primary chat and Telegram rejects it. It only "works" today because the fitness/therapy chat_ids are unset and thus coincide with primary. `_token_for_bot` should resolve a `(token, chat_id)` pair and the send functions should use the bot's chat_id when no explicit `chat_id` is passed.

### Recommended
- **[Requirements]** The scheduler `bot` field is plumbed into `ScheduleEntry`, its `[bot:: …]` serialization, and `_fire_entry` (`scheduler_store.py:911,924`), but there is **no API path to set or read it**. `CreateScheduleRequest`/`UpdateScheduleRequest`/`ScheduleResponse` (`api/routes/scheduler.py:35,48,61`) lack a `bot` field, and `create_schedule` (line 133-144) never passes it to `store.create`. So a bot-routed schedule can only be created by hand-editing `Inbox.md` — the orchestrator/API can't do it. This leaves the #318 scheduler acceptance criterion only half-met.
- **[Correctness]** Second, stale `/send` endpoint: `api/routes/scheduler.py:159-170` defines a duplicate `send_adhoc_message` with its own `SendMessageRequest` (line 104) that was **not** updated — no `bot` field, calls `send_message_async(request.text)`. The MCP tool targets `/api/reminders/send` (the updated one, `mcp_server.py:248`), so this path is currently unused by the new feature, but it's a divergent parallel endpoint that will surprise the next reader. Either update it for parity or note why it's intentionally left alone.

### Minor
- **[Test coverage]** None of the 19 new tests assert on `chat_id` — they only check the token in the resolved URL. This is precisely why the chat_id gap above slipped through. A test that a named bot sends to that bot's chat_id (once the pair is resolved) would lock in the fix.
- **[Standards]** `bot: str = None` on the three send functions should be `bot: Optional[str] = None` to match the type hints used elsewhere in the file (`chat_id: str = None` is pre-existing, but the new params are a chance to follow `Optional`).

### Summary
Needs changes. Token routing, MCP passthrough, fallback safety (unknown→primary, never raises, no token logging), and scheduler serialization round-trip are all correct and well-tested — but the feature resolves the bot token without the bot's chat_id, so it will misroute the moment specialized bots get their own chats (the actual goal of #318), and the scheduler `bot` field has no API surface to set it.

nbramia:

| 1 | Routing resolves the bot token but not its chat_id — sends always go to the primary chat_id | Action Required | **Accept** | Confirmed. `_token_for_bot` swaps only the token; send functions still default `chat_id` to `settings.telegram_chat_id`, ignoring `TelegramBotConfig.chat_id`. Coincides today only because DM chat_ids equal the user id; the moment a distinct `TELEGRAM_FITNESS_CHAT_ID` is set it misroutes. Fix: resolve `(token, chat_id)` together and default to the bot's chat_id. |
| 2 | Scheduler `bot` field has no API surface | Recommended | **Accept** | Confirmed — `CreateScheduleRequest`/`UpdateScheduleRequest`/`ScheduleResponse` and the MCP `lifeos_schedule_create`/`update` schemas lack `bot`, so it's only settable by hand-editing Inbox.md. Adding the field makes the scheduler criterion usable end-to-end. |
| 3 | Duplicate `/api/scheduler/send` endpoint not updated | Recommended | **Accept** | Confirmed parallel adhoc-send path with its own `SendMessageRequest`. Add `bot` there too for consistency. |
| 4 | No test asserts on chat_id | Minor | **Accept** | Directly tied to #1 — add chat_id assertions when fixing it. |
| 5 | `bot: str = None` should be `Optional[str]` | Minor | **Reject** | The adjacent `chat_id: str = None` in the same functions uses identical non-Optional style. Matching local convention (AGENTS.md "match existing style") beats changing only `bot` and making it inconsistent with its sibling. The pydantic `SendMessageRequest.bot` is already `Optional[str]`. |

**Findings forwarded to fix:** 4 (addressing directly in main context)

nbramia:
## Review Round 1 — Changes Applied

| 1 | Token routed without chat_id | **Applied** | `_resolve_bot(bot)` now returns `(token, chat_id)`; all three send functions default `chat_id` to the bot's own chat (explicit arg still wins, then bot chat, then primary). A specialized bot's message now lands in its own chat. `_token_for_bot` retained as a thin wrapper. |
| 2 | Scheduler bot field had no API surface | **Applied** | Added `bot` to `CreateScheduleRequest`, `UpdateScheduleRequest`, `ScheduleResponse` (+`from_entry`), `create_schedule` passthrough, and the `lifeos_schedule_create`/`_update` MCP schemas. |
| 3 | Parallel `/api/scheduler/send` not updated | **Applied** | Added `bot` to that endpoint's `SendMessageRequest` and passthrough, matching `/api/reminders/send`. |
| 4 | No chat_id test coverage | **Applied** | Added tests asserting named-bot chat_id routing, explicit-override precedence, `_resolve_bot` pair resolution, and scheduler API field coverage. |
| 5 | `bot: str = None` → `Optional[str]` | **Rejected** | Matches the adjacent `chat_id: str = None` in the same functions; changing only `bot` would be locally inconsistent. |

**Tests:** `./scripts/test.sh` — 2716 passed, 12 skipped. The 5 failures are the known pre-existing/environmental set (verified on clean `main`). 11 new tests added this round (30 total in the new file).

Self-verification: re-read the full diff — every change traces to a forwarded finding, no unexpected files. Review loop complete.

PR #327, opened 2026-06-07
feat(fitness): mirror the workout log to a Google Sheet

nbramia:
## Summary

Mirrors the SQLite fitness log (#320) into a Google Sheet so it's viewable on a phone. Opt-in via `LIFEOS_FITNESS_SHEET_ID` — a no-op when unset (fresh clones unaffected).

Closes #321

## What's included

- **`api/services/sheets.py`** — write-back methods (`update_values`, `append_values`, `clear_values`, `ensure_sheets`); only reads existed before.
- **`api/services/fitness_sheet_mirror.py`** — full rebuild of `Sessions` + `Sets` tabs from the store. **Dedup is by construction** — the sheet is overwritten, never appended, so re-running can't duplicate rows; a content-hash short-circuits no-op rewrites. Triggered in the **background** after a `log`/`update`/`log_metric` (debounced + serialized via a dirty/running guard) so it never blocks the bot reply; failures (e.g. a 403) are caught and logged with a re-auth hint, never crashing logging.
- **`api/services/google_auth.py`** — bump the personal Sheets scope `spreadsheets.readonly` → `spreadsheets` (read+write).
- **settings** `LIFEOS_FITNESS_SHEET_ID` + `.env.example`.

## ⚠️ Requires a one-time Google re-auth
The scope bump means the existing OAuth token (read-only Sheets) must be re-consented. After merge + setting the sheet id, re-run the Google auth flow so the read-write scope is granted; until then the mirror logs a 403 and no-ops (logging is unaffected).

## Heads-up: pre-existing lint touched
`google_auth.py` had pre-existing **duplicate definitions** of `resolve_account` / `get_configured_accounts` (the richer earlier versions are shadowed by simpler later ones — a latent bug). Since the pre-commit ruff hook blocks on any staged-file error, I silenced them with `# noqa: F811` **without changing behavior** rather than "fixing" auth in a fitness PR, and dropped one genuinely-unused import. Recommend a separate cleanup issue to resolve the duplication properly.

## Test evidence

Per request, the full ~16-min suite is deferred to one run at the end of this fitness build batch. Targeted suites pass and ruff is clean:
- `tests/test_fitness_sheet_mirror.py` (8) — off-by-default guardrail, tab build (None→""), write-both-tabs, hash short-circuit, change-triggers-rewrite, force, failure-no-raise.
- `tests/test_fitness_store.py` / `test_agent_tools_workouts.py` (37) — unaffected.
[...]

PR #334, opened 2026-06-08
refactor(google-auth): remove shadowed duplicate account helpers

### LINKED ISSUE #330: refactor(google-auth): remove shadowed duplicate account helpers
## Problem

`api/services/google_auth.py` defines `resolve_account` and `get_configured_accounts` **twice**. The richer earlier versions (lines ~44-59) are silently shadowed by simpler later duplicates (lines ~217-225) — Python uses the last definition, so the richer behavior never runs.

## Context

Found while implementing the fitness Sheet mirror (#327). The pre-commit ruff hook flags these as F811; to avoid changing auth behavior inside a fitness PR, they were silenced with `# noqa: F811` (behavior-preserving) and flagged for this dedicated cleanup.

The **live** (later) `get_configured_accounts` returns a hardcoded `[WORK, PERSONAL]`, ignoring WORK2 and credential-file existence. The **shadowed** (earlier) version detects accounts by checking for `config/credentials-{account}.json` and includes WORK2. So the intended richer behavior is currently dead.

## Suggested approach

- Decide which behavior is correct (almost certainly the earlier credential-file-detecting version, which handles WORK2).
- Keep one definition of each, delete the duplicate, and remove the two `# noqa: F811` markers added in #327.
- Verify `get_configured_accounts` callers (e.g. `agent_system_prompt.py`) still behave correctly — this may *change* which accounts are considered configured, so confirm against the actual `config/` contents.

## Files involved

- `api/services/google_auth.py` (the two duplicate definitions + the `# noqa: F811` markers)

## Acceptance criteria

- [ ] Each of `resolve_account` / `get_configured_accounts` is defined once
- [ ] The retained `get_configured_accounts` correctly reflects configured accounts (WORK2 + credential-file detection), confirmed against `config/`
- [ ] The `# noqa: F811` markers are removed; ruff is clean
- [ ] No regression in Google account resolution across callers

nbramia:
## Summary

`resolve_account` / `get_configured_accounts` were each **defined twice** in `google_auth.py`. Python used the simpler later copies, so the richer earlier versions (WORK2-aware, credential-file-detecting) were dead code. This removes the later duplicates so the intended behavior is live, and drops the `# noqa: F811` markers added in #327.

Closes #330

## Behavior change (intended)

- `resolve_account("work2")` → `WORK2` (was incorrectly `WORK`).
- `get_configured_accounts()` → reflects which `config/credentials-*.json` files actually exist, including WORK2 (was a hardcoded `[WORK, PERSONAL]`).

On a machine with only personal+work credentials the live account *set* is unchanged (just ordering); the fix matters when WORK2 is configured or work credentials are absent.

## Test evidence

`./scripts/test.sh` — **2811 passed**, 12 skipped (the 5 failures are the known pre-existing/environmental set, verified on `main` throughout this batch). No caller across calendar/gmail/drive/scheduler/agent_tools regressed. New `tests/test_google_auth_accounts.py` (7) locks in the WORK2-aware behavior and guards against re-shadowing. Ruff clean (F811 gone).

## Review focus

- The behavior change is desirable (the richer versions were the intended ones) — confirm no caller relied on the buggy hardcoded `[WORK, PERSONAL]`.
- Test isolation: `get_configured_accounts` tests `chdir` to a tmp dir with a fake `config/`.

### DIFF
--- a/api/services/google_auth.py
+++ b/api/services/google_auth.py
@@ -218,16 +218,6 @@ class GoogleAuthService:
             return False
 
 
-def resolve_account(account: str) -> GoogleAccount:  # noqa: F811 (pre-existing duplicate; see PR #327 note)
-    """Resolve a string account name ('personal'/'work') to a GoogleAccount enum."""
-    return GoogleAccount.PERSONAL if account == "personal" else GoogleAccount.WORK
-
-
-def get_configured_accounts() -> list[GoogleAccount]:  # noqa: F811 (pre-existing duplicate; see PR #327 note)
-    """Return the list of configured Google account types."""
-    return [GoogleAccount.WORK, GoogleAccount.PERSONAL]
-
-
 # Singleton instances for each account
 _auth_services: dict[GoogleAccount, GoogleAuthService] = {}
 
--- a/tests/test_google_auth_accounts.py
+++ b/tests/test_google_auth_accounts.py
@@ -1,0 +1,50 @@
+"""
+Tests for google_auth account helpers (issue #330).
+
+Locks in the WORK2-aware behavior of resolve_account / get_configured_accounts
+that was previously dead — shadowed by simpler duplicate definitions. Guards
+against the duplicates (or a re-simplification) coming back.
+"""
+import pytest
+
+from api.services.google_auth import resolve_account, get_configured_accounts, GoogleAccount
+
+pytestmark = pytest.mark.unit
+
+
+class TestResolveAccount:
+    def test_personal(self):
+        assert resolve_account("personal") == GoogleAccount.PERSONAL
+
+    def test_work(self):
+        assert resolve_account("work") == GoogleAccount.WORK
+
+    def test_work2_is_not_collapsed_to_work(self):
+        # The bug fixed in #330: the shadowing duplicate mapped anything != "personal"
+        # to WORK, losing WORK2.
+        assert resolve_account("work2") == GoogleAccount.WORK2
+
+    def test_unknown_defaults_to_personal(self):
+        assert resolve_account("nonsense") == GoogleAccount.PERSONAL
+
+
+class TestGetConfiguredAccounts:
+    def test_personal_always_present(self):
+        assert GoogleAccount.PERSONAL in get_configured_accounts()
+
+    def test_detects_credential_files_including_work2(self, tmp_path, monkeypatch):
+        # get_configured_accounts reads ./config relative to cwd.
+        monkeypatch.chdir(tmp_path)
+        (tmp_path / "config").mkdir()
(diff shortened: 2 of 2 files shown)


nbramia (commented):
## PR Review: refactor(google-auth): remove shadowed duplicate account helpers

Reviewed adversarially (Correctness, Standards, Requirements). Verified the live behavior, audited all callers, and ran the new tests + ruff locally.

### Verification performed
- **All 9 callers audited** — `resolve_account` (drive/gmail/calendar routes, agent_tools, scripts/google_auth) and `get_configured_accounts` (agent_tools, agent_system_prompt, meeting_prep, calendar_indexer, scheduler_store). Every `get_configured_accounts` caller iterates the list and wraps each account's service in try/except (e.g. `agent_tools.py:799-813`, `meeting_prep.py:382-393`, `calendar_indexer.py:149-160`). **None** index by position or assert length, so the new dynamic set is strictly safe — and more correct: previously WORK was always returned even without `credentials-work.json`, causing per-account failures swallowed by the try/except. No caller relied on the buggy hardcoded `[WORK, PERSONAL]`.
- **WORK2 end-to-end** — now that `resolve_account("work2")` returns `WORK2` and `get_configured_accounts` can include it, `get_google_auth` (`google_auth.py:241-255`) resolves `credentials-work2.json`/`token-work2.json` and `__init__` falls back to WORK scopes (line 95). The path is coherent.
- **Tests** — `tests/test_google_auth_accounts.py` 7/7 pass locally. Ruff clean, F811 gone, exactly one definition of each function remains.

### Action Required
None.

### Minor
- **[Correctness]** `get_configured_accounts` reads `Path("./config")` relative to cwd (`google_auth.py:59`). This is **not a regression** — it mirrors the pre-existing `get_google_auth` convention (line 243), and the systemd unit + `server.sh` both pin cwd to the project root. Worth being aware of only if a future caller invokes it from a non-root cwd. The new test correctly accounts for this by `chdir`-ing to a tmp dir.

### Summary
Merge-ready. The change deletes genuinely dead duplicate definitions, the WORK2-aware versions were the intended behavior, every caller tolerates the dynamic account set, and the new tests lock in the fix while guarding against re-shadowing.

nbramia:

| 1 | `get_configured_accounts` reads `Path("./config")` relative to cwd | Minor | **Reject** | Not a regression — mirrors the pre-existing `get_google_auth` convention; cwd is pinned to project root by the systemd unit and `server.sh`. The new test `chdir`s correctly. |

The reviewer audited all 9 call sites and confirmed none relied on the buggy hardcoded `[WORK, PERSONAL]` (all iterate with per-account try/except); the WORK2 path is coherent end-to-end.

Review Round 1: no actionable findings — review loop complete.

Issue #333, opened 2026-06-08
build swift app for health data sync


nbramia:
Build a **HealthBridge** iOS app that reads HealthKit and delivers Apple Health/Fitness data to LifeOS, superseding the hand-built iOS Shortcut from #323 as the recommended collector. The server-side import already shipped (#332) and consumes the data unchanged — this issue is the on-device collector plus a small server ingest endpoint.

## Build environment & approach

- **Build on a Mac (the MacBook).** Xcode is macOS-only; the iPhone pairs to whatever Mac you build from. (Mac Mini works but the laptop is handier for interactive Xcode + signing.)
- **Runs on the iPhone, not the Mac** — HealthKit data lives on the phone and the Simulator has no real health data. Loop: code on Mac → deploy to iPhone → app reads Health → POSTs to LifeOS over Tailscale.
- **Use XcodeGen, not a hand-authored `.xcodeproj`.** The project is defined by a committed `project.yml`; `xcodegen generate` produces the Xcode project. This keeps the whole app as plain text in the repo (authorable + diffable) and avoids the generated `project.pbxproj`.

## Division of labor

**Automatable (an `/implement` run can do this — text + LifeOS server, no Mac needed):**
1. **Server target** — refactor `import_health()` (`scripts/apple_data_import.py`) so a core `_ingest_health(data: dict) -> dict` does the parse+upsert, with the file path and an in-memory payload as the two callers. Add an authenticated `POST /api/fitness/health/ingest` (new `api/routes/fitness.py`, registered in `api/main.py`) that hands the posted JSON to that core.
2. **Author the full Swift app** as source under `apple/HealthBridge/`: `project.yml`, the `.swift` files, `Info.plist`, README.
3. Doc updates: `docs/guides/apple-health.md` (app = primary collector, Shortcut = fallback) and an ADR-015 (or amend ADR-014) noting the app supersedes the Shortcut.

**On-device (yours — GUI + iPhone, cannot be automated):**
- Open in Xcode, set a bundle id + **free Apple-ID signing team**, pair/trust the iPhone.
- Build & run to the phone, **grant the HealthKit permission prompt on the iPhone**, tap **Sync now**.
- Re-deploy from Xcode every ~7 days (free-ID cert expiry) to keep background delivery alive.

[...]

PR #331, opened 2026-06-08
chore: parallelize test.sh (-n auto) + diff-aware 'auto' scope mode

nbramia:
## Summary

Speeds up the local test loop. `scripts/test.sh` now runs unit tests in parallel via pytest-xdist (`-n auto --dist loadscope`) — the unit suite drops from minutes to **~44–50s**. Adds a `./scripts/test.sh auto` mode that inspects the git diff and runs only the tests a change can affect, and points `/implement` Phase 2 at it.

Closes #329

## What changed

- **`scripts/test.sh`**
  - `PYTEST_PARALLEL=(-n auto --dist loadscope)` threaded into `run_unit_tests`, the unit portion of `smoke`/`all`, plus new `run_slow_tests` / `run_changed_test_files`. Browser/integration stay serial (single shared server + Playwright).
  - New `auto` mode: `compute_changed_files` (merge-base with `origin/main` + uncommitted + untracked) → `decide_plan` (pure, unit-testable mapping) → dispatch.
    - docs-only → **skip**; `tests/`-only → **just the changed test files** (conftest/fixture changes fall back to full unit); frontend (`*.html`, `static/**` js, `api/routes/**`, templates) → **+critical browser**; sync/index (`run_all_syncs`, `*sync*`, indexer/embeddings/vectorstore/bm25) → **+slow**; else → **unit**. Mixed changes widen *additively* — scope is never narrowed below safe.
  - Extracted `run_critical_browser_test` from `run_smoke_tests` (reused by `auto`).
- **`tests/test_test_sh_auto.py`** — 20 cases asserting the mapping, driven through a new `LIFEOS_TEST_PLAN_ONLY` plan-only hook (no real pytest, runs in <1s).
- **`.claude/skills/implement/SKILL.md`** — Phase 2 step 4 now runs `./scripts/test.sh auto`.

## Test evidence

- New mapping tests: `20 passed`.
- Full parallel unit run via modified `test.sh unit`: `2760 passed` in **50.8s**.
- Pre-push selection (`-m "unit and not slow" -n auto`): `1951 passed, 0 failed` in 43.9s.
- The 11 failures seen in the raw worktree run are **pre-existing data/environment failures** (this worktree has no `data/` dir or `.env`): verified identical with my tracked changes stashed (pristine main) and run serially. Parallelism introduces no new failures, and none are `unit`-marked.

## Review focus

- `decide_plan` regex correctness and ordering (docs/tests-only are all-match; frontend/sync are any-match additive).
[...]

PR #335, opened 2026-06-08
feat(fitness): Apple Health ingest endpoint (#333 server side)

nbramia:
## Summary

Server half of #333: an authenticated `POST /api/fitness/health/ingest` so the on-device HealthBridge app has a delivery target, sharing **one ingest code path** with the nightly file importer.

Relates to #333 (the Swift app + on-device build are the follow-up PR; this issue stays open until that's deployed).

## What's included

- **`api/services/health_import.py` (new)** — `ingest_health(data)`: the parse+upsert core (+ timestamp/kind/summary helpers), extracted from the import script so both a route and the script call it (clean layering — a FastAPI route shouldn't import a private fn from `scripts/`).
- **`scripts/apple_data_import.py`** — `import_health()` now just reads the file and delegates to `ingest_health()`. Behavior unchanged; existing tests pass through the delegation.
- **`api/routes/fitness.py` (new)** — `POST /api/fitness/health/ingest`: bearer-auth via a **dedicated** `LIFEOS_HEALTH_INGEST_TOKEN` (`hmac.compare_digest`; **503 when unset** so it's off by default), hands the body to `ingest_health`, returns counts. Registered in `main.py`.
- **settings + `.env.example`** — `LIFEOS_HEALTH_INGEST_TOKEN`.

Idempotent (workout `uuid`; metric `(type, start)`); empty/absent payloads are safe.

## Also fixes a repo-wide push-gate bug (separate commit)

While adding `api/routes/fitness.py` I hit `test_auto_scope_mapping[empty]` failing in the **pre-push hook** — blocking *all* local pushes. Root cause in #331's `scripts/test.sh`: `compute_changed_files` used `${VAR:-}` (truthy only when non-empty), so the test's explicit `LIFEOS_TEST_CHANGED_FILES=""` ("no changes") fell through to the real git diff and picked up the working tree. Fixed with `${VAR+x}` (honors a set-but-empty override). Isolated in its own commit.

## Test evidence

`./scripts/test.sh auto` — **2818 passed** (after the fix), 12 skipped. The 5 failures are the known pre-existing/environmental set. New: `tests/test_health_ingest_endpoint.py` (7 — auth gate 503/401/200, idempotent ingest, empty payload) + `tests/test_test_sh_auto.py` now fully green (24, incl. `[empty]`). Existing `test_apple_health_import.py` (11) still green through the refactor.

## Review focus

- The service extraction (`import_health` → `ingest_health`) — behavior parity with the old inline version.
- The bearer-auth gate (503-when-unset, hmac compare).
- The `test.sh` `${VAR+x}` fix — correct shell semantics for set-but-empty.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -304,6 +304,9 @@ LIFEOS_RELATIONSHIP_FOLDER=Relationship
 # (see docs/guides/apple-health.md). Default: data/apple-imports/health.json.
 # Point at a synced location (e.g. ~/Code/Sync/health/health.json) for automation.
 # LIFEOS_HEALTH_EXPORT_PATH=
+# Bearer token for the HealthBridge app's POST delivery (POST /api/fitness/health/ingest).
+# Empty disables that endpoint. Generate with: openssl rand -hex 32
+# LIFEOS_HEALTH_INGEST_TOKEN=
 
 # =============================================================================
 # FINANCIAL DATA (optional)
--- a/api/main.py
+++ b/api/main.py
@@ -32,7 +32,7 @@ from fastapi.exceptions import RequestValidationError
 from pathlib import Path
 from fastapi.staticfiles import StaticFiles
 from fastapi.responses import FileResponse
-from api.routes import search, ask, calendar, gmail, drive, people, chat, briefings, admin, conversations, memories, imessage, crm, slack, photos, reminders, scheduler, tasks, monarch, jobs, perf, agents, vault
+from api.routes import search, ask, calendar, gmail, drive, people, chat, briefings, admin, conversations, memories, imessage, crm, slack, photos, reminders, scheduler, tasks, monarch, jobs, perf, agents, vault, fitness
 from config.settings import settings
 
 logger = logging.getLogger(__name__)
@@ -217,6 +217,7 @@ app.include_router(jobs.router)
 app.include_router(perf.router)
 app.include_router(agents.router)
 app.include_router(vault.router)
+app.include_router(fitness.router)
 
 # Serve static files
 web_dir = Path(__file__).parent.parent / "web"
--- a/api/routes/fitness.py
+++ b/api/routes/fitness.py
@@ -1,0 +1,56 @@
+"""
+Fitness API routes.
+
+Currently just the authenticated Apple Health ingest endpoint (#333) — the
+on-device HealthBridge app POSTs its payload here over Tailscale, and it lands
+in fitness.db via the same ingest core the nightly file importer uses.
+"""
+import hmac
+import logging
+
+from fastapi import APIRouter, HTTPException, Request
+from pydantic import BaseModel, Field
+
+from config.settings import settings
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/api/fitness", tags=["fitness"])
+
+
+class HealthIngestRequest(BaseModel):
+    """Apple Health payload — matches the health.json schema (see #333 / guide)."""
+    workouts: list[dict] = Field(default_factory=list)
+    metrics: list[dict] = Field(default_factory=list)
+
+
+def _check_ingest_auth(request: Request) -> None:
+    """Bearer-token gate. Disabled (503) until LIFEOS_HEALTH_INGEST_TOKEN is set —
+    a dedicated token, separate from the MCP transport's bearer."""
+    expected = settings.health_ingest_token
+    if not expected:
+        raise HTTPException(
+            status_code=503,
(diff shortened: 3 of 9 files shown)


nbramia (commented):
## PR Review: feat(fitness): Apple Health ingest endpoint (#333 server side)

Verified locally: branch `feat/health-ingest-endpoint` checked out, targeted suite green (`test_health_ingest_endpoint.py` + `test_apple_health_import.py` + `test_test_sh_auto.py` = 42 passed). Refactor parity, auth gate, and the `test.sh` fix all independently re-derived. No Action Required items.

### Recommended
- **[Security/Correctness]** `api/routes/fitness.py:46` — Pydantic body validation runs **before** `_check_ingest_auth`, because `payload: HealthIngestRequest` is a function parameter while the auth check is the first statement in the body. An unauthenticated caller posting a malformed body gets `422` with field-level validation detail instead of `401`/`503`. Verified: with the token unset, `POST {"workouts":"not-a-list"}` returns `422`, not the expected `503` "disabled" response — i.e. the gate can be probed/observed pre-auth. No ingest occurs and nothing sensitive leaks, but for a gated endpoint the auth check should fire before request-body parsing (e.g. a dependency that runs first, or read/validate the body inside the handler after the auth call).

### Minor
- **[Performance]** `api/services/health_import.py:65,90` — Each workout/metric row opens a fresh SQLite connection in `has_workout_ref` / `has_metric` (and again in `add_session`/`log_metric`), so a large POST is O(N) connection open/close. This is **inherited** from the pre-refactor inline importer, not introduced here, so it's out of scope for this PR — flagging only because the new endpoint can now receive arbitrarily large payloads on demand (vs. the nightly file path). Consider a request-size/row-count bound on `HealthIngestRequest` if device payloads can be large.
- **[Naming]** The #333 spec calls the core `_ingest_health` (private); the PR exposes it as public `ingest_health`. This is a justified, better choice (cross-module public API shared by a route + a script) and the PR description calls it out — noting only for spec-vs-impl traceability.

### Verification notes (no action needed)
- **Refactor parity:** `ingest_health` is byte-for-byte the old inline loop bodies; only difference is `data.get("workouts", [])` → `data.get("workouts") or []` (more robust — tolerates explicit `null`). Dedup keys match: the same normalized UTC-ISO `start` string is passed to both `has_metric` (check) and `log_metric` (insert), and `has_workout_ref`/`has_metric` query the verbatim stored form. Idempotency proven by tests.
- **Auth gate:** 503-when-unset, scheme check, and `hmac.compare_digest` are correct; tests cover 503/401(missing)/401(wrong)/200/empty-payload.
- **test.sh `${VAR+x}` fix:** Correct three-way semantics confirmed — unset → git-diff fallback (prod default), set-empty → honored as "no changes" (the `[empty]` case that was failing), set-nonempty → honored. `set -e` (not `set -u`) is active, so the `+x` guard is safe; set-empty yields zero changed-file lines.

### Summary
Merge-ready. Clean extraction with verified behavior parity, a correct auth gate, and a correct shell fix. The one Recommended item (auth running after body validation) is a hardening nit on a Tailscale-internal endpoint, not a blocker.

nbramia:

| 1 | Pydantic body validates before `_check_ingest_auth` (unauth + malformed body → 422 not 503) | Recommended | **Accept** | A gated endpoint should authenticate before parsing attacker-controlled input. Take the raw `Request`, auth first, then manually validate the body. |
| 2 | No row-count bound on the ingest payload | Minor | **Accept** | Cheap DoS hygiene now that the endpoint takes on-demand payloads. Add a generous cap (allows multi-year backfill, rejects absurd payloads with 413). The inherited N+1 is out of scope. |
| 3 | `_ingest_health` → public `ingest_health` | Minor | **Reject** | Justified cross-module public API; reviewer agreed. No action. |

**Findings forwarded to fix:** 2 (addressing directly in main context)

nbramia:
## Review Round 1 — Changes Applied

| 1 | Auth after body validation | **Applied** | `health_ingest` takes the raw `Request`, runs `_check_ingest_auth` first, then manually validates the body. Unauth + malformed → 503/401; valid auth + malformed → 422. Tests added. |
| 2 | No payload size bound | **Applied** | Cap at 100k items → 413. Test added (with a patched low cap). |
| 3 | Public `ingest_health` name | **Rejected** | Justified cross-module API. |

**Tests:** `tests/test_health_ingest_endpoint.py` — 10 passed (4 new). Ruff clean. Pre-push gate green.

Self-verification: both fixes trace to forwarded findings; no new files. Review loop complete.

PR #337, opened 2026-06-08
fix(orchestrator): verify current facts before negative claims

nbramia:
## Summary

Reviving a `/tune` commit (from 2026-06-02) that was pushed but never PR'd. Adds two lines to the orchestrator system prompt so it stops confidently asserting that something "doesn't exist / hasn't been released / isn't available" from stale training data:

1. Never make a negative claim about the current world (a release, price, schedule, roster) from memory — call `search_web` first; only say it's unavailable *after* an empty search, and say you searched.
2. On user pushback ("you're wrong" / "do research" / "look it up"), you MUST `search_web` before replying — don't repeat the prior claim, and never say "my research confirms" without an actual search this turn.

Rebased onto current `main`; the change is the only diff (+4 lines in `api/services/agent_system_prompt.py`).

## Test evidence

Rebased cleanly. `tests/test_agent_system_prompt.py` — 9 passed; ruff clean; pre-push unit suite green (60). Prompt-only change, no logic touched.

## Review focus

- The new guidance is additive and consistent with the adjacent "you have web search, don't claim a knowledge cutoff" line — confirm no contradiction with other prompt sections.

### DIFF
--- a/api/services/agent_system_prompt.py
+++ b/api/services/agent_system_prompt.py
@@ -93,6 +93,10 @@ Don't use tools for general knowledge, definitions, coding help, math, or anythi
 
 **Never say you "can't access" live data, "can't browse the web", or reference a "knowledge cutoff."** You have web search — use it.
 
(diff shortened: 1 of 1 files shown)


nbramia:
Self-reviewed (prompt-only, +4 lines, additive and consistent with the adjacent 'you have web search / no knowledge cutoff' guidance; no logic touched). A review-subagent would be disproportionate for a 4-line tune the user authored via /tune. Merging.

PR #339, opened 2026-06-10
feat(search): make retrieval durably recency-aware

nbramia:
## Summary

"Recent" / "last week" queries (chat, Telegram, MCP) frequently returned stale results — last year or months old — even when fresher matches existed. The model almost always *knows* today's date; the staleness was in **retrieval**, which the model can't see or override. This fixes it durably, with no hardcoded notion of "today" (now is computed fresh per request in the operator's timezone).

Two root causes:

1. **The recency boost was a silent no-op.** `hybrid_search` read the doc date from `metadata["date"]`, but `vectorstore.search` spreads the date to the **top-level** `modified_date` key — so the lookup always returned `None`, the boost was always `0.0`, and RRF fusion ranked purely on similarity with **zero weight on age**. A perfectly-on-topic 2025 note outranked a slightly-less-similar one from last week, and the model faithfully reported it as "recent." Now reads the correct key (top-level for vector hits; surfaced from a new BM25 sidecar table for keyword-only hits).

2. **Nothing translated relative time into a filter.** Added a pure, deterministic `resolve_relative_time(text, now)` (today/yesterday, this·last week/month/year, past N days/weeks/months, recent/lately) and wired `date_from`/`date_to` through `HybridSearch.search`, `/api/search`, `/api/ask`, and the `lifeos_search` / `lifeos_ask` MCP tools. The routes auto-resolve a window from the query when no explicit bounds are passed — so recency works even when the LLM forgets to pass dates. The orchestrator prompt also now instructs the model to resolve relative phrases against the injected current date and prefer recent matches.

### Design notes
- **No forced reindex.** BM25 dates live in a non-destructive sidecar table `doc_dates` (`CREATE TABLE IF NOT EXISTS`) rather than rebuilding the FTS5 virtual table. The vector-path fix is effective immediately; BM25 recency self-heals as docs are re-indexed on the nightly sync.
- **Undated docs are never silently dropped** by a date filter (matches the long-standing `/api/search` post-filter behavior).
- `resolve_relative_time` is a pure function of injected `now` — never reads the clock — so it's deterministic and testable.

## Test evidence
`./scripts/test.sh auto` → **2860 passed**, 12 skipped. The 4 failures are pre-existing/environmental and unrelated to this diff (verified by re-running on a clean `git stash`): two `test_settings` defaults overridden by this machine's `.env`, one vault data-integrity check, one order-dependent slack-sync test that passes in isolation.

New/changed tests:
- `test_hybrid_search.py`: rewrote `test_hybrid_search_with_recency` — it was a **false positive** (fed `metadata:{"date":...}` matching the *buggy* code, and wrapped its assert in `if len(results) >= 2:` so it passed vacuously). Now uses the real production result shape and asserts unconditionally that the newer doc overtakes the older. Added date-range filter tests (fused + vector-only paths) and an undated-passthrough test.
- `test_relative_time.py` (new): resolver edge cases against a pinned reference date, bounded-phrase-beats-vague precedence, datetime acceptance, purity.
- `test_bm25_index.py`: date round-trip, update/clear-on-readd, delete cleanup, bulk_add.

## Review focus
- `hybrid_search.py` recency-boost key fix + date filtering on **both** the fused and vector-only return paths.
- `bm25_index.py` sidecar `doc_dates` table — lifecycle correctness across add / bulk_add / delete / delete_by_path / clear.
- Whether route-level **auto-resolution** of a date window from the query is too aggressive (e.g. "recent" → trailing 90-day hard filter). Explicit params always win; undated docs pass through.
- The two AGENTS.md "ask first" surfaces: MCP tool schema additions and the BM25 schema change (chosen non-destructive).

### DIFF
--- a/api/routes/ask.py
+++ b/api/routes/ask.py
@@ -6,12 +6,12 @@ POST /api/ask - Ask questions and get synthesized answers from the vault.
 import time
 import logging
 from typing import Optional
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter
 from pydantic import BaseModel, Field, field_validator
 
 from api.services.hybrid_search import HybridSearch
 from api.services.synthesizer import get_synthesizer, construct_prompt
-from config.settings import settings
+from api.utils.date_parser import resolve_effective_dates
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api", tags=["ask"])
@@ -32,6 +32,12 @@ class AskRequest(BaseModel):
     """Ask request schema."""
     question: str = Field(..., min_length=1, description="Question to answer")
     include_sources: bool = Field(default=True, description="Include source citations")
+    date_from: Optional[str] = Field(
+        default=None,
+        description="Inclusive lower bound (YYYY-MM-DD). When omitted, a relative-time "
+                    "phrase in the question (e.g. 'last week') is auto-resolved.")
+    date_to: Optional[str] = Field(
+        default=None, description="Inclusive upper bound (YYYY-MM-DD).")
 
     @field_validator('question')
     @classmethod
@@ -93,11 +99,17 @@ async def ask(request: AskRequest) -> AskResponse:
     # Step 1: Retrieve relevant context using hybrid search (vector + BM25)
     retrieval_start = time.time()
 
+    date_from, date_to = resolve_effective_dates(
+        request.question, request.date_from, request.date_to
+    )
+
     try:
         hybrid_search = get_hybrid_search()
         chunks = hybrid_search.search(
             query=request.question,
-            top_k=10  # Get top 10 chunks for context
+            top_k=10,  # Get top 10 chunks for context
+            date_from=date_from,
+            date_to=date_to,
         )
     except Exception as e:
         logger.warning(f"Hybrid search error: {e}")
--- a/api/routes/search.py
+++ b/api/routes/search.py
@@ -12,7 +12,7 @@ from pydantic import BaseModel, Field, field_validator
 
 from api.services.vectorstore import VectorStore
 from api.services.hybrid_search import HybridSearch
-from config.settings import settings
+from api.utils.date_parser import resolve_effective_dates
 
 logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api", tags=["search"])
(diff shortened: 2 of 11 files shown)


nbramia:

Reviewer found **no Action Required** items (assessed merge-ready). Decisions on the rest:

| 1 | Vague "recent"/"lately" auto-resolves to a *hard* 90-day filter with no empty-result fallback → could drop a >90-day-old relevant doc | Recommended | **Accept** | Fixed in `a66dee5`. `resolve_relative_time` gains `include_vague`; routes call it with `include_vague=False` so only *bounded* phrases ("last week") become hard filters. Vague recency is left to the now-working boost (orders without excluding). |
| 2 | `calculate_recency_boost` uses naive `now()` while `resolve_effective_dates` is tz-aware — can disagree ~1 day | Recommended | **Downgrade → Minor, no change** | Doc dates are naive **date-only** strings; the decay granularity is days over a 365-day window. Making `now` tz-aware there would risk a naive/aware subtraction bug for a sub-day difference that can't change the boost. Naive `now()` is the correct match for that computation. |
| 3 | "past N months" = `30*N` days, not calendar months | Minor | **Reject (no action)** | Documented approximation; reviewer agreed it's acceptable. |
| 4 | Two parallel date-filter mechanisms in `search.py` (top-level params vs `filters`) — latent footgun | Minor | **Accept** | Added a comment in `a66dee5` noting they're independent and compose. |

**Findings forwarded / addressed:** 2 (both in `a66dee5`).

nbramia:
## Round 1 — Changes Applied (`a66dee5`)

| 1 | Over-aggressive auto-filter on vague "recent" | Applied | `include_vague` param on `resolve_relative_time`; routes opt out via `include_vague=False`. Added 4 tests (vague→None, bounded still resolves, route wrapper ignores vague "recent", explicit params win). |
| 4 | Date-filter redundancy unclear | Applied | Clarifying comment on the two independent knobs in `search.py`. |

**Tests:** `pytest test_relative_time test_hybrid_search test_bm25_index test_search_api test_ask_api` → **99 passed**. Pre-commit unit run → 60 passed. ruff clean.

nbramia:
Review loop complete: Round 1 found no Action Required items; the one substantive Recommended finding (over-aggressive vague-recency filter) was addressed in `a66dee5` with tests, and self-verification surfaced no new concerns. Proceeding to merge.

PR #343, opened 2026-06-14
feat(scheduler): full CRUD in manage_schedules + editable instructions in markdown

### LINKED ISSUE #342: feat(scheduler): full CRUD in manage_schedules + editable instructions in markdown
Two related gaps in the scheduler. Both are about closing the loop between how schedules are managed (chat tool) and how they're persisted/edited (vault markdown).

---

### Problem

The agentic chat tool `manage_schedules` (`api/services/agent_tools.py`) only exposes `create` and `list` actions. There is no way for the orchestrator to edit an existing schedule (rename, re-time, enable/disable, change action) or delete one — so in-chat requests like "turn off my morning briefing" or "move that reminder to 9am" can't be fulfilled. The underlying `SchedulerStore` already implements `update()` and `delete()`, and the MCP server (`mcp_server.py`) already exposes full CRUD via `lifeos_schedule_update` / `lifeos_schedule_delete`. The gap is purely in the chat tool layer — an asymmetry between the MCP surface and the chat surface.

### Suggested approach

- Extend the `action` enum to `["create", "list", "update", "delete"]` and update the tool description.
- Add a `schedule_id` input property (required for update/delete; obtained from `list`).
- Add `_schedule_update` and `_schedule_delete` handlers that call `store.update(...)` / `store.delete(...)`, and route them in `_tool_manage_schedules`.
- For `update`, only pass through provided fields (mirror the MCP `lifeos_schedule_update` "only provided fields change" semantics).
- Add progress-label strings for `manage_schedules.update` / `manage_schedules.delete` (alongside the existing `.create` / `.list` labels).
- Decide whether the deprecated `manage_reminders` alias needs the same treatment or stays create/list-only.

### Files involved

- `api/services/agent_tools.py` — tool schema (~L367), handlers (`_schedule_create`/`_schedule_list`/`_tool_manage_schedules` ~L1230-1271), progress labels (~L1823)
- `api/services/scheduler_store.py` — `update()` / `delete()` already exist (reference only)

---

### Problem

The scheduler's markdown source of truth (`LifeOS/Scheduler/Inbox.md`) carries only the schedule *definition* — name, trigger, action, mtype, executor, enabled checkbox (see `_format_entry_line` in `api/services/scheduler_store.py`). The actual **instruction payload** (`message_content` — the static text, natural-language prompt, or agent task description, plus `endpoint_config`) is persisted **only** in the index cache (`data/scheduler_index.json`) and merged back by ID via `_carry_payload`. As a result, the raw instructions for each scheduled task do not appear anywhere in the vault markdown and cannot be manually viewed or edited there.

### Goal

The instruction text for each schedule should appear in `Inbox.md` (or its per-schedule markdown) and be manually editable, with edits round-tripping back through the parser into the index cache — consistent with the "markdown is the source of truth" model documented in AGENTS.md.

### Suggested approach

- Serialize `message_content` (and `endpoint_config` where relevant) into the markdown in a way that survives round-trip parsing. Inline `[field:: value]` fields don't handle multi-line prompts well, so consider an indented continuation / fenced block beneath the schedule's checkbox line.
- Update `_format_entry_line` (write) and `_parse_entry_line` (read) as a matched pair so the payload is lossless across reindex.
- Once the payload lives in markdown, reconsider whether `_carry_payload` is still needed or should defer to the markdown value (markdown wins on conflict, per source-of-truth model).
- Verify the dashboard regeneration (`_write_dashboard`) and reindex paths still behave.

### Files involved

- `api/services/scheduler_store.py` — `_format_entry_line` (~L199), `_parse_entry_line` (~L222), `_carry_payload` (~L558), reindex paths (~L508-557)

---

## Acceptance criteria

- [ ] `manage_schedules` accepts `action="update"` with a `schedule_id` and changes only the provided fields
- [ ] `manage_schedules` accepts `action="delete"` with a `schedule_id` and removes the schedule
- [ ] Both return a clear confirmation string (and a sensible error when `schedule_id` is missing/unknown)
- [ ] Progress labels exist for the new actions
- [ ] A schedule's instruction text (`message_content`) appears in `Inbox.md` and is human-readable
- [ ] Manually editing that instruction text in the markdown round-trips into the index cache on reindex
- [ ] Tests cover update + delete via the chat tool, and markdown round-trip of `message_content`

nbramia:
## Summary

Closes #342. Two related scheduler gaps:

**Part 1 — `manage_schedules` chat tool CRUD.** The agentic chat tool only supported `create`/`list`. It now also supports `update` and `delete`, matching the full CRUD the MCP surface (`lifeos_schedule_*`) already exposed. `update` is a partial update (only the fields you pass change); both require `schedule_id` (from `list`) and return clear errors when it's missing or unknown. Also registers `manage_schedules` in `_CONSOLIDATED_TOOLS` so its per-action status labels actually fire (they were previously dead — `manage_schedules` wasn't in the set, so only the generic "Managing schedules..." ever showed).

**Part 2 — editable instructions in the vault markdown.** Previously `message_content` (the instruction/prompt text) lived only in the index cache (`data/scheduler_index.json`), so it never appeared in `Inbox.md` and couldn't be hand-edited. It now serializes beneath each schedule's checkbox line as an indented `> ` blockquote body, which is human-readable, multi-line-safe, and round-trips through reindexing. The line helpers became block-aware (insert/rewrite/remove + reindex parse over `(start, end, entry)` blocks).

### Migration note
`_merge_prior` keeps a cache fallback for `message_content`: schedules created before this change have no body in markdown, so they keep their cached content and migrate lazily — a body is written the next time the schedule is updated or fires. New schedules get a body immediately. `endpoint_config` and run history remain cache-only (no markdown representation). One consequence of the fallback: clearing a message by deleting its body alone won't take (the cache restores it); changing the text works as expected.

## Test evidence

Run against an up-to-date isolated checkout on the server venv (the local box has no venv; the on-box `~/Code/lifeos` is stale):

- `tests/test_scheduler_store.py tests/test_scheduler_api.py tests/test_scheduler_watcher.py` → **81 passed**
- Related modules (agent tools/loop, telegram routing, migrate-reminders, audit e2e, mcp exposure) → **237 passed**
- Full unit suite → **946 passed, 1 failed** — the one failure is `test_claude_code_ingest.py::test_live_process_cache_avoids_repeated_scans`, which errors at `import psutil` (dependency missing from that venv), unrelated to this change.

New tests: tool-level `update`/`delete` (happy path, partial update, missing/unknown id) and markdown body round-trip (write, external edit, multi-line rebuild, delete removes body).

## Review focus

- `scheduler_store._iter_entry_blocks` / `_find_block_span` and the block-aware rewrite/remove — correctness of body span detection (blank lines in content, content lines that look like checkboxes).
- The `_merge_prior` lazy-migration tradeoff (clearing-by-emptying-body limitation) — acceptable, or should markdown win unconditionally?
- `_schedule_update` field forwarding (`schedule_action` → `action`; `enabled=False` handling).

### DIFF
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -29,7 +29,7 @@ from config.settings import settings
 logger = logging.getLogger(__name__)
 
 # Consolidated tools that use sub-action status messages
-_CONSOLIDATED_TOOLS = {"manage_tasks", "manage_reminders", "person_info"}
+_CONSOLIDATED_TOOLS = {"manage_tasks", "manage_reminders", "manage_schedules", "person_info"}
 
 # Patterns that indicate the model is giving up without trying tools
 _GIVE_UP_PATTERNS = re.compile(
--- a/api/services/agent_tools.py
+++ b/api/services/agent_tools.py
@@ -366,21 +366,26 @@ TOOL_DEFINITIONS = [
     {
         "name": "manage_schedules",
         "description": (
-            "Manage schedules: create or list. A schedule binds a trigger "
-            "(once/cron) to an action (notify/prompt/endpoint/agent). Use "
-            "action='agent' to run autonomous work on a schedule."
+            "Manage schedules: create, list, update, or delete. A schedule binds "
+            "a trigger (once/cron) to an action (notify/prompt/endpoint/agent). Use "
+            "action='agent' to run autonomous work on a schedule. For update/delete, "
+            "pass schedule_id from a prior list; update changes only the fields you supply."
         ),
         "input_schema": {
             "type": "object",
             "properties": {
                 "action": {
                     "type": "string",
-                    "enum": ["create", "list"],
+                    "enum": ["create", "list", "update", "delete"],
                     "description": "Operation to perform.",
+                },
+                "schedule_id": {
+                    "type": "string",
+                    "description": "ID of the schedule to update or delete (from action='list').",
                 },
                 "name": {
                     "type": "string",
-                    "description": "Short schedule name/title (for create).",
+                    "description": "Short schedule name/title (for create; optional for update).",
                 },
                 "schedule_type": {
                     "type": "string",
@@ -404,6 +409,10 @@ TOOL_DEFINITIONS = [
                     "type": "string",
                     "enum": ["local", "cloud", "cloud-haiku", "cloud-sonnet"],
                     "description": "For schedule_action='agent': which executor the spawned #agent task targets.",
+                },
+                "enabled": {
+                    "type": "boolean",
+                    "description": "Enable (true) or pause (false) the schedule (for update).",
                 },
             },
             "required": ["action"],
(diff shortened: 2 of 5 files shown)


nbramia:

| 1 | Indented `> ` lines under a schedule are absorbed into message_content (reserved syntax) | Recommended | Accept | Valid and confirmed. Cheap, in-scope doc note in the module docstring so future editors know the syntax is reserved. |
| 2 | Lazy-migration: clearing a message by deleting only the body line doesn't take | Minor | Reject (no code change) | Already documented in `_merge_prior` docstring and the PR body; clearing via the tool works fully. Inherent, no data-loss risk. |
| 3 | Empty-body marker `    >` vs `    > ` whitespace asymmetry | Minor | Reject | Round-trip is lossless (both parse to ""); only literal trailing whitespace may differ. Reviewer agreed no action. |
| 4 | Unrelated codex files in working tree | Minor | Reject | Confirmed correctly excluded from this PR's diff (they ship in a separate PR). |

**Findings forwarded to addresser:** 1 (applied directly by orchestrator — single-line doc note)

PR #352, opened 2026-06-21
feat(agent-worker): always write an Agent Output note on task completion

### LINKED ISSUE #350: feat(agent-worker): always write an Agent Output .md on task completion
## Problem

The agent worker only writes a file to the vault's Agent Output folder (`LifeOS/Tasks/Agent Output`, `settings.agent_output_dir`) as a *spillover* — and only when the agent's final reply exceeds 2000 chars (`worker.py` `_completion_summary` → `_spill_to_vault`). Short completions and recurring schedule runs leave no durable artifact; the only record is a Telegram message. There is also no link from a scheduler-driven recurring task back to its schedule, so repeated fires of the same schedule can't be grouped into one note.

## Goal

Every successfully completed root `#agent` task should write an agent-output `.md`:
- **One-off task** → a new `.md` per task.
- **Recurring task** (a `cron` Scheduler entry with `action: agent`) → all fires of that schedule **prepend** to one shared `.md`, newest run on top.

## Confirmed decisions

- "Recurring" = Scheduler `action: agent` entries whose `schedule_type == "cron"`. `once` schedules are one-shot → treated as one-off.
- Write **only on successful completion** (not on failed / blocked / budget-exceeded).
- Prepend format: newest run on top, each under a `## YYYY-MM-DD HH:MM` heading, separated by a horizontal rule, with frontmatter kept at the top of the file.

No DB schema change, no new API endpoint, no new dependency, no MCP/API contract change. The schedule→task link is carried via a task **tag** (`sched-<id>`), which round-trips through the markdown source of truth and is already plumbed end-to-end.

## Suggested approach

**1. Tag recurring tasks at handoff — `api/services/scheduler_store.py`**

In `_hand_off_to_agent` (~line 987), when `entry.schedule_type == "cron"`, append a `sched-<id>` tag to the created task. Tag charset `[\w-]` (`task_manager.py:573` `_TAG_RE`) means `sched-ab12cd34` round-trips create → markdown → `GET /api/tasks/{id}` → `task["tags"]`. `once` schedules get no extra tag → normal one-off path. Presence of a `sched-<id>` tag therefore means "recurring (cron)", so the worker needs no schedule-type lookup to decide prepend-vs-new.

**2. Always write the output file — `api/services/agent_worker/worker.py`**

Refactor `_spill_to_vault` (line 2030) into `_write_agent_output(session, task, final_text)` that **always** writes (for non-empty `final_text`) and branches one-off vs recurring. Keep the return contract `tuple[rel_path, obsidian_url] | None` (None → callers keep the inline fallback).

- **Detect recurring:** scan `task.get("tags", [])` for `^sched-(\w+)$`; capture `schedule_id`.
- **One-off** (no `sched-` tag): new file `<output_dir>/<YYYY-MM-DD>-<slug>-<sid4>.md` (existing date+slug naming plus a 4-char `session_id` suffix to avoid same-day/same-slug clobbering now that every task writes). Frontmatter as today (`task`, `session_id`, `routing`, `created`, `source: agent-worker`) + body.
- **Recurring** (`sched-<id>` present): shared file keyed on the schedule.
  - Resolve a readable name via new `_fetch_schedule(schedule_id)` → `GET {api_base}/api/scheduler/{id}` (route exists, `scheduler.py:183`), using `entry["name"]`. On 404/error, fall back to `recurring-<id>` (the `id` is stable, so grouping still works).
  - Target file: `<output_dir>/<name-slug>.md`.
  - **First run** (file absent): write frontmatter (`schedule`, `schedule_id`, `created`, `updated`, `source: agent-worker-recurring`) + the run block.
  - **Subsequent runs** (file present): split the leading `---…---` frontmatter from the body, bump `updated:`, then write `frontmatter + new_run_block + old_body`.
  - Run block: a `## YYYY-MM-DD HH:MM` heading, the `final_text`, then a `---` rule.

Update `_completion_summary` (line 1669): drop the length-gated call and **always** call `_write_agent_output` when `final_text` is non-empty. Telegram blurb behavior preserved: empty final text → transcript-pointer path (no file); short → inline full text + `Saved to vault:` line; long → preview + link (as today). `_handle_outcome` needs no structural change (the always-write lives inside `_completion_summary`, which only runs on `STATUS_COMPLETED && not is_spawned`; spawned children, operator root-spawns, and non-completed outcomes are unaffected).

Reuse existing helpers: `settings.vault_path`, `settings.agent_output_dir`, `build_obsidian_link`, the slug regex already in `_spill_to_vault`.

## Notes / non-goals

- The agent may still write its own `lifeos_vault_write` deliverables into the same folder (different filenames). These coexist with the always-write journal and are **not** deduped — same as today's spillover-vs-agent-write coexistence.
- A renamed cron schedule starts a fresh `<name-slug>.md` (the `id` is the stable key but the filename is name-derived for readability). Acceptable; rare.
- Verify the extra `sched-<id>` tag is ignored by preflight executor routing (it should only match known executor tags).

## Files involved

- `api/services/scheduler_store.py` — `_hand_off_to_agent`: add `sched-<id>` tag for cron schedules.
- `api/services/agent_worker/worker.py` — refactor `_spill_to_vault` → `_write_agent_output`; add `_fetch_schedule`; update `_completion_summary`.
- `tests/test_scheduler_store.py` — update `test_agent_action_creates_tagged_task` to expect `sched-<id>` for cron; add a `once`-gets-no-`sched`-tag case.
- `tests/test_agent_worker_loop.py` — update the four `test_completion_summary_*` tests (now always mention a saved path); ensure the harness sets `vault_path`.
- `tests/test_agent_output_write.py` (new) — one-off creates a dated file; recurring first run creates `<name-slug>.md`; second recurring run prepends above the first (newest-on-top, frontmatter intact, `updated:` bumped); schedule-resolution failure falls back to `recurring-<id>.md`.

## Acceptance criteria

- [ ] Every successfully completed root `#agent` task (one-off and recurring) writes a `.md` to `LifeOS/Tasks/Agent Output`, regardless of reply length.
- [ ] One-off tasks produce a new dated file per task; filenames don't clobber on same-day/same-slug.
- [ ] A `cron` `action: agent` schedule tags its handoff task with `#sched-<id>`; a `once` schedule does not.
- [ ] Repeated fires of the same cron schedule prepend to a single `<name-slug>.md`, newest run on top under a `## <timestamp>` heading, prior runs below the rule, frontmatter preserved and `updated:` bumped.
- [ ] Schedule-name resolution failure falls back to `recurring-<id>.md` and still groups correctly.
- [ ] Only successful completions write files; failed/blocked/budget-exceeded outcomes are unchanged.
- [ ] The Telegram completion message links the saved file for both short and long outputs.
- [ ] `./scripts/test.sh` passes, including new/updated tests.

nbramia:
## Summary

Every successfully completed root `#agent` task now lands a Markdown note in the vault's Agent Output folder (`LifeOS/Tasks/Agent Output`), regardless of reply length — previously a note was only written as spillover for replies >2000 chars.

- **One-off tasks** → a new dated note `<YYYY-MM-DD>-<slug>-<sid>.md`. The short session-id suffix prevents same-day/same-slug clobbering now that every task writes.
- **Recurring (cron) scheduler agent tasks** → all fires append to **one shared note per schedule** (`<schedule-slug>.md`), newest run on top under a `## YYYY-MM-DD HH:MM` heading, with frontmatter `created` preserved and `updated` bumped.

The schedule→task link rides on a `sched-<id>` tag stamped by the scheduler handoff for **cron schedules only** (`once` schedules stay one-off). On completion the worker reads the tag, resolves the schedule's human name via `GET /api/scheduler/{id}` for a readable filename (falling back to `recurring-<id>.md` on 404/error), and prepends the run. `_spill_to_vault` is generalized into `_write_agent_output`, now always invoked from `_completion_summary`.

No DB schema change, no new API endpoint, no new dependency, no MCP/HTTP contract change.

Closes #350

## Test evidence

- New `tests/test_agent_output_write.py` (10 tests): one-off dated file; no same-day clobber; vault-unset → None; recurring first run creates `<name-slug>.md`; second run prepends newest-on-top; `created` preserved + `updated` bumped; name-resolution failure → `recurring-<id>.md`; tag detection; `_split_frontmatter`/`_slugify` helpers.
- Updated `tests/test_scheduler_store.py` (cron handoff now expects `sched-<id>`; added `once`-gets-no-tag case) and `tests/test_agent_worker_loop.py` (autouse vault redirect; inline-summary test now expects the always-write pointer).
- `./scripts/test.sh auto` + pre-push full unit suite: **2051 passed, 8 skipped**. The 14 unrelated failures in the broad run (CRM data-integrity, meeting-prep, settings/me-page config defaults) were verified pre-existing on the clean base with this branch stashed.

## Review focus

- `_write_agent_output` recurring branch in `api/services/agent_worker/worker.py` — frontmatter `created` preservation across fires and the prepend ordering in `_recurring_content` / `_split_frontmatter`.
- `_hand_off_to_agent` in `api/services/scheduler_store.py` — that only `cron` schedules get the `sched-<id>` tag, and that the extra tag is inert to preflight routing.

### DIFF
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -24,9 +24,13 @@ import signal
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import httpx
+
+if TYPE_CHECKING:
+    from datetime import datetime
+    from pathlib import Path
 
 from api.services.agent_worker.preflight import (
     ROUTE_ASK,
@@ -2129,7 +2133,9 @@ class Worker:
                 # fire still maps to the same file.
                 sched_name = self._resolve_schedule_name(schedule_id)
                 slug = _slugify(sched_name) if sched_name else ""
-                filename = f"{slug}.md" if slug else f"recurring-{schedule_id}.md"
+                # Append the schedule id so two distinct schedules that share a
+                # human name don't interleave into the same note.
+                filename = f"{slug}-{schedule_id}.md" if slug else f"recurring-{schedule_id}.md"
                 file_path = target_dir / filename
                 content = self._recurring_content(
                     file_path, schedule_id, sched_name or title, now, final_text,
@@ -2162,7 +2168,8 @@ class Worker:
         return (rel_path, obsidian_url)
 
     def _recurring_content(
-        self, file_path, schedule_id: str, label: str, now, final_text: str,
+        self, file_path: Path, schedule_id: str, label: str,
+        now: datetime, final_text: str,
     ) -> str:
         """Build the new contents for a recurring schedule's shared note by
         prepending this fire above any existing runs, newest first. Frontmatter
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import re
 import signal
 import threading
 import time
@@ -61,6 +62,30 @@ logger = logging.getLogger(__name__)
 # (init-failed MCPs). Above this we spill the body to a vault note and
 # put just a 1-line preview + obsidian:// link in the Telegram message.
 _INLINE_SUMMARY_MAX_CHARS = 2000
+
+# A recurring (cron) schedule stamps its handed-off #agent task with a
+# `sched-<id>` tag (see scheduler_store._hand_off_to_agent). The worker reads
+# it on completion to append every fire's output to one shared note per
+# schedule instead of a new note per fire.
+_SCHED_TAG_RE = re.compile(r"^sched-(\w+)$")
+
+
+def _slugify(text: str) -> str:
+    """Lowercase, hyphenated, filesystem-safe slug capped at 60 chars.
+    Returns '' for empty/symbol-only input."""
+    return re.sub(r"[^A-Za-z0-9]+", "-", text or "").strip("-").lower()[:60]
+
+
+def _split_frontmatter(text: str) -> tuple[str, str]:
+    """Split a leading `---\\n...\\n---` YAML frontmatter block from the body.
+    Returns (frontmatter_with_delimiters, body). When there's no frontmatter,
+    returns ('', text) so the whole document is treated as body."""
+    if not text.startswith("---"):
+        return "", text
+    m = re.match(r"^---\n.*?\n---\n?", text, re.DOTALL)
+    if not m:
+        return "", text
+    return text[: m.end()], text[m.end():]
 
 
 def _worker_label(routing: str | None) -> str:
@@ -1709,21 +1734,32 @@ class Worker:
                     f"(agent idled without a final text reply — check transcript at "
                     f"`data/agent_transcripts/{session.session_id}.jsonl` for tool-use detail)"
                 )
-        elif len(final_text) <= _INLINE_SUMMARY_MAX_CHARS:
-            result_blurb = final_text
         else:
-            # Spill the full body to a vault note and link to it instead of
-            # truncating mid-answer. The agent's system prompts now ask the
-            # agent itself to create artifacts for long outputs, but the
-            # operator-facing UX shouldn't depend on the agent following
-            # that guidance: any over-length response gets spilled here.
-            spillover = self._spill_to_vault(session, task, final_text)
-            if spillover is None:
-                # Vault not configured or write failed — preserve the old
-                # behavior so the operator still gets *something* readable.
+            # Every completed task now lands a durable note in the vault's
+            # Agent Output folder — one-off tasks get a new note, recurring
+            # (cron-scheduled) tasks append to one shared note per schedule.
+            # The agent's system prompts also ask it to create artifacts for
+            # long outputs, but the operator-facing record shouldn't depend on
+            # the agent following that guidance: the worker always writes.
+            written = self._write_agent_output(session, task, final_text)
+            if len(final_text) <= _INLINE_SUMMARY_MAX_CHARS:
+                # Short answer: show it inline. Append a pointer to the saved
+                # note when the write succeeded (vault may be unconfigured).
+                result_blurb = final_text
+                if written is not None:
+                    rel_path, obsidian_url = written
+                    result_blurb += (
+                        f"\n\nSaved to vault: `{rel_path}`\n"
+                        f"[Open in Obsidian]({obsidian_url})"
+                    )
+            elif written is None:
+                # Over-length but vault not configured / write failed —
+                # preserve the old behavior so the operator still gets
+                # *something* readable instead of an empty message.
                 result_blurb = final_text[:_INLINE_SUMMARY_MAX_CHARS] + "…"
             else:
-                rel_path, obsidian_url = spillover
+                # Over-length: link to the note instead of truncating mid-answer.
+                rel_path, obsidian_url = written
                 preview = final_text.split("\n\n", 1)[0].strip()
                 if len(preview) > 400:
                     preview = preview[:400].rsplit(" ", 1)[0] + "…"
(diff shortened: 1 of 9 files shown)


nbramia (commented):
## PR Review: feat(agent-worker): always write an Agent Output note on task completion

Adversarial review with Correctness, Security/Privacy, Requirements, and Standards specialists. I verified the test claims myself (104/104 in the three affected files pass; clean `_spill_to_vault`→`_write_agent_output` rename with no dangling refs) and traced the control flow rather than trusting the description.

### Action Required
_None._ All 8 acceptance criteria in #350 are implemented and behave correctly. The recurring prepend / frontmatter logic is sound — `_split_frontmatter`'s non-greedy `^---\n.*?\n---\n?` correctly stops at the first close delimiter, so run blocks ending in `---\n` are never mis-parsed and `created` is preserved across fires (verified by 3-fire repro). `sched-<id>` is inert to preflight routing (only exact executor tags match). `_slugify` strips `/`, `\`, `..`, and null bytes, so no path traversal from task descriptions or schedule names. The write is reachable only via `_completion_summary` (single call site, `worker.py:1640`, gated on `STATUS_COMPLETED and not is_spawned`), so failed/blocked/budget/spawned outcomes never write — criterion #6 holds.

### Recommended
- **[Requirements] No negative test for criterion #6** (`tests/test_agent_worker_loop.py`). Nothing asserts that a FAILED / BLOCKED / BUDGET_EXCEEDED outcome leaves the Agent Output dir empty. The behavior is correct by construction, but #6 is a load-bearing "only on success" guarantee that a future refactor could silently regress with the suite still green. The autouse `_redirect_agent_output` fixture already points the vault at `tmp_path/vault`, so this is a one-line add to an existing failed/budget test: `assert not list((tmp_path / "vault" / _settings.agent_output_dir).glob("*.md"))`.
- **[Correctness] Two distinct cron schedules with the same human name collide** (`worker.py:2130-2135`). Both slugify to the same `<slug>.md`, so their fires interleave into one shared note and the `schedule_id` frontmatter gets overwritten by whichever fired last — silent cross-contamination. This is the exact failure mode the PR *defends against* for one-off tasks (the `-<sid>` suffix), so the inconsistency is notable. Either key the filename on `schedule_id` (e.g. `<slug>-<id>.md`) or document the collision as accepted (the issue covers renamed-schedule but is silent on same-name-two-schedules). Low probability; silent when it happens.
- **[Standards] Docs don't reflect the new behavior.** `docs/specs/product/agent-worker.md:45` and `docs/specs/technical/agent-worker.md:69,114` still describe completion as "mark done, swap tag, Telegram summary" with no mention that *every* completion now writes a durable note, that short answers gain a "Saved to vault" pointer, or that recurring schedules group into one shared note via the `sched-<id>` tag. `docs/guides/scheduler.md` also says nothing about per-schedule output notes. (The old over-cap spillover was never documented, so this is a net-new gap, not a stale removal.)

### Minor
- **[Correctness] `created` reset if a hand-edited note loses its `created:` line** (`worker.py:2185-2187`). The preservation regex silently falls back to today, defeating the durability guarantee. Self-inflicted; worth a one-line comment noting the dependency.
- **[Correctness] `_split_frontmatter` is LF-only** (`worker.py:85`). A CRLF-saved note won't match → existing frontmatter gets demoted to body and a second `---` block is prepended on the next fire. Obsidian writes LF so real-world risk is low; latent gap with no guard.
- **[Requirements] Criterion #4 "prior runs below the rule" only asserted indirectly.** Tests check newest-on-top ordering and heading count but not the `---` separator between runs. `text.count("\n---\n") >= 2` would close it.
- **[Standards] Stale "spill"/"spillover" vocabulary after the rename** — test name `test_completion_spills_to_vault_when_over_cap` + comments (`test_agent_worker_loop.py:1182,1194,1215,1225`), and prose comments at `worker.py:62`, `local_executor.py:207,289`. Tests still valid; terminology is now orphaned.
- **[Standards] Missing type hints** on `_recurring_content` params `file_path` and `now` (`worker.py:2164-2166`) while the three siblings (and all other new methods) are annotated. `file_path: Path, now: datetime`.

### Summary
Merge-ready modulo nits — the feature is correctly implemented, well-tested on the happy paths, privacy-clean (no new logging of personal data, no path traversal, same already-private write target), and surgically scoped. The two most worthwhile follow-ups are a negative test for criterion #6 and deciding the same-name-schedule collision (fix vs. document); the doc refresh is the main process item.

nbramia:

| 1 | Same-name cron schedules collide on `<slug>.md` | Recommended | **Accept** | Real correctness bug, and inconsistent with the one-off `-<sid>` defense. Fix: recurring filename → `<slug>-<id>.md` (stable per schedule, collision-proof). |
| 2 | No negative test for criterion #6 (failed/blocked/budget writes no file) | Recommended | **Accept** | Load-bearing guarantee a refactor could silently regress; cheap to add given the autouse vault-redirect fixture. |
| 3 | Docs still describe old completion flow | Recommended | **Accept** | Behavior change to documented flows. Update product/technical agent-worker specs, scheduler guide, and the `LIFEOS_AGENT_OUTPUT_DIR` reference. |
| 4 | Missing type hints on `_recurring_content` | Minor | **Accept** | Cheap clarity win (add `Path`/`datetime` hints under a TYPE_CHECKING guard). |
| 5 | Criterion #4 "prior runs below the rule" only asserted indirectly | Minor | **Accept** | Fold a `---`-separator assertion into the existing second-run test. |
| 6 | `created` resets to today if a hand-edited note loses its `created:` line | Minor | **Reject** | Acceptable fallback — an absent date can't be recovered; only affects manually-corrupted notes. |
| 7 | `_split_frontmatter` is LF-only | Minor | **Reject** | `Path.read_text()` applies universal-newline translation (CRLF→LF); the real path never hits it (reviewer's own Correctness agent confirmed). |
| 8 | Orphaned "spill/spillover" vocabulary in test names | Minor | **Reject** | The over-cap path genuinely still spills-and-links; renaming is pure churn. No dangling `_spill_to_vault` refs remain. |

**Findings forwarded to addresser:** 5 (handled directly by the implementer in this context).

nbramia:
## Review Round 1 — Addressed

| 1 | Same-name cron schedule collision | Applied | Recurring filename → `<slug>-<id>.md` (`worker.py`). Distinct schedules sharing a name no longer interleave. Tests updated to the new filename. |
| 2 | No negative test for criterion #6 | Applied | Added `test_failed_task_writes_no_agent_output` — a FAILED outcome notifies but leaves the Agent Output folder empty. |
| 3 | Docs describe old completion flow | Applied | Updated product + technical agent-worker specs (new "Agent Output notes" section + ToC), scheduler guide hand-off section, and the `LIFEOS_AGENT_OUTPUT_DIR` config row; bumped `Last Updated`. |
| 4 | Missing type hints on `_recurring_content` | Applied | Added `Path`/`datetime` hints under a `TYPE_CHECKING` guard. |
| 5 | Criterion #4 "prior runs below the rule" indirect | Applied | Second-run test now asserts two `---` rules in the body. |
| 6 | `created` resets on hand-corrupted note | Rejected | Acceptable fallback — an absent `created:` line can't be recovered; only affects manually-edited notes. |
| 7 | `_split_frontmatter` LF-only | Rejected | `Path.read_text()` applies universal-newline translation; the real path never sees `\r\n`. |
| 8 | "spill/spillover" test vocabulary | Rejected | The over-cap path genuinely still spills-and-links; renaming is churn. No dangling `_spill_to_vault` refs remain. |

**Tests:** `./scripts/test.sh` unit suite (pre-push) — 2052 passed, 8 skipped. Affected files (`test_agent_output_write`, `test_agent_worker_loop`, `test_scheduler_store`) — 105 passed.
**Commits:** `a0d4ef5` (fix), `6055c9d` (docs).

PR #355, opened 2026-06-21
feat(personas): HTTP persona discovery and persona_id for chat API

### LINKED ISSUE #351: feat: HTTP persona discovery and persona_id for chat API
## Problem

LifeOS already supports multi-bot personas for Telegram (`config/telegram_bots.json` + `config/personas/*.md`, forwarded as `persona` on `POST /api/ask/stream`). HTTP clients — especially [whisper-relay](https://github.com/nbramia/whisper-relay) in LifeOS mode — have no way to discover configured personas or select one per turn. Voice and web chat therefore always hit the primary (no persona) path, while Telegram fitness/therapist/doctor bots stay isolated.

whisper-relay cannot read LifeOS config files locally (HTTP-only boundary per its ADR-002). Persona selection needs a LifeOS API surface that auto-picks up new registry entries when tokens are configured.

## Context

- Telegram: each bot listener keeps its own `chat_id → conversation_id` map and passes `persona=self._persona` into `chat_via_api` (`api/services/telegram.py`).
- `AskStreamRequest` already accepts optional `persona: str` (full preamble text, max 8k) and injects it in `run_agent_loop` (`api/routes/chat.py`).
- Registry loader: `settings.telegram_primary_bot` + `settings.telegram_bots` from `config/telegram_bots.json` — skips bots whose `token_env` is unset.
- Conversations are global in SQLite today (`api/services/conversation_store.py`) with no `persona_id` column; Telegram isolation is client-side only.
- `docs/specs/technical/client-surfaces.md` documents whisper-relay as an HTTP consumer; this extends that contract.

### 1. Persona discovery — `GET /api/personas`

Return configured chat personas for HTTP clients (no secrets):

```json
{
  "personas": [
    { "id": "primary", "label": "LifeOS", "capabilities": ["handoff", "agent"] },
    { "id": "fitness", "label": "Fitness" },
    { "id": "therapist", "label": "Therapist" }
  ]
}
```

- Build from `telegram_primary_bot` + `telegram_bots` (same rules as Telegram startup).
- Optional `label` / `description` fields in `telegram_bots.json` later; default label from capitalized `id` is fine for v1.
- Only `primary` advertises `capabilities: ["handoff", "agent"]` — specialized bots are pure chat (matches Telegram primary-only `/claude`, `/agent`, reply-thread hooks).

### 2. Persona selection on ask — `persona_id` on `POST /api/ask/stream`

Add optional `persona_id: str` to `AskStreamRequest`. Server resolves `persona_id` → persona preamble text from the registry (same source as Telegram). Reject unknown ids with 400.

- Keep existing `persona` string field for backward compatibility (Telegram internal client unchanged).
- Precedence: if both sent, define one rule (recommend `persona_id` wins or 400 on conflict).

### 3. Conversation scoping (required for whisper-relay sidebar)

Tag conversations with persona at creation/first turn and filter list/detail:

- Add nullable `persona_id` column to `conversations` (default `"primary"` for existing rows).
- Set on first message in a new thread when `persona_id` is present on ask/stream.
- `GET /api/conversations?persona_id=fitness` — filter to that persona (default `primary` when omitted preserves current web behavior).
- `GET /api/conversations/{id}` — unchanged shape; 404 if id exists but persona mismatch (or document cross-persona access as allowed — pick one).

This lets whisper-relay show persona-specific threads without mixing Telegram/web/voice history incorrectly.

### 4. Documentation

- `docs/specs/product/api-reference.md` — new endpoint + request fields.
- `docs/specs/technical/client-surfaces.md` — whisper-relay persona contract.
- Bump `Last Updated` per `docs/AGENTS.md`.

## Files involved

- `config/settings.py` — helper to list HTTP-visible personas (reuse registry loader)
- `config/telegram_bots.json` — optional `label` field (v1 optional)
- `api/routes/chat.py` — `AskStreamRequest.persona_id`, resolution
- `api/routes/conversations.py` — `persona_id` query param + store filtering
- `api/services/conversation_store.py` — schema migration + persona_id on create
- `tests/test_telegram_multibot.py` (patterns) — new tests for HTTP persona API
- `docs/specs/product/api-reference.md`, `docs/specs/technical/client-surfaces.md`

## Acceptance criteria

- [ ] `GET /api/personas` returns `primary` plus each configured specialized bot from `telegram_bots.json` whose token env is set; omits unset bots
- [ ] Adding a new registry entry + env vars surfaces the new persona after LifeOS restart without code changes
- [ ] `POST /api/ask/stream` with `persona_id: "fitness"` applies the same persona preamble as the fitness Telegram bot
- [ ] Unknown `persona_id` returns HTTP 400 with a clear error
- [ ] New conversations created via ask/stream store `persona_id`; existing conversations backfill as `primary`
- [ ] `GET /api/conversations?persona_id=<id>` returns only that persona's threads
- [ ] Primary persona remains the default when `persona_id` is omitted (no regression for web chat)
- [ ] Contract tests or unit tests cover discovery, resolution, and filtered list
- [ ] API docs and client-surfaces updated

## Downstream

Blocks [whisper-relay persona selector issue](https://github.com/nbramia/whisper-relay) (to be filed after this).

nbramia:
## Summary

Exposes LifeOS's configured chat personas to HTTP clients (web chat, voice/whisper-relay) so they can select a persona per turn instead of always hitting the primary path. Until now only Telegram could route to the fitness/therapist/doctor bots; this extends the same registry to the HTTP contract.

Closes #351

## What changed

- **`GET /api/personas`** — discovery endpoint returning `primary` plus each configured specialized bot whose token env is set. Built from the same registry as Telegram startup (`settings.list_http_personas()`), so adding a `config/telegram_bots.json` entry + its token env surfaces a new persona after restart with no code change. No secrets exposed; only `primary` advertises `handoff`/`agent` capabilities.
- **`persona_id` on `POST /api/ask/stream`** — resolved **before** the SSE stream opens (clean 400, not a mid-stream error) to the exact preamble the matching Telegram bot uses. Unknown id → 400; sending both `persona` and `persona_id` → 400. The internal Telegram `persona` (raw text) path is unchanged.
- **Conversation scoping** — additive `persona_id` column on `conversations` (existing rows backfill to `'primary'` via an idempotent `ALTER TABLE` migration), set on creation via ask/stream, and `GET /api/conversations?persona_id=<id>` filtering. The query param defaults to `primary`, so web chat (which omits it) sees no behavior change.
- **Docs** — `api-reference.md` (new endpoint + request/response fields) and `client-surfaces.md` (whisper-relay persona contract); dates bumped.
- Dropped two pre-existing unused imports (`dataclasses.field`, `format_conversation_history`) in the touched files to pass the pre-commit ruff gate.

## Design decisions

- **Conflict rule:** `persona` + `persona_id` together → 400 (explicit over silent precedence). The two surfaces are mutually exclusive in practice (Telegram sends `persona`, HTTP sends `persona_id`).
- **Detail access:** `GET /api/conversations/{id}` is intentionally **not** persona-scoped (unchanged shape) — documented in api-reference.md.
- **Schema change:** the `persona_id` column is additive and backward-compatible; flagged here per AGENTS.md "Ask first: DB schema changes" — it's exactly what the issue specifies.

## Test evidence

New `tests/test_persona_api.py` (18 tests): discovery helper, resolution (`primary`→`""`, known→preamble, unknown→`None`), `GET /api/personas`, ask/stream 400s (unknown id + conflict), end-to-end `persona_id`→`run_agent_loop` preamble + conversation tagging, persona-scoped store create/get/list, and the backfill migration.

- `tests/test_persona_api.py` — **18 passed**
- Related suites (conversation_store, conversations_api, telegram_multibot, chat_api) — **123 passed**
- Full pre-push unit selection (`-m "unit and not slow"`) — **2058 passed, 8 skipped**

## Review focus

- Persona resolution + the two 400 guards in `ask_stream` (executed before streaming).
- The `conversations` `persona_id` migration / backfill and the `list_conversations` filter (f-string interpolates only a fixed literal; the value is parameterized).
- Default-`primary` behavior on `GET /api/conversations` (no regression for web).

## Notes

- **MCP `lifeos_conversations_list`**: this tool calls `GET /api/conversations`, which now defaults to `persona_id="primary"`. Today every conversation is tagged `primary` (legacy backfill + web + all Telegram bots send raw `persona` text, never `persona_id`), so the effective output is unchanged; once an HTTP client starts sending `persona_id`, this tool will list only the primary persona's threads. The tool definition is unchanged.

### DIFF
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -791,6 +791,35 @@ def find_task_by_topic(
 
 router = APIRouter(prefix="/api", tags=["chat"])
 
+
+class PersonaInfoResponse(BaseModel):
+    """A single HTTP-visible chat persona (no secrets)."""
+    id: str
+    label: str
+    capabilities: list[str] = []
+
+
+class PersonasResponse(BaseModel):
+    """Response for the persona discovery endpoint."""
+    personas: list[PersonaInfoResponse]
+
+
+@router.get("/personas", response_model=PersonasResponse)
+async def list_personas():
+    """List chat personas available to HTTP clients (web, voice/whisper-relay).
+
+    Returns the primary persona plus each configured specialized bot. Adding a
+    registry entry + its token env var surfaces a new persona after restart with
+    no code change. Only the primary advertises handoff/agent capabilities.
+    """
+    return PersonasResponse(
+        personas=[
+            PersonaInfoResponse(id=p.id, label=p.label, capabilities=p.capabilities)
+            for p in settings.list_http_personas()
+        ]
+    )
+
+
 # Attachment configuration
 ALLOWED_MEDIA_TYPES = {
     # Images - 5MB each
@@ -849,6 +878,11 @@ class AskStreamRequest(BaseModel):
     conversation_id: Optional[str] = None
     attachments: Optional[list[Attachment]] = None
     persona: Optional[str] = None
+    # HTTP clients (web, voice) select a persona by id; the server resolves it to
+    # the same preamble the matching Telegram bot uses. The raw `persona` field
+    # above is the internal Telegram path (full preamble text); the two are
+    # mutually exclusive (sending both is a 400 in the route handler).
+    persona_id: Optional[str] = None
 
     @field_validator("persona")
     @classmethod
@@ -981,6 +1015,27 @@ async def ask_stream(request: AskStreamRequest):
     if not request.question.strip():
         raise HTTPException(status_code=400, detail="Question cannot be empty")
 
+    # Resolve persona BEFORE the SSE stream opens so an invalid selection is a
+    # clean 400 rather than a mid-stream error. `persona` (raw preamble text) is
+    # the internal Telegram path; `persona_id` is the HTTP-client path resolved
+    # against the same registry. They are mutually exclusive.
+    persona_preamble = request.persona or ""
+    new_conversation_persona_id = "primary"
+    if request.persona_id is not None:
+        if request.persona is not None:
+            raise HTTPException(
+                status_code=400,
+                detail="Provide either persona_id or persona, not both",
+            )
+        resolved = settings.resolve_persona(request.persona_id)
+        if resolved is None:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Unknown persona_id: {request.persona_id!r}",
+            )
+        persona_preamble = resolved
+        new_conversation_persona_id = request.persona_id
+
     async def generate():
         try:
             # Get or create conversation
@@ -988,8 +1043,9 @@ async def ask_stream(request: AskStreamRequest):
             conversation_id = request.conversation_id
 
             if not conversation_id:
-                # Create new conversation
-                conv = store.create_conversation()
+                # Create new conversation, tagged with the selected persona so
+                # persona-scoped listing (e.g. the voice sidebar) can filter it.
+                conv = store.create_conversation(persona_id=new_conversation_persona_id)
                 conversation_id = conv.id
                 # Generate title from question
                 title = generate_title(request.question)
@@ -1224,7 +1280,7 @@ async def ask_stream(request: AskStreamRequest):
                 model_tier=orchestrator_model,
                 max_tool_rounds=5,
                 model=orchestrator_model if escalated else "",
-                persona=request.persona or "",
+                persona=persona_preamble,
             ):
                 if event["type"] == "text":
                     yield f"data: {json.dumps({'type': 'content', 'content': event['content']})}\n\n"
--- a/api/routes/conversations.py
+++ b/api/routes/conversations.py
@@ -13,7 +13,7 @@ from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
 
 from api.services.conversation_store import (
-    get_store, generate_title, format_conversation_history
+    get_store, generate_title
 )
 from api.services.hybrid_search import HybridSearch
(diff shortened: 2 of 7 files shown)


nbramia:

Independent adversarial review completed; findings refereed against the code.

| 1 | `lifeos_conversations_list` MCP tool (→ `GET /api/conversations`) now defaults to `persona_id="primary"`, no longer returning all conversations | Recommended | Accept as doc note | Verified accurate. But the `primary` default is exactly what #351 specifies, the MCP tool *definition* is unchanged, and practical impact today is nil — every conversation is tagged `primary` (legacy backfill + web + all Telegram bots, which send raw `persona` text, never `persona_id`). Documented in the PR body rather than changing code (an "all-personas" option would be scope creep beyond #351). |
| 2 | Conflict guard `if request.persona is not None` trips 400 on `persona=""` + `persona_id` | Minor | Reject | Unreachable: `chat_via_api` only adds `persona` to the body when truthy, so no client sends `persona=""`. Strict guard is defensible. |

Reviewer verified solid (no action): persona resolution + dual 400 guards run before the SSE stream opens; additive SQLite migration is idempotent and backfills existing rows to `primary`; the `list_conversations` filter is parameterized (no injection); web chat and the internal Telegram client are unaffected; `GET /api/personas` leaks no secrets; orchestrates→capabilities is consistent across code/docs/tests.

**Findings forwarded to addresser:** 0 (one documentation note applied to the PR body).

PR #373, opened 2026-06-22
feat(memories): hybrid semantic + keyword recall for the memory store

### LINKED ISSUE #370: feat(memories): hybrid (semantic + keyword) recall for the memory store
## Problem

Persistent-memory recall (`MemoryStore.search_memories` / `get_relevant_memories` in `api/services/memory_store.py:349`) is **keyword-overlap only**, gated at 15% term overlap. A memory saved as "I prefer terse replies" never surfaces for a query phrased "keep it short" — no shared literal tokens, no match. Because relevant memories are injected into the system prompt every chat turn (`api/services/agent_loop.py:416`), a missed match silently degrades every downstream answer. We want **semantic recall in addition to keyword**, not as a replacement.

## Context

Pure-semantic would *regress* the store's flagship use case: alias/name-spelling memories ("Sarah may also be spelled Sara"), where exact-token keyword matching is the strength and dense vectors are weak. The correct target is **hybrid recall** — the same vector + keyword, RRF-fused pattern LifeOS already runs on its main corpus (ADR-004).

The infrastructure to reuse already exists and is proven in-use:
- `get_embedding_service()` → `embed_text()` / `embed_texts()` (`api/services/embeddings.py`) — singleton with GPU→CPU fallback; single-query embedding is already the `/api/search` hot path.
- `reciprocal_rank_fusion(vector_results, bm25_results, k=60)` (`api/services/hybrid_search.py:156`) — standalone, tested, fuses two ranked lists by rank.

## Suggested approach

Upgrade `MemoryStore.search_memories` to hybrid; keep the public signature so callers (`get_relevant_memories`, the agent `search_memories` tool, `/api/memories/search`) are unchanged.

1. **Embed in-process — do NOT use ChromaDB.** The store is tiny (single-digit to low-hundreds of memories). In-process numpy cosine over all active memories is sub-100ms even at 1000. A ChromaDB collection would add a sync burden on every create/update/delete/soft-delete plus a network hop on a per-turn path, for no benefit at this scale.
2. **Cache per-memory embeddings in a sidecar** keyed by `memory_id + content_hash + model_name`. Do **not** inline float vectors into `memories.json` — its human-readability is a deliberate design choice (`memory_store.py:1-9`). Embed lazily on create/update; one-time backfill for existing memories; the content-hash + model-name key auto-invalidates on edit or embedding-model change.
3. **Query-time:** embed the query once → cosine vs. cached vectors → semantic ranking; keep the existing keyword-overlap ranking; **fuse the two rankings with the existing `reciprocal_rank_fusion`** (rank-based fusion avoids the cosine-vs-overlap score-scale mismatch — do not weighted-sum the raw scores).
4. **Apply a precision floor.** Memories are injected unconditionally, so a junk match pollutes the prompt. Replace the current 15% keyword gate with an RRF top-N plus a minimum cosine floor (~0.3–0.4, tuned). Favor precision over recall.
5. **Graceful degradation:** if the embedding service is unavailable, fall back to the current keyword-only path (mirrors the rest of the codebase).

### Risks / decisions to settle during implementation
- **Per-turn embed cost:** `get_relevant_memories` runs every turn. Skip the embed entirely when the store is empty; optionally reuse the query embedding if the turn also runs a corpus search.
- **VRAM contention:** when the local LLM is loaded, embedding can OOM (GPU→CPU fallback covers correctness; CPU embed of one short query is a few hundred ms).
- **Threshold tuning:** the new RRF top-N + cosine floor needs tuning against real memories.

## Files involved

- `api/services/memory_store.py` — hybrid search, sidecar embedding cache, backfill (primary change)
- `api/services/embeddings.py` — reuse only (`get_embedding_service`)
- `api/services/hybrid_search.py` — reuse only (`reciprocal_rank_fusion`)
- `tests/test_memory_store.py` — new tests for semantic matches, hybrid fusion, precision floor, keyword fallback, cache invalidation

## Acceptance criteria

- [ ] A semantically-related query with zero literal token overlap (e.g. "keep it short" vs. a "I prefer terse replies" memory) returns the memory
- [ ] An alias/spelling memory ("Sara"/"Sarah") still matches via the keyword signal (no regression vs. current behavior)
- [ ] Ranking is produced by RRF over the semantic and keyword rankings, not a raw weighted score-sum
- [ ] A vague/unrelated query does NOT inject low-relevance memories (precision floor enforced)
- [ ] Per-memory embeddings are cached in a sidecar (not inlined into `memories.json`) and recomputed only on content change or model change
- [ ] Existing memories are backfilled with embeddings on first use; `memories.json` remains human-readable
- [ ] Embedding-service failure degrades to the existing keyword-only path
- [ ] Public signatures of `search_memories` / `get_relevant_memories` are unchanged; existing callers and tests pass
- [ ] `./scripts/test.sh` passes

nbramia:
## Summary

Memory recall (`MemoryStore.search_memories`) was keyword-overlap only, so a query like "keep it short" never surfaced a "prefers terse replies" memory — and these memories are injected into the chat prompt every turn, so each miss silently degrades answers. This adds **hybrid recall**: embedding cosine similarity fused with the existing keyword overlap via Reciprocal Rank Fusion, reusing the same pattern LifeOS already uses for its main corpus (ADR-004).

Closes #370

## What changed

- **`search_memories` is now hybrid**: keyword candidates (≥ `min_relevance`) ∪ semantic candidates (cosine ≥ `MEMORY_SEMANTIC_FLOOR`), fused with the existing `reciprocal_rank_fusion`. Keyword keeps its edge on the alias/spelling memories this store is built for; semantic catches paraphrases. Memories matched by **both** signals rank highest.
- **Precision floor** (`MEMORY_SEMANTIC_FLOOR = 0.35`, tunable): memories are injected every turn, so vague queries with no keyword overlap and sub-floor cosine return nothing.
- **Sidecar embedding cache** `<memories>_embeddings.json`, keyed by content hash + model name. Vectors never touch the human-readable `memories.json`; recomputed only on content change (per-entry hash) or model swap (whole-cache). Pruned to active memory IDs.
- **Graceful degradation**: if the embedding service is unavailable (e.g. no model cached on a fresh install), search falls back to the original keyword-only path unchanged.
- Reuses `get_embedding_service` and `reciprocal_rank_fusion` — no new dependencies, no ChromaDB (the store is tiny; in-process numpy cosine is sub-100ms).
- Public signatures of `search_memories` / `get_relevant_memories` are unchanged.

Also removed pre-existing unused imports/locals in the two touched files to satisfy the ruff pre-commit hook (it lints full changed files).

## Test evidence

- `tests/test_memory_store.py`: **72 passed** (8 new hybrid tests + 64 existing). New tests cover semantic match with zero keyword overlap, alias/keyword no-regression, RRF dual-match ranking, precision floor, keyword fallback on embedding failure, `memories.json` stays vector-free, and cache invalidation on content/model change. New tests use a deterministic fake embedding service; legacy search tests are forced keyword-only so they stay hermetic.
- Full `unit and not slow` suite: **2123 passed, 8 skipped**. This includes `tests/test_memories.py` and `tests/test_audit_e2e.py::TestMemoryRelevanceFiltering` which exercise `search_memories` against the **real** embedding model — confirming the hybrid path and the 0.35 floor behave correctly in production, not just under the fake.

## Review focus

- **`MEMORY_SEMANTIC_FLOOR = 0.35`** — the precision/recall knob. Validated by the real-model relevance-filtering test, but may want tuning against a larger memory set; instruction-tuned embedding models can have elevated baseline cosine.
- **Per-turn cost**: `get_relevant_memories` runs every turn and now embeds the query (one forward pass). Short-circuits when the store is empty; the warm cache means memory vectors aren't recomputed. In production the embedding model is already resident for corpus search.
- **Embedding cache concurrency**: read-modify-write of the sidecar mirrors the existing non-atomic `_save` pattern; worst case is a redundant recompute, not corruption.

### DIFF
--- a/api/services/memory_store.py
+++ b/api/services/memory_store.py
@@ -8,11 +8,12 @@ Storage: Human-readable JSON file at ~/.lifeos/memories.json
 - Can be pre-populated with personal context
 - Auto-loaded on startup
 """
+import hashlib
 import json
 import logging
 import re
 import uuid
-from dataclasses import dataclass, asdict
+from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Optional
@@ -21,6 +22,11 @@ logger = logging.getLogger(__name__)
 
 # Default storage path - JSON file for human readability
 DEFAULT_MEMORIES_PATH = Path.home() / ".lifeos" / "memories.json"
+
+# Minimum cosine similarity for a memory to count as a semantic match. Because
+# relevant memories are injected into the chat prompt every turn, this floor
+# keeps vague/unrelated queries from pulling in low-relevance memories. Tunable.
+MEMORY_SEMANTIC_FLOOR = 0.35
 
 # Memory categories and their trigger patterns
 CATEGORY_PATTERNS = {
@@ -81,8 +87,6 @@ def categorize_memory(content: str) -> str:
     Returns:
         Category string (people, preferences, facts, decisions, reminders, context)
     """
-    content_lower = content.lower()
-
     # Check each category's patterns
     for category, patterns in CATEGORY_PATTERNS.items():
         for pattern in patterns:
@@ -168,6 +172,14 @@ class MemoryStore:
 
         # Ensure directory exists
         self.file_path.parent.mkdir(parents=True, exist_ok=True)
+
+        # Sidecar cache of memory embeddings, kept next to the memories file.
+        # Pure cache (regenerable from memories.json) — kept separate so the
+        # human-readable memories file never carries raw float vectors.
+        self.embeddings_path = self.file_path.with_name(
+            self.file_path.stem + "_embeddings.json"
+        )
+        self._embedding_cache: dict = self._load_embedding_cache()
 
         # Load or initialize memories
         self._memories: dict[str, Memory] = {}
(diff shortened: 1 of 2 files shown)


nbramia (commented):
## PR Review: feat(memories): hybrid semantic + keyword recall for the memory store

Reviewed adversarially with independent verification (executable scenarios against a fake embedding service + temp stores, and the real model in this env). The 72 memory tests pass and the core design (in-process numpy cosine, sidecar cache, RRF reuse, graceful fallback) is sound and faithful to issue #370 and ADR-004. Findings below; none are merge-blockers, but #1 and the floor/test gaps are worth addressing.

### Action Required
- **[Correctness]** Silent total degradation via the broad `except` — `_semantic_scores` (`api/services/memory_store.py:418-420, 438, 444-446`). `fresh = service.embed_texts([...])` is zipped against `pending` with no length check; if `embed_texts` short-returns (partial GPU/CPU encode), `zip` truncates silently, the un-embedded memory is never written to `vectors`, then line 438 `pruned[memory.id]["vector"]` raises `KeyError`, which the catch-all `except Exception` at 444 converts to `return None` → the entire search silently falls back to keyword-only with a log line indistinguishable from "service unavailable." Verified: a fake returning 1 vector for 2 pending disables semantic recall for *all* queries until content/model changes. Add `assert len(fresh) == len(pending)` (or guard the lookup) and consider narrowing/branching the `except` so an internal bug (zip truncation, dimension mismatch — see Minor) is not silently swallowed as "embeddings unavailable."

### Recommended
- **[Correctness]** No precision gate on semantic-only injections beyond the floor — `api/services/memory_store.py:500-510`. A memory matching *only* semantically at barely-above-floor cosine lands at semantic rank 0 → RRF `1/61`, tying or outranking a strong keyword-only match (also rank 0 → `1/61`); tie broken by dict insertion order. Verified: a no-keyword memory at cos 0.40 sorts ahead of an exact keyword hit. Since these go into the prompt every turn, a cluster of mediocre semantic matches (0.36-0.40) can crowd out precise keyword hits. Consider a semantic-only sub-cap, keyword corroboration for injection, or a higher floor.
- **[Correctness]** Non-atomic sidecar write on a process-wide singleton — `api/services/memory_store.py:382-389`. Two concurrent chat-turn searches both mutate `self._embedding_cache` and `open("w")+json.dump` the same file. Lost-update is benign (redundant recompute), but a torn write can leave invalid JSON on disk mid-stream — so the PR's "no corruption" claim is too strong for the on-disk file. It self-heals (next `_load_embedding_cache` reads it as empty and rebuilds — verified), but `tempfile + os.replace` removes the window cheaply.
- **[Requirements/Tests]** Floor value is essentially untested — `MEMORY_SEMANTIC_FLOOR = 0.35` (`api/services/memory_store.py:29`). All 8 hybrid tests pass for any floor in **(0.30, 0.96]**; only the lower edge (`test_precision_floor_blocks_unrelated`, ~0.30) and upper edge (the 0.96 semantic match) are pinned. The constant could drift to 0.80 undetected. Add a test asserting a cosine just above the floor (e.g. 0.40, no keyword overlap) IS returned.
- **[Tests]** `test_rrf_ranks_dual_matches_first` (`tests/test_memory_store.py:639-659`) does not actually distinguish RRF from a naive union/intersection ranker — it asserts only `ids[0]` and the result *set*, never the relative order of the two single-signal matches (the one place `Σ1/(k+rank)` diverges from a trivial heuristic). A naive "intersection-first" ranker passes it unchanged. Strengthen it to assert single-signal ordering or a scenario where RRF and intersection-first diverge.
- **[Standards]** `MEMORY_SEMANTIC_FLOOR` is a module-level magic constant, but the comment calls it "Tunable" — every other threshold in the codebase (e.g. `reranker_enabled`, `LIFEOS_AGENT_COST_CONFIRM_THRESHOLD_DOLLARS` at `config/settings.py:263`) is a `settings.py` `Field(..., alias="LIFEOS_...")`. Move it to `config/settings.py` to make it tunable without a source edit.

### Minor
- **[Correctness]** No embedding-dimension guard — `api/services/memory_store.py:438-442`. The model-name check (line 408) does not catch dimension drift under a stable name (or a corrupt cache entry). A mismatched cached vector makes `np.dot` raise `ValueError` → same silent keyword-only downgrade as the Action Required item.
- **[Tests]** Real-model coverage is shallow. `test_low_relevance_memories_excluded` (`tests/test_audit_e2e.py:1105-1116`) loads the real model (confirmed) but asserts only `any("budget" ...)` — it proves the pipeline runs without crashing, not exclusion or ranking quality, and the assertion is satisfied by keyword overlap alone. On a fresh machine without the model cached, this test (and the PR's "confirms the 0.35 floor in production" claim) silently exercises the keyword-only fallback. The floor's production behavior is effectively unverified by an assertion.
- **[Tests]** Legacy keyword tests (`TestMemoryStoreSearch`, `TestMemoryWorkflow`) now run with `get_embedding_service` forced to raise, so they exercise the *fallback* branch, not the keyword half of the live hybrid path. Documented in docstrings and acceptable under "Tests Are Sacred," but the keyword-within-RRF path is only covered by fake vectors.
- **[Tests]** `temp_json_file` fixture cleans up `memories.json` but not the sibling `_embeddings.json` sidecar, orphaning files in the temp dir. No correctness impact (unique mkstemp path per test); cosmetic.

Verified clean: keyword-only fallback ordering is byte-equivalent to the prior behavior; the `dirty` flag has no stale-sidecar sequence (delete+add is caught because new memories are always in `pending`); zero-norm query/vector edge cases are handled; empty-query `return []` is safe for all callers; removed imports/locals are genuinely unused; no MCP/API/dependency/schema changes.

### Summary
Solid, well-scoped implementation that meets all 8 acceptance criteria functionally — **needs minor changes**: tighten the catch-all that silently disables semantic recall (#1), and close the floor/RRF test gaps so the precision knob and fusion are actually pinned by assertions rather than passing under a wide band.

nbramia:

| 1 | Catch-all `except` + unchecked `zip`: a short `embed_texts` return drops a memory from the cache, the cosine loop `KeyError`s, and the broad except masks it as a full keyword-only fallback | Action Required | **Accept** | Verified the path. Low probability (the embed contract is reliable) but it's the exact silent-degradation this design avoids, and cheap to harden. |
| 2 | Non-atomic sidecar write can leave invalid JSON on a torn write | Recommended | **Accept** | Self-heals on reload (pure cache), but `tempfile + os.replace` removes the transient broken state entirely. |
| 3 | Floor inclusion side untested — suite passes for any floor in (0.30, 0.96] | Recommended | **Accept** | Adding a just-above-floor (0.40) semantic-only test brackets the floor to (0.30, 0.40) with the existing 0.30-excluded test. |
| 4 | `test_rrf_ranks_dual_matches_first` doesn't fully distinguish RRF from a naive ranker | Recommended | **Accept (downgraded to Minor)** | `ids[0]==both` already rules out a naive semantic-sort, but I'll add explicit rank-order assertions. |
| 5 | `temp_json_file` fixture leaks the `_embeddings.json` sidecar | Minor | **Accept** | Test hygiene; trivial fix. |
| 6 | A barely-above-floor semantic-only match can tie/outrank an exact keyword hit after RRF | Recommended | **Downgrade to Minor — no code change** | Inherent to RRF-by-rank and consistent with the main corpus hybrid search; the cosine floor + tuning is the intended precision lever. Overriding it would diverge from the established pattern. |
| 7 | Migrate `MEMORY_SEMANTIC_FLOOR` to a `config/settings.py` `LIFEOS_*` Field | Recommended | **Reject** | "Unlike every other threshold" is inaccurate — `min_relevance=0.15`, RRF `k=60`, and recency `max_boost=0.5` are all hardcoded constants. A universal hardcoded default was the chosen design; a setting adds surface the issue didn't request. |
| 8 | Dimension guard; real-model test asserts only non-crash; legacy tests exercise the fallback branch | Minor | **Reject / note** | Dimension drift can't occur under model-name cache keying; the real-model test is pre-existing and out of scope. |

**Findings forwarded to addresser:** 5 (items 1–5). Addressing directly in the orchestrator context given their small, well-understood scope, then re-running the full unit suite to verify.

nbramia:
## Review Round 1 — Addressed (commit b2d628d)

| 1 | Catch-all `except` + unchecked `zip` → silent total fallback | Applied | `embed_texts` return is length-checked (warns on mismatch); cosine loop uses `pruned.get(id)` and skips any un-embedded memory, so a partial service response degrades to "score what we have" instead of `KeyError`→full keyword fallback. |
| 2 | Non-atomic sidecar write | Applied | `_save_embedding_cache` now writes a temp file then `os.replace` (atomic on same dir) — no truncated JSON on a torn write. |
| 3 | Floor inclusion side untested | Applied | Added `test_semantic_match_just_above_floor_is_returned` (cosine 0.40, no keyword overlap → returned); with the existing 0.30-excluded test this brackets `MEMORY_SEMANTIC_FLOOR` to (0.30, 0.40). |
| 4 | RRF test didn't distinguish naive ranker | Applied | Added explicit rank-order assertions: `both` strictly outranks both single-signal matches. |
| 5 | `temp_json_file` fixture leaked the sidecar | Applied | Fixture now removes the `_embeddings.json` sidecar too. |
| 6 | Semantic-only can tie/outrank exact keyword hit | Not changed (downgraded) | Inherent to RRF-by-rank, consistent with the corpus hybrid search; floor + tuning is the precision lever. |
| 7 | Migrate floor to `LIFEOS_*` setting | Rejected | Hardcoded constant matches the `min_relevance` / `k=60` / `max_boost` precedent; chosen design. |

**Tests:** `pytest -m "unit and not slow" tests/` — 2124 passed, 8 skipped. `tests/test_memory_store.py` — 73 passed. ruff clean.

Self-verified from the orchestrator context: only the two original files changed, every change traces to a forwarded finding, no partial fixes. Review loop complete.

Issue #383, opened 2026-06-22
refactor(prompts): consolidate orchestrator instruction layer


nbramia:
## Problem

A full audit of how the chat orchestrator assembles its system prompt / instructions / tools surfaced four "consolidate & simplify" issues: prompt caching is effectively **off** (so ~10k tokens of stable scaffolding are re-sent uncached every turn), the static prompt has accumulated redundancy and a couple of stale/hard-coded details, the inter-agent delegation blurb is copy-pasted across three worker executors, and a second answer-generation path (`synthesizer` + `query_router`) runs in parallel to the agentic loop with unclear remaining scope. None of this changes behavior, but it costs tokens/latency and raises the maintenance burden.

This issue is the **consolidate/simplify track**. Persona-level specificity (per-persona tool scoping, voice-vs-text, `primary.md`, worker persona-awareness) is intentionally **out of scope** and will be tracked in a separate issue.

## Context

From the instruction/prompt audit (verified against the code):

**1. Prompt caching is dead on the Anthropic path.**
- `api/services/llm_client.py::_extract_system` (~L433-440) flattens the system block list into a single `"\n\n".join(...)` **string** before the SDK call (`astream` ~L512-514, `create` ~L457-459), so the `cache_control: {"type":"ephemeral"}` markers set in `agent_system_prompt.py::build_system_prompt` never reach the API.
- `api/services/agent_loop.py` (~L480-488) then **explicitly strips** `cache_control` from every tool definition (`# Strip cache_control from tool definitions (Anthropic-specific)`), so the marker on the last tool in `agent_tools.py` (~L727) is also dead.
- Net: the ~2.6k-token static prompt **and** the ~7.5k-token tool definitions are re-sent uncached on every turn *and* every tool round.
- ⚠️ Treat as **restore/verify + measure**, not a blind "turn it on" — it may have been disabled deliberately. Anthropic prompt caching requires `system` as a list-of-blocks carrying `cache_control`; the static prompt + tool defs are stable across turns so they're good cache candidates. Must not break the local/OpenAI backend (`LocalLLMClient` legitimately flattens — OpenAI has no prompt caching).

**2. Static-prompt redundancy/staleness** in `agent_system_prompt.py` (`_STATIC_PROMPT_TEMPLATE`, ~L17-135):
- `person_info` documented twice; the "never say you can't search the web" directive repeated ~3×; the email `draft → show → confirm → send` flow stated twice (tool desc + multi-tool patterns).
[...]

PR #384, opened 2026-06-22
feat(chat): restore orchestrator prompt caching on the Anthropic backend

nbramia:
## Summary

Restores Anthropic prompt caching for the chat orchestrator. The static system prompt (~2.6k tokens) and tool definitions (~7.5k tokens) were being re-sent **uncached on every turn and every tool round** because two cache breakpoints were silently dead:

1. `AnthropicLLMClient` flattened the `system` block list into a single `"\n\n".join(...)` string before the SDK call, dropping the `cache_control` markers set in `build_system_prompt`.
2. `agent_loop` explicitly stripped `cache_control` off every tool definition before the call, killing the marker on the last tool in `agent_tools.py`.

On top of that, `LLMUsage` carried no cache-token fields, so even working caching was invisible — and `AgentResult` already declared `total_cache_read_tokens`/`total_cache_creation_tokens` that nothing ever populated.

This is **Phase 1** of #383 (the isolated, measurable phase). Phases 2–4 follow in separate PRs.

Relates to #383.

## What changed

- `AnthropicLLMClient._prepare_system` (was `_extract_system`) now forwards the `system` value to the SDK **unchanged** — a block list keeps its `cache_control` markers. `LocalLLMClient` still flattens to a string (OpenAI has no prompt caching).
- `agent_loop` stops stripping tool `cache_control`. The local backend strips it itself in `_anthropic_tools_to_openai`, so only the Anthropic path keeps the marker.
- `LLMUsage` gains `cache_creation_input_tokens` / `cache_read_input_tokens`, populated from the Anthropic usage object in both the streaming and non-streaming paths. `_track_usage` feeds them into the existing `AgentResult` cache totals, making caching measurable.

## Test evidence

`tests/test_llm_client.py` + `tests/test_agent_loop_caching.py` (new):
- system block list forwarded to the SDK **unflattened**, `cache_control` intact (the core bug);
- plain-string system still forwarded as-is (backward compat);
- streaming path forwards tool `cache_control` and reports `cache_read` tokens on `done`;
- `run_agent_loop` surfaces `cache_read` tokens on `AgentResult` (a warm 2nd turn → `cache_read > 0`).

Pre-push unit suite: **2140 passed, 8 skipped**. (14 repo-wide `require_db`/`slow` data-integrity failures are pre-existing and env-dependent — they fail identically on a clean tree and are outside the `unit` selection.)

## Review focus

- Correctness of passing `system` as a block list straight to the Anthropic SDK (`messages.create` / `messages.stream`) with `cache_control` preserved.
- That the local/OpenAI backend is genuinely unchanged (still flattens; `_anthropic_tools_to_openai` still strips `cache_control`).
- `tools = TOOL_DEFINITIONS` now passes the module-level list by reference — verified nothing mutates it downstream.

### DIFF
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -474,18 +474,15 @@ async def run_agent_loop(
     def _track_usage(usage: LLMUsage):
         result.total_input_tokens += usage.input_tokens
         result.total_output_tokens += usage.output_tokens
+        result.total_cache_read_tokens += usage.cache_read_input_tokens
+        result.total_cache_creation_tokens += usage.cache_creation_input_tokens
         # Local model has no cost
         result.total_cost_usd = 0.0
 
-    # Strip cache_control from tool definitions (Anthropic-specific)
-    tools = []
-    for t in TOOL_DEFINITIONS:
-        tool_copy = dict(t)
-        tool_copy.pop("cache_control", None)
-        schema = dict(tool_copy.get("input_schema", {}))
-        schema.pop("cache_control", None)
-        tool_copy["input_schema"] = schema
-        tools.append(tool_copy)
+    # Pass tool definitions through with their cache_control marker intact so
+    # Anthropic caches the large, stable tool schema across turns and rounds.
+    # The local backend strips cache_control itself in _anthropic_tools_to_openai.
+    tools = TOOL_DEFINITIONS
 
     for round_num in range(1, max_tool_rounds + 1):
         print(f"[agent] Round {round_num}/{max_tool_rounds} starting")
--- a/api/services/llm_client.py
+++ b/api/services/llm_client.py
@@ -23,10 +23,16 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class LLMUsage:
-    """Token usage from an LLM response."""
+    """Token usage from an LLM response.
+
+    The cache fields are populated only on the Anthropic backend (prompt
+    caching); they stay 0 for the local/OpenAI backend, which has no caching.
+    """
     input_tokens: int = 0
     output_tokens: int = 0
     total_tokens: int = 0
+    cache_creation_input_tokens: int = 0
+    cache_read_input_tokens: int = 0
 
 
 @dataclass
@@ -430,13 +436,15 @@ class AnthropicLLMClient:
         self._sync_client = anthropic.Anthropic(api_key=self._api_key)
         self._async_client = anthropic.AsyncAnthropic(api_key=self._api_key)
 
-    def _extract_system(self, system: str | list | None) -> str | None:
-        if system is None:
-            return None
-        if isinstance(system, list):
-            return "\n\n".join(
-                block["text"] for block in system if block.get("type") == "text"
-            )
+    def _prepare_system(self, system: str | list | None) -> str | list | None:
+        """Return the ``system`` value for the Anthropic SDK unchanged.
+
+        A list of content blocks is forwarded as-is so the ``cache_control``
+        markers set in build_system_prompt survive to the API and prompt
+        caching stays active across turns. A plain string is returned as-is.
+        (LocalLLMClient flattens a block list to a string instead — the
+        OpenAI-compatible backend has no prompt caching.)
+        """
         return system
 
     def create(
@@ -454,9 +462,9 @@ class AnthropicLLMClient:
             "messages": messages,
             "max_tokens": max_tokens,
         }
-        sys_text = self._extract_system(system)
-        if sys_text:
-            kwargs["system"] = sys_text
+        system_param = self._prepare_system(system)
+        if system_param:
+            kwargs["system"] = system_param
         if tools:
             kwargs["tools"] = tools
         if temperature is not None:
@@ -480,9 +488,9 @@ class AnthropicLLMClient:
             "messages": messages,
             "max_tokens": max_tokens,
         }
-        sys_text = self._extract_system(system)
-        if sys_text:
-            kwargs["system"] = sys_text
+        system_param = self._prepare_system(system)
+        if system_param:
+            kwargs["system"] = system_param
         if tools:
             kwargs["tools"] = tools
         if temperature is not None:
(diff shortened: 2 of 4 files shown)


nbramia (commented):
## PR Review: feat(chat): restore orchestrator prompt caching on the Anthropic backend

Reviewed adversarially with correctness + standards/test-quality specialists. Branch checked out, anthropic SDK 0.84.0 confirmed, all 41 tests in the two touched test files pass locally. The core fix is correct and surgical.

### Action Required
None. The change is correct on every axis I checked.

### Recommended
- **[Test fidelity]** `tests/test_agent_loop_caching.py:25-26` — `_FakeClient.astream(...)` has a trailing `**kwargs`, but the real `AnthropicLLMClient.astream` (`api/services/llm_client.py:502-510`) has no `**kwargs` and no `timeout` param. The fake silently swallows extra/renamed kwargs, so a signature drift between `run_agent_loop` and the real client would still pass. It already masks a latent (pre-existing, out-of-scope) mismatch: the synthesis round calls `client.astream(..., timeout=180)` at `agent_loop.py:635`, which the real Anthropic `astream` doesn't accept. Dropping `**kwargs`/`timeout` from the fake to mirror the real keyword-only signature would turn this into a real contract test. (The `test_llm_client.py` boundary tests do independently exercise the real client, which is why this is Recommended, not Action Required.)
- **[Test coverage]** `tests/test_agent_loop_caching.py` — only `total_cache_read_tokens` is asserted on `AgentResult` (line 75). `_track_usage` accumulates both cache fields (`agent_loop.py:477-478`) and `_FakeClient` already accepts a `cache_creation` kwarg, but no test passes it through to assert `total_cache_creation_tokens` surfaces. One-line symmetric-coverage gap.

### Minor
- **[Docs/naming]** `test_cache_read_tokens_surface_in_agent_result` (`tests/test_agent_loop_caching.py:69-71`) and the module docstring (lines 6-8) frame the assertion as proving a "warm 2nd turn." It actually proves the *plumbing* (a `done` event reporting `cache_read=2600` flows into `AgentResult`), via one synthetic round — it does not prove Anthropic caches across two real turns. That's the correct scope for a unit test; the wording just over-reads it.
- **[Test precision]** The local-backend tool strip is covered for the nested `input_schema.cache_control` (`tests/test_llm_client.py:41-55`); the top-level tool `cache_control` is dropped by construction (`_anthropic_tools_to_openai` rebuilds from `name`/`description`/`parameters` only, `llm_client.py:61-67`) rather than by explicit assertion. Acceptable.

### Verified clean
- **Block-list `system` is the SDK's native shape.** SDK 0.84.0 `MessageCreateParams.system` is `Union[str, Iterable[TextBlockParam]]`, and both `TextBlockParam` and `ToolParam` carry `cache_control`. Forwarding the unflattened list with markers intact is correct, not a hack.
- **`tools = TOOL_DEFINITIONS` (by reference) is safe.** Empirically confirmed `_anthropic_tools_to_openai` shallow-copies `input_schema` before `.pop` (`llm_client.py:59-60`) and builds fresh dicts — it never mutates the originals. The Anthropic path hands the list to the SDK, which serializes (no in-place mutation). The only write to `TOOL_DEFINITIONS` is at import time (`agent_tools.py:727`). No deep copy needed.
- **`getattr(usage, attr, 0) or 0`** correctly yields 0 for None (Anthropic's null cache fields when caching is off), 0 for a missing attribute, and passes real integers through. No None-handling bug.
- **No double-counting / no lost synthesis usage.** `usage_this_round` resets per round outside the retry loop (`agent_loop.py:492`), so retried attempts count once; `_track_usage` is called once per round (line 535) and once for the synthesis round (line 646). The synthesis round still cache-reads the marked system prompt; it omits `tools` (pre-existing behavior on `main`, not a regression).
- **Local/OpenAI backend genuinely unchanged** — `LocalLLMClient._build_messages_list` still flattens the block list to a string (`llm_client.py:211-214`) and `_anthropic_tools_to_openai` still strips `cache_control` (lines 59-60). Both invariants are test-covered.
- **Boundary assertions are genuine, not tautological** — `TestAnthropicCaching` swaps the real `_sync_client`/`_async_client` for a fake and calls the real `create`/`astream`, capturing the kwargs handed to `messages.create`/`messages.stream`, proving `cache_control` reaches the SDK call boundary on both paths plus the string-system backward-compat path.
- **Surgical & scoped to Phase 1** — touches only the 4 stated files; ~26 net production lines, the rest tests. No unrelated refactors. `_extract_system` → `_prepare_system` rename is complete (zero stragglers). MCP "Ask-first" guardrail not tripped — this changes whether the existing `cache_control` marker is *forwarded*, not any tool's schema.

### Summary
Merge-ready. The two Recommended items (tighten the fake `astream` signature; add a `cache_creation` surfacing assertion) are cheap test hardening, not blockers.

nbramia:

| 1 | `_FakeClient.astream` `**kwargs`/`timeout` masks signature drift vs. the real `AnthropicLLMClient.astream` | Recommended | **Accept** | Tightened the fake to mirror the real Anthropic `astream` signature exactly (no `**kwargs`, no `timeout`), so future drift surfaces as a test failure. |
| 1b | Underlying pre-existing bug: synthesis round passes `timeout=180` to `AnthropicLLMClient.astream`, which rejects it | (noted, out-of-scope) | **Accept → filed #385** | Real latent degradation on the Anthropic synthesis path, but orthogonal to caching. Filed as #385 rather than expanding this PR's scope. The now-faithful fake will catch it once a synthesis-path test exists. |
| 2 | Only `total_cache_read_tokens` asserted on `AgentResult`; `total_cache_creation_tokens` not | Recommended | **Accept** | Added a symmetric assertion (fake now reports `cache_creation=512`; test asserts both totals). |
| 3 | Test name/docstring over-read "warm 2nd turn" — unit test proves plumbing, not cross-turn caching | Minor | **Accept** | Renamed test and reworded the module docstring to scope the claim to the wiring; cross-turn SDK shapes are covered by `test_llm_client.py`. |
| 4 | Local-backend top-level tool `cache_control` strip covered by construction, not an explicit assertion | Minor | **Reject** | Reviewer marked it acceptable; existing `test_strips_cache_control` already covers `_anthropic_tools_to_openai`'s stripping. Adding more is low value. |

**Findings forwarded to addresser:** 0 — all accepted items were cheap test-only hardening applied directly in the next commit. No Action Required items.

PR #386, opened 2026-06-22
refactor(prompts): consolidate orchestrator instruction layer (Phases 2+3)

nbramia:
## Summary

Two consolidation phases of #383, bundled because both are low-risk, behavior-preserving cleanups of the orchestrator's instruction layer.

**Phase 2 — lean the static prompt** (`agent_system_prompt.py`)
- Merged the web-search directive (don't claim a cutoff / can't-browse / negative-from-memory / search-on-pushback) from **four** repeated paragraphs into one Exception + one pushback rule — every behavior preserved.
- The email `draft → show → confirm → send` flow was stated twice; the tool descriptions now point to the single authoritative Multi-tool pattern.
- Folded the three parallel calendar patterns into one (confirm-before-write stated once) and dropped the `read_vault_file` pattern that duplicated the How-to bullet.
- **Templated `max_tool_rounds`**: `build_system_prompt(max_tool_rounds=…)` renders the number into an *uncached dynamic block* (the cached static prefix stays byte-stable) so the prompt and the loop's actual budget can't drift. `run_agent_loop` passes its real value.
- Documented the per-request existing-task-tags block in the `manage_tasks` description so the model expects it.
- Verified the `relationship strength (0-100)` note is **accurate, not stale** (`format_relationship_context` emits `Strength/100`) → kept.

**Phase 3 — unify the delegation blurb** (`agent_worker/`)
- New `agent_worker/delegation.py` is the single source: `lifeos_agent_*` tool-name constants, a `delegation_preamble(session_id, trigger, model)` helper (claude_code + codex), and `INTER_AGENT_BLOCK` (the local worker's richer protocol). Renaming a tool now touches one file.
- All three executors import from it. The local executor's `_SYSTEM_PROMPT_STATIC` embeds `INTER_AGENT_BLOCK` and is **byte-for-byte identical** to before (verified), preserving its prompt cache.

Relates to #383 (Phases 2 & 3).

## Test evidence
- New: `tests/test_agent_system_prompt.py::test_max_tool_rounds_templated_not_hardcoded`, `tests/test_agent_worker_delegation.py` (3 cases).
- 229 passed across the prompt + caching + all agent-worker executor + persona test files.
- Pre-push unit suite: **2144 passed, 8 skipped**.
- `_SYSTEM_PROMPT_STATIC` byte-equality vs `origin/main` asserted programmatically; static prompt ~10497 → ~9800 chars.

## Review focus
- Phase 2 prose: confirm no behavior was dropped in the web-search / email / calendar consolidations (the goal is no behavior change, not maximal cuts — I prioritized that over the issue's aspirational 20-30%).
- `max_tool_rounds` lives in an *uncached* block so caching (PR #384) isn't disturbed — confirm the cached static block is still byte-stable across turns.
- Phase 3: `delegation_preamble`'s `trigger`/`model` params faithfully reproduce each executor's intent; local block unchanged.

## Not run here
- `/routing-health` (Phase 2 acceptance criterion): it validates the model-routing matrix (orthogonal to these prompt-text-only changes) and needs the live server running this code. Recommend running it post-merge once main is deployed.

### DIFF
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -429,7 +429,7 @@ async def run_agent_loop(
         Dicts with "type" key: "text", "status", or "result".
     """
     client = _select_client(model, force_local=force_local)
-    system_prompt = build_system_prompt(persona=persona)
+    system_prompt = build_system_prompt(persona=persona, max_tool_rounds=max_tool_rounds)
 
     # Bind a fresh per-turn email-draft set. The send gate uses this to refuse
     # sending any draft created during this same turn — sends require the user
--- a/api/services/agent_system_prompt.py
+++ b/api/services/agent_system_prompt.py
@@ -89,7 +89,7 @@ Searches saved memories by keyword. Use to recall previously saved information o
 
 Don't use tools for general knowledge, definitions, coding help, math, or anything that doesn't require {name}'s personal data or current/live information. Just answer directly.
 
-**Exception:** If a question touches anything that can change over time (rankings, prices, current events, "best X right now", latest versions, schedules, rosters, releases), ALWAYS call search_web first — even if it seems like general knowledge. Your training data is stale, so never claim from memory that you "can't access" live data, that you have a "knowledge cutoff," or that something "hasn't been released / doesn't exist / isn't available yet." State that something isn't available only *after* a web search comes up empty, and say you searched.
+**Exception:** If a question touches anything that can change over time (rankings, prices, current events, "best X right now", latest versions, schedules, rosters, releases), ALWAYS call search_web first — even if it seems like general knowledge. Your training data is stale, so never claim from memory that you "can't access" live data, "can't browse the web," or have a "knowledge cutoff," and never assert that something "hasn't been released / announced / happened yet," doesn't exist, or isn't available — call search_web and let the results decide. State that something isn't available only *after* a web search comes up empty, and say you searched.
 
 **On pushback** ("do research," "you're wrong," "look it up"), you MUST call search_web before replying — don't repeat your previous claim, and never say "my research confirms" unless you actually searched this turn.
 
--- a/api/services/agent_system_prompt.py
+++ b/api/services/agent_system_prompt.py
@@ -56,7 +56,7 @@ Returns iMessage and WhatsApp chat logs with a specific person. Shows actual mes
 Web search for any current or real-time information — weather, news, prices, rankings, benchmarks, reviews, technical specs, documentation, public facts, or anything that may have changed since your training. Use whenever the answer benefits from up-to-date data. Only skip if the answer is purely in {name}'s personal data.
 
 **manage_tasks (action: create/list/complete):**
-Create, list, or complete Obsidian tasks.
+Create, list, or complete Obsidian tasks. When tagging a task and an existing-tags list is provided below this prompt, prefer a tag that already exists over inventing a near-duplicate.
 
 **manage_reminders (action: create/list):**
 Create or list timed Telegram notification reminders.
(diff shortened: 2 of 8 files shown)


nbramia (commented):
## PR Review: refactor(prompts): consolidate orchestrator instruction layer (Phases 2+3)

Adversarial round-1 review. I independently re-verified every load-bearing claim against `origin/main` (rendered both prompts and diffed; reconstructed the old constants via AST and compared byte-for-byte; ran the affected tests). The structural/cache/Phase-3 claims all hold. The only genuine regressions are a handful of prose nudges dropped from the web-search and email directives — none touch a code-enforced invariant.

### Action Required
- _None._ No behavior backed by code or a safety gate was lost.

### Recommended
- **[Correctness/Behavior]** Web-search directive dropped three nudges that the other two-thirds of each rule kept (verified absent from the rendered prompt, not just the diff): (1) the **"can't browse the web"** forbidden phrase — `agent_system_prompt.py:91`'s Exception keeps "can't access" live data and "knowledge cutoff" but no longer forbids "can't browse the web"; (2) **"announced" / "happened yet"** — old para C forbade asserting from memory that something "hasn't been released / **announced** / **happened** yet"; only "released" survives. These read like incidental casualties of the merge rather than deliberate cuts, since their sibling phrases were retained. If "no behavior change" is the bar, fold "can't browse the web" back into the can't-access list and restore the announced/happened verbs. Low blast radius (prompt prose), but it's the clearest deviation from the stated goal.

### Minor
- **[Behavior]** Web-search: **"let the results decide"** framing dropped (`agent_system_prompt.py:91`). "Call search_web first" carries most of the intent; the explicit don't-pre-judge nudge is gone.
- **[Behavior]** Email: the **"show the draft first"** precondition was removed from the system-prompt `send_email_draft` description (`agent_system_prompt.py:71`) and now lives only in the cross-referenced Multi-tool pattern. De-rated to Minor because the safety-critical invariant (no same-turn send) is **hard-enforced in code** (`agent_tools.py:1355` `_draft_created_this_turn`) and the always-in-context tool-call description still states it (`agent_tools.py:514-516`). This is redundancy reduction, not a send-gate weakening.
- **[Behavior]** Calendar move flow: old "**present proposed changes**" collapsed into the generic "present the details" (`agent_system_prompt.py:112`). The present-and-confirm-before-write gate survives for all three actions; only the "show the delta" specificity softened.
- **[Test fidelity]** The PR states `_SYSTEM_PROMPT_STATIC` byte-equality vs `origin/main` is "asserted programmatically," but no committed test pins it — `test_agent_worker_delegation.py:45` only checks `INTER_AGENT_BLOCK in _SYSTEM_PROMPT_STATIC`, which would still pass if the surrounding restructure had altered other bytes. (I verified the actual constant IS byte-identical: 4024 chars, char-for-char.) Consider an explicit equality test if the byte-stable cache prefix is meant to be a regression-guarded invariant.
- **[Consistency]** `codex_executor._delegation_header` is no longer byte-identical to main: "Monitor with … read **the** result" became "Monitor **the child** with … read **its** result" (it now shares claude_code's phrasing via the helper). Behavior-neutral and arguably an improvement; flagging only because the local block's byte-equality was called out as a property and codex's wasn't — confirm the codex wording change is intentional.

### Verified (claims that hold)
- **Cache integrity ✓** — `max_tool_rounds` renders only into the uncached dynamic block; `block[0]` is byte-stable across both `max_tool_rounds` (5/9/12) and persona variation, and contains no numeric tool-round literal. #384's cache prefix is intact. Well covered by `test_max_tool_rounds_templated_not_hardcoded` + `test_static_block_still_cached`.
- **Local executor byte-equality ✓** — `_SYSTEM_PROMPT_STATIC` is char-for-char identical to main (4024 chars).
- **claude_code `.format()` ✓** — no leftover `{session_id}`; `delegation_preamble` emits no `{`/`}`, so no KeyError/ValueError risk; real `_build_command` format path renders correctly with `caller_session_id` embedded.
- **Phase 3 fidelity ✓** — `delegation_preamble` faithfully reproduces each executor's model recommendation and trigger condition.
- **`read_vault_file` "duplicate" drop ✓** — genuinely redundant; the guidance survives in the How-to section.
- **Tests ✓** — 132 affected tests pass (prompt + delegation + all three executors + caching + persona).

### Summary
Merge-ready as a behavior-preserving refactor — cache integrity, byte-equality, and Phase-3 fidelity all verified. Recommend restoring the "can't browse the web" / "announced" / "happened yet" web-search nudges before merge to fully honor the "no behavior change" goal; everything else is optional polish.

nbramia:

| 1 | Web-search consolidation dropped the "can't browse the web" forbidden phrase and "announced"/"happened yet" from the negative-assertion triad | Recommended | **Accept** | Incidental casualties of the merge; for a no-behavior-change PR the exact refusal phrases matter (the model pattern-matches them). Folded all three back into the single consolidated Exception paragraph. |
| 2 | "let the results decide" framing dropped | Minor | **Accept** | Cheap to restore while fixing #1; folded back. |
| 3 | "show the draft first" precondition removed from the `send_email_draft` description | Minor | **Reject** | This is the *intended* dedup — the full flow lives in the authoritative "Sending an email" pattern, the description cross-references it, and the same-turn-send invariant is hard-enforced in code (`agent_tools.py`). Restoring it re-introduces the duplication the phase exists to remove. |
| 4 | Calendar "present proposed changes" specificity lost on the move flow | Minor | **Reject** | Intended consolidation. The present-and-confirm-before-write gate is stated once for all three actions and is also in each calendar tool's description. No behavior lost. |
| 5 | Byte-equality of `_SYSTEM_PROMPT_STATIC` claimed but not pinned by a committed test | Minor | **Accept** | Added `test_inter_agent_block_text_is_pinned` asserting the exact `INTER_AGENT_BLOCK` text, so accidental edits to the shared constant (which the cached local prompt embeds) are caught. |
| 6 | `codex_executor._delegation_header` no longer byte-identical to main ("Monitor with" → "Monitor the child with") | Minor | **Acknowledge (intentional)** | This is the point of Phase 3 — codex and claude_code now share one canonical phrasing via `delegation_preamble`. Behavior-neutral; codex byte-equality was never a property (unlike the local block, which is preserved exactly). |

**Findings forwarded to addresser:** 0 — accepted items (#1, #2, #5) were applied directly in the next commit; no Action Required.

PR #388, opened 2026-06-22
fix(chat): accept timeout= on AnthropicLLMClient.astream

### LINKED ISSUE #385: bug(chat): synthesis round passes timeout= to AnthropicLLMClient.astream, which rejects it
## Problem

`run_agent_loop`'s final **synthesis round** (the `else` branch after tool rounds are exhausted) calls:

```python
# api/services/agent_loop.py:631-635
async for event in client.astream(
    messages,
    system=system_prompt,
    max_tokens=4096,
    timeout=180,
):
```

`LocalLLMClient.astream` accepts a `timeout` kwarg, but **`AnthropicLLMClient.astream` does not** (`api/services/llm_client.py:502-510` — its signature is `(messages, *, system, max_tokens, tools, temperature)`).

So on the Anthropic backend (the default), whenever a query exhausts all `max_tool_rounds` and falls through to the synthesis round, the call raises `TypeError: astream() got an unexpected keyword argument 'timeout'`. It's caught by the synthesis round's `try/except`, so the loop doesn't crash — but the user gets `"(Error during synthesis: ...)"` **instead of the synthesized answer**. This silently degrades exactly the long, multi-round queries that most need a final synthesis.

## Discovery

Surfaced during adversarial review of #384 (Phase 1 of #383). Pre-existing; not introduced by that PR. #384's test fake originally had a `**kwargs` that masked the signature drift; that fake was tightened to mirror the real `AnthropicLLMClient.astream` signature exactly, so a future test that exercises the synthesis round on the Anthropic path will catch this.

## Fix

Add a `timeout: float | None = None` parameter to `AnthropicLLMClient.astream` and apply it as the per-request timeout (the Anthropic SDK accepts a `timeout` request option), so both backends honor the synthesis-round timeout and the signatures stay consistent. Add a test that drives `run_agent_loop` to the synthesis round against a fake with the real (no-`**kwargs`) signature.

## Acceptance criteria
- [ ] On the Anthropic backend, an exhausted-rounds query produces a synthesized answer, not an `(Error during synthesis: ...)` message.
- [ ] `AnthropicLLMClient.astream` and `LocalLLMClient.astream` share a compatible signature for the kwargs `agent_loop` passes.
- [ ] A regression test exercises the synthesis round with a faithful client signature.

nbramia:
## Summary

`run_agent_loop`'s synthesis round (the fallback after all tool rounds are exhausted) calls `client.astream(..., timeout=180)`. `LocalLLMClient.astream` accepts a `timeout` kwarg, but `AnthropicLLMClient.astream` did not — so on the Anthropic backend (the default) an exhausted-rounds query raised `TypeError`, caught by the synthesis round's `try/except` and surfaced to the user as `"(Error during synthesis: ...)"` instead of the synthesized answer.

Adds a `timeout` parameter to `AnthropicLLMClient.astream` and forwards it as the per-request timeout (the Anthropic SDK's `messages.stream` accepts `timeout`). Both backends now honor the synthesis-round timeout and share a compatible `astream` signature.

Closes #385.

## Test evidence
- `test_astream_forwards_timeout_to_sdk` (new) — the real client accepts `timeout` and passes it to the SDK (the direct regression guard).
- `test_synthesis_round_passes_timeout_and_completes` (new) — `run_agent_loop` exhausts its rounds, hits the synthesis round, and produces the answer; the faithful fake's `astream` signature now includes `timeout`, mirroring the fixed client.
- `tests/test_llm_client.py` + `tests/test_agent_loop_caching.py`: 43 passed. Pre-push unit suite green. (14 repo-wide `require_db`/`slow` data-integrity failures are pre-existing and outside the `unit` selection.)

## Review focus
- Confirm the Anthropic SDK accepts `timeout=` on `messages.stream` (verified: it's in the `AsyncMessages.stream` signature on anthropic 0.84.0).
- The fix is scoped to `astream` only (matching `LocalLLMClient`, where only `astream` carries `timeout`); `create`/`acreate` are never called with a timeout in-repo.

### DIFF
--- a/api/services/llm_client.py
+++ b/api/services/llm_client.py
@@ -528,6 +528,8 @@ class AnthropicLLMClient:
         if temperature is not None:
             kwargs["temperature"] = temperature
         if timeout is not None:
+            # The SDK accepts float | httpx.Timeout; a bare float sets a uniform
+            # per-request timeout. (LocalLLMClient wraps it with connect=10s.)
             kwargs["timeout"] = timeout
 
         async with self._async_client.messages.stream(**kwargs) as stream:
--- a/api/services/llm_client.py
+++ b/api/services/llm_client.py
@@ -507,10 +507,13 @@ class AnthropicLLMClient:
         max_tokens: int = 4096,
         tools: list[dict] | None = None,
         temperature: float | None = None,
+        timeout: float | None = None,
     ) -> AsyncGenerator[dict, None]:
         """Async streaming via Anthropic API.
 
-        Yields the same event format as LocalLLMClient.astream().
+        Yields the same event format as LocalLLMClient.astream(). ``timeout``
+        (seconds) sets a per-request timeout — the agent loop's synthesis round
+        passes one, and LocalLLMClient.astream accepts the same kwarg.
         """
         kwargs: dict[str, Any] = {
             "model": self._model,
@@ -524,6 +527,8 @@ class AnthropicLLMClient:
             kwargs["tools"] = tools
         if temperature is not None:
             kwargs["temperature"] = temperature
+        if timeout is not None:
+            kwargs["timeout"] = timeout
 
         async with self._async_client.messages.stream(**kwargs) as stream:
             async for event in stream:
--- a/tests/test_agent_loop_caching.py
+++ b/tests/test_agent_loop_caching.py
@@ -83,49 +83,75 @@ async def test_cache_tokens_surface_in_agent_result():
     assert result.full_text == "The answer is 42."
 
 
-class _ExhaustThenSynthesize:
-    """Returns a tool call on every tool round (so the loop exhausts its rounds
-    and falls through to the synthesis round), then a text answer on the
-    synthesis round. astream's signature mirrors the real client exactly,
-    including the ``timeout`` kwarg the synthesis round passes."""
+class _MockAnthropicStream:
+    """Minimal async context manager mimicking anthropic's streaming response,
+    so the REAL AnthropicLLMClient.astream can be exercised against it."""
 
-    def __init__(self):
-        self.calls = []
+    def __init__(self, final_message, deltas=None):
+        self._final = final_message
+        self._deltas = deltas or []
 
-    async def astream(self, messages, *, system=None, max_tokens=4096,
-                      tools=None, temperature=None, timeout=None):
-        from api.services.llm_client import LLMUsage
-        self.calls.append({"has_tools": tools is not None, "timeout": timeout})
-        if tools:
-            yield {"type": "tool_calls", "calls": [
-                {"id": "c1", "type": "function",
-                 "function": {"name": "search_vault", "arguments": "{}"}}]}
-            yield {"type": "done", "usage": LLMUsage(), "finish_reason": "tool_use"}
-        else:
-            yield {"type": "text", "content": "Synthesized answer."}
-            yield {"type": "done", "usage": LLMUsage(), "finish_reason": "end_turn"}
+    async def __aenter__(self):
+        return self
+
+    async def __aexit__(self, *exc):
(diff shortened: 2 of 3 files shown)


nbramia (commented):
## PR Review: fix(chat): accept timeout= on AnthropicLLMClient.astream

Verified hands-on: inspected the anthropic 0.84.0 SDK signature, ran the affected suites (43 passed), and reverted the fix to confirm which tests actually catch the regression.

### Action Required
_None._ The fix is correct and the bug is guarded by a faithful regression test.

### Recommended
- **[Test Quality]** The integration test `test_synthesis_round_passes_timeout_and_completes` (`tests/test_agent_loop_caching.py:110`) does **not** catch the production bug. I reverted the `timeout` param from `AnthropicLLMClient.astream` and this test still **passed** — because it patches `_select_client` to return the hand-written `_ExhaustThenSynthesize` fake, so it never instantiates or calls the real Anthropic client. It only verifies that `run_agent_loop` reaches the synthesis `else` branch and forwards `timeout=180` with no tools (useful for pinning the call shape, but not a regression guard for #385). Its protection depends entirely on someone keeping the fake's signature in sync with the real client. The PR description frames this test as "mirroring the fixed client" — that's overstated. The **real** regression guard is the unit test `test_astream_forwards_timeout_to_sdk` (`tests/test_llm_client.py:639`), which correctly **fails** (`TypeError: ... unexpected keyword argument 'timeout'`) against the buggy code. Consider tightening the integration test to drive the real `AnthropicLLMClient.astream` (e.g. monkeypatch `_async_client.messages.stream`) so signature drift on the real client surfaces here too. Not blocking, since the unit test already covers the regression.

### Minor
- **[Correctness/Parity]** Forwarding a bare `float` is correct — the SDK accepts `timeout: float | httpx.Timeout | None | NotGiven`, no `httpx.Timeout` wrapper needed. But note a subtle backend asymmetry: a bare float makes httpx apply 180s uniformly to connect/read/write/pool, whereas `LocalLLMClient.astream` (`api/services/llm_client.py:321`) builds `httpx.Timeout(timeout, connect=10.0)` — so Local pins connect to 10s while Anthropic gets connect=180s. Also, the SDK default read timeout is 600s, so the synthesis round intentionally tightens read 600→180 on the Anthropic path. All intended/harmless, but the two backends aren't byte-for-byte identical in timeout shape; a one-line comment noting that would help the next reader.

### Summary
Merge-ready. The fix is minimal, correct against anthropic 0.84.0, scoped to exactly the 3 files described (+77/-2), and the normal non-exhausted path is unchanged (`timeout=None` skips the kwarg, SDK default applies). The one substantive note is that of the two new tests, only the unit test actually guards the regression — the integration test passes even on buggy code — but the bug IS covered, so this is a recommendation, not a blocker.

nbramia:

| 1 | The integration test `test_synthesis_round_passes_timeout_and_completes` used a hand-written fake, so it didn't actually guard the regression (reverting the fix left it passing); only the unit test caught the bug | Recommended | **Accept** | Reworked it into `test_synthesis_round_drives_real_client_with_timeout`, which drives the **real** `AnthropicLLMClient.astream` (only the SDK is mocked). Empirically verified: with the `timeout` param reverted, **both** new tests now fail; with the fix, both pass. The integration test now guards the bug end-to-end, not just the call shape. |
| 2 | Forwarding a bare `float` is correct but the parity nuance vs `LocalLLMClient`'s `httpx.Timeout(..., connect=10.0)` is worth a comment | Minor | **Accept** | Added a comment at the forwarding site noting the SDK accepts `float \| httpx.Timeout` and that a bare float is a uniform per-request timeout (Local wraps with connect=10s). |

**Findings forwarded to addresser:** 0 — both accepted items were applied directly in the next commit.

Issue #390, opened 2026-06-22
feat(personas): persona schema — frontmatter, primary.md, voice-awareness, personal-context, web-spawn


nbramia:
## Problem

Personas are free-form markdown of uneven depth, all granted every tool, with no machine-readable config and no way to differ by modality. PR #389 shipped the *content* improvements (therapist/primary/doctor) and the `/persona-check` skill, but deliberately left the **schema** out because it needs a loader change. This issue tracks that schema.

## Context

`resolve_persona` currently returns the raw persona-file text, so adding YAML frontmatter today would leak into the system prompt; and `primary` has no file (its personality lives in `agent_system_prompt.py`). The design was prototyped and reviewed with the user; their decisions are baked into the phases below.

## Schema (agreed)

Frontmatter holds **only what code acts on** (routing fields stay in `telegram_bots.json`):

```yaml
---
id: therapist        # validated == bot name
model: ""            # optional per-persona model default; empty = orchestrator default
voice:               # rules applied ONLY on voice turns (needs the modality flag)
  - Speak in plain sentences — no markdown or bullet lists.
  - Calmer cadence; one idea at a time.
---
```

Prose skeleton (checklist, include what applies): Opening line → Tone → What you do → Sourcing → Tools you lean on → Response shape → When data is thin → Out of scope.

## Suggested approach (phased)

**Phase 1 — loader + frontmatter.** Parse/strip frontmatter in `settings.telegram_bots` using the existing `python-frontmatter` dep (used by `chunker.py`); store `voice`/`model` on `TelegramBotConfig`; `resolve_persona` returns the **body only**. Add `config/personas/README.md` (the schema doc) and migrate the 3 existing personas onto frontmatter. ⚠️ `fitness.md` has literal `{…}` — do **not** `str.format()` the body; strip frontmatter and append resolved blocks instead. `/persona-check`'s frontmatter validation lights up automatically.

**Phase 2 — `primary.md`.** Read `config/personas/primary.md` if present; lift `primary`'s personality (the `## Response format` / tone block) out of the static prompt into it so it's a tunable peer, with the global tool *mechanics* staying in the static prompt (de-dupes with #386's leaner prompt).

**Phase 3 — voice-awareness.** Thread a modality flag (text vs voice) from the chat/voice routes into `run_agent_loop`, and apply each persona's frontmatter `voice` rules on voice turns. (whisper-relay already posts to `/api/ask/stream`; the flag rides along.)

[...]

PR #391, opened 2026-06-22
feat(personas): frontmatter loader — strip YAML, parse voice/model

nbramia:
## Summary

#390 **Phase 1** — the persona-schema foundation. Persona files may now carry a YAML frontmatter block (`id` / `model` / `voice`); the loader strips it so only the **body** reaches the system prompt, and parses `voice`/`model` onto `TelegramBotConfig` for later phases to consume. Migrates the three existing personas onto the schema and adds the schema doc.

Relates to #390. (Phases 2–5 — `primary.md`, voice-awareness, personal-context, web-spawn — are follow-up PRs.)

## What changed

- **`settings._parse_persona(text, name)`** → `(body, voice, model)` using the existing `python-frontmatter` dep (same one `chunker.py` uses). `telegram_bots` uses it, so `resolve_persona` and the Telegram path return **body-only** — no YAML leak.
- Body is returned **verbatim** (no `str.format`), so literal `{…}` examples in `fitness.md` survive — the gotcha called out in the issue.
- `TelegramBotConfig` gains `voice: tuple[str, ...]` + `model: str` (parsed now, inert until the voice/model phases).
- Frontmatter added to `fitness` / `therapist` / `doctor`; `config/personas/README.md` documents the schema. Files **without** frontmatter still load unchanged — purely additive.
- Fixed `fitness.md`'s future-dated "once Apple Health lands" wording.

## Test evidence

- New `TestPersonaFrontmatter` (in `test_persona_api.py`): frontmatter stripped + literal braces survive + `voice`/`model` parsed; no-frontmatter passthrough; `resolve_persona` returns body-only with `bot.voice`/`bot.model` populated; every real persona file parses with no frontmatter leaking into the body.
- Full unit gate: **2151 passed, 8 skipped**. Existing persona/telegram/static-prompt tests unchanged. `ruff` clean.
- Sanity: the 3 migrated files parse cleanly (doctor 0 / fitness 2 / therapist 3 voice rules; bodies start with the prose).

## Review focus

- `_parse_persona` edge cases (no frontmatter, non-list `voice`, `id` mismatch warning).
- That nothing `str.format`s a persona body (would break `fitness.md` braces).
- Backward-compat: files without frontmatter behave exactly as before.

### DIFF
--- a/config/personas/README.md
+++ b/config/personas/README.md
@@ -14,7 +14,7 @@ The loader (`settings._parse_persona`, used by `settings.telegram_bots`) strips 
 
 The frontmatter holds **only what code acts on** — never real names, vault paths, or other personal values (the project's open-source rule; `/persona-check` enforces it). The schema leaves room to add hard tool allow-lists later (a `tools:` field) without restructuring.
 
-Files with no frontmatter still load fine (the whole file is the body), so frontmatter is purely additive.
+A file with no leading `---` block loads whole — frontmatter is purely additive. A file that *starts* with a `---…---` block is always parsed as frontmatter (standard YAML-frontmatter semantics), so don't open a persona body with a raw `---` horizontal rule. A malformed frontmatter block doesn't crash loading — the loader logs a warning and falls back to using the raw file as the preamble.
 
 ## Prose skeleton
 
--- a/config/personas/README.md
+++ b/config/personas/README.md
@@ -1,0 +1,38 @@
+# Persona files
+
+Each persona is a markdown file: an optional YAML **frontmatter** block (machine-read config) followed by the **prose body** (the personality the model reads). The bot registry (`config/telegram_bots.json`) maps a bot `name` → its `persona_file`; this file owns *how the persona behaves*, the registry owns *routing* (token, chat, `orchestrates`). Frontmatter never duplicates registry fields.
+
+The loader (`settings._parse_persona`, used by `settings.telegram_bots`) strips the frontmatter and returns the **body only** as the system-prompt preamble — so the YAML never leaks into the prompt. The body is used **verbatim** (no `str.format`), so a persona may safely contain literal `{...}` examples (e.g. `fitness.md`).
+
+## Frontmatter (YAML, optional)
+
+| Field | Meaning |
+|---|---|
+| `id` | Should equal the bot `name` in `telegram_bots.json` (a mismatch logs a warning). |
+| `model` | Optional per-persona model preference (e.g. `opus`); empty/omitted = orchestrator default + normal escalation. |
+| `voice` | A list of behaviour rules applied **only on voice (spoken) turns**; ignored for text. |
+
+The frontmatter holds **only what code acts on** — never real names, vault paths, or other personal values (the project's open-source rule; `/persona-check` enforces it). The schema leaves room to add hard tool allow-lists later (a `tools:` field) without restructuring.
+
+Files with no frontmatter still load fine (the whole file is the body), so frontmatter is purely additive.
+
+## Prose skeleton
+
+Not every persona needs every section — include what applies. Suggested order:
+
+1. **Opening line** — role + surface, one sentence.
+2. `## Tone` — personality in text.
+3. `## What you do` — core behaviour (for `doctor`, the pipeline).
+4. `## Sourcing` — where to ground answers (omit if nothing special).
+5. `## Tools you lean on` — advisory prose: which tools this persona mainly uses (not enforced).
+6. `## Response shape` — length/format norms for text (voice norms go in frontmatter `voice`).
+7. `## When data is thin` — sparse-data fallback.
+8. `## Out of scope` — redirect behaviour (omit for `primary`, the catch-all).
+
+## Status / roadmap
+
+This file documents the schema as loaded today (frontmatter parsed + stripped; `voice`/`model` stored on `TelegramBotConfig`). Consuming the parsed fields and the rest of the schema is tracked in #390:
+
+- `voice` rules are parsed now but **applied on voice turns** in a follow-up (needs a text-vs-voice modality flag).
+- `model` preference is parsed now; wiring it into model selection is a follow-up.
+- A `primary.md` (lifting `primary`'s personality out of the static prompt) and a resolved **personal-context block** (partner / therapists / inner circle / folders, from existing config — never hardcoded here) are follow-up phases.
--- a/config/personas/doctor.md
+++ b/config/personas/doctor.md
@@ -1,3 +1,8 @@
+---
+id: doctor
+model: ""
+---
+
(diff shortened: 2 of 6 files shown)


nbramia (commented):
## PR Review: feat(personas): frontmatter loader — strip YAML, parse voice/model

Reviewed adversarially with Correctness, Security/Requirements, and Standards specialists. All claims independently re-verified against the worktree (commit `87db42c`) using the project venv (`python-frontmatter` 1.1.0, PyYAML 6.0.3). Ruff clean; all 24 tests in `test_persona_api.py` pass. The headline claims hold — frontmatter is genuinely stripped, the body is never `str.format()`'d (fitness.md's literal `{...}` survives end-to-end), and no-frontmatter files are byte-identical to the old behavior. Two regressions, however, warrant fixes before merge.

### Action Required

- **[Correctness/Standards] Malformed YAML in any persona file crashes loading for ALL bots and 500s the chat/persona endpoints.** `config/settings.py:61` calls `frontmatter.loads()` with no try/except inside `_parse_persona`, and the call site (`config/settings.py:826-829`) catches only `OSError`. `yaml.ParserError`/`ScannerError` are **not** `OSError` subclasses, so they propagate. Reproduced empirically: with bots `[good, malformed, good]`, `settings.telegram_bots` (a plain, uncached `@property`) raised mid-loop — one bad file takes out *all* bots, not just the offender. Blast radius: `resolve_persona` (`chat.py:442`) and `list_http_personas` (`chat.py:240`) both hit this property, so every `/chat` persona resolution and the persona-discovery endpoint would 500 instead of degrading. This is a **new failure mode** — the old `read_text().strip()` could never raise a parse error. Hand-edited YAML with a typo (tab indent, unclosed `[`, stray `:`) is a plausible, not impossible, scenario. **Fix:** catch `yaml.YAMLError` (or broad `Exception`) inside `_parse_persona`, log a warning, and fall back to treating the raw text as the body. No test covers this — a regression test asserting `telegram_bots`/`list_http_personas` don't raise on a malformed file is the highest-value addition.

- **[Correctness] No-frontmatter bodies that open with (or contain an early) `---` block are silently truncated — directly contradicting the README.** The loader now runs frontmatter parsing on *every* persona file. `config/personas/README.md:21` promises *"Files with no frontmatter still load fine (the whole file is the body)"* — false whenever the first block is `---`-fenced. Reproduced: `"---\nThe Coach\n---\n\nYou are a coach..."` → body becomes only `"You are a coach..."` (title block eaten as metadata); an early `---`-delimited section truncates similarly. The old `.read_text().strip()` preserved the file verbatim. The 3 committed files are safe (migrated to real frontmatter), but any registry-referenced file authored before this PR with a leading rule/title/quote block is silently shortened — no error, no warning, changed system prompt. Either guard the truncation case or soften the README claim.

### Recommended

- **[Correctness/Standards] Silent type-coercion of `voice`/`model` should warn.** `config/settings.py:62-65`: a `voice:` scalar (rule written without a list) or a dict → `()` (rules silently dropped); `model: 5` → `"5"`. An easy authoring mistake vanishes with no signal, while the adjacent `id`-mismatch path *does* warn (line 67) — inconsistent. Log a warning on structural schema violations.
- **[Correctness] BOM silently disables frontmatter parsing.** A UTF-8 BOM before `---` makes `frontmatter.loads` treat the whole file (including the literal `---` + YAML) as body, leaking `id:`/`model:`/`voice:` into the system prompt and dropping the parsed fields — no warning. Low likelihood; noting the silent hole.
- **[Standards] Missing edge-case test coverage.** Beyond malformed-YAML (A1), no test exercises: the id-mismatch warning (no `caplog` assertion — the only observable effect of `name`/`id` is unverified), non-list `voice` → `()`, empty file, frontmatter-only/no-body, or the `OSError`-at-call-site graceful-degradation path (`settings.py:828`). Behaviors are individually correct today; tests guard against silent regression.
- **[Security — follow-up, not this PR] `model` is unvalidated against the allowed tier set.** `settings.py:65` accepts any string; `resolve_model_alias` passes unknowns through unchanged. Inert now (trusted committed files), but the phase that wires `bot.model` into selection should validate against `_MODEL_ALIASES` and fall back + warn, else a typo flows to the Anthropic client as a bogus model id.

### Minor

- **[Standards] Redundant string-literal return annotation** at `settings.py:51` (`-> "tuple[...]"`) — no forward reference needed. Harmless; matches local style (neighbors quote too). No change required.
- **[Standards] `model: ""` in the migrated files** is equivalent to omitting the key — slightly redundant but arguably a clearer schema placeholder.

### Clean (verified, no action)
- No code `str.format()`s a persona body (grep across `api/`+`config/` empty); both consumer paths (chat → `build_system_prompt` appends it as a separate block at `agent_system_prompt.py:192`; doctor f-string at `telegram.py:680`) interpolate the stripped body safely.
- No-frontmatter backward-compat byte-identical to old behavior (verified across plain/padded/multiline/empty samples).
- No PII / personal-values leak in the 3 migrated files or README (only `~/Code/LifeOS`, the canonical checkout — acceptable).
- `python-frontmatter>=1.1.0` already declared in `requirements.txt` — no new dependency.
- Commit message well-formed; the `fitness.md` wording fix is a disclosed one-line drive-by (called out in the commit body), not scope creep.

### Summary
Needs changes before merge: two genuine regressions — (1) a malformed-YAML persona file raises an uncaught non-`OSError` exception that takes down all bots and 500s the chat/persona endpoints, and (2) a no-frontmatter body opening with `---` is silently truncated, contradicting the README. Both are new failure modes this PR introduces, both are small fixes (broaden the except + fall back to raw body; guard or re-word the README), and both lack test coverage. Everything else — stripping correctness, brace-survival, backward-compat, privacy, hygiene — is solid.

nbramia:

| 1 | Malformed YAML in a persona file crashes all bots (uncached `telegram_bots` → `resolve_persona`/`list_http_personas` 500) | Action Required | **Accept** | Real new regression — `frontmatter.loads` raises `yaml.YAMLError`, not `OSError`. Guarded in `_parse_persona`; degrade to raw body + warn. |
| 2 | No-frontmatter body opening with `---` is truncated | Action Required | **Downgrade → doc** | That's correct frontmatter semantics, not a bug; a guard would fight the library. Clarified README instead. |
| 3 | Non-list `voice` silently dropped | Recommended | **Accept** | Now warns + drops safely (consistent with the id-mismatch warning). |
| 4 | Missing tests (malformed/non-list/id-mismatch) | Recommended | **Accept** | Added all three. |
| 5 | BOM-before-`---` edge case | Recommended | **Reject** | Theoretical; persona files don't carry a BOM. |
| 6 | `model` not validated against tier set | Recommended | **Reject (defer)** | `model` isn't consumed until a later #390 phase. |
| 7 | Quoted return annotation; `model:""` | Minor | **Reject** | Matches codebase style (`resolve_persona` is identically annotated). |

**Forwarded to addresser:** 4 (findings 1, 2, 3, 4).

### Addressed (round 1) — done directly given the small scope
| 1 | YAML-crash blast radius | Applied | `_parse_persona` catches parse errors → raw-body fallback + warning (`config/settings.py`). |
| 2 | `---` truncation | Applied (doc) | README clarifies leading-`---` semantics + the malformed-fallback behaviour. |
| 3 | Non-list `voice` | Applied | Warns and drops to `()`. |
| 4 | Test coverage | Applied | `test_malformed_frontmatter_falls_back_to_raw_body`, `test_non_list_voice_is_ignored`, `test_id_mismatch_warns`. |

**Tests:** `-m "unit and not slow"` → **2154 passed, 8 skipped**. ruff clean. Commit `c95642b`.

Issue #411, opened 2026-06-25
fix(agent-worker): codex dispatch has same crash-before-init re-execute window as #400


nbramia:
## Problem
`_dispatch_codex_session` shares the same spawn-before-init window that #400 (PR #408) guarded for `_dispatch_claude_code_session`: it decides resume on the same `claude_code_session_id` column and writes its spawn event before launching the subprocess. If the worker dies after spawn but before init persists, a redispatch re-executes the original prompt.

## Context
#408 added a re-execute guard (`_cli_subprocess_launch_count` + the FAILED-tail terminal-status persistence) to the claude_code path only — the doctor routes through claude_code, so codex was out of scope. A one-line comment near the #408 guard notes this.

## Suggested approach
Mirror #408's guard + the FAILED/BUDGET terminal-status persistence in `_dispatch_codex_session`.

## Acceptance criteria
- [ ] A codex session that spawned but never persisted its session id is not re-executed on redispatch.
- [ ] A codex FAILED/BUDGET outcome finalizes the operator/child session row (no CLAIMED re-dispatch loop).
- [ ] Test covering the codex crash-before-init path.

PR #408, opened 2026-06-25
fix(agent-worker): crash-safe doctor dispatch — resume not re-execute

nbramia:
## Summary

Two crash-safety holes in the doctor's worktree-first lifecycle (`Part of #397`, `Closes #400`):

**(a) Resume-vs-re-execute (the important one).** `_dispatch_claude_code_session` decides fresh-vs-resume solely on `is_resume = bool(session.claude_code_session_id)`. That UUID only persists on the CLI `init` event, which fires *after* the subprocess spawns (`claude_code_executor.py` `_run` :448 emits `claude_code_spawn` → subprocess launches → init persists at :615). If the worker dies in that window, the field stays NULL and the next dispatch **re-executes the original prompt** — for a doctor turn that can re-file the GitHub issue or restart `/implement`.

Fix: guard the fresh-spawn fork with the existing `claude_code_spawn` transcript event (written immediately before the subprocess launches). Present + NULL session id ⇒ a mid-flight crash, so we do **not** re-execute. The session is finalized `FAILED`, a `claude_code_reexecute_prevented` event is recorded, and the operator is notified to re-trigger deliberately. No DB schema change — reuses the spawn signal already emitted at exactly the right point (per the issue's suggested approach).

**(b) Idempotent worktree pruning.** `scripts/cleanup-worktrees.sh` is now safe to call pre-flight. It tolerates already-pruned state and clears a specific stale worktree dir + branch left by a crashed run — which `git worktree prune` alone does **not** remove when the directory survives on disk — so the next `git worktree add` can't fail with "already exists" / "already checked out". Every step no-ops cleanly when there's nothing to do.

## Mechanism choice (resume-vs-re-execute)

Preferred the existing `claude_code_spawn` transcript event over a new DB column because: (1) the issue explicitly suggests reusing the spawn signal; (2) DB schema changes are "ask first" per AGENTS.md; (3) the event is already written at exactly the crash boundary (right before subprocess launch). On detection we finalize `FAILED` + notify rather than silent no-op because the dead subprocess left no resumable UUID (`-r` needs it) and may have produced partial side effects the operator should inspect — mirroring the existing undelivered-prompt escalation path (`worker.py` :1423).

## Test evidence

Targeted (nathan-linux, isolated worktree off this branch):
[...]

PR #416, opened 2026-06-25
fix(agent-worker): codex dispatch crash-before-init guard + terminal-status persistence

nbramia:
## Summary
Closes **#411** — mirrors the #408/#400 fix (claude_code path) into `_dispatch_codex_session`, which the #408 review flagged as having the same spawn-before-init re-execute window unguarded.

- **Re-execute guard**: a codex session with a `codex_spawn` transcript event but a NULL session id (subprocess launched, `codex_init` never persisted) is finalized FAILED + notified rather than re-executed. Reuses `_cli_subprocess_launch_count` with codex's event kinds, so a missing codex binary (`codex_spawn` + `codex_binary_not_found`) is correctly excluded.
- **Terminal-status persistence**: the FAILED/BUDGET tail now persists the session-row status so an operator codex session can't linger CLAIMED and re-dispatch.

## Test evidence
3 new codex tests (crash-before-init → FAILED, no re-execute; binary-not-found excluded; FAILED finalizes the row). Targeted codex + claude_code dispatch suites: 14 passed. Full suite running.

## Review focus
Does the guard correctly mirror #408 for codex's structure (direct `_telegram_send`, the reused `claude_code_session_id` column for codex's thread id)? Any codex-specific edge the claude_code path didn't have.

### DIFF
--- a/api/services/agent_worker/worker.py
+++ b/api/services/agent_worker/worker.py
@@ -1506,9 +1506,8 @@ class Worker:
         # `claude` binary (spawn event written, but no subprocess) is NOT
         # misdiagnosed as a side-effecting interruption — that case re-executes
         # safely once Fix A (terminal-status persistence below) stops the loop.
-        # Scope: claude_code only. The codex path (_dispatch_codex_session) has
-        # the same spawn-before-init window unguarded — intentionally out of #400
-        # (doctor == claude_code); tracked as a follow-up.
+        # The codex path (_dispatch_codex_session) carries the same guard via the
+        # shared _cli_subprocess_launch_count helper (#411).
         prior_launches = self._cli_subprocess_launch_count(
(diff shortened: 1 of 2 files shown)


nbramia:
## Review complete — merging

Single focused reviewer (proportionate — direct mirror of the 2-reviewer-vetted #408): **merge-ready, no Action Required.** Verified the guard fires only on the genuine non-restart re-dispatch window, the binary-not-found exclusion is correct, FIX A mirrors #408's tail, and the codex-specific differences (direct `_telegram_send`, reused session-id column) are handled. Applied the two recommended cleanups: dropped the now-stale 'codex tracked as a follow-up' comment and aligned the event verb to `codex_reexecute_averted`. Full suite green (2234 passed); targeted codex tests green on the fixed commit.

PR #409, opened 2026-06-25
feat(chat): web/voice parity for orchestrating-persona dialogue (#403)

nbramia:
## Summary

Implements **Option A (parity)** for #403: a web/voice client can now answer a `[CLARIFY]`/`[GOAL]` from a web/voice-spawned orchestrating-persona (doctor) session **without Telegram**, and the session resumes via the **existing** worker resume path. This unblocks goal-first doctor's clarify/goal-approval conversation on web/voice.

The design adds a web-side way to **deposit an answer** for an already-open question — it does **not** add a second resume engine. Resumption still flows through `_process_clarification_answers` → `_resume_goal` / `_resume_as_followup`, which read `pending_questions` exactly as before.

### How the open question is keyed to the web conversation
1. On a successful orchestrating-persona spawn, `chat.py` links the conversation to the spawned agent session via a new additive `conversations.agent_session_id` column.
2. When the session emits `[CLARIFY]`/`[GOAL]`, the worker registers an open `pending_questions` row (unchanged) — every such row already carries `session_id`.
3. The new endpoint resolves conversation → linked `session_id` → the session's open question, and deposits the answer onto **that existing row** (preserving its `kind`) via a new **session-keyed** deposit. The worker's existing tick then resumes it through the right path.

`session_id` is unique per session, so there's no bot-collision concern (the #348 message-id problem) — no `bot` scoping needed on the session-keyed path.

### Endpoint shape (additive, non-breaking)
- `POST /api/conversations/{id}/answer` `{ "answer": "..." }` → `{ ok, session_id, status }`. `400` empty answer · `404` no conversation · `409` no spawned session / no open question (never asked, already answered, or timed out).
- `GET /api/conversations/{id}` now includes `pending_question: { session_id, question, kind } | null` — present **only** while the spawned session is awaiting an answer. The client renders an answer affordance when present.

The existing Telegram reply-to-resume round-trip and the message-id `deposit_answer` / `get_open_question_by_message_id` paths are **untouched**.

## Test evidence
Run on nathan-linux (`~/.venvs/lifeos`), isolated worktree `/tmp/wt-403`:

- Targeted: `tests/test_conversations_api.py tests/test_agent_worker_claude_code_resume.py` → **39 passed** (incl. 9 new).
- Adjacent: `test_doctor_bot.py test_agent_worker_clarifications.py test_conversation_store.py test_agent_worker_codex_dispatch.py` → **109 passed**.
[...]

Issue #311, opened 2026-06-03
feat(chat): stream engine-handoff results back into the web thread


nbramia:
## Problem

The web `/chat` engine handoff (#310) spawns the Codex/Claude Code worker session and acknowledges it in the thread, but the **result** is delivered via Telegram + the `/agents` page — it never streams back into the web conversation. A web user who triggers "use codex to …" sees "Handed off — result will arrive via Telegram," then has to switch surfaces to see the outcome.

## Context

`POST /api/chat/handoff` spawns the session with `chat_id = settings.telegram_chat_id`, so the worker's completion notification routes to Telegram (the existing notification path). The web thread has no mechanism to receive an async result minutes later. This was a deliberate, documented MVP limitation in #310.

## Suggested approach

Give the web thread a way to receive the spawned session's terminal result:
- Have the handoff record the spawned `session_id` against the `conversation_id` (a mapping table or a placeholder assistant message).
- When the worker finalizes the session, write the result into the conversation store for that `conversation_id` (in addition to / instead of the Telegram notification when the origin was web).
- The web UI polls (or subscribes via SSE) for new messages on the conversation and renders the result into the thread when it lands — replacing the "running in the background" placeholder.

Mind: the worker runs out-of-process; it already has the session row + `chat_id`. A web-origin marker (e.g. `chat_id="web:<conversation_id>"` or a dedicated origin field) lets the worker route the result back to the conversation store instead of Telegram.

## Files involved

- `api/routes/chat.py` — `/api/chat/handoff` (record session↔conversation), plus a poll/stream endpoint for new messages
- `api/services/agent_worker/worker.py` — completion notification routing for web-origin sessions
- `api/services/conversation_store.py` — write the worker result into the thread
- `web/index.html` — poll/subscribe and render the late-arriving result

## Acceptance criteria

- [ ] A web-triggered engine handoff's result appears in the same web conversation thread (not only Telegram), replacing the "running in the background" placeholder.
- [ ] Telegram-origin handoffs are unaffected (still notify Telegram).
- [ ] No polling storm — bounded/backoff polling or SSE.

## Related

- Follow-up to #310 (web `/chat` engine handoff). Mentioned as the known limitation there.

nbramia:
## In Progress

Implementation PR opened: #418 — feat(chat): stream web-spawned session output into the conversation thread. Entering adversarial review.

PR #418, opened 2026-06-25
feat(chat): stream web-spawned session output into the conversation thread

nbramia:
## Summary
Closes **#311**. Completes the web/voice ↔ Telegram parity: a web-spawned agent session's progress + result now appear in the same `/chat` conversation thread, not only Telegram + `/agents`. Builds directly on the #409/#412 infra (the `conversations.agent_session_id` link + the existing 4s pending-question poll).

**Mechanism (additive — the Telegram/notify path is byte-identical):**
- `conversation_store.get_conversation_id_by_agent_session_id()` — the reverse of #409's link.
- Worker `_mirror_to_conversation(session_id, text)` — writes operator output into the linked conversation as an assistant message; **no-op when the session isn't linked** (every Telegram-origin session), which is what keeps AC2.
- **Incremental `[NOTIFY]` streaming** via an *additive* `conversation_mirror(session_id, body)` callback on `ClaudeCodeExecutor` (the existing `notification_callback` signature is unchanged; session_id is passed per-call, so it's concurrency-safe across the CLI thread pool). Terminal result/failure are mirrored at the dispatch handlers (covers codex too).
- `/api/chat/handoff` now links the spawned session to the conversation, so codex/claude handoffs mirror via the same path; the client starts the existing poll after a handoff.
- Client renders late-arriving assistant messages from the existing poll (dedup by message id — no new poll loop).

## Test evidence (nathan-linux, isolated worktree)
- Full unit suite: **2246 passed, 8 skipped**.
- Targeted dispatch/store/executor: **116 passed**.
- Browser (`test_pending_question_ui_browser.py`, incl. a new render-once-no-duplicate test): **6 passed**; prod :8000 healthy before/after.

## Review focus
- **AC2** — confirm the mirror is a strict no-op for unlinked (Telegram-origin) sessions and the Telegram send text is unchanged.
- **Client dedup** — the `seenMessageIds` seeding (first poll seeds from already-rendered messages; only assistant roles render; the user answer-echo isn't double-rendered).
- **Concurrency** — session_id passed per-call, no shared mutable state.
- **Test hygiene** — the dispatch test helpers now construct a default `ConversationStore()` (prod db path); read-only no-op there, but worth deciding whether the helpers should inject a temp-db store.

### DIFF
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -922,8 +922,11 @@ async def chat_handoff(request: HandoffRequest):
     from api.services.agent_worker.session_store import SessionStore
 
     working_dir = resolve_working_directory(task)
-    # Route the worker's completion notification to the operator's Telegram (the
-    # web thread can't receive an async result yet). None if Telegram isn't set.
+    # Also route the worker's completion notification to the operator's Telegram.
+    # As of #311 the spawned session's progress + result are mirrored back into
+    # this web/voice thread too (the conversation is linked to the session
+    # below), so Telegram is now an additional delivery channel, not the only
+    # one. None if Telegram isn't set.
     chat_id = getattr(settings, "telegram_chat_id", "") or None
 
     if engine == "codex":
--- a/api/routes/chat.py
+++ b/api/routes/chat.py
@@ -946,8 +946,8 @@ async def chat_handoff(request: HandoffRequest):
     session_id = result.get("session_id", "")
     ack = (
         f"🤝 Handed off to **{label}** — running in the background "
-        f"(session `{session_id[:12]}`). The result will arrive via Telegram and "
-        f"appear on the /agents page."
+        f"(session `{session_id[:12]}`). The result will appear right here, "
+        f"and also via Telegram and on the /agents page."
     )
     if request.conversation_id:
         try:
@@ -957,6 +957,16 @@ async def chat_handoff(request: HandoffRequest):
             )
         except Exception:
             logger.warning("failed to record handoff acknowledgment", exc_info=True)
+        # #311: link the conversation to the spawned session so the worker can
+        # mirror the session's progress + terminal result back into this thread
+        # (in addition to Telegram). Best-effort — a link failure only loses the
+        # web round-trip, not the Telegram delivery. Mirrors the orchestrating-
+        # persona path (chat.py set_agent_session_id after a doctor spawn).
+        if session_id:
+            try:
+                get_store().set_agent_session_id(request.conversation_id, session_id)
+            except Exception:
+                logger.warning("failed to link conversation to handoff session", exc_info=True)
 
     return {
         "ok": True,
--- a/api/routes/conversations.py
+++ b/api/routes/conversations.py
@@ -67,6 +67,13 @@ class ConversationDetailResponse(BaseModel):
     # conversation is currently awaiting an answer (#403). The client renders
     # an answer affordance and POSTs to `{id}/answer`. Absent/None otherwise.
     pending_question: Optional[PendingQuestionResponse] = None
+    # Whether a session spawned by this conversation is still running (#311).
+    # True iff the conversation links a session whose status is non-terminal.
+    # False when there's no linked session or it has reached a terminal status
+    # (completed/failed/budget_exceeded). The client's result-streaming poll
+    # uses this to STOP once the session is done and nothing awaits an answer —
+    # otherwise the 4s poll would run forever after the session finishes.
+    agent_session_active: bool = False
 
 
 class ConversationListResponse(BaseModel):
@@ -144,19 +151,34 @@ async def get_conversation(conversation_id: str):
     messages = store.get_messages(conversation_id)
 
     # Surface an open [CLARIFY]/[GOAL] from a session this conversation spawned
-    # (#403) so the client can show an answer affordance. Best-effort: a lookup
-    # failure (or no spawned session) just omits it — never breaks the read.
+    # (#403) so the client can show an answer affordance, and report whether that
+    # session is still running (#311) so the client's result-streaming poll can
+    # stop once it's done. Both are best-effort: a lookup failure (or no spawned
+    # session) just leaves pending_question=None / agent_session_active=False and
+    # never breaks the read.
     pending_question = None
+    agent_session_active = False
     if conv.agent_session_id:
         try:
-            from api.services.agent_worker.session_store import SessionStore
-            q = SessionStore().get_open_question_by_session_id(conv.agent_session_id)
+            from api.services.agent_worker.session_store import (
+                SessionStore,
+                TERMINAL_STATUSES,
+            )
+            session_store = SessionStore()
+            q = session_store.get_open_question_by_session_id(conv.agent_session_id)
             if q:
                 pending_question = PendingQuestionResponse(
                     session_id=conv.agent_session_id,
                     question=q.get("question") or "",
                     kind=q.get("kind") or "clarification",
                 )
+            # Active = the linked session row exists and hasn't reached a
+            # terminal status. A missing session row (never created, or pruned)
+            # counts as not-active, so the client stops polling.
+            session = session_store.get_by_session_id(conv.agent_session_id)
+            agent_session_active = bool(
+                session is not None and session.status not in TERMINAL_STATUSES
+            )
         except Exception:  # noqa: BLE001
             logger.warning("pending-question lookup failed", exc_info=True)
 
@@ -177,6 +199,7 @@ async def get_conversation(conversation_id: str):
             for m in messages
         ],
         pending_question=pending_question,
+        agent_session_active=agent_session_active,
     )
 
 
--- a/api/services/agent_worker/claude_code_executor.py
+++ b/api/services/agent_worker/claude_code_executor.py
@@ -660,7 +660,10 @@ class ClaudeCodeExecutor:
                     except Exception as exc:  # pragma: no cover — defensive
                         logger.warning("notification callback raised: %s", exc)
                     if self._conversation_mirror:  # #311: stream into the web thread
-                        self._conversation_mirror(session.session_id, body)
+                        try:
+                            self._conversation_mirror(session.session_id, body)
+                        except Exception as exc:  # pragma: no cover — defensive
+                            logger.warning("conversation mirror raised: %s", exc)
                     self.transcript_store.append(sid, "claude_code_clarify", {
                         "body": body, "body_chars": len(body),
                     })
@@ -676,7 +679,10 @@ class ClaudeCodeExecutor:
                         except Exception as exc:  # pragma: no cover — defensive
                             logger.warning("notification callback raised: %s", exc)
                         if self._conversation_mirror:  # #311: stream into the web thread
-                            self._conversation_mirror(session.session_id, body)
+                            try:
+                                self._conversation_mirror(session.session_id, body)
+                            except Exception as exc:  # pragma: no cover — defensive
+                                logger.warning("conversation mirror raised: %s", exc)
                     self.transcript_store.append(sid, "claude_code_notify", {
                         "body": body, "body_chars": len(body),
                     })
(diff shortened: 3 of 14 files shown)


nbramia:

| 1 | 4 pre-existing dispatch test helpers build a default `ConversationStore()` → open/migrate the **prod** `data/conversations.db` on the test machine | Action Required | **Accept (AR)** | Real footgun — exactly what the autouse `_isolate_vault_indexer_stores` fixture exists to prevent. Inject a tmp-db store into the 4 helpers (and/or an autouse conftest redirect). |
| 2 | Poll never terminates — runs every 4s forever after the session completes (AC3 "bounded" only partially met; #311 widens the trigger via handoff) | Action Required→Rec | **Accept (Rec)** | Add `agent_session_active` to GET; client stops when session terminal AND no pending_question. Feasible — the route already looks up the linked session. |
| 3 | Handoff poll uses post-`.then()` `state.currentConversationId` → can poll the wrong thread if user navigates mid-POST | Recommended | **Accept** | Capture the handoff conversation_id in a local const; poll that. |
| 4 | Executor `_conversation_mirror` calls not wrapped in try/except (unlike adjacent `_notify`) | Recommended | **Accept** | A raising sink must not abort the assistant-event loop; match the established guard. |
| 5 | Seed correctness rests on an undocumented ordering invariant (ack persisted before first poll) | Recommended | **Accept** | Add a defensive comment. |
| 6 | Handoff ack promises "appear right here" but link is best-effort | Recommended | **Reject** | The copy already hedges ("**also** via Telegram and on the /agents page"), so delivery isn't lost if the link fails. No change. |
| 7 | User-role messages silently dropped in renderNewMessages | Minor | **Reject** | Documented + correct for current behavior (worker only mirrors assistant roles; user echo handled by submitAnswer). |
| 8 | Ack string is a client-surface contract | Recommended-confirm | **Reject (verified)** | Confirmed display-only — `ask-stream.js` renders `d.message`; Telegram doesn't parse it. |
| 9 | No uniqueness constraint on agent_session_id / codex-child invariant / lookup-write not atomic | Minor | **Reject** | Single-user, last-write-wins by design; guarded; no action. |

Privacy/SQL posture clean (local-only, fully parameterized, no content logged). **Findings forwarded to addresser: 5** (1 AR + 4 Recommended).

nbramia:
## Review Round 1 — Referee Decisions (supplement: correctness reviewer)

A third specialist (correctness) surfaced two more accepted items + one minor, forwarded to the addresser:

| 10 | Codex completion + failure mirrors not gated on `not parent_session_id`; the "codex isn't spawned as a child" comment is **factually wrong** (`inter_agent.SPAWN_MODELS` includes codex) | Recommended | **Accept** | No-op today only because children are never conversation-linked — safety resting on an unstated invariant. Gate both codex mirrors (parity with claude_code) + fix the comment. |
| 11 | No `PRAGMA busy_timeout` on conversation_store connections; this PR adds the worker as a **second concurrent writer** to conversations.db | Recommended | **Accept** | Concurrent writers can hit "database is locked"; the mirror's try/except would silently drop the message. Add a busy_timeout. |
| 12 | Stale comment at `chat.py:925` ("web thread can't receive an async result yet") | Minor | **Accept** | This PR makes it false — update it. |
| — | spawn→link microsecond race; index on `agent_session_id`; first-poll seed swallowing an ultra-early mirror; pre-existing CLARIFY child-gating asymmetry | Minor | **Reject** | Best-effort by design / negligible at personal scale / inherent to the seed (covered by a comment) / pre-existing + faithfully mirrored. |

Confirmed clean by this reviewer: AC2 no-op airtight (SQLite NULL≠param), Telegram byte-identical, concurrency safe (per-call session_id, re-entrant connections), no duplicate-render. **Total forwarded to addresser this round: 8** (1 AR + 7 Recommended/Minor).

nbramia:
## Review Feedback Addressed — Round 1

A referee validated these findings; each was independently verified against the code before applying. Changes are surgical + additive — the Telegram/notify path is unchanged.

| 1 | Test-DB isolation: 4 pre-existing dispatch helpers built a default `ConversationStore()` → opened/migrated the **prod** `data/conversations.db` | **Applied** | Injected `ConversationStore(db_path=tmp_path/conversations.db)` into all four helpers (`_make_worker` ×2, `_recording_worker`, `_capturing_worker`), matching the `_mirroring_*` helpers. **Plus** an autouse conftest fixture `_isolate_conversation_store_db` that redirects `get_conversation_db_path()` to a per-test tmp file so no test can ever construct a prod-pathed store (mirrors `_isolate_vault_indexer_stores`). Hoisted the two redundant mid-file imports to the module top. |
| 2 | Poll never terminates — runs every 4s forever after the session completes (AC3) | **Applied** | Server: `GET /api/conversations/{id}` now returns `agent_session_active` — `true` iff the linked session row exists and its status ∉ `TERMINAL_STATUSES` (reuses the `SessionStore` already built for `pending_question`; a missing row → not-active). Client: `pollOnce` stops the loop when `agent_session_active === false` AND no `pending_question` (explicit `=== false` so an older server omitting the field keeps prior behavior; `renderNewMessages` runs first so a terminal-poll result still renders). Tests: 3 route unit tests (active/terminal/missing-row) + a browser test asserting the detail-GET count plateaus once terminal. |
| 3 | Handoff poll could target the wrong conversation if the user navigated mid-POST | **Applied** | Captured the handoff conversation id into a local `const handoffConversationId` **before** the fetch; use it for both the request body and `startPendingQuestionPolling`. |
| 4 | Executor `_conversation_mirror` calls not crash-guarded (unlike adjacent `_notify`) | **Applied** | Wrapped each of the 3 mirror calls in `try/except` + `logger.warning`, matching the notify guard. New test: a raising mirror is swallowed — session still completes, both notifies still stream. |
| 5 | Seed correctness rests on an undocumented ordering invariant | **Applied** | Added a defensive comment noting the seed assumes the handoff ack is persisted server-side (`chat_handoff` writes it synchronously before returning) before the first poll — must be revisited if ack persistence becomes async. Verified `chat_handoff` is synchronous. |

REJECTED findings (6–9) confirmed correct as-is and left unchanged per the referee decisions.

**Tests (nathan-linux, isolated `/tmp` worktree at the pushed commit `047daf8`, venv `~/.venvs/lifeos/bin/python`):**
- `pytest -m unit -q` → **2250 passed, 8 skipped** (0 failures; no `test_scheduler_watcher` flake). Run time 14:49.
- Browser: localhost-only server on a spare port + `LIFEOS_TEST_BASE_URL`, `tests/test_pending_question_ui_browser.py` → **7 passed**. Prod `:8000` healthy (200) before and after; test instance torn down.

**New commits:**
- `ac31e5e` fix: address review round 1 — isolate test conversations DB from prod
- `04dfff7` fix: address review round 1 — bound the web-thread result poll (AC3)
- `8ea3218` fix: address review round 1 — pin handoff conversation for the result poll
- `047daf8` fix: address review round 1 — crash-guard executor conversation-mirror

nbramia:
## Review Feedback Addressed — Round 1 (findings 6-8, third reviewer)

Three additional referee-accepted findings, folded into the same round-1 pass. Each independently verified before applying.

| 6 | Codex completion + failure mirrors not gated on `not session.parent_session_id` (claude_code path is); comment "codex isn't spawned as a child" is factually wrong | **Applied** | Gated both codex mirrors on non-child for parity + defense-in-depth. Verified the wrong comment: `codex` IS in `inter_agent.SPAWN_MODELS` (`CLI_ROUTINGS = ("claude_code", "codex")`), so a parent agent can dispatch a codex child via `_dispatch_codex_session` — it's a no-op today only because children are never conversation-linked. Replaced the comment with the real reason. Added 2 codex child-no-mirror tests (completion + failure). |
| 7 | No `busy_timeout` on `conversations.db` — new concurrent worker writer could hit "database is locked" and the mirror's try/except silently drops the message | **Applied (with correction)** | Added a shared `ConversationStore._connect()` setting `timeout=10.0` + `PRAGMA busy_timeout=10000` (matches `SessionStore._connect`), routed all 10 connection sites through it. Correction for the record: a bare `sqlite3.connect()` already sets `busy_timeout=5000` via Python's driver default (not 0) — so this widens the window from 5s→10s and makes the intent explicit rather than fixing a literal `busy_timeout=0`. The robustness concern (new concurrent writer) is real and addressed. |
| 8 | Stale comment in `chat_handoff` (~925): "the web thread can't receive an async result yet" | **Applied** | Now false — #311 is exactly what makes the web/voice thread receive results. Updated the comment to note Telegram is now an additional channel, not the only one. Routing unchanged. |

REJECTED findings (spawn-then-link race, agent_session_id index, first-poll ultra-early-mirror, pre-existing CLARIFY child-gating asymmetry) left untouched per the referee decisions.

**Tests (nathan-linux, isolated `/tmp` worktree at pushed commit `8f8c224`):**
- `pytest -m unit -q` → **2251 passed, 8 skipped, 1 failed**. The 1 failure is `test_scheduler_watcher.py::TestEndToEnd::test_external_edit_propagates_to_store` — the known timing-flake (passes in isolation: re-ran → 1 passed; unrelated to these changes, which don't touch the scheduler).
- Browser: localhost-only server on a spare port + `LIFEOS_TEST_BASE_URL`, `tests/test_pending_question_ui_browser.py` → **7 passed**. Prod `:8000` healthy (200) before and after; test instance torn down.

**New commits (this batch):**
- `a8c99fb` fix: address review round 1 — gate codex mirrors on non-child + fix comment
- `0efe14e` fix: address review round 1 — explicit 10s busy_timeout on conversations.db
- `8f8c224` fix: address review round 1 — refresh stale chat_handoff Telegram comment

Full round-1 set is now 8 findings (1-5 + 6-8), all applied and verified.

nbramia:
## Review Feedback Addressed — AC3 terminal-stop race

Fixes a real race in the AC3 terminal-stop logic (introduced in `04dfff7`): the executor flips the session row to a terminal status (`claude_code_executor.py:555`, `codex_executor.py:365`) BEFORE the dispatch handler writes the terminal-result mirror into the conversation (the mirror lands ~1s later, after a `_telegram_send_with_id` network call). So `GET` can report `agent_session_active=false` while the result message is not yet stored. A single terminal poll landing in that window would `stopPendingQuestionPolling()` before the result rendered — and for the **codex handoff path the terminal mirror is the only web output**, so the result would be lost until a manual reload.

| Finding | Action | Details |
|---------|--------|---------|
| AC3 terminal-stop race — single terminal poll can stop before the result mirror is stored | **Applied** | Client (`web/chat/pending-question.js`): require **two consecutive terminal polls** before stopping (`terminalPollCount`, reset in start/stop, reset by any active poll or pending question). `renderNewMessages` runs before the stop check every poll, so a result written during the race is stored + rendered on the second terminal poll (≥1 full ~4s cycle after the status flip) before the loop stops. New browser test: a terminal poll with the result not-yet-stored does NOT stop; the next poll renders the late result and only then stops. Updated the existing stop test for the two-poll timing. |

**Tests (nathan-linux, isolated `/tmp` worktree at pushed commit `cf0f7d0`):**
- `pytest -m unit -q` → **2252 passed, 8 skipped, 0 failed** (clean run; no scheduler flake this time).
- Browser: localhost-only server on a spare port + `LIFEOS_TEST_BASE_URL`, `tests/test_pending_question_ui_browser.py` → **8 passed** (the 7 prior + the new race-guard test). Prod `:8000` healthy (200) before and after; test instance torn down.

**New commit:** `cf0f7d0` fix: address review round 1 — require 2 terminal polls to close AC3 stop race

nbramia:
## Review Round 1 — complete (with a self-verify fix)

All 8 referee-forwarded findings addressed and verified. During the post-address self-verify, the orchestrator found one **additional concrete bug** in the AC3 terminal-stop and forwarded it:

- **AC3 stop race**: the executor flips the session row to `STATUS_COMPLETED` *before* the dispatch handler writes the terminal-result mirror (which lands after a Telegram network send). A client poll landing in that sub-second window saw `agent_session_active=false` with the result not yet stored and stopped early — losing the **codex handoff** result (its only web output) ~10–15% of the time until reload. Fixed (`cf0f7d0`): require **two consecutive terminal polls** before stopping (`renderNewMessages` runs first, so the late result renders on the 2nd poll), with a direct regression test.

Final verification at `cf0f7d0`: **2252 unit passed / 0 failed**, **8 browser passed**. No further concrete findings — review loop complete, proceeding to merge.

nbramia:
## Merging

Review loop complete (1 round + a self-verify race fix), 8 browser + 2252 unit green at cf0f7d0. Squash-merging to main.

Issue #357, opened 2026-06-21
test: flaky test_telegram_multibot test_primary_uses_legacy_state_file (order-dependent)


nbramia:
## Problem

`tests/test_telegram_multibot.py::TestListenerWiring::test_primary_uses_legacy_state_file` fails intermittently in the **full** unit-suite run but passes in isolation and when its own file runs alone (23/23). It blocked a `git push` via the pre-push hook (`1 failed, 2084 passed`), then passed on re-run — a classic test-isolation / ordering flake (likely shared state-file path leaking across tests).

## Context

Surfaced during the #349 merge. The test is part of the multibot/doctor-bot feature (#348/#354) and is unrelated to #349 — confirmed by running it in isolation (passes) and the whole multibot file (23 passed). Because the pre-push hook runs the full suite, an order-dependent flake here can block unrelated pushes.

## Suggested approach

Isolate the test's state: ensure the legacy state-file path is a per-test `tmp_path` (or monkeypatched) rather than a process-global/default path, so a prior test in the run can't leave a file that flips this assertion. Verify with `pytest -p no:randomly` vs randomized ordering, or run the full suite a few times to reproduce.

## Files involved

- `tests/test_telegram_multibot.py` — `TestListenerWiring::test_primary_uses_legacy_state_file`
- whichever module owns the legacy listener state-file path it asserts on.

## Acceptance criteria

- [ ] The test passes deterministically in the full suite across repeated runs and randomized ordering.
- [ ] No reliance on process-global/default state-file paths in the test.

nbramia:
## In Progress

Implementation PR created: #420 — test: isolate Telegram state file from production. Note: the rare flake could not be reproduced across many real-mode runs; this adds the project's established autouse state-file isolation (removing the global-default-path gap). Entering review.

nbramia:
[...]

PR #420, opened 2026-06-25
test: isolate Telegram state file from production (#357)

nbramia:
## Summary
Closes **#357**. Adds an autouse fixture that redirects `TelegramBotListener._STATE_FILE` to a per-test tmp path, so no test relies on or leaks the process-global default `data/telegram_state.json`. Mirrors the existing `_isolate_vault_indexer_stores` / `_isolate_conversation_store_db` autouse fixtures.

## Reproduction note (transparency)
I could **not** reproduce the original intermittent failure across many real-mode (`-n auto --dist loadscope`) runs (6 real-mode + 1 sequential + 3 forced-order, all green). Both of the test's asserts are **deterministic** — an explicit `primary` config makes `_is_primary` true, and its own `patch.object(_STATE_FILE, state)` makes `_state_file == state` — so the failure is a cross-module isolation leak, not the test's logic.

The most likely leak is reliance on the shared default state-file path: any listener built without patching `_STATE_FILE` (e.g. the ones `api.main`'s lifespan starts for a real `TestClient`) reads and (via `_save_last_update_id`) writes the real `data/telegram_state.json`. This PR removes that gap.

**Honest scope:** AC #2 ("no reliance on process-global/default state-file paths") is fully met. AC #1 ("deterministic across repeated/randomized runs") is *supported* by 10/10 green real-mode runs on the branch, but since the rare flake never reproduced, I can't prove it's eliminated — this hardens the established cause rather than fixing a reproduced failure.

## Test evidence
10/10 real-mode full-suite runs green on the branch (2252 passed, 8 skipped each). No test asserts the default `_STATE_FILE` value; telegram tests that patch it themselves nest over the fixture.

## Review focus
The autouse fixture affects all tests — confirm it can't alter telegram-test behavior (those that patch `_STATE_FILE` nest fine; named-bot tests use the else-branch path) and that `monkeypatch` restores cleanly.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -259,6 +259,27 @@ def _isolate_vault_indexer_stores(tmp_path, monkeypatch):
 
 
 @pytest.fixture(autouse=True)
+def _isolate_telegram_state_file(tmp_path, monkeypatch):
+    """Keep tests off the production Telegram offset file (#357).
+
+    ``TelegramBotListener._STATE_FILE`` defaults to the process-global relative
+    path ``data/telegram_state.json``. Any listener built without patching it —
+    notably the ones ``api.main``'s lifespan starts for a real ``TestClient``
+    app — reads and (via ``_save_last_update_id``) writes that shared file, which
+    is the suspected source of the order-dependent flake in
+    ``test_primary_uses_legacy_state_file``. Redirect the legacy/primary state
+    file to a per-test tmp path so no test touches or leaks the shared default;
+    the telegram tests that patch ``_STATE_FILE`` themselves simply nest over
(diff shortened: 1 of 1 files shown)


nbramia:
## Review Round 1 — no actionable findings, review loop complete

Single reviewer verified every claim directly (nesting of the tests' own `patch.object`, no test asserts the default value, Path-type correctness, monkeypatch restoration, no import side effects) and confirmed **merge-ready** — no Action Required or Recommended. One **Minor** (already disclosed): the inline named-bot path (`data/telegram_state_{name}.json`, `telegram.py:467`) isn't isolated — sufficient for this primary-path flake; noted as a potential future hardening, not a change to this PR. Proceeding to merge.

Issue #419, opened 2026-06-25
fix(agent-worker): verify deploy landed before doctor reports 'Shipped'


nbramia:
## Problem

The goal-first doctor finishes a fix by pulling the merged code into the canonical checkout and restarting, then reports success — but it never verifies the deploy actually took effect. If the `git pull` is a no-op or fails, the doctor still sends `[NOTIFY] Shipped: PR #<n> merged to main, server restarted` while the running server is still on the old code. A confident false success.

This isn't hypothetical: it is exactly the failure mode hit this session. Separately, there is no signal anywhere when the running server drifts behind `main`.

## Context

Found while verifying the goal-first doctor + #311 end-to-end. Two compounding facts:

1. **No deploy verification.** `config/personas/doctor.md` (line ~36) does `git -C ~/Code/LifeOS checkout main && git -C ~/Code/LifeOS pull`, then restarts, then sends the "Shipped" `[NOTIFY]` (line ~38). Nothing checks that the checkout's HEAD advanced to the merged SHA, or that the restart picked it up.
2. **A real way the pull silently fails.** The production checkout was found with `core.bare=true` (a misconfiguration, likely from a worktree operation), which makes `git pull`/`checkout` fail with `fatal: this operation must be run in a work tree`. Under that state the doctor's pull is a silent no-op and its "Shipped" notice is false.

Also observed: production sat **10 commits behind `origin/main`** for an entire session (the whole goal-first epic + 4 follow-ups merged on GitHub but never deployed), caught only because an end-to-end check failed. `core.bare` was reset to `false` and prod fast-forwarded to current `main` this session, but nothing prevents recurrence or detects the false-success.

Single-user self-hosted system; deploys are expected to be either a manual `git pull` or the doctor's self-deploy.

## Suggested approach

**Primary — make the doctor's deploy step self-verifying:**
- After the post-merge `git pull`, confirm the canonical checkout's HEAD == the merged commit SHA **and** the checkout is a real work tree (`core.bare` false / `git rev-parse --is-inside-work-tree` true) **before** sending "Shipped".
- On failure, `[NOTIFY]` a clear deploy-failure naming the stuck SHA and reason (e.g. `merged #N but the server is still on <old-sha> — pull failed: <reason>`) and stop, rather than reporting success.
[...]

PR #421, opened 2026-06-25
feat(server): verify-deployed guard so the doctor can't falsely report 'Shipped'

nbramia:
## Summary
Closes **#419**. The goal-first doctor pulls `main` then restarts and reports "Shipped" — but never verified the pull landed. A silent pull failure (notably a bare/misconfigured checkout: `core.bare=true` makes `git pull`/`checkout` a no-op error — hit in practice this past session) let it claim success while the **old code kept running**.

- `scripts/server.sh`: new `verify-deployed [<sha>]` — exits 0 only if `PROJECT_DIR` is a real work tree (not bare) whose HEAD matches `<sha>` (or `origin/main` when omitted); else prints `not-deployed: <reason>` to stderr and exits 1.
- `config/personas/doctor.md`: after the post-merge pull, run `verify-deployed` **before** "Shipped"; on failure `[NOTIFY]` a deploy failure (naming the stuck SHA + reason) and stop.
- `docs/guides/doctor-bot.md`: document it in the pick-up-after-merge step.

The issue's optional secondary item (a "prod is N commits behind `origin/main`" health signal) is **deferred** to keep this focused on the high-value guard.

## Test evidence
Targeted: **40 passed** (`test_agent_worker_self_restart.py` + `test_doctor_bot.py`), incl. the new `test_verify_deployed_checks_worktree_and_head` (deployed / abbreviated-sha / wrong-sha / **bare-repo** / no-arg-vs-origin-main) and a persona-contract needle so the verification can't silently regress. Full unit suite running. Local smoke: `verify-deployed <HEAD>` → exit 0; wrong sha → exit 1.

## Review focus
The `verify_deployed` bash function (abbreviated-sha prefix match; the `is-inside-work-tree` / bare-repo guard; no-arg origin/main fallback) and that the doctor.md wiring makes "Shipped" strictly conditional on a passing verify.

### DIFF
--- a/config/personas/doctor.md
+++ b/config/personas/doctor.md
@@ -33,7 +33,7 @@ First **file the work as GitHub issue(s)** via `/draft-issue` (capture the numbe
 
 **Supervise; don't hand-code.** You decompose the goal and review; **subagents do the implementation** (each `/implement` run is itself a supervised implementer with its own adversarial review). Keep sub-tasks bounded — a single supervised `/implement` run can exceed the headless background-wait ceiling, so size the pieces or drive `/implement` runs sequentially rather than spawning one giant waited subagent.
 
-**After the final merge to `main`:** bring the canonical checkout to the merged code before restarting — `git -C ~/Code/LifeOS checkout main && git -C ~/Code/LifeOS pull` — then clean up any remaining worktree/branch, then restart (see Restarting below).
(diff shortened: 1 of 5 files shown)


nbramia:

Reviewer independently reproduced all behavior (exit codes, bare-repo/destroyed-repo guard, git-style prefix match, detached HEAD, strict "Shipped" gating). **Merge-ready** — no Action Required/Recommended. Folding in the two Minor defensive items (cheap hardening for any direct caller, even though the doctor's path can't hit them):

| 1 | A too-short explicit `<sha>` (e.g. `verify-deployed 5`) could prefix-collide and false-match `deployed` | Minor | **Accept** | Add a `>=7` min-length guard (return 2) so an ambiguous sha is rejected, not silently matched. Removes a real footgun for direct CLI use. |
| 2 | Tests don't cover the short-sha rejection or the no-origin-main (`fetch first`) branch | Minor | **Accept** | Add both cases to the existing test. |

Applying now (small, in-PR).

nbramia:
## Review addressed — merging

Both Minor items applied (`3db51b8`): min-length (`>=7`) guard rejecting an ambiguously-short sha (return 2), + tests for the short-sha rejection and the no-origin-main branch. Targeted suite green (10 passed); full unit suite was green on the base (2253). Review loop complete.

Issue #356, opened 2026-06-21
fix(agent-worker): spawned child [CLARIFY] still messages the operator


nbramia:
## Problem

After #349, a spawned `claude_code` child stays silent to the operator on the **completed** path (live `[NOTIFY]` + heartbeat + terminal send all suppressed for sessions with a `parent_session_id`). But a child that emits a `[CLARIFY]` block still streams the question to the operator and goes `BLOCKED` with a `pending_question` registered against the operator — inconsistent with the "children report to their parent, not the operator" design.

## Context

Surfaced during the #349 review (round 1 finding #6, re-confirmed round 2). Plan-mode `BLOCKED` is unreachable for spawned children (`inter_agent.spawn` never sets `plan_mode`), so only the `[CLARIFY]` edge remains.

Naively gating the `[CLARIFY]` notify on `is_child` (mirroring the `[NOTIFY]` gate) is **wrong**: the child would still go `BLOCKED` but the operator would never see the question, stranding the child with no way to get an answer. The correct fix routes the child's clarification **to its parent** (e.g. surface it as the child's terminal output / a pending message the parent reads on resume, or have the parent answer it), so the parent — which owns the operator-facing conversation — decides how to respond.

## Files involved

- `api/services/agent_worker/claude_code_executor.py` — `[CLARIFY]` handling (~line 558) and the `awaiting_clarification` → BLOCKED outcome.
- `api/services/agent_worker/worker.py` — `_dispatch_claude_code_session` BLOCKED branch (~line 1234) and how a spawned child's BLOCKED state is surfaced to its parent.

## Acceptance criteria

- [ ] A spawned `claude_code` child that emits `[CLARIFY]` does not send the question directly to the operator.
- [ ] The child's clarification reaches its parent (parent can answer or relay), and the child can resume once answered — no permanent stranding.
- [ ] Operator `/claude` and vault `#claude` clarifications are unchanged.

nbramia:
## In Progress

Implementation PR created: #422 — route a spawned child's [CLARIFY] to its parent. Note a scope decision in the PR (child completes + parent re-engages vs. literal same-session resume). Entering review.

nbramia:
## Follow-up

The re-spawn-only limitation of the shipped fix is being addressed in #428 — a parent can now answer a completed child's `[needs clarification]` question via `lifeos_agent_send`, which reopens the child's persisted CLI session (`-r`) with full prior context instead of requiring a fresh re-spawn. Entering adversarial review.

[...]

Issue #349, opened 2026-06-21
feat(agent-worker): Haiku tier + one message for delegated tasks


nbramia:
## Problem

When a `#local` agent-worker task delegates mid-run to a Claude Code child agent (via `lifeos_agent_spawn`), the operator gets **two** Telegram messages: the child streams its findings live as `[NOTIFY]` blocks, then the parent local worker sends its own completion summary. The two messages duplicate each other with different framing, and the child often splits its answer across multiple `[NOTIFY]` sends. There is also no way for the worker to escalate cheaply — the Claude Code child is always Opus.

This surfaced via the "Daily World Cup Games Synopsis" schedule (`68f80d9f`, `#local`, daily 9am ET), which looked like it was "firing twice." It is **one** schedule; the doubling is the parent + child both messaging. On days the worker did not spawn a child, only one message arrived — matching the "some days two" observation. (A separate `#cloud` schedule `e38dc3ca` existed only on Jun 14 and is already deleted; not the ongoing cause.)

## Context

Root-cause trace (code as of this branch):

- `lifeos_agent_spawn` (`api/services/agent_worker/inter_agent.py:142-150`, `215-333`) is an explicit tool the agent calls. Its `model` enum is `{claude, local, claude_code, codex}` and is stored in the child session's `routing`. There is **no LLM-tier choice** — claude_code is always Opus.
- The claude_code CLI model is hardcoded `--model opus` at `api/services/agent_worker/claude_code_executor.py:295` (built in `_build_command`, which already receives `session_id`).
- `[NOTIFY]` bodies are stripped from `final_text` (`claude_code_executor.py:509-512`) and streamed **only** to Telegram + transcript (`526-538`). The parent receives **only** the child's `final_text` (`worker.py:1946-1966` `_child_final_text` reads `claude_code_completed.payload["final_text"]`), so the child's rich `[NOTIFY]` content never reaches the parent — which is why the parent re-derived a plainer list.
- Spawned children (`parent_session_id` set) already skip the parent-style `_notify_terminal`, but their live `[NOTIFY]` stream is **not** gated on being a child, so it still reaches the operator.
- The parent completion message is built in `worker.py:1669-1752` `_completion_summary`; `_worker_label` (`worker.py:66-75`) only says "Local/Cloud agent worker" — no escalation indicator.
- `Session` carries `parent_session_id` but has no model field; `session_store` already uses additive `if col in row.keys()` migrations.
[...]

PR #422, opened 2026-06-25
fix(agent-worker): route a spawned child's [CLARIFY] to its parent, not the operator

nbramia:
## Summary
Closes **#356**. A spawned `claude_code` child that emitted `[CLARIFY]` (a) streamed the question to the **operator** (executor's clarify loop was ungated, unlike `[NOTIFY]`/`[GOAL]`) and (b) went **BLOCKED** with a `pending_question` against the child — which **permanently stranded its yielded parent** (a parent only resumes once every child is terminal, and BLOCKED isn't terminal).

**Fix (executor-only, surgical):** for a child, fold the `[CLARIFY]` question into the child's output (`notify_bodies`, prefixed `[needs clarification]`, mirroring the #349 `[NOTIFY]` folding) and **don't** set `pending_clarification` — so no operator notify and no BLOCKED outcome. The child completes; the parent (which owns the operator conversation) reads the question on resume via `_child_final_text` and decides. Operator `/claude` and vault `#claude` clarifications are **unchanged** (the non-child branch is byte-identical).

## Scope decision (please weigh in)
This removes the leak + the strand and routes the question to the parent. The child **completes** rather than literally resuming its paused session — the parent re-engages (re-spawn with the answer / relay / proceed). True mid-flight **same-session** child resume is a larger inter-agent protocol change: a yielded parent only resumes on *terminal* children, and `inter_agent.send` rejects *terminal* targets, so resuming a paused child mid-flight would require extending the yield/resume model. I judged that out of scope for the bug fix and worth a separate issue if the in-flight-resume capability is wanted. **Is this proportionate, or do you want the fuller protocol change?**

## Test evidence
Targeted (executor + dispatch + inter_agent): **89 passed**, incl. the new `test_child_clarify_folds_into_final_text_and_does_not_block` and the existing `test_clarify_returns_blocked_with_question` (non-child path, unchanged — AC #3). Full unit suite running.

## Review focus
- Does the child path correctly **not** notify (Telegram + web mirror), **not** BLOCK, and fold the question into `final_text` so `_child_final_text` surfaces it to the parent?
- Is the non-child path byte-identical (AC #3)?
- **The scope decision above** — is "child completes, parent re-engages" an acceptable read of AC #2 ("the child can resume once answered"), or is the fuller same-session-resume protocol needed?

### DIFF
--- a/api/services/agent_worker/claude_code_executor.py
+++ b/api/services/agent_worker/claude_code_executor.py
@@ -652,9 +652,25 @@ class ClaudeCodeExecutor:
                 if scan.narrative.strip():
                     state.final_text = scan.narrative.strip()
                 for body in scan.clarify:
-                    state.pending_clarification = body
                     state.notifications_sent += 1
                     state.last_notify_at = time.time()
+                    if state.is_child:
+                        # #356: a spawned child must not pause on an operator reply.
+                        # The operator owns no thread to a child, and a BLOCKED child
+                        # would strand its yielded parent (which only resumes once
+                        # every child is terminal). So fold the question into the
+                        # child's output — prefixed so the parent recognizes it as a
+                        # request, like [NOTIFY] folding (#349) — and let the turn
+                        # complete. The PARENT, which owns the operator conversation,
+                        # reads it on resume and decides (answer via a re-spawn,
+                        # relay to the operator, or proceed). Crucially DON'T set
+                        # pending_clarification: no BLOCKED outcome, no operator notify.
+                        state.notify_bodies.append(f"[needs clarification] {body}")
+                        self.transcript_store.append(sid, "claude_code_child_clarify_folded", {
+                            "body": body, "body_chars": len(body),
+                        })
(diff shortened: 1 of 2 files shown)


nbramia:
## Review Round 1 — no actionable findings, review loop complete

Reviewer traced the full chain (executor fold → `_effective_final_text` → `_child_final_text` → parent resume turn) and confirmed: child path skips operator notify/mirror, avoids BLOCKED (no `pending_clarification`), and surfaces the question to the parent; non-child path byte-identical (AC #3). **Merge-ready, no Action Required.** One Minor (clarify folds ahead of notify within a turn — cosmetic, both reach the parent) → no change. **Scope decision independently validated as defensible**: terminal-completion is the only way to unblock the strand, and same-session mid-flight resume is genuinely a separate protocol change (blocked by `inter_agent.send` rejecting terminal targets + the terminal-only resume gate) — correctly deferred. Full suite green (2254). Merging.

Issue #400, opened 2026-06-24
fix(agent-worker): integration-branch + worktree crash-safety


nbramia:
## Problem

The doctor's lifecycle makes the **integration branch + worktree a first-class, long-lived artifact** (created up front off `main`, sub-PRs land on it, deleted at the end — steps 2 and 5 of the goal pipeline). Two crash-safety holes make that lifecycle fragile:

**(a) Worktree/branch collision.** Worktree setup is a linear `git worktree add` with cleanup (`git worktree remove`) only on the happy path. A prior crashed run leaves the worktree dir and branch behind, so the next `git worktree add` fails ("already exists" / "already checked out") with no pre-flight prune or reuse.

**(b) Resume-vs-re-execute.** Fresh-vs-resume dispatch is decided solely by whether `session.claude_code_session_id` is set, and that UUID is only persisted when the CLI emits its `init` event. If the worker dies after the subprocess spawns but before the init event persists, the field stays NULL and the next dispatch **re-executes the original prompt** instead of resuming — for a doctor turn that can re-file an issue or restart `/implement`.

## Context

- `config/personas/doctor.md` implement pipeline — worktree created/cleaned only on the happy path.
- `scripts/cleanup-worktrees.sh` already wraps `git worktree prune` but is never called pre-flight.
- `api/services/agent_worker/worker.py:1239` — `is_resume = bool(session.claude_code_session_id)` is the execute-vs-resume fork; NULL → re-runs the spawn prompt (`:1252-1271`).
- `api/services/agent_worker/claude_code_executor.py:515-521` — init handler is the only place the UUID persists during a run; the subprocess spawns in `_run` (`:350`) before that — the crash window.
- `api/services/agent_worker/session_store.py:503-517` — `set_claude_code_session_id`.

## Suggested approach

- **Worktree idempotency:** pre-flight `git worktree prune` (reuse `scripts/cleanup-worktrees.sh`), reuse-or-allocate-unique path, and guarantee cleanup on **any** exit (success/failure/crash), expressed as an idempotent recipe — not a linear add/remove pair. Generalize to the integration branch (delete branch + worktree at end-of-goal = step 5).
- **Resume safety:** persist a "started/spawning" marker before launching the subprocess (or detect partial progress), so the dispatch fork resumes (or no-ops) rather than re-running a side-effecting prompt.

## Files involved

- `config/personas/doctor.md` (lifecycle steps 2 & 5)
[...]

Issue #379, opened 2026-06-22
fix(agent-worker): operator kill leaves Claude Code subprocess running until next poll


nbramia:
## Problem

`POST /api/agents/sessions/{session_id}/kill` returns success immediately (e.g. `{"killed": ["sess_demo123"], "failures": []}`) but the underlying Claude Code worker OS subprocess (`claude -p ... --model opus ...`) keeps running. The endpoint marks the session row killed and tears down the *managed/cloud* remote, but it never signals the local CLI subprocess — so a "successfully killed" session can keep consuming compute/tokens and can still act (e.g. send user-facing `[NOTIFY]` messages) until the worker's next managed poll reconciles it.

## Context

Observed empirically during testing: a Claude Code session was spawned via a voice handoff, the kill endpoint was called and returned `killed: ["sess_demo123"]`, but the `claude -p` subprocess was still alive and investigating in plan mode on Opus until it was manually `kill -9`'d. The session row was flipped to killed while the OS process was never signalled.

The endpoint's own docstring/comments acknowledge this is a "local-only kill" that relies on the worker's next managed poll to reconcile the remote side (see the `_maybe_managed_driver` docstring in `api/routes/agents.py`). The shared `teardown_session` helper only (a) best-effort kills the managed cloud session and (b) flips local DB status + appends a transcript event — it has no path to terminate the local Claude Code CLI subprocess.

This matters because an operator killing a runaway / expensive / misbehaving agent reasonably expects compute to stop promptly. Today the OS process can linger until the next poll, continuing to spend tokens and potentially sending user-facing notifications after a "successful" kill.

**Cross-process constraint:** the kill endpoint runs in the API process, while the `claude -p` subprocess is owned and spawned (via `subprocess.Popen`) by the separate agent-worker process. The session store currently records no OS PID / process-group id for the subprocess, so the API process has nothing to signal even if it wanted to.

## Suggested approach

Directional, not prescriptive: the operator-kill path should signal/terminate the worker subprocess promptly rather than waiting for the next managed poll. Options to weigh given the cross-process boundary:

- Record the worker subprocess PID / process-group id alongside the session (e.g. in the session store) at spawn time so the kill can `SIGTERM`/`SIGKILL` it (process-group kill if the CLI spawns children).
- Or have the kill endpoint send an IPC nudge to the agent-worker so the worker reaps the subprocess immediately on the kill rather than at the next poll tick.

[...]

PR #423, opened 2026-06-26
fix(agent-worker): operator kill terminates the local CLI subprocess promptly

nbramia:
## Summary
Closes **#379**. `POST /api/agents/sessions/{id}/kill` flipped the DB row FAILED + tore down the managed remote but **never signalled the local `claude -p` (or codex) OS subprocess** — owned by the separate worker process — so it kept running (compute + could still notify) until the next poll. Now the kill terminates it within ~seconds.

**Mechanism (no schema change — the PID rides in the append-only transcript):**
- The executor spawns the subprocess with `start_new_session=True` (its own process-group leader, so `killpg` hits the CLI + its children but **not** the worker) and records a `claude_code_pid` / `codex_pid` `{pid,pgid}` transcript event after spawn.
- `teardown_session` (which the kill endpoint loops over the subtree) reads the latest pid event and, for LOCAL routings, does a liveness check → `killpg(SIGTERM)` → ~2s bounded grace → `killpg(SIGKILL)`. Best-effort: a missing/stale pid or signal error never breaks teardown.
- **No post-kill `[NOTIFY]`**: an externally-killed subprocess would otherwise hit the FAILED path and emit a spurious "session failed" notice. A guard in the executor — if the row is already FAILED when `proc.wait()` returns (only the operator kill flips it mid-run) — returns `REASON_KILLED`, and the dispatch suppresses the notice + web mirror.
- Cascade (`operator_killed`/`cascade_killed`), managed/cloud teardown, and the `{"killed":[...],"failures":[...]}` response are preserved. Full **codex parity**.

## Test evidence
Targeted (executor + inter_agent + dispatch + kill-api): **106 passed**; codex: **26 passed**; full unit suite: **2263 passed, 8 skipped**. New tests cover the pid event + `start_new_session`, the silent guard (REASON_KILLED, no `claude_code_failed`/COMPLETED), teardown SIGTERM→SIGKILL/stale-pid/clean-exit, managed-session skips local kill but still flips status, dispatch notice-suppression, and the endpoint driving a 2-node subtree kill.

## Review focus
- **Cross-process signal safety**: `start_new_session` ⇒ `killpg` can't reach the worker; the pid-reuse liveness check; best-effort never breaks teardown.
- **The kill-detection race**: the guard keys on the DB row (FAILED set only by the operator kill while the executor holds RUNNING) — not the returncode. Is there any other concurrent FAILED-writer (watchdog/timeout, #400 guard)?
- **No post-kill notify** for both claude_code and codex.
- **Dogfooding note**: the real cross-process signal is mocked in tests (`os.kill`/`os.killpg`); the live "kill a real `claude -p` and the OS process is gone in <~3s" path will be dogfooded on the worker after merge.

### DIFF
--- a/api/routes/agents.py
+++ b/api/routes/agents.py
@@ -859,36 +859,44 @@ async def operator_kill_session(session_id: str, body: KillRequest | None = None
     try:
         from api.services.agent_worker.inter_agent import teardown_session as _teardown
 
-        killed: list[str] = []
-        failures: list[dict[str, str]] = []
-        for idx, s in enumerate(subtree):
-            if s.status in TERMINAL_STATUSES:
-                continue
-            if idx == 0:
-                kind = "operator_killed"
-                payload = {
-                    "reason": reason,
-                    "managed_remote": s.managed_agent_session_id,
-                }
-            else:
-                kind = "cascade_killed"
-                payload = {
-                    "root": target.session_id,
-                    "reason": reason,
-                    "managed_remote": s.managed_agent_session_id,
-                }
-            result = _teardown(
-                session_store, transcript_store, s,
-                transcript_kind=kind,
-                transcript_payload=payload,
-                managed_driver=driver,
-            )
-            killed.append(s.session_id)
-            if result.get("managed_failure"):
-                failures.append({
-                    "session_id": s.session_id,
-                    "reason": result["managed_failure"],
-                })
+        def _run_teardown() -> tuple[list[str], list[dict[str, str]]]:
+            # #379: the per-CLI-child subprocess reap inside teardown_session does
+            # a blocking SIGTERM→grace(up to ~2s)→SIGKILL. Run the whole subtree
+            # teardown off the event loop (mirror spawn_agent's asyncio.to_thread)
+            # so a multi-node kill can't stall all HTTP for N×grace seconds.
+            killed: list[str] = []
+            failures: list[dict[str, str]] = []
+            for idx, s in enumerate(subtree):
+                if s.status in TERMINAL_STATUSES:
+                    continue
+                if idx == 0:
+                    kind = "operator_killed"
+                    payload = {
+                        "reason": reason,
+                        "managed_remote": s.managed_agent_session_id,
+                    }
+                else:
+                    kind = "cascade_killed"
+                    payload = {
+                        "root": target.session_id,
+                        "reason": reason,
+                        "managed_remote": s.managed_agent_session_id,
+                    }
+                result = _teardown(
+                    session_store, transcript_store, s,
+                    transcript_kind=kind,
+                    transcript_payload=payload,
+                    managed_driver=driver,
+                )
+                killed.append(s.session_id)
+                if result.get("managed_failure"):
+                    failures.append({
+                        "session_id": s.session_id,
+                        "reason": result["managed_failure"],
+                    })
+            return killed, failures
+
+        killed, failures = await asyncio.to_thread(_run_teardown)
         return {"killed": killed, "failures": failures}
     finally:
         if driver is not None:
--- a/api/services/agent_worker/claude_code_executor.py
+++ b/api/services/agent_worker/claude_code_executor.py
@@ -216,6 +216,10 @@ REASON_AWAITING_CLARIFICATION = "awaiting_clarification"
 REASON_AWAITING_GOAL_APPROVAL = "awaiting_goal_approval"
 REASON_TIMEOUT = "timeout"
 REASON_BINARY_NOT_FOUND = "binary_not_found"
+# #379: the operator kill endpoint flips the session row to FAILED and signals
+# our subprocess. When `proc.wait()` returns after such a kill, we exit silently
+# under this reason so the worker doesn't fire a spurious "session failed" notice.
+REASON_KILLED = "killed"
 
 
 @dataclass
@@ -466,6 +470,10 @@ class ClaudeCodeExecutor:
                 cwd=working_dir,
                 text=True,
                 env=self._clean_env(),
+                # #379: own session/process-group leader so the operator kill can
+                # `os.killpg(pgid, ...)` the CLI + every child it spawns WITHOUT
+                # touching this worker process (which shares the worker's group).
+                start_new_session=True,
             )
         except FileNotFoundError as exc:
             self.transcript_store.append(sid, "claude_code_binary_not_found", {"error": str(exc)})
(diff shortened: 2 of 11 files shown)


nbramia:

| 1 | The kill-guard assumes "only the operator kill flips the row FAILED while we run", but `LocalExecutor._cascade_kill_lineage` is a 2nd routing-agnostic FAILED-writer that races a CLI child's successful `proc.wait()` → guard returns REASON_KILLED → **`final_text` silently lost** | **Action Required** | **Accept** | Real data-loss race (verified the cascade writer). Fix via **option (ii)**: guard fires only when `status==FAILED AND proc.returncode != 0`, so a returncode-0 completion always preserves its output; an operator kill (signal → negative rc) still maps to REASON_KILLED. Fix the two false-invariant comments. Add a regression test. (Option (i) — transcript marker — is racy: the executor can read before teardown writes the killed event.) |
| 2 | `operator_kill_session` is `async def` but runs the subtree teardown (incl. the 2s grace `time.sleep` per CLI child) synchronously → blocks the FastAPI event loop up to N×2s | Recommended | **Accept** | Wrap the teardown loop in `asyncio.to_thread` (mirror `reply_to_thread`, agents.py:761) so signalling/grace runs off the event loop. |
| 3 | `os.kill(pid,0)` liveness check is a best-effort *narrowing* of the PID-reuse TOCTOU, not a guarantee; the comment overstates it | Minor | **Accept** | Correct the comment (best-effort narrowing, not a guarantee). |
| 4 | `_on_timeout` correctly does NOT write FAILED (so a timeout isn't mis-tagged REASON_KILLED) — but the guard ordering depends on that invariant | Minor | **Accept** | Add a comment on `_on_timeout` (both executors) that it must not write a terminal status, to protect the guard. |
| 5 | Best-effort contract holds; non-CLI sessions skipped; genuine failures still notify | — | **Confirmed good** | No action. |

**Findings forwarded to addresser: 4** (1 AR + 3 Rec/Minor).

nbramia:
## Review Round 1 — Referee Decisions (supplement: standards/security + consolidation)

Two more specialist passes landed. Net forwarded to addresser: **5 items** (2 AR + 3 Rec/Minor), folding in the earlier table.

| 6 | **Codex parity hunks have ZERO test coverage** — `start_new_session`, `codex_pid`, the REASON_KILLED guard, and dispatch suppression all shipped untested; the two `REASON_KILLED` constants are defined per-executor, so a codex regression ships green | **Action Required** | **Accept** | Mirror the claude_code tests into codex: the mid-run silent-guard, the pid-event + `start_new_session` capture, and a codex dispatch-suppression test. |
| 7 | SIGTERM grace loop polls only the group LEADER pid; if the leader exits but a group child lingers, it skips the SIGKILL escalation → children can survive | Minor | **Accept** | Best-effort SIGKILL sweep of the group after the grace so lingering children are reaped (no-op/ProcessLookupError if already gone). |
| — | `_kill_local_subprocess` reads the whole transcript then `reversed()` (vs `iter_events`) | Minor | **Reject** | Kills are rare; negligible. No change. |

Security/trust boundary confirmed sound (pid/pgid are worker-written, not attacker-supplied; same-user killpg correct). Combined with the earlier 4 (cascade-race data-loss [AR], event-loop blocking, liveness comment, `_on_timeout` comment), the addresser plan is: **#1 (returncode!=0 guard + comments + test), #6 (codex tests), #2 (asyncio.to_thread), #7 (SIGKILL sweep), #3/#4 (comments)**.

nbramia:
## Review Feedback Addressed — Round 1

Each finding was independently verified against the code before applying. All applied.

| 1 | Cascade-race data loss: the REASON_KILLED guard assumed the operator kill is the only mid-run FAILED-writer, but `LocalExecutor._cascade_kill_lineage` is a second, routing-agnostic one — racing a CLI child's clean `proc.wait()` and silently dropping its `final_text` | **Applied** | Verified `_cascade_kill_lineage` (local_executor.py:609-616) flips non-terminal descendants (CLI children included) to FAILED. Gated the guard on `status == FAILED AND proc.returncode != 0` in **both** executors: an operator kill signals the subprocess (non-zero/negative rc → still REASON_KILLED); a clean completion (rc 0) falls through to the COMPLETED path and persists `final_text`. Rewrote both false-invariant comments. Regression tests added (claude_code + codex): rc-0 race → COMPLETED, `*_completed` event with `final_text`, NOT REASON_KILLED. |
| 2 | `operator_kill_session` ran the subtree teardown (per-CLI-child SIGTERM→2s grace→SIGKILL) synchronously on the FastAPI event loop, blocking all HTTP for up to N×grace s | **Applied** | Wrapped the subtree-teardown loop in `await asyncio.to_thread(...)` (mirrors `spawn_agent`). Behavior, transcript events, cascade, and response shape unchanged. |
| 4 | The kill-guard ordering depends on a timed-out session still being RUNNING when the guard checks — `_on_timeout` must not write a terminal status | **Applied** | Added a comment on `_on_timeout` in both executors documenting it must NOT write a terminal status (it only sets the flag + terminates the OS process; `execute()` owns the row transition). |
| 5a | `os.kill(pid,0)` liveness comment overstated — it's a best-effort *narrowing* of the pid-reuse TOCTOU, not a guarantee | **Applied** | Reworded to state it narrows (not eliminates) the TOCTOU, and why killpg-on-pgid makes that acceptable. |
| 6 | Codex parity (`start_new_session`, `codex_pid` event, REASON_KILLED guard, dispatch suppression) shipped with ZERO test coverage; per-executor REASON_KILLED constants mean a codex regression ships green | **Applied** | Mirrored the claude_code coverage into the codex test files: spawn records `codex_pid` + passes `start_new_session=True`; the mid-run silent guard (FAILED + non-zero rc → REASON_KILLED, plus the #1 rc-0 case); dispatch suppresses the operator notice + web mirror on REASON_KILLED. |
| 7 | The SIGTERM grace loop polled only the group LEADER pid; if the leader exits but a group child lingers it `break`'d and skipped SIGKILL → children could survive | **Applied** | After the grace loop, always run a best-effort `os.killpg(pgid, SIGKILL)` sweep (ProcessLookupError = group already fully gone = no-op success path; otherwise survivors are reaped). The `local_subprocess_killed` audit signal stays accurate. New regression for the leader-exits-but-child-lingers case; existing clean-exit test updated to model a fully-gone group (sweep no-op → only SIGTERM recorded). |

**Tests (nathan-linux, isolated worktree at the pushed commit):**
- Targeted (claude_code executor + dispatch, inter_agent, kill-api, codex executor + dispatch): **139 passed**
- Full unit suite `pytest -m unit -n auto --dist loadscope`: **2270 passed, 8 skipped** (+7 new tests over the prior 2263; known-flaky scheduler test passed this run)

**New commits:**
- `35cf278` — #1 cascade-race guard + #4 timeout-invariant comment + executor regression tests
- `5862a7d` — #6 codex dispatch notice-suppression tests
- `25485d7` — #2 offload teardown off the event loop
- `73fb617` — #7 SIGKILL sweep + #5a liveness comment + inter_agent tests

nbramia:
## Review Round 1 — addressed, review loop complete

All 5 referee-forwarded items applied + verified (2 AR + 3 Rec/Minor): the **cascade-race data-loss** fix (guard now `status==FAILED and proc.returncode != 0` in both executors → a clean completion preserves its final_text; operator kill still → REASON_KILLED), **codex parity tests**, the `asyncio.to_thread` **offload** of the blocking teardown, the **SIGKILL group sweep**, and the comment corrections. Self-verified each hunk traces to its finding. Full suite **2270 passed** (+7 new), targeted 139. Merging.

The real cross-process signal (mocked in tests) will be **dogfooded on the live worker** post-merge.

PR #428, opened 2026-07-08
feat(agent-worker): parent can reopen a completed CLI child to answer its clarification

nbramia:
## Summary

Follow-up to #356/#422. A spawned CLI child that emits `[CLARIFY]` folds the question into its output as `[needs clarification] …` and completes — correct for the strand bug, but the parent's only path to get the question answered was re-spawning a fresh child, throwing away all the context the first child built up. That context was never actually lost: the child's CLI session persists on disk under `claude_code_session_id`, and the worker already resumes completed CLI sessions via `-r` for operator follow-ups.

This PR exposes that path to parents (**reopen-on-send**): `lifeos_agent_send` to your own direct COMPLETED `claude_code`/`codex` child enqueues the answer as the child's next turn and flips it back to CLAIMED, which the existing spawned-session dispatcher resumes with full prior context. The parent then just `yield_until`s the child again. All other terminal sends still reject: FAILED/BUDGET_EXCEEDED targets, non-direct-parent callers, `local`/`claude` routings, and children with no persisted CLI session id (`-r` needs the UUID; reopening would strand them CLAIMED).

Also aligns the prompts with reality:
- The CLI system prompt's CLARIFICATION section is now per-audience. Children were still being promised "your session will pause and the user's answer will be relayed" — false since #422. The operator variant is byte-identical to before.
- The delegation preamble, local-worker inter-agent block, and `lifeos_agent_send` tool description now tell parents what `[needs clarification]` means and how to answer it (send first, then yield — a yielded session can't send).

Relates to #356.

## Test evidence

On nathan-linux (worktree of this branch):
- `pytest tests/test_agent_worker_inter_agent.py tests/test_agent_worker_delegation.py tests/test_agent_worker_claude_code_executor.py -q` — **97 passed** (12 new: reopen happy path for both CLI routings, enqueue-before-flip ordering, five rejection cases, per-audience prompt wording, delegation guidance + updated pinned block text).
- Full `pytest -m unit` on the final branch head (cfc7e28): **2286 passed, 0 failed** (8 skipped, 1079 deselected) on nathan-linux.
- Post-fix targeted sweep (review round 1): 174 passed across the six agent-worker test files.

## Review focus

- The reopen gate in `inter_agent.send` — is the eligibility condition (COMPLETED + direct parent + CLI routing + persisted CLI session id) airtight against stranding a child in CLAIMED with no resume path?
- Ordering: message is enqueued before the status flip so a dispatch tick can't claim the child with an empty resume prompt.
- Prompt-text changes are behavior-adjacent: operator CLARIFY wording must remain byte-identical (test-pinned), and the INTER_AGENT_BLOCK pin in test_agent_worker_delegation.py was intentionally updated.

### DIFF
--- a/api/services/agent_worker/claude_code_executor.py
+++ b/api/services/agent_worker/claude_code_executor.py
@@ -169,9 +169,7 @@ the surrounding system. If you discover the task is bigger than expected
 the minimal safe change.
 
 CLARIFICATION:
-- Use [CLARIFY] to ask a question. Your session will pause and the user's
-  answer will be relayed back to you. After sending [CLARIFY], STOP and do
-  not continue working.
+{clarification}
 
 PERSISTENCE:
 - If your first approach doesn't work, try alternatives before giving up.
@@ -198,6 +196,23 @@ NOTIFICATIONS — use [NOTIFY] for:
 DELEGATION:
 You already have a browser (--chrome), filesystem, and shell. {delegation}
 """
+
+
+# Operator sessions really do pause on [CLARIFY] — the worker goes BLOCKED and
+# relays the Telegram reply. A spawned child does NOT (#356): its question is
+# folded into its output as "[needs clarification] …" and the turn completes;
+# the parent may answer via lifeos_agent_send, which resumes the child's CLI
+# session. Each variant states only what actually happens to that session.
+_CLARIFY_OPERATOR = """\
+- Use [CLARIFY] to ask a question. Your session will pause and the user's
+  answer will be relayed back to you. After sending [CLARIFY], STOP and do
+  not continue working."""
+
+_CLARIFY_CHILD = """\
+- Use [CLARIFY] to ask a question. Your turn will end and the question is
+  delivered to the parent agent that spawned you; if it answers, you will be
+  resumed with the answer as your next turn. After sending [CLARIFY], STOP
+  and do not continue working."""
 
 
 _PLAN_PREFIX = """\
@@ -384,6 +399,7 @@ class ClaudeCodeExecutor:
         resume_session_id: Optional[str],
         session_id: str = "",
         model: str = "opus",
+        is_child: bool = False,
     ) -> list[str]:
         platform_desc = (
             "Linux server running Ubuntu"
@@ -404,6 +420,7 @@ class ClaudeCodeExecutor:
                 user_name=settings.user_name,
                 code_dir=settings.code_dir,
                 platform_desc=platform_desc,
+                clarification=_CLARIFY_CHILD if is_child else _CLARIFY_OPERATOR,
                 delegation=delegation_preamble(
                     session_id,
                     trigger="To run background work in parallel,",
@@ -454,6 +471,7 @@ class ClaudeCodeExecutor:
         cmd = self._build_command(
             prompt, resume_session_id, session_id=sid,
             model=session.claude_code_model or "opus",
+            is_child=bool(session.parent_session_id),
         )
 
         self.transcript_store.append(sid, "claude_code_spawn", {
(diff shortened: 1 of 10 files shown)


nbramia:

| 1 | Multi-message silent drop on resume: both CLI dispatchers drain all pending messages but resume with only `pending[0]` — reopen-on-send makes multi-enqueue likelier | Recommended | **Accept** | Verified: `worker.py:1564` (claude_code) and `worker.py:1808` (codex) both discard extra drained messages while `send` reports `delivered: true`. Fix: concatenate all drained messages into the resume prompt, with a test. |
| 2 | `docs/specs/technical/agent-worker.md` stale: send described as queue-only; clarification flow doesn't cover the #422 fold + reopen loop | Recommended | **Accept** | Repo standard: docs reflect current state. The reopen contract currently lives only in code comments and tool descriptions. |
| 3 | Stale `[needs clarification]` re-delivery: `_child_final_text` keeps the last *non-empty* completed final_text, so an empty second run re-carries the first run's question | Recommended | **Accept — downgrade to Minor** | Real but a narrow edge (child completes its second turn with entirely empty output). Cheap guard: let the latest completed event win even when its final_text is empty (only for event kinds that carry the field). |
| 4 | MCP tool definition change (send description) is Ask-first per AGENTS.md | Minor | **Reject** | The operator explicitly approved this plan — including the send tool description change — before implementation; schema untouched, disclosed in the PR body. |
| 5 | Worker restart between reopen flip and dispatch marks the child FAILED though it's safely resumable | Minor | **Reject** | Pre-existing `resume_pending()` semantics shared with the operator-followup reopen path; fails safe (FAILED + notice), never strands. Out of scope. |
| 6 | Clarify-centric guidance overstates for codex children (they never emit `[needs clarification]`) | Minor | **Reject** | The guidance is conditional ("If a child's output contains…") and accurate; codex children simply never trigger it. Reopen remains useful for codex as a general follow-up path, as the tool description states. |

**Findings forwarded to addresser:** 3

Also surfaced during verification (pre-existing, separate issue to be filed): the codex completion send (`worker.py:1841`) is not child-gated, so a codex *child*'s completion still streams to the operator and registers a followup anchor — the same inconsistency #349 fixed for claude_code.

nbramia:
## Review Feedback Addressed (Round 1)

| 1 | Multi-message silent drop on CLI resume | Applied | Both dispatchers' resume branches (`_dispatch_claude_code_session`, `_dispatch_codex_session`) now concatenate ALL drained pending messages in order (`"\n\n".join(...)`) as the resume prompt, instead of resuming with `pending[0]` and discarding the rest. Fresh-spawn branches untouched (their first message is the structured spawn payload). Tests: `test_resume_delivers_all_pending_messages` (claude_code) and `test_codex_resume_delivers_all_pending_messages` assert a two-message resume delivers both, in order, via `resume()` not `execute()`. |
| 2 | `docs/specs/technical/agent-worker.md` stale on send/reopen + child clarify fold | Applied | `lifeos_agent_send` table row now documents the lifecycle transition (reopen of a COMPLETED CLI child, enqueue-before-flip ordering, `-r` resume with full prior context, other terminal sends still rejecting). New **Child clarifications (#422 / #428)** subsection under the clarification flow describes the fold (`[needs clarification] …` → child completes) and the reopen answer loop (parent sends answer → child resumes → parent yields again), contrasting operator sessions' unchanged pause-and-reply. Verified the product spec's clarification text is operator-flow only and still accurate — left alone. |
| 3 | Stale `[needs clarification]` re-delivery via `_child_final_text` | Applied | The latest completed event's `final_text` now wins even when empty, so a reopened child's empty second run can't re-carry the first run's question into the parent's resume turn. Verified all three scanned event kinds (`completed`, `managed_completed`, `claude_code_completed`) always persist a `final_text` key today; a key-presence guard keeps legacy pre-#349 events (`final_chars` only) from clobbering a real value. Tests: `test_child_final_text_latest_completed_event_wins_even_when_empty`, `test_child_final_text_legacy_event_without_key_does_not_clobber`. |

**Tests:** targeted suite on nathan-linux (`test_agent_worker_inter_agent / _delegation / _claude_code_executor / _claude_code_dispatch / _codex_dispatch / _loop`) — **174 passed**.
**New commits:** `f2f1a62` (code fixes 1+3), `cfc7e28` (docs fix 2).

PR #430, opened 2026-07-08
fix(agent-worker): codex child completions stay silent to the operator

### LINKED ISSUE #429: fix(agent-worker): codex child completions still stream to the operator (missing #349 parity)
## Problem

#349 made spawned `claude_code` children silent to the operator: their completion send is gated on `not session.parent_session_id` (`worker.py` claude_code completion branch), and their output reaches the parent via `_child_final_text`. The codex dispatch path never got the same gate: in `_dispatch_codex_session`'s COMPLETED branch, the Telegram completion send + `create_pending_question` followup registration run for ALL codex sessions, including spawned children — only the web/voice conversation mirror below it is child-gated.

So a codex child spawned via `lifeos_agent_spawn(model="codex")`:
- streams its final output to the operator's Telegram even though the parent will relay it in its own completion message (duplicate/confusing), and
- registers an operator-replyable followup anchor against a child session — so an operator threaded reply and a parent reopen-on-send (#428) can both enqueue messages against the same child.

## Context

Surfaced during the #428 review (round 1, finding 1 verification). Pre-existing — not introduced by #428, which mitigated the message-drop side by concatenating all drained pending messages into the resume prompt.

## Files involved

- `api/services/agent_worker/worker.py` — `_dispatch_codex_session` COMPLETED branch (the `_telegram_send_with_id` + `create_pending_question` calls, ~line 1841): gate on `not session.parent_session_id`, mirroring the claude_code branch (~line 1671).

## Acceptance criteria

- [ ] A spawned codex child's completion does not send to the operator's Telegram and does not register a `pending_question` followup anchor.
- [ ] Operator-origin `/codex` sessions are unchanged.
- [ ] The child's final output still reaches its parent (transcript `codex_handled_completion` / `_child_final_text` path).

nbramia:
## Summary

A spawned codex child's completion still streamed to the operator's Telegram and registered an operator-replyable followup anchor — the codex dispatch path never got the #349 child gate the claude_code path has. This gates the completion send + `create_pending_question` on `not session.parent_session_id`, and — because the operator leak was previously the *only* place a codex child's output went — gives the output a real path to its parent: `codex_completed` now persists `final_text` (parity with `claude_code_completed`, #349) and `_child_final_text` scans it.

Closes #429

## Test evidence

Pre-commit on nathan-linux (rsynced tree):
- Full `pytest -m unit`: **2290 passed, 0 failed** (8 skipped, 1079 deselected) on the final tree including review-round-1 fixes; the pre-fix tree ran 2289 passed, 0 failed.
- Targeted (codex dispatch/executor, claude_code dispatch, inter-agent, worker loop): 149 passed, including 3 new tests — child completion silent to operator + no anchor; `codex_completed` carries `final_text`; `_child_final_text` reads `codex_completed`.

## Review focus

- The gate moves the `#311` web-mirror call inside the `body and not parent` condition (the inner child gate was redundant once the outer one exists) — confirm operator-origin web-spawned codex sessions still mirror.
- `_child_final_text` scan-set addition interacts with the #428 latest-event-wins + key-presence semantics — legacy `codex_completed` events (no `final_text` key) must not clobber.
- The codex FAILED/BUDGET notice path intentionally remains un-gated (failure notices go to the operator on both CLI paths today — out of scope here).


nbramia (commented):
## Adversarial review — round 1

Independent verification performed on the checked-out branch (`c6c71c2`); the 3 new tests were run by name on nathan-linux against the rsynced copy of this exact tree (3 passed), plus the three touched test files (56 passed).

### Verified correct

- **Gate semantics**: the new `if body and not session.parent_session_id:` (worker.py:1859) matches the claude_code branch (worker.py:1678) exactly in effect. For a non-child, behavior is byte-identical to before — including the exception path: a failed `_telegram_send_with_id` is caught and `_mirror_to_conversation` still runs, same as pre-PR.
- **#311 mirror move**: moving the mirror inside the gate changes nothing for operator-origin web-spawned codex sessions (children are never conversation-linked, so the inner gate it replaced was equivalent). Regression-covered by the existing `test_codex_completion_mirrors_into_linked_conversation`.
- **Child output path is complete end-to-end**: codex_executor.py:400 persists `STATUS_COMPLETED` before appending `codex_completed`, `_resume_yielded_for_children` (worker.py:594) is engine-agnostic on terminal child status, and `_child_final_text` (worker.py:2535-2538) now reads `codex_completed.final_text`. So the child's output genuinely reaches the parent — it doesn't just stop leaking to the operator.
- **#428 interaction**: `resume()` routes through the same `_run` as `execute()` (codex_executor.py:195/214), so a reopened codex child's second run re-persists `final_text` and the latest-event-wins scan can't re-carry a stale first-run clarification. The key-presence guard correctly skips legacy `codex_completed` events (final_chars only).
- **Comment accuracy**: "All four kinds persist a final_text key today" verified — local_executor.py:625, managed_executor.py:580, claude_code_executor.py:620, codex_executor.py:410.
- **Parity claim on the ungated failure path**: true — the claude_code FAILED/BUDGET branch also sends the notice un-gated for children (worker.py:1719-1720), so leaving codex the same is genuine parity, not a new gap.
- **New tests are non-vacuous**: the positive-control test proves message id 777 is what a non-child registers, so the child test's `withid_sends == []` + absent-anchor assertions are meaningful.

### Action Required

None.

### Recommended

- **[Follow-up tracking]** A FAILED or BUDGET_EXCEEDED child (codex *and* claude_code) still sends its "⚠️ … failed/hit budget" notice to the operator un-gated (worker.py:1897-1898 and worker.py:1719-1720). The PR correctly scopes this out as pre-existing parity, but the same duplicate-noise rationale from #429 applies (the parent's resume summary already carries the child's `[failed]` status header). File a follow-up issue covering both paths so it doesn't get lost.

### Minor

- **[Consistency]** `codex_completed` persists the raw, unstripped `state.final_text` (codex_executor.py:410) while the operator-facing `body` is stripped (worker.py:1858). A whitespace-only final_text is truthy in `_children_summary` (worker.py:2502), yielding an "output:" block of blank lines in the parent's resume turn. Marginal; matches the other executors' behavior class.
- **[Tests]** No codex-specific test for the legacy key-presence guard (a `codex_completed` event *without* a `final_text` key must not clobber a real value). The guard is kind-agnostic and exercised by the #428 claude_code tests, so this is optional hardening.

### Summary

Merge-ready. The gate is an exact semantic mirror of the claude_code branch, the child→parent output path was verified complete end-to-end (executor status persistence → engine-agnostic parent wake → transcript scan), and the mirror restructure is behavior-preserving with existing regression coverage. Findings are a follow-up-issue nit and two minor polish items — nothing blocking.

nbramia:

| 1 | FAILED/BUDGET child notices still reach the operator un-gated on both CLI paths | Recommended (file follow-up) | **Accept as process action** | Verified at worker.py:1712-1717 (claude_code) and :1892-1899 (codex). Out of scope for this parity fix — the parent's resume header already carries `[failed]`. Filing a follow-up issue covering both paths. |
| 2 | `codex_completed` persists unstripped `final_text`; whitespace-only text yields a blank "output:" block in the parent's resume turn | Minor | **Accept** | Normalize `state.final_text` once (strip) after the `-o` fallback so the transcript payload, `final_chars`, and the outcome all agree with the worker's stripped `body`. |
| 3 | No codex-specific legacy key-presence-guard test | Minor | **Accept** | Cheap mirror of the existing claude_code guard test; hardens the #429 scan-set addition directly. |

**Findings forwarded to addresser:** 2 (fast path — both Minor and mechanical, applied by the orchestrator directly; finding 1 handled as an issue, not code)

nbramia:
## Round 1 fixes applied (fast path — orchestrator-direct)

| 2 | Unstripped `final_text` persistence | Applied | `state.final_text` is normalized (stripped) once after the `-o` fallback in `CodexExecutor`, so the `codex_completed` payload, `final_chars`, and the outcome agree with the worker's stripped `body`; whitespace-only reads as empty. |
| 3 | Codex legacy key-guard test | Applied | `test_child_final_text_legacy_codex_event_without_key_does_not_clobber` mirrors the claude_code guard test. |
| 1 | Child failure/budget notices un-gated | Tracked | Filed #431 covering both CLI paths. |

**Tests:** targeted (codex dispatch/executor + claude_code dispatch) — 57 passed.

Issue #438, opened 2026-07-09
fix(sync): nightly credential syncs silently no-op — slack since Jul 1, gmail since Jun 26


nbramia:
## Problem

The nightly sync (lifeos-sync.timer, 03:30 ET) has been silently failing for credential-dependent sources while recording `success`:

- **slack**: last real run Jun 30 (450s, 9 interactions). Since Jul 1 it completes in **0.28s** with zero records. Newest indexed Slack message is 2026-06-30.
- **gmail_personal / gmail_work**: last real run Jun 25 (~1200s). Since Jun 26 they complete in 0.6–0.8s with `errors: 1`, still recorded `success`.
- **google_docs**: dropped from ~12s to 0.5s on Jul 1.

No alert fired because the skip paths exit 0 and `run_all_syncs.run_sync()` maps returncode 0 → SUCCESS.

## Diagnosis so far (2026-07-09 investigation)

The slack 0.28s signature exactly matches the `is_slack_enabled() == False` early-return in `scripts/sync_slack.py` (timed at 0.235s with SLACK_USER_TOKEN emptied). That requires `SLACK_USER_TOKEN` to be missing/empty in the child process env AND not restored by `load_dotenv` (which uses `override=False`, so a present-but-empty inherited var wins over the file).

**Everything works outside the 03:30 window.** Verified on nathan-linux:
- `scripts/sync_slack.py --execute` run interactively: syncs normally.
- Same under `systemd-run --user` with the production `EnvironmentFile=/home/nathanramia/Code/LifeOS/.env`: syncs normally. systemd parses the current .env correctly (token present, 98 vars).
- The API service (same EnvironmentFile) has the token in `/proc/<pid>/environ`.
- `.env` is a symlink to the Syncthing-synced `~/Code/Sync/envs/LifeOS/.env`; all three machine copies currently contain the token (Mac Mini's copy is stale — May 27, 4144 bytes vs 5572 — Syncthing on the Mini appears broken/paused, worth checking separately).
- No relevant code changed in the Jun 25–Jul 8 window (reflog gap).

So the poisoning happens only in the live 03:30 context — most plausibly the .env differs at that hour (something rewrites it overnight; it is written nightly ~22:26 by claude-sync/Syncthing) — or something else entirely that only observation will catch.

## Tracer installed — CHECK THIS FIRST

A self-removing diagnostic cron was installed on nathan-linux (100.68.0.120) on 2026-07-09:

[...]

PR #442, opened 2026-07-09
fix(sync): skipped credential syncs fail loudly; detect duration collapse

nbramia:
## Summary

Hardening half of #438: the nightly slack sync silently no-oped since Jul 1 (gmail since Jun 26) because skip/failure paths exited 0 and were recorded as SUCCESS. This makes every silent-failure mode loud: skipped credential syncs exit nonzero (→ FAILED + Telegram alert), account-level gmail/calendar failures exit nonzero, `.env` deterministically wins over inherited env (`load_dotenv(override=True)`), and a duration-collapse detector flags any source that historically takes >60s "succeeding" in <2s — the exact signature of this outage — as a backstop for unknown failure modes.

Relates to #438 (root-cause tracer runs tonight at 3:24 AM; this PR is the fix-regardless-of-root-cause half).

## Test evidence

19 new tests in `tests/test_sync_script_exit_codes.py`, `tests/test_sync_health.py`, `tests/test_run_all_syncs.py` — all pass on nathan-linux. Full unit suite (`-m "not browser and not requires_server and not integration and not slow"`, xdist): 3107 passed; the 14 failures are byte-identical to clean `origin/main` in the same isolated environment (data/.env-dependent tests, pre-existing — verified by diffing failure sets).

## Review focus

- Exit-code semantics: slack skip → exit 2 (even dry-run); gmail fatal (account-level `except`) → exit 1 after `SYNC_STATS` emission so partial stats still record. Per-message errors intentionally do NOT fail the run.
- `get_typical_duration_seconds` excludes sub-2s "successes" from the median so consecutive silent no-ops can't normalize themselves into the baseline.
- `load_dotenv(override=True)` tradeoff: inherited env can no longer intentionally override `.env` for these three sync entry points.

### DIFF
--- a/api/services/sync_health.py
+++ b/api/services/sync_health.py
@@ -587,19 +587,21 @@ def get_typical_duration_seconds(
     Returns None when there is no eligible history.
     """
     conn = get_sync_health_db()
-    rows = conn.execute(
-        """
-        SELECT duration_seconds FROM sync_runs
-        WHERE source = ?
-          AND status = ?
-          AND duration_seconds IS NOT NULL
-          AND duration_seconds >= ?
-        ORDER BY started_at DESC
-        LIMIT ?
-        """,
-        (source, SyncStatus.SUCCESS.value, min_duration_seconds, n),
-    ).fetchall()
-    conn.close()
+    try:
+        rows = conn.execute(
+            """
+            SELECT duration_seconds FROM sync_runs
+            WHERE source = ?
+              AND status = ?
+              AND duration_seconds IS NOT NULL
+              AND duration_seconds >= ?
+            ORDER BY started_at DESC
+            LIMIT ?
+            """,
+            (source, SyncStatus.SUCCESS.value, min_duration_seconds, n),
+        ).fetchall()
+    finally:
+        conn.close()
 
     if not rows:
         return None
--- a/api/services/sync_health.py
+++ b/api/services/sync_health.py
@@ -6,6 +6,7 @@ health check APIs. Sources must sync at least daily or be flagged as stale.
 """
 import sqlite3
 import logging
+import statistics
 from dataclasses import dataclass
 from datetime import datetime, timezone, timedelta
 from enum import Enum
@@ -566,6 +567,43 @@ def record_sync_error(
     conn.commit()
     conn.close()
     logger.error(f"Recorded sync error for {source}: {error_message}")
+
+
+def get_typical_duration_seconds(
+    source: str,
+    n: int = 5,
+    min_duration_seconds: float = 2.0,
+) -> Optional[float]:
+    """Median duration of the last ``n`` eligible successful runs for ``source``.
+
+    Used by run_all_syncs to detect duration collapse: a sync that historically
+    takes minutes suddenly completing in a fraction of a second is the
+    signature of a silent no-op (e.g. credentials missing from the child env —
+    issue #438). Runs shorter than ``min_duration_seconds`` are excluded from
+    the history because they are exactly the pathology being hunted; letting
+    them into the median would make consecutive silent no-ops look "typical"
+    after a few days.
+
+    Returns None when there is no eligible history.
+    """
+    conn = get_sync_health_db()
+    rows = conn.execute(
+        """
+        SELECT duration_seconds FROM sync_runs
+        WHERE source = ?
+          AND status = ?
+          AND duration_seconds IS NOT NULL
+          AND duration_seconds >= ?
+        ORDER BY started_at DESC
+        LIMIT ?
+        """,
+        (source, SyncStatus.SUCCESS.value, min_duration_seconds, n),
+    ).fetchall()
+    conn.close()
+
+    if not rows:
+        return None
+    return float(statistics.median(row["duration_seconds"] for row in rows))
 
 
 def get_sync_health(source: str) -> SyncHealth:
--- a/docs/specs/technical/observability.md
+++ b/docs/specs/technical/observability.md
@@ -2,7 +2,7 @@
 
 > **Status:** Complete
 > **Owner:** Platform
-> **Last Updated:** 2026-02-19
+> **Last Updated:** 2026-07-09
 
 Performance tracing and alerting for LifeOS.
 
@@ -126,6 +126,10 @@ While in maintenance mode, CRITICAL alerts from `service_health` are suppressed.
 When a service fails and a fallback is used, this is recorded as a "degradation event". These are collected and reported in the nightly health check if there are 5+ in 24 hours.
 
 Services are tracked on-use, not by polling. Status updates when a service is actually called.
+
+### Sync Duration-Collapse Detection
+
+`run_all_syncs` flags a successful sync run that completed suspiciously fast — the source typically takes >60s but finished in under max(2s, 5% of typical) — the signature of a silent no-op. Detection records a `duration_collapse` row in `sync_errors` and adds a "Suspiciously fast" section to the Telegram sync summary; the `sync_runs` row stays `success`, so the health dashboard still shows green for that run.
 
 ## Related Documents
 
--- a/scripts/run_all_syncs.py
+++ b/scripts/run_all_syncs.py
@@ -745,6 +745,9 @@ def get_disabled_work_sources() -> set[str]:
 # no-op (e.g. credentials missing from the child env). Exit-code hardening in
 # the sync scripts catches the known skip paths; this catches unknown ones.
 DURATION_COLLAPSE_MIN_TYPICAL_SECONDS = 60.0
+# Floor of the relative elapsed threshold: a run is "collapsed" when it
+# finishes under max(this floor, 5% of typical), shrinking the blind spot
+# for slow sources (e.g. a 450s-typical sync no-oping in 10s).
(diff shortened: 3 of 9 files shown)


nbramia (commented):
## PR Review: fix(sync): skipped credential syncs fail loudly; detect duration collapse (round 1)

Independently verified before findings: ran the three touched test files on nathan-linux against a byte-identical copy of this branch (md5-matched all 5 changed files) — 74/74 pass, including the 19 new tests. Confirmed the collapse check runs before `record_sync_complete` (current run is still `status=running`, so `get_typical_duration_seconds`'s `status='success'` filter excludes it), sub-2s "successes" are excluded from the median so consecutive no-ops can't self-normalize, `SYNC_STATS` is emitted before `sys.exit(1)` in the gmail script and `run_all_syncs.run_sync` parses stats from combined output *before* the returncode check (scripts/run_all_syncs.py:830-833) so partial stats are recorded on FAILED, `load_dotenv(override=True)` executes before any `api.*`/`config.*` import in both scripts (matters: `slack_integration.SLACK_USER_TOKEN` is captured at module import, api/services/slack_integration.py:34), no schema changes, stdlib-only additions.

### Action Required
- **[Requirements]** google_docs is still fully silent — the third documented casualty of #438 ("google_docs dropped from ~12s to 0.5s on Jul 1") is covered by neither half of this fix. `scripts/sync_google_docs.py:46-48` swallows every exception into `{"status": "error"}` and `__main__` (line 56) ignores the result → exit 0 → SUCCESS; and the collapse detector can't catch it because its typical ~12s is below `DURATION_COLLAPSE_MIN_TYPICAL_SECONDS = 60.0` (scripts/run_all_syncs.py:747). #438's acceptance criterion "a disabled/skipped credential sync produces a FAILED run + Telegram alert, not silent success" is unmet for this credential-dependent source, and the PR summary's "makes every silent-failure mode loud" overclaims. Either harden `sync_google_docs.py` the same way (error status → nonzero exit) in this PR, or explicitly scope it out with a follow-up issue before merge.

### Recommended
- **[Correctness]** Slack real-run errors still exit 0. `run_slack_sync` only logs `results["errors"]` (scripts/sync_slack.py:82-85) and `main()` only checks `status == "skipped"` (line 116). A run where the Slack API dies mid-sync (errors present, 0 interactions) is still recorded SUCCESS — the same class of silence #438 is about, one failure mode over. Consider failing (or at least flagging) when `errors` is non-empty and nothing was indexed.
- **[Correctness]** Detection dead zone between the two thresholds: a no-op taking 2–60s of "typical", or a broken run taking ≥2s, escapes the detector — and a ≥2s broken run *enters* the median history, ratcheting the baseline down within `n=5` runs. A relative test (e.g. `elapsed < max(2.0, 0.05 * typical)`) would shrink the blind spot; at minimum worth a comment on the constants (scripts/run_all_syncs.py:747-748) acknowledging the gap.

### Minor
- **[Correctness]** `get_typical_duration_seconds` leaks its connection if the query raises (api/services/sync_health.py:589-603 — no try/finally). Callers swallow the exception (`_detect_duration_collapse`), so the only cost is an unclosed WAL handle. Consistent with the file's existing style, but `contextlib.closing` would be cheap.
- **[Correctness]** Behavior change for the job-queue manual path: `_handle_sync_source` runs `scripts/sync_slack.py` with no args (api/services/job_queue.py:358-366), so a disabled integration now raises `RuntimeError` where it previously "succeeded". That loudness is the point, but note the manual `sync_source` job for slack has always been an argless dry-run no-op — pre-existing oddity this PR makes visible.
- **[Standards]** A collapsed run is still recorded `SUCCESS` in `sync_runs` (only a `sync_errors` row + Telegram ⚠️). Deliberate per the PR, but the health dashboard will show green for a source that just alerted as a possible no-op; worth a sentence in docs/specs/technical/observability.md or data-and-sync.md when documenting the detector.
- **[Correctness]** `GmailService`/`CalendarService` are constructed *outside* the account-level try (scripts/sync_gmail_calendar_interactions.py:158, and the calendar equivalent), so constructor-time credential failures crash uncaught — still a nonzero exit (loud, fine) but skips `SYNC_STATS` emission and `conn.close()`. Fine to leave; noting the fatal_error path doesn't cover it.

### Summary
Core mechanism is correct and well-tested — I verified the exit-code plumbing, stats-on-failure, median hygiene, and dotenv ordering against the real code, and the tests pass on the server. Needs one change (or an explicit scope-out) before merge: google_docs, named in #438 as silently broken, is still silent after this PR.

nbramia:

| 1 | google_docs still silent (sync_google_docs.py swallows exceptions → exit 0; collapse detector blind below 60s typical) | Action Required | Accept | Verified: `except → {"status": "error"}` at line 46-48, `__main__` ignores the result. google_docs is named in #438 as broken since Jul 1; the acceptance criterion covers it. |
| 2 | Slack real-run errors still exit 0 (`main()` only checks `status == "skipped"`) | Recommended | Accept (upgraded) | Verified worse than claimed: `full_sync`/`incremental_sync` catch exceptions from BOTH steps → a dead token at API-call time = `status: "partial"`, zero work, exit 0, recorded SUCCESS. Fix: exit nonzero when errors present AND zero work done; tolerate partial errors with real work (consistent with gmail per-message tolerance). |
| 3 | Detection dead zone (2–60s escapes; ≥2s no-ops poison the median) | Recommended | Accept detection half; accept residual risk on median | Make elapsed threshold relative: `elapsed < max(2.0, 0.05 × typical)`, keep >60s typical gate. Median poisoning by ≥2s broken runs is accepted residual risk — the detector is a backstop; exit codes are the primary defense. |
| 4 | `get_typical_duration_seconds` leaks DB connection if query raises | Minor | Accept | try/finally. |
| 5 | job_queue runs sync_slack.py argless — disabled slack now fails the job | Minor | Reject (no change) | That failure is loud and correct — exactly the point of this PR. The argless-therefore-dry-run oddity is pre-existing and out of scope. |
| 6 | Collapsed runs stay SUCCESS in sync_runs — note in docs | Minor | Accept | One line in the observability spec. |
| 7 | Gmail/Calendar service constructors outside the account-level try | Minor | Reject (no change) | Constructor-time failure crashes with a traceback → nonzero exit → FAILED + stderr recorded. Loud path already; wrapping adds nothing. |

**Findings forwarded to addresser:** 5 (items 1, 2, 3, 4, 6)

nbramia:
## Review Feedback Addressed (round 1)

All five referee-forwarded findings verified against the code before applying — every premise held (google_docs `__main__` ignored the error dict; slack `full_sync`/`incremental_sync` map total failure to `status: "partial"`; fixed 2s collapse cutoff; no try/finally in `get_typical_duration_seconds`; no doc coverage of the detector).

| 1 | google_docs still silent | Applied | `scripts/sync_google_docs.py` `__main__` refactored into `main(argv=None)`; `status == "error"` → `logger.error` + `sys.exit(1)`. 3 new tests in `tests/test_sync_script_exit_codes.py` (`TestGoogleDocsSyncExitCodes`): error → SystemExit 1, healthy execute → clean, dry-run → clean; patched at `api.services.gdoc_sync.sync_gdocs` since the import is function-local. Note: per-doc failures (`sync_all` catching individual doc errors) still return success — that path reports `failed` counts and was out of scope per the finding. |
| 2 | slack total failure exits 0 | Applied | Verified: both sync steps' exceptions land in `results["errors"]` with status "partial", so a dead token at API-call time = zero work + exit 0. `scripts/sync_slack.py` `main()` now exits 1 when `errors` non-empty AND `channels_synced`/`messages_indexed` both falsy; partial errors with real work stay success (consistent with gmail per-message tolerance). No change to `slack_sync.py` status semantics. 2 new tests. |
| 3 | Fixed 2s collapse threshold dead zone | Applied | `_detect_duration_collapse` now uses `elapsed < max(2.0, 0.05 * typical)`; the `typical > 60s` gate is unchanged. Constant comment updated. 2 new tests: typical=450/elapsed=10 → flagged (10 < 22.5); typical=450/elapsed=30 → not flagged. All pre-existing collapse tests still pass. |
| 4 | DB connection leak in `get_typical_duration_seconds` | Applied | Query wrapped in try/finally so the connection closes if the query raises. |
| 5 | Docs: collapsed runs stay SUCCESS | Applied | Added a "Sync Duration-Collapse Detection" subsection to `docs/specs/technical/observability.md` (under Alerting): >60s typical + elapsed under max(2s, 5% of typical) → `duration_collapse` row in `sync_errors` + "Suspiciously fast" Telegram section; `sync_runs` row stays `success` so the dashboard shows green. Frontmatter `Last Updated` bumped. |

**Tests:** full unit suite on nathan-linux (isolated copy, `-m "not browser and not requires_server and not integration and not slow"`, xdist) — 3113 passed, 35 skipped, 14 failed; the 14 failures are exactly the documented pre-existing data/.env-dependent set (identical on clean main). The three touched test files: 81/81 pass (74 prior + 7 new).

**New commits:**
- `9b641df` fix: address review round 1 — google_docs and slack total failures exit nonzero
- `2b31d5b` fix: address review round 1 — duration-collapse elapsed threshold scales with typical
- `94eb277` fix: address review round 1 — close sync_health conn on query error; document collapse detector

PR #451, opened 2026-07-09
fix(gpu-watchdog): actively reclaim VRAM on sustained saturation

nbramia:
## Why

The nightly-sync "crash" on **2026-07-09** was a GPU/network wedge, not an app bug. Reconstructed from the kernel log of the prior boot:

- **03:04** WiFi drops (`4WAY_HANDSHAKE_TIMEOUT`, the known MT7925 SPOF) → **03:10** the network `page_pool` wedges permanently (`18 inflight`, never clears → dead network for 6h).
- **03:11–04:02** GPU **SDMA queue exhaustion**: `amdgpu: No more SDMA queue to allocate (8 total queues)` / `DQM create queue ... ret -12`, from short-lived processes that grabbed GPU queues while network-blocked and never released them.
- **04:04** VRAM saturates: `VRAM SATURATED: 97% (62.4/64.0 GB)`.
- The `gpu-watchdog` **logged the saturation and did nothing else** — it only alerts. The box stayed wedged until a manual reboot at 09:56.

The watchdog's own header documents this exact failure (issue #199 follow-up: "the embedding model load pushed past the ceiling and the kernel locked up ... until the operator force-shut the machine"). It was built to **warn**, never to **act**. The sync's existing LLM-yield guard (`run_all_syncs.py`) couldn't help: the embedding phases were dependency-skipped last night, and it can't reclaim from *hung* processes anyway.

## What

Turn the watchdog into an auto-healing safety valve:

- When VRAM stays **≥ `RECLAIM_PCT` (default 92%)** for **`RECLAIM_STRIKES` (default 2)** consecutive 5-min ticks, **stop `lifeos-llm.service`** to force-release its ~35 GB VRAM + GPU queues.
- Once VRAM drains **< `RESTART_PCT` (default 60%)**, **restart it automatically** — no operator action.
- Runs *before* the alert cooldown, so a recent ⚠️ warning can't suppress a reclaim.
- Uses the same passwordless `sudo -n systemctl` allowlist as `run_all_syncs.py`'s embedding path. Where that allowlist is absent (fresh clones), it **degrades to today's alert-only behaviour** — no new hard dependency.
- Disable entirely with `LIFEOS_VRAM_RECLAIM_PCT=101`.

This won't stop the *upstream* WiFi wedge (that's the wired-ethernet / driver fix, tracked separately) — but it breaks the terminal VRAM-lockup spiral regardless of what triggers it.

## Tests

`tests/test_gpu_watchdog.py` — 10 cases driving the strike → stop → restore state machine through fake sysfs VRAM + `systemctl`/`sudo`/`curl` shims: single-tick arms but doesn't act, two ticks reclaim, no double-stop while marked, restore on drain, marker held above threshold, strike reset below ceiling, `sudo`-denied degradation, reclaim-disabled, LLM-not-running skip, and a regression that the original alert still fires. All pass.

Full unit suite: **3214 passed**; the 5 failures are pre-existing and environmental (server `.env` sets `local_llm_model` to Gemma vs. the test's `gpt-oss` default; p91 data-integrity + slack tests read the live vault/DB) — untouched by this change (diff is 2 files).

## Testability hooks added
`LIFEOS_VRAM_STATE_DIR`, `LIFEOS_VRAM_CARD_DIR`, `LIFEOS_WATCHDOG_CURL` — override state dir / sysfs card / curl binary so tests run hermetically.


nbramia:
Rebased onto `main` (712c715) and reviewed before merge.

## Conflict resolution

Conflicted with #522 (merged today), which added SDMA-queue-exhaustion detection to this same file. The two are complementary, not duplicative — #522 *detects* a failure mode VRAM% can't see and alerts; this PR *acts* on sustained VRAM saturation. Main had no reclaim behaviour of any kind before this.

The conflict was confined to the state-file declaration block. Resolved by keeping this PR's `STATE_DIR` indirection and routing #522's two SDMA stamps through it as well. `STATE_DIR` defaults to `$PROJECT_DIR/logs`, so production paths are byte-identical to before; the gain is that SDMA state is now hermetically testable too.

## Verified

- **Safe verbs.** Uses `sudo -n systemctl stop` then `start`, never `restart`. That matters: `setup-systemd.sh` deliberately allowlists only `start stop reset-failed` for `lifeos-llm` ("restart intentionally omitted because of GPU memory cleanup concerns"), so a `restart` would have silently failed on a password prompt.
- **No crash-loop reinforcement.** Restart is gated on VRAM draining below `RESTART_PCT` (60%), so it won't restart llama-server into a still-full GPU — the known self-reinforcing failure.
- **Degrades safely.** `sudo -n` denied → logs and no-ops, so a fresh clone without the allowlist keeps today's alert-only behaviour.
- **SDMA path still works after the reroute.** #522 shipped no tests for it, so I smoke-tested the merged script directly with a stubbed `journalctl`, fake sysfs and a scratch state dir: `SDMA queue exhaustion detected: 1 occurrence(s)` logged, stamp written to the scratch dir, nothing leaked into repo `logs/`.
- `bash -n` clean; `tests/test_gpu_watchdog.py` 10 passed.

## One defect found and fixed

The test module carried **no pytest marker**, so all 10 tests were deselected by the `unit and not slow` scope the pre-push hook runs — the reclaim state machine would have shipped with tests that never ran automatically. Added `pytestmark = pytest.mark.unit`. Pre-push scope: **2401 → 2411**.

## Residual risk (accepted, not eliminated)

This is not a passive change: at ≥92% VRAM for two consecutive 5-min ticks it **stops `lifeos-llm`**, interrupting any in-flight local-model work. Baseline VRAM here is ~35.5/64 GB (~55%), so tripping 92% requires a genuinely abnormal state. Escape hatch is `LIFEOS_VRAM_RECLAIM_PCT=101`.

Full unit scope after the merge: **2411 passed, 8 skipped**.

PR #469, opened 2026-07-10
feat(chat): make search_web a real web search (DuckDuckGo + Claude native)

### LINKED ISSUE #467: feat(chat): make search_web a real web search (Claude native tool)
## Problem

The `search_web` tool is a **misnomer** — it doesn't search the web. `api/services/web_search.py` just calls the LLM (`get_anthropic_llm().create(...)`) with **no web access, no search API, no browsing** — it asks Claude to *"answer this query using your knowledge."* Its own docstring admits it can't get real-time data.

So for anything current — prices, weather, news, latest versions — it structurally cannot get real data, even though the system prompt (`api/services/agent_system_prompt.py:92`) forcefully tells every persona to *"ALWAYS call search_web first"* and to *"never claim you can't access live data."* The agent obeys, calls `search_web`, gets "I don't have live data," and reports a "technical error." Discovered when the Financial Advisor couldn't answer "which position moved most today."

## Approach (settled with Nathan): backend-agnostic, hybrid

The orchestrator can route to **Anthropic, Codex (OpenAI), or the local llama-server** (`LIFEOS_LLM_BACKEND` + the escalation ladder). Web search must work on **all three** — so the search itself is done in LifeOS's own code (real results fetched), then synthesized by whichever LLM is active (all three already synthesize text). Two-tier:

1. **Non-Anthropic backends (Codex / local), and the universal fallback:** **DuckDuckGo, no API key** (via the `ddgs` package) — fetch real current results in LifeOS, format them (the existing `format_web_results_for_context`), and synthesize with the active LLM. Works identically on Codex and local. New dependency `ddgs` (approved, no key).
2. **Anthropic backend (quality upgrade):** use Claude's **native web-search server tool** (iterative, cited) — `web_search_20260209` (name `web_search`) on Opus 4.8/4.7/4.6 + Sonnet 5/4.6, basic `web_search_20250305` on Haiku 4.5 (LifeOS's default). No beta header; with `_20260209` do NOT also declare `code_execution`. Server-tool errors arrive as an HTTP-200 `web_search_tool_result` block (not an exception); handle `stop_reason: "pause_turn"` (re-send).

Keep the `search_web` / `search_web_with_synthesis` signatures stable so `_tool_search_web` (`agent_tools.py:934`) and other callers are unchanged. Degrade honestly on failure (DDG rate-limit / no results → say so, never fabricate). Reconcile the system prompt — web search is now genuinely real on every backend, so the "always search / never say you can't" language is finally honest (no per-backend caveat needed).

**Cost:** DuckDuckGo is free; Claude's native web-search is billed per search on the Anthropic path.

## Files involved

- `requirements.txt` (add `ddgs`)
- `api/services/web_search.py` (stub → hybrid real search: DDG fetch + Anthropic-native path)
- `api/services/llm_client.py` (pass the native web-search server tool + surface `web_search_tool_result` on the Anthropic backend; expose the active backend)
- `api/services/agent_system_prompt.py` (the "always search_web" language is now honest on every backend)
- Tests: DDG path returns real-shaped results (mock `ddgs`); Anthropic-native path declares the correct model-gated tool type (mock the client); honest degradation on failure; tool name/signature unchanged

## Acceptance criteria

- [ ] Real web results on **all** backends: Anthropic (native tool), Codex, and local (DuckDuckGo) — verifiable with a current-events / live-price query returning real, current info
- [ ] Honest degradation when a search genuinely returns nothing (no fabricated "live" data)
- [ ] The `search_web` tool name and `_tool_search_web` signature are unchanged (no orchestrator-contract break)
- [ ] Model-gated native tool-type selection (Haiku 4.5 → basic variant) on the Anthropic path
- [ ] Fixes the capability for every persona, not just the Advisor

nbramia:
## Summary

`search_web` didn't search the web — it called the LLM with no web access and asked it to answer *"from your knowledge,"* so it could never return live data (prices, weather, news), even though the system prompt tells every persona to *always* search the web first. (The old stub also routed through `get_anthropic_llm()`, pinned to a now-retired model — the "technical error" users saw was that client's 404.) Discovered when the Financial Advisor couldn't answer "which position moved most today."

Now real on **every backend**:
- **DuckDuckGo** (`ddgs`, no API key) fetches real current results *in LifeOS* — backend-independent, so it works identically on Anthropic, Codex, and the local llama-server. Offloaded via `asyncio.to_thread`.
- **Anthropic backend:** upgrades to Claude's **native server-side web search** (`web_search_20250305`) via the active orchestrator model (`get_local_llm()`, current + web-search-capable — sidestepping the retired specialist model) — iterative, cited — with DuckDuckGo as the fallback.
- **Degrades honestly:** an empty/failed search returns nothing to say, never a fabricated "from training data" answer.

`search_web` / `search_web_with_synthesis` signatures are unchanged, so `_tool_search_web` and the chat caller are untouched.

Closes #467

## Test evidence

`./scripts/test.sh auto`: **3242 passed** (failures are the known pre-existing env/data-dependent set only). 11 web-search tests: formatter, DuckDuckGo field-mapping + non-fatal failure, `search_web` thread offload, and `search_web_with_synthesis` across backends (native on Anthropic, DuckDuckGo on local, native→DuckDuckGo fallback, honest empty). **Live smoke tests:** DuckDuckGo returned real AMD quote results; the native path returned a real current-dated answer (both verified against live services).

## Review focus

- **Backend gating** (`_use_native_anthropic` on `settings.llm_backend`) — native only on Anthropic; DuckDuckGo everywhere else and as the fallback.
- **Honest degradation** — no path fabricates an answer when the search is empty/fails.
- **Native path model** — uses `get_local_llm()` (current orchestrator model), not the retired `get_anthropic_llm()` pin.

## Follow-up (separate)

`get_anthropic_llm()` is pinned to a **retired model** (`claude-sonnet-4-20250514` → 404), which also breaks relationship-insights / fact-extraction / tone-analysis specialist calls. Out of scope here (this PR routes around it); worth a separate fix — I'll file an issue.

### DIFF
--- a/api/services/llm_client.py
+++ b/api/services/llm_client.py
@@ -572,7 +572,10 @@ class AnthropicLLMClient:
         tool_calls = None
         for block in resp.content:
             if block.type == "text":
-                text = block.text
+                # Accumulate: a response may carry multiple text blocks (e.g. a
+                # native web-search answer split at citation boundaries). Keeping
+                # only the last would truncate the answer to its final fragment.
+                text += block.text
             elif block.type == "tool_use":
                 if tool_calls is None:
                     tool_calls = []
--- a/api/services/web_search.py
+++ b/api/services/web_search.py
@@ -1,65 +1,58 @@
 """
 Web search service for LifeOS.
 
-Uses the Anthropic API to answer web-style queries. The model doesn't
-have real-time web access, but can answer many factual questions from
-its training data. For truly real-time queries (weather, stock prices,
-live scores), it will indicate the limitation.
-"""
-import logging
-from typing import Optional
+Real web search on every LLM backend. The search itself runs in LifeOS via
+DuckDuckGo (no API key), so it works identically whether the orchestrator is on
+Anthropic, Codex, or the local llama-server. On the Anthropic backend we upgrade
+to Claude's native server-side web search (iterative, cited) for higher quality.
 
-from api.services.llm_client import get_anthropic_llm
+Both entry points degrade honestly: a failed or empty search returns nothing to
+say (empty results / empty answer), never a fabricated "from my training data"
+response — that stub behavior is exactly what this module replaces.
+"""
+import asyncio
+import logging
+
+from config.settings import settings
 
 logger = logging.getLogger(__name__)
 
 
-async def search_web(query: str, max_results: int = 5) -> list[dict]:
-    """
-    Answer a web-style query using the local LLM.
+def _ddg_search(query: str, max_results: int = 5) -> list[dict]:
+    """Real web results via DuckDuckGo (no API key), synchronous.
 
-    Args:
-        query: The search query
-        max_results: Maximum number of results to return
-
-    Returns:
-        List of results, each with: title, url, snippet
+    Best-effort: returns [] on any failure (rate-limit, network, parse) so the
+    caller degrades to "nothing found" rather than inventing an answer.
     """
     try:
-        client = get_anthropic_llm()
-        response = client.create(
-            messages=[{
-                "role": "user",
-                "content": (
-                    f"Answer this query using your knowledge: {query}\n\n"
-                    "If this requires real-time data (live prices, current weather, "
-                    "today's news), say so clearly. Otherwise, provide the best "
-                    "factual answer you can."
-                ),
-            }],
-            system="You are a helpful assistant answering factual questions. Be concise and accurate.",
-            max_tokens=1024,
-        )
-        return [{
-            "title": "Knowledge Base Answer",
-            "url": "",
-            "snippet": response.text[:500],
-        }]
+        from ddgs import DDGS
+
+        with DDGS() as ddgs:
+            hits = ddgs.text(query, max_results=max_results)
     except Exception as e:
-        logger.error(f"Web search failed: {e}")
+        logger.warning(f"DuckDuckGo search failed: {e}")
         return []
+    out: list[dict] = []
+    for h in hits or []:
+        out.append({
+            "title": h.get("title") or "",
+            "url": h.get("href") or h.get("url") or "",
+            "snippet": h.get("body") or h.get("snippet") or "",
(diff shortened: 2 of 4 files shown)


nbramia:

| 1 | Native answer truncated to the last text block (`_parse_anthropic_response` overwrites) | Action Required | **Accept** | Confirmed — my own live smoke returned a `".\n\nNote:…"` leading fragment (post-citation tail). Cited answers span multiple text blocks. Fix: accumulate (`text += block.text`) — behavior-preserving for the single-block case, correct for multi-block. |
| 2 | Native path has no real coverage (fully mocked) | Recommended | **Accept** | Added a `_parse_anthropic_response` multi-text-block accumulation test. |
| 3 | `pause_turn` unhandled | Recommended | **Downgrade → Minor** | `max_uses=5` < the 10-iteration cap; empty→DuckDuckGo fallback covers it, partial is real (no fabrication). Documented, no resume logic. |
| 4 | New `ddgs` dependency (Ask-first) | Recommended | **Confirmed approved** | Nathan chose DuckDuckGo (no key) explicitly. No action. |
| 5 | `_fetch_web` dead code / consumer contract | Minor | **Note** | Pre-existing dead code; the live consumer `_tool_search_web` feeds tool output back to the orchestrator to synthesize → no UX regression. Won't touch. |

**Applied:** the accumulation fix (shared `_parse_anthropic_response`) + regression test + a `pause_turn` note.

nbramia:
## Review Round 1 — Complete (self-verified)

Addressal committed (`54dbb48`): the Action-Required truncation is fixed in the shared `_parse_anthropic_response` (accumulate text blocks) — **verified live** (native answer now returns in full, not the trailing fragment) and the full gate is green (3244 passed; the shared change caused no orchestrator regression). Added the multi-block accumulation regression test and a `pause_turn` note. All changes trace to accepted findings — **no Round 2**.

PR #477, opened 2026-07-10
fix(fitness): set durations + guard against phantom 'Logged' replies

nbramia:
## Problem

Comparing the fitness bot's Telegram transcripts against the SQLite store / Google Sheet mirror surfaced two defects:

**1. Silent data loss — phantom "Logged" replies.** Once a few `Logged …` confirmations accumulate in the conversation history, the orchestrator (Haiku) starts replying `Logged 6/20: Pec Fly Machine 3×12 @100 lb.` **without calling `manage_workouts` at all** (`tool_rounds: 0` in routing metadata). Nine exercises across 6/10 and 6/20 were confirmed to the user but never written. The sheet was never the problem — it faithfully mirrors a store that never got the data. Notably it is *not* exercise-name filtering: unknown names pass through title-cased by design.

**2. No structured time for timed work.** Stairs/cardio time lived in free-text notes (`"500 stairs in 7:01"`), so it can't render as a column or trend.

## Fix

- **Phantom-write guard** (`agent_loop.py`): if the final reply opens with `Logged/Updated/Recorded/Saved` and the turn made **zero** tool calls, retract it and nudge once ("you made no tool call — record it now or rephrase"). Mirrors the existing give-up self-correction; fires at most once per turn. Plus an explicit persona rule that every log needs a fresh tool call in the current turn.
- **`duration_seconds` on sets**: new nullable column on `workout_sets` (ALTER-on-open migration), `manage_workouts` sets schema field, summary/history rendering (`Stairs 500 in 7:01`), and a human-readable `time` column in the sheet mirror's Sets tab. Convention: count (steps/meters) goes in `reps`, time in `duration_seconds`.

## Testing

- 71 tests pass across the four touched test files (7 new phantom-write tests incl. nudge-fires-once and plain-answers-untouched; store migration + roundtrip; mirror time column; tool rendering).
- Full unit suite green via pre-push hook; the CRM/p91 data-integrity failures seen in a bare worktree are environmental (need live `data/`) and pass from the server checkout.

## Follow-up (same PR): `unit` column + Metrics tab

- `workout_sets.weight_unit` → **`unit`** (RENAME COLUMN migration): `lb`/`kg` only when a weight is present (default lb); counted work can carry its own unit (`steps`, `m`); weightless rows get `''` — the migration clears the old unconditional `lb` on them. The legacy `weight_unit` key is still accepted from the LLM.
- Summaries render counted work as `Stairs 500 steps in 7:01`.
- Sheet mirror: Sets header renamed to `unit`; new **Metrics** tab mirroring manually reported metrics (body weight etc.) — device imports (intraday Apple Health samples) stay out of the sheet.
- Note: unlike the additive duration column, the rename is **not** pre-applied to the live DB — it would break the currently-deployed writer. It runs automatically at first store use after deploy.

### DIFF
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -63,16 +63,21 @@ def _looks_like_giving_up(text: str) -> bool:
 # Phantom-write guard: a reply that opens with "Logged …" / "Updated …" claims a
 # state change. If the turn made ZERO tool calls, nothing was written — the model
 # pattern-matched earlier confirmation lines in the conversation history instead
-# of calling the tool (observed with the fitness bot on Haiku: replies like
-# "Logged 6/20: Pec Fly Machine 3×12" with tool_rounds=0 silently lost sets).
+# of calling the tool (observed with the fitness bot on a weak model: "Logged …"
+# replies with tool_rounds=0 silently lost sets). Scope: only the zero-tool-call
+# turn is caught; a turn that makes a READ call and then claims a write still
+# passes — that variant hasn't been observed and distinguishing reads from
+# writes here isn't worth the tool-registry coupling yet.
 _WRITE_CLAIM_PATTERN = re.compile(r"(?i)^\s*(logged|updated|recorded|saved)\b")
 
 PHANTOM_WRITE_NUDGE = (
     "Stop — you replied as if something was recorded, but you made NO tool call "
-    "this turn, so nothing was saved. If the user's message contains data to "
-    "record (a workout, a metric, a task…), call the appropriate tool NOW to "
-    "actually record it, then confirm. If nothing needed recording, rephrase "
-    "your answer without claiming anything was logged or updated."
+    "this turn, so nothing was saved. If the user's CURRENT message contains new "
+    "data to record (a workout, a metric, a task…), call the appropriate tool NOW "
+    "to actually record it, then confirm. If it was already recorded in a "
+    "PREVIOUS turn, do NOT record it again — just answer the question. If "
+    "nothing needed recording, rephrase your answer without claiming anything "
+    "was logged or updated."
 )
 
 
--- a/api/services/agent_loop.py
+++ b/api/services/agent_loop.py
@@ -60,6 +60,27 @@ def _looks_like_giving_up(text: str) -> bool:
     return bool(_GIVE_UP_PATTERNS.search(text))
 
 
+# Phantom-write guard: a reply that opens with "Logged …" / "Updated …" claims a
+# state change. If the turn made ZERO tool calls, nothing was written — the model
+# pattern-matched earlier confirmation lines in the conversation history instead
+# of calling the tool (observed with the fitness bot on Haiku: replies like
+# "Logged 6/20: Pec Fly Machine 3×12" with tool_rounds=0 silently lost sets).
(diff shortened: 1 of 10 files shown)


nbramia (commented):
## PR Review: fix(fitness): set durations + guard against phantom 'Logged' replies

Adversarial review by three specialists (correctness, security/privacy, standards). All 76 tests independently re-run and verified green; phantom-write tests confirmed to exercise the real `run_agent_loop` (only the LLM client and tool executor are mocked).

### Action Required
None.

### Recommended (applying before merge)
- **[Correctness]** `PHANTOM_WRITE_NUDGE` can induce a duplicate write: on "did you log that?" answered from history with zero tool calls, the nudge tells the model to "record it NOW" — a weak model may re-log sets sitting in history. Harden wording: if already recorded in a previous turn, don't log again. (`agent_loop.py`)
- **[Correctness]** Guard scope: a read-then-phantom turn (round 1 `history`, round 2 "Logged …") evades the zero-tool-calls check. Observed incident was tool_rounds=0 so ship as scoped, but document the boundary in the guard comment.
- **[Correctness]** Migration lb-clear isn't crash-idempotent: the `UPDATE … SET unit=''` is gated on the rename predicate; a crash between RENAME (DDL, autocommit) and commit strands weightless rows at `lb` forever. The UPDATE is idempotent — run it unconditionally. (`fitness_store.py:149-153`)
- **[Correctness]** Enforce the unit invariant on write: strip `lb`/`kg` when no weight is present, so a model that habitually sends `unit:'lb'` can't produce "Stairs 500 lb". (`_insert_sets`)
- **[Security/Privacy]** Test fixtures and the guard comment embed the real lost workout (real exercise, date, load). AGENTS.md requires obviously synthetic data in code/tests — keep the provenance explanation, swap in invented data. Also use the repo's established synthetic body weight (178.4) instead of the new one-off 212.0 in the mirror test.
- **[Standards]** `_fmt_duration` (agent_tools) and `_time_cell` (sheet mirror) are byte-identical — consolidate into one helper in `fitness_store.py`.

### Minor
- **[Correctness]** `_WRITE_CLAIM_PATTERN` applies to every persona; legit zero-tool replies starting "Updated version: …" get one retraction+retry. Bounded (once/turn, all three surfaces reset buffered text on self_correction) — acceptable.
- **[Standards]** 474-line diff is in pr-check's 400–600 WARN band (306 lines are tests). No linked issue. `feat` commit rides in a `fix/` PR — declared in the body.
- **[Correctness]** `_fmt_duration` H:MM:SS branch untested.
- **[Security]** Metrics tab expands what the opt-in sheet mirror exports (manually-reported health metrics) — gating unchanged and correct; noted in PR body.

### Verified clean
Parameterized SQL throughout (migrations are fixed literals); no new logging of message content; nudge path not reachable from user text (matches assistant reply only); no test weakened or deleted — both renamed store tests assert intentionally changed semantics; migration handles fresh/pre-duration/mid-state DBs; SQLite ≥3.25 requirement satisfied everywhere supported; no stale `weight_unit` consumers outside the migration shim; no doc/ADR staleness (column rename within ADR-013 scope).

### Summary
Merge-ready after the Recommended batch above — no blocking defects found.

PR #481, opened 2026-07-10
fix(scripts): close remote-test.sh secret-leak, injection, and first-run blockers

nbramia:
## Summary

Fixes the three Action Required findings on `scripts/remote-test.sh` from the #480 review, plus two related correctness items. Stacked onto `harness-audit-tier2` so merging it updates #480. Scoped to `remote-test.sh` only — the `claude-session-pane.sh` dead-code finding and the branch-name-convention finding on #480 are untouched and still need a decision.

Relates to #480.

## What changed

- **Secret / personal-data leak → world-readable `/tmp` (Action Required).** rsync ignores `.gitignore`, so the old build-artifact-only exclude list mirrored `.env`, `config/token-*.json`, `config/credentials-*.json` and ~15 GB of personal data under `data/` onto the server at 0644. Now the exclude list is generated from `git ls-files --others --ignored` — the sync carries only tracked + untracked-unignored files, the same boundary `test.sh auto`'s diff logic uses. `.git` is still synced (needed for the merge-base diff).
- **Branch-name command injection → RCE (Action Required).** A refname may legally contain `'`/`;`/`$`/backticks and was interpolated into the remote ssh command. Sanitized with `tr -c 'A-Za-z0-9._-'` **and** the remote command is now assembled with `printf %q` for the dir and every arg.
- **First-run failure (Action Required).** rsync only creates the final path component, so the first run — and every run after `/tmp` is cleared on reboot — failed rc 11. Added `ssh mkdir -p` before rsync.
- **DONE marker on interrupt (Recommended).** Added an `INT`/`TERM` trap so the `[remote-test] DONE` marker is always emitted, so the `until grep -q "[remote-test] DONE"` wait loop this PR adds to `implement/SKILL.md` can't hang forever on a killed run.
- **Same-branch collision (Recommended).** Remote dir is now keyed by branch **and** a hash of the checkout path, so two agents on one branch in different worktrees don't clobber each other; same-checkout re-runs stay incremental.

## Test evidence

Full end-to-end `ssh` transport wasn't runnable from the review environment (no agent SSH creds to `nathan-linux-ts`), so each fix was verified by isolating its mechanism:

- **Privacy:** rsynced the live checkout (which has the real secrets + 15 GB `data/`) through the new gitignore-derived exclude to a local dir — `.env`, all `config/token-*`/`credentials-*`/personal JSONs, and `data/` absent; `.git` + sources present (15 GB → 82 MB). Ran `./scripts/test.sh auto` in that pruned copy: **14 failed, 3243 passed, 35 skipped — byte-identical failing set to the pre-change run with `data/` present**, so no test depends on the excluded data.
- **Injection:** with a malicious branch (`x';touch …;'`) *and* a malicious arg together, the assembled `REMOTE_CMD` neutralizes both — the injected `touch` never executes.
- **First-run:** reproduced the rc-11 missing-parent failure, confirmed `mkdir -p` first → rc 0.
- `bash -n` clean.

## Review focus

- The `git ls-files --others --ignored --exclude-standard --directory` exclude generation + `/`-anchoring, and the fail-closed guard (refuses to sync if the git enumeration fails, rather than syncing everything).
- `printf %q` remote-command assembly portability (targets stock macOS bash 3.2).

### DIFF
--- a/scripts/remote-test.sh
+++ b/scripts/remote-test.sh
@@ -6,7 +6,7 @@
 #
 # Runs the test suite on nathan-linux for checkouts that have no local venv
 # (the MacBook). Rsyncs the CURRENT WORKING TREE — uncommitted and untracked
-# changes included — to a branch-keyed isolated dir on the server, then runs
+# changes included — to an isolated dir on the server, then runs
 # ./scripts/test.sh there. No commit or push is required, so this satisfies
 # AGENTS.md § Testing ("get a green run before committing; rsync the tree to an
 # isolated temp dir") without the contradiction the old push-then-worktree
@@ -17,9 +17,15 @@
 # copy includes .git plus your working-tree edits, so the remote diff matches
 # the Mac's. Pass any test.sh mode to override, e.g. `remote-test.sh unit`.
 #
+# Privacy: rsync does NOT honor .gitignore. We build the exclude list from git
+# so the sync carries only what git tracks plus untracked-unignored files —
+# secrets (.env, config/token-*.json, config/credentials-*.json) and personal
+# data (data/, ~15 GB) are gitignored and therefore never leave the machine.
+#
 # Streaming: remote stdout/stderr stream back live. A final marker line
 #   [remote-test] DONE rc=<code>
-# is always printed, so a backgrounded run can be waited on with:
+# is always printed — even on Ctrl-C / kill — so a backgrounded run can be
+# waited on with:
 #   ./scripts/remote-test.sh > "$OUT" 2>&1 &   # (or run_in_background)
 #   until grep -q "\[remote-test\] DONE" "$OUT"; do sleep 5; done
 #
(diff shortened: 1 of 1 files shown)


nbramia:
## Review Round 1 — self-review (referee, main context)

Focused single-file shell change (~60 lines), authored and verified empirically in this session, so reviewed from the main context rather than fanning out subagents (per the usage-limit protocol this stack adds to `implement/SKILL.md`).

**Verified clean:**
- `git ls-files --others --ignored --exclude-standard --directory` is correctly paired (`--ignored` requires an exclude source; `--exclude-standard` provides it). Fail-closed: if the enumeration errors the script refuses to sync rather than falling back to sending everything.
- `/`-anchoring via `sed 's#^#/#'` is injection-safe (path content is the subject line, not part of the sed expression) and correct for both dir (`/data/`) and file (`/.env`) entries.
- Empty-exclude case (pristine clone, nothing ignored) degrades gracefully to the inline artifact excludes.
- INT/TERM trap + EXIT cleanup compose correctly (single DONE marker on interrupt, temp file always removed under `set -u`).
- `printf %q` assembly neutralizes a malicious branch **and** arg simultaneously (tested); `tr -c` sanitization is a redundant second layer.

**Minor edge (not blocking):** a gitignored filename containing a literal rsync glob metacharacter (`*`, `[`) would be treated as a pattern — irrelevant for the actual secrets here (all plain names). Not worth complicating the generator.

No actionable findings — review loop complete.

PR #489, opened 2026-08-04
Replace wildcard CORS with an explicit allowlist

nbramia:
Closes the CORS item tracked as nbramia/ramia#15.

`api/main.py` configured `allow_origins=["*"]` with `allow_credentials=True`, under a comment saying "for local development".

Two problems:

1. **It never did what it looked like.** Browsers reject a wildcard origin on credentialed requests, so the combination is incoherent — it granted nothing while appearing permissive.
2. **It widened the security boundary.** This app has no authentication of its own; the tailnet is the boundary. Wildcard CORS extended that to "any page a tailnet device happens to have open", against 79 non-GET routes.

## The change

An explicit list: `localhost`, `127.0.0.1`, and the `tailscale serve` HTTPS front. The tailnet hostname comes from a new `tailnet_https_url` setting (`TAILNET_HTTPS_URL`, already in `.env`) rather than being written into source; unset simply drops it from the list.

## Why the UI cannot break

The web UI is served by this app and addresses it with root-relative paths only — verified by sweeping `web/` for `fetch(` call sites, all of which are `/api/...`. Those requests are same-origin, and same-origin requests never consult CORS rules. Nothing else calls this API from a browser at another origin; the MCP bridge and the `local-processing` watchers are server-side, where CORS does not apply.

So the practical blast radius is close to zero — this closes a hole rather than changing behaviour.

## Verification status — please read

**This is untested against a running server.** nathanramia-linux has been offline throughout, so I could not exercise it. What I did verify: both files compile, and the origin list builds correctly with and without `TAILNET_HTTPS_URL` set.

Worth a click-around after merge, since `lifeos-autodeploy` will ship it within ~10 minutes of the host coming back. If anything did depend on cross-origin access, the symptom is requests failing in the browser while `curl` still succeeds.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -184,10 +184,27 @@ app = FastAPI(
     lifespan=lifespan
 )
 
-# CORS middleware for local development
+# CORS: an explicit allowlist of the addresses this app answers on.
+#
+# The web UI does not depend on any of this — it is served by this app and
+# addresses it with root-relative paths, so its requests are same-origin and
+# never consult these rules. The list covers a page loaded at one of these
+# addresses that calls the API at another.
+#
+# A wildcard here was both ineffective and dangerous. Browsers reject `*` on
+# credentialed requests, so it never granted what it appeared to; and because
+# this app has no authentication of its own, `*` widened the security boundary
+# from "the tailnet" to "any page a tailnet device happens to have open".
+_cors_origins = [
+    f"http://localhost:{settings.port}",
+    f"http://127.0.0.1:{settings.port}",
+]
+if settings.tailnet_https_url:
+    _cors_origins.append(settings.tailnet_https_url.rstrip("/"))
+
 app.add_middleware(
     CORSMiddleware,
-    allow_origins=["*"],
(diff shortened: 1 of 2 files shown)


nbramia:
Closing as already shipped.

The exact change in this PR is on `main` — `api/main.py` builds `_cors_origins` from localhost/127.0.0.1 plus `settings.tailnet_https_url`, with `allow_origins=_cors_origins`. Verified byte-identical: `git diff main origin/cors-allowlist -- api/main.py` contains zero CORS hunks.

It landed via `6487db5` (2026-08-03), which was squashed under the unrelated title *"chore: remove project-level lifecycle skills superseded by the bootstraps plugin (#490)"* — one day before this PR was opened, which is why this one was never closed.

The branch is now 23 commits behind and, beyond CORS, would reintroduce code that predates the Telegram-token log redaction (#520) and the server host guard (#506). Nothing here is worth salvaging.

The stated blocker at the time — "untested against a running server" because the host was offline — no longer applies; the host is up and the allowlist is live.

Issue #566, opened 2026-08-15
feat(llm): make the local LLM target swappable per caller and handle reasoning-model output


nbramia:
## Context

Qwen3.8-27B (released 2026-08-14, Apache 2.0, dense 27.78B, hybrid Gated DeltaNet + vision encoder) was evaluated as a replacement for the current local model (Gemma 4 26B-A4B Q8_0). The evaluation surfaced two pre-existing weaknesses in the local LLM layer that are worth fixing on their own merits, independent of whether Qwen ever becomes the default.

**Why Qwen is not being made the default:** it is dense, so every token streams all ~17 GB (Q4_K_M) of weights through the UMA. The same architecture family measured **7.5 tok/s at Q8_0 on this exact SoC** (Ryzen AI MAX+ 395 / gfx1151) versus ~38-40 tok/s for the current MoE, which only activates ~3.8B params per token. Q4_K_M roughly doubles that, but it is still materially slower in the latency-critical routing path. Tracking the "should it become the default?" question separately — see the companion issue.

This issue covers the code changes only. Host-side work (llama.cpp rebuild for the Gated DeltaNet operators, GGUF download, benchmarking) is operational and not part of any PR.

## Problem 1 — `LocalLLMClient` silently discards reasoning output

`api/services/llm_client.py` has no handling for `reasoning_content` anywhere (verified: zero matches in the file). Reasoning models return chain-of-thought in a separate `reasoning_content` field on the OpenAI-compatible response, or inline in `<think>…</think>` tags — not in `content`.

The failure mode is silent and already reachable with the current model:

1. `query_router.py:420` calls `generate_text(prompt)` on the default `max_tokens=2048`.
2. Reasoning consumes the budget; `content` comes back empty.
3. The caller receives `""`, which is indistinguishable from a legitimate empty answer.
4. It falls through to the generic failure path at `query_router.py:379` — logged as a malformed response, root cause invisible.

This is a live bug for the current model, not a hypothetical one for Qwen.

## Problem 2 — the local LLM target is a single hardcoded endpoint

`_get_local_routing_client()` and `get_local_llm()` both construct `LocalLLMClient()` against one URL and one model name. There is no way to point the latency-critical routing callers (`query_router`, `person_facts`, `agent_viz_summary`) at a different local server than the agent worker's `local_executor`, and no way to disable thinking or set reasoning depth per request.

[...]

PR #568, opened 2026-08-15
fix(llm): surface reasoning_content separately instead of silently eating the token budget

nbramia:
## Summary

PR 1 of 2 for #566. Fixes `api/services/llm_client.py`, which had zero handling for reasoning-model output (`reasoning_content` field / inline `<think>...</think>` tags). The concrete failure this caused: the local model (Gemma 4 26B-A4B via llama-server) burns its `max_tokens` budget on chain-of-thought, `content` comes back empty with `finish_reason="length"`, and `query_router.py`'s `generate_text()` caller gets `""` — indistinguishable from a legitimate empty answer, with the root cause invisible in logs.

## Changes

- **`LLMResponse.reasoning`** — new field, populated by `_parse_response`, never merged into `.text`.
- **`LLMResponse.reasoning_starved`** — a **computed property** (not a stored flag): `not text and bool(reasoning) and finish_reason == "length"`. Chose a property over an exception because it's the simplest mechanism that fits the existing dataclass style, and being computed means it can never drift out of sync with the fields it derives from.
- **`_parse_response`** extracts `reasoning_content` and strips inline `<think>...</think>` blocks out of `content`, routing both into `.reasoning`. An unterminated `<think>` (truncated mid-thought, no closing tag) is treated as reasoning through to the end of the string rather than leaking a fragment into the answer. Ordinary non-reasoning responses are byte-for-byte unchanged — `content` is only touched when a `<think>` tag is actually found (no incidental `.strip()` on the normal path).
- **`generate_text`** (and `generate_json`, which wraps it) now log a warning when `reasoning_starved` is true, instead of silently returning `""`.
- **`astream()`** strips inline `<think>` tags out of streamed content deltas — buffered across chunk boundaries (via `_consume_think_stream`/`_longest_partial_tag_suffix`) so a tag split across two SSE chunks isn't half-leaked as text — and never reads `reasoning_content` deltas at all, so neither path reaches a `"text"` event.
  - **No new SSE event type.** Checked both `astream` consumers (`agent_loop.py`, `synthesizer.py`) and `docs/specs/technical/client-surfaces.md`: no consumer needs the reasoning text itself, and the documented SSE contract has no reasoning event to add one to. Simplest fix is to just not surface it.

## Tests

Added to `tests/test_llm_client.py`, all synthetic OpenAI-format payloads — no live llama-server, network, or GPU:
- Plain non-reasoning response unaffected (regression guard)
- Separate `reasoning_content` field kept out of `.text`
- Inline `<think>` stripped from a single-chunk response
- Unterminated `<think>` (truncation) routed to `.reasoning`, not leaked into `.text`
- `reasoning_starved` true/false cases (budget-exhausted vs. legitimate empty vs. wrong finish_reason)
- `generate_text` warns on starvation, stays silent on normal responses
- Streaming: plain stream unaffected, `reasoning_content` delta never surfaced, `<think>` split across multiple chunks fully stripped, unterminated `<think>` at truncation discarded, literal `<` in ordinary text not swallowed

`HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ./scripts/test.sh` — green (53/53 in `test_llm_client.py`; the 13 failures elsewhere in the full suite are pre-existing, reproduced identically on `main`, and depend on live local CRM/vault data unrelated to this file).

## Test plan

- [x] `./scripts/test.sh` (CPU-only) green
- [x] Confirmed pre-existing unrelated failures reproduce on `main` too
- [ ] PR 2 wires this into the actual caller-visible behavior (query_router logging / whatever #566's second half covers)

### DIFF
--- a/api/services/llm_client.py
+++ b/api/services/llm_client.py
@@ -43,6 +43,23 @@ class LLMResponse:
     model: str = ""
     finish_reason: str = ""
     tool_calls: list[dict] | None = None
+    reasoning: str = ""
+
+    @property
+    def reasoning_starved(self) -> bool:
+        """True when the model spent its whole token budget on reasoning and
+        never got to an answer.
+
+        Reasoning-capable models (e.g. Gemma 4 26B-A4B via llama-server) can
+        return chain-of-thought separately from the answer — via a dedicated
+        ``reasoning_content`` field or inline ``<think>...</think>`` tags —
+        both stripped into ``reasoning`` by ``_parse_response``. If the token
+        budget runs out before the model reaches its answer, ``text`` ends up
+        empty with ``finish_reason == "length"``, which is otherwise
+        indistinguishable from a legitimate empty response. Computed rather
+        than stored so it can't drift out of sync with the fields it reads.
+        """
+        return not self.text and bool(self.reasoning) and self.finish_reason == "length"
 
 
 def _anthropic_tools_to_openai(tools: list[dict]) -> list[dict]:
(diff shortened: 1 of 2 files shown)


nbramia:
## Adversarial review — DO NOT MERGE as-is

Reviewed by codex (independent pass, GitHub file/diff reads; local sandbox blocked it from running the suite). Three defects found in the `<think>`-tag handling. All three reproduce from the code as written.

### 1. CRITICAL — closing-only `</think>` leaks reasoning into `.text` and SSE

Many llama.cpp jinja templates **pre-fill the opening `<think>` into the prompt**, so the model emits only a closing tag. Input:

```python
content = "internal reasoning\n</think>{\"sources\":[\"calendar\"],\"reasoning\":\"ok\"}"
```

`_extract_inline_thinking` matches only a complete `<think>…</think>` pair or a lone *opening* tag, so it returns this untouched. The reasoning lands in `.text`.

Streaming is hit the same way — `in_think` starts `False` and `_consume_think_stream` only looks for `<think>` in that state, so chunks `["internal", " reasoning</thi", "nk>answer"]` emit the reasoning **plus a stray `</think>`** as text events. That breaks the `client-surfaces.md` SSE contract and can poison `query_router.py:420`, which parses JSON out of the returned string.

Not covered by the new tests, which only exercise complete opening+closing pairs.

### 2. MAJOR — legitimate literal `<think>` text is corrupted

```python
content = "Use `<think>draft</think>` as the example tag."
```

becomes ``"Use `` as the example tag."`` with `draft` silently moved into `.reasoning`. Anyone asking about HTML-like tags, chat templates, or code snippets gets mangled output — plausible in this repo. In streaming, chunks `["<thi", "nk>literal</think>"]` can make `astream()` emit **no text at all** where it previously emitted text.

### 3. MAJOR — `reasoning_starved` misses the closing-only starvation case

```python
{"choices": [{"message": {"content": "long hidden reasoning\n</think>"}, "finish_reason": "length"}]}
```

`.text` is non-empty, `.reasoning` is empty, so `reasoning_starved` is `False`, `generate_text` skips its new warning, and the leaked reasoning is returned as the answer — the exact failure this PR exists to eliminate.

(Checked separately: llama.cpp's OpenAI chat path does use `finish_reason="length"` for budget exhaustion, so the property's reliance on that value is sound. No finding there.)

## Required fix — one rule resolves all three

**Reasoning is always a PREFIX of the response, never mid-text.**

- `<think>` at the start (modulo leading whitespace) opens a reasoning prefix, running to the matching `</think>` or to end-of-string if unterminated.
- `</think>` with no preceding `<think>` means the template pre-filled the opener — everything before it is reasoning, everything after is the answer.
- Tags that are *not* part of a leading reasoning prefix are left completely alone, in content and in streamed deltas.

That fixes #1 (closing-only now classified), #2 (mid-prose tags untouched — there is text before them, so they aren't a prefix), and #3 (the payload now has non-empty `.reasoning` and empty `.text`).

Fix in progress on the same branch, with failing-before/passing-after regression tests for each finding.

PR #576, opened 2026-08-17
fix(scheduler): reject unknown notification bot names instead of silently using the primary chat

### LINKED ISSUE #575: fix(scheduler): reject unknown notification bot names instead of silently using the primary chat
## Problem

A schedule whose notification bot name doesn't match any configured bot delivers to the primary chat instead — with no signal to the user that the routing was ignored. The specialized bots exist precisely to keep domains separate, so a misroute doesn't just mislabel a message: it puts finance, health, or therapy content into the general feed.

This is not hypothetical. A weekday portfolio alert named a bot that had been renamed by an earlier persona-rename change. The schedule kept firing successfully, kept recording a "sent" status, and delivered every alert to the main channel for weeks. Nothing in the system said anything was wrong.

Two properties combine to make this invisible. Nothing validates a bot name when a schedule is created or updated, so a typo or an orphaned name is accepted as readily as a real one. And at delivery time the unresolvable name degrades to the primary bot, which is a reasonable failure mode on its own but is reported only to a log file nobody reads.

A rename of any bot silently orphans every schedule that referenced the old name. The failure is delayed, silent, and points at the wrong subsystem when it finally surfaces.

## Solution

Bot names become a validated field rather than a free-form string, checked against the live bot registry at every point where a schedule is written.

Creating or updating a schedule with an unrecognized bot name fails with an error that names the valid options. The rejection happens on both the scheduler HTTP surface and the equivalent scheduler tools exposed over MCP, so an agent writing a schedule gets the same feedback a direct API caller does. An empty or absent name remains valid and continues to mean the primary bot.

At fire time the behavior stays fail-open — an unresolvable name must never cost the user a notification — but stops being silent. The notification is still delivered via the primary bot, and it carries a visible marker that the intended routing could not be honored. The user sees the misroute in the channel where it lands, not in a log.

The registry is the single source of truth for valid names. Documentation and tool descriptions that currently enumerate example bot names should point at the registry instead of restating a list, so the next rename cannot leave them stale.

## Acceptance Criteria

- [ ] Given a bot registry containing a bot named `fitness`, when `POST /api/scheduler` is called with `bot` set to a name absent from the registry, then the request is rejected with HTTP 422 and the error message enumerates the valid names.
- [ ] Given the same registry, when `PUT /api/scheduler/{schedule_id}` is called with `bot` set to an unregistered name, then the request is rejected with HTTP 422 and the stored schedule is left unmodified.
- [ ] When `POST /api/scheduler` or `PUT /api/scheduler/{schedule_id}` is called with `bot` omitted, or set to the empty string, the request succeeds and the schedule resolves to the primary bot.
- [ ] When `POST /api/scheduler` or `PUT /api/scheduler/{schedule_id}` is called with `bot` set to the literal `primary`, the request succeeds.
- [ ] When `POST /api/scheduler` is called with `bot` set to a name that exists in the registry, the request succeeds and the persisted entry retains that name.
- [ ] When the MCP tools `lifeos_schedule_create` and `lifeos_schedule_update` are invoked with an unregistered bot name, each returns an error naming the valid options rather than writing the schedule.
- [ ] Given a stored schedule carrying a bot name absent from the registry, when that schedule fires, then the notification is still delivered — the message is never dropped or suppressed on account of the unresolvable name.
- [ ] Given that same stored schedule, when it fires, then the delivered message states that the configured bot could not be resolved and names it, so the misroute is visible in the delivered notification and not only in the logs.
- [ ] Given a stored schedule whose bot name is present in the registry, when it fires, then the delivered message contains no misroute marker.
- [ ] Given a configuration with no specialized bots registered at all, when a schedule is created with `bot` omitted and later fires, then both operations succeed — validation must not break an installation that has only a primary bot.
- [ ] No bot name, chat identifier, or token value is hardcoded in the validation path; the set of valid names is read from the registry at call time.

## Verification

- **Automated tests:** Every criterion above maps to a test. Route-level criteria are covered by API tests that patch the bot registry to a synthetic set — a create with a bad name, an update with a bad name, and the empty/`primary`/valid-name accept cases. MCP criteria are covered by calling the two tool handlers directly with a patched registry. Fire-time criteria are covered by constructing a schedule entry with an orphaned bot name, firing it against a mocked send function, and asserting both that a send occurred and that the sent text carries the marker; the paired positive case asserts a resolvable name produces no marker. The empty-registry criterion is covered by patching the registry to an empty list.
- **Manual checks:** None required.
- **Existing test suite:** `tests/test_scheduler*.py`, `tests/test_persona_api.py`, and the MCP tool tests must continue to pass. The existing fire-path tests that assert plain message bodies for resolvable bots must remain green unmodified — if the marker leaks into the resolvable case, those tests catch it.

## Scope

**In scope:**
- Bot-name validation on the scheduler create and update HTTP routes.
- Bot-name validation in the `lifeos_schedule_create` and `lifeos_schedule_update` MCP tool handlers.
- A visible misroute marker on notifications sent by a schedule whose bot name cannot be resolved.
- Correcting the stale example bot names in the scheduler request models, the schedule entry model comment, and the MCP tool descriptions, replacing enumerated examples with a pointer to the registry.
- Test coverage for the orphaned-name case at create, at update, and at fire time.

**Out of scope:**
- Migrating or repairing existing schedules that carry orphaned bot names. The one known instance was corrected out of band before this issue was filed; a general migration is a separate concern.
- Validating bot names on any surface other than the scheduler — the send endpoint, the agent worker, and the persona layer each accept bot names and are unchanged here.
- Changing the fail-open fallback itself. Delivering via the primary bot remains correct behavior; only its silence is being fixed.
- Any change to the bot registry format or to how tokens and chat identifiers are configured.

## Technical Context

- **Key files:**
  - `api/routes/scheduler.py` — `CreateScheduleRequest` (line 34) and `UpdateScheduleRequest` (line 48) carry the `bot` field; the create route is at line 132 and the update route at line 193. The `bot` field description at line 44 and the send-endpoint description at line 112 both cite `'therapy'`, which is not a registry name.
  - `mcp_server.py` — the `lifeos_schedule_create` and `lifeos_schedule_update` schemas are at lines 741 and 760, their `bot` descriptions at lines 751 and 770 both cite `'therapy'`, another stale citation sits at line 824, and the shared handler for both tools is at line 1744.
  - `api/services/telegram.py` — `_resolve_bot` at line 48 is the existing resolution logic and already returns `(None, None)` for an unknown name after logging a warning; it is the natural place to source the valid-name set from, and its contract should not change.
  - `api/services/scheduler_store.py` — `ScheduleEntry.bot` is declared at line 76 with the stale `'therapy'` comment; the fire-time send is at line 985 and the failure-path send at line 998.
  - `config/telegram_bots.json` — the tracked registry, loaded by the `telegram_bots` property in `config/settings.py` at line 949.
- **Patterns to follow:** Validation belongs in Pydantic field validators on the two request models, matching how the scheduler routes already validate other enumerated fields. The registry must be read inside the validator rather than captured at import time, since `telegram_bots` is an uncached property that reflects the current environment. The MCP handler should reuse the same validation helper rather than reimplementing the check.
- **Constraints:** No new dependencies. Do not hardcode bot names, chat identifiers, or tokens anywhere in the validation path. The repository is open source and must work for a fresh clone that has configured only a primary bot, so an empty registry has to be a valid state rather than a rejection. The misroute marker must be generic text that does not embed anyone's chat identifier.

## References

- `docs/guides/scheduler.md` — the scheduler guide, which currently documents no `bot` field at all and should describe the validated field and its source of truth.
- `docs/guides/personas.md` and `config/personas/README.md` — the persona layer that owns the bot names being validated against.
- PR #471 — the persona rename whose orphaned schedule exposed this gap.
- PR #464, PR #461 — earlier steps of the same persona rename.

## History

Surfaced when a weekday portfolio alert was reported arriving in the general channel rather than the finance channel. The alert's schedule still named the pre-rename bot introduced in #461 and renamed in #471. The rename passed its test suite because persona configuration is covered but schedule entries referencing bot names are not — the missing coverage is part of this defect, not incidental to it.

nbramia:
## TL;DR

A schedule can name which Telegram bot delivers its notification, but nothing checked that the name existed. A name that didn't match anything — typically one left orphaned by a bot rename — was accepted, kept reporting successful fires, and quietly delivered every notification to the general channel instead. Now an unrecognized name is rejected when the schedule is written, with the accepted names in the error, and a schedule already carrying a bad name still gets delivered but arrives with a visible routing warning naming the bot it could not use. An installation with only a primary bot configured is unaffected.

Closes #575

## Design

The bot registry is the single source of truth and is read at the moment of each check, never captured earlier, because it reflects the current environment and may be empty. An empty registry is a valid configuration — a fresh clone with only a primary bot — so it accepts an unset name and rejects only a name it doesn't have.

**Validation lives in the API and only in the API.** Rejection happens on the scheduler create and update routes. The MCP tools do *not* pre-check: they send the write and let the route's 422 come back as the tool's error. That is a deliberate correction made during review — see below.

Fire-time behavior deliberately stays fail-open: an unresolvable name must never cost a notification, so the send still goes out from the primary bot. Only the silence is fixed — the message gains a warning banner naming the unresolvable bot. The banner is generic text carrying nothing but that name, it appears on both the normal and the failure notification, and building it can never abort the send it annotates. A resolvable name produces no banner at all.

### Why the MCP pre-check was removed

The first cut of this PR validated the bot name inside `mcp_server.py` before its HTTP round-trip. That was wrong, and review caught it: it made the MCP client's view of the registry authoritative over the API's, and the two are not the same view.

- `settings.telegram_bots` reads `Path("config/telegram_bots.json")` — **relative to the process cwd**. The stdio MCP child is spawned with the calling agent's cwd, not the repo, so the registry read returned nothing and `valid_bot_names()` was just `['primary']`. A `lifeos_schedule_create(bot="finance")` that worked before this PR would have been hard-rejected and never written.
- Anchoring the path to the repo root would only half-fix it. A client pointed at a remote `LIFEOS_API_URL` has the tracked registry file but not the token env vars that make an entry count, so it would still reject names the server accepts.

So the pre-check is gone rather than hardened — a net code removal. `_call_api`'s existing `httpx.HTTPStatusError` branch returns `{"error": "API error 422: …Unknown Telegram bot 'x'. Configured names: …"}` and `_format_response` renders it as `Error: …`, which satisfies the issue's MCP acceptance criterion with the API as the one validator.

**One intentional deviation from the issue.** The issue specifies HTTP 422 and suggests Pydantic field validators. Those are mutually exclusive in this app: `api/main.py` registers a global `RequestValidationError` handler that rewrites every Pydantic failure to 400, and it cannot serialize a `ValueError` in the error context, so a field validator returns 500 rather than a rejection. This is pre-existing and reproducible on `POST /api/ask` with a blank question on `main`. The 422 requirement therefore won, and the check sits in the route bodies — which is also where these routes already validate their other enumerated fields.

## Implementation Notes

`api/services/telegram.py` gains three small registry-backed helpers next to the existing `_resolve_bot`, whose contract is untouched: `valid_bot_names`, `is_known_bot`, and `validate_bot_name`. Everything else calls into these. `validate_bot_name` returns the name to store, trimming surrounding whitespace so a tool argument that arrived as `' ledger'` stores as `'ledger'` and resolves at fire time; case is **not** folded, keeping exact-match parity with `_resolve_bot`.

- `api/routes/scheduler.py` — `_require_known_bot` translates a `ValueError` into an `HTTPException(422)` and returns the trimmed name; called in `create_schedule` after the action check and as the first statement of `update_schedule`, so a rejected update never reaches the store and a trimmed name is what gets persisted.
- `mcp_server.py` — tool descriptions only. No validation logic (see Design).
- `api/services/scheduler_store.py` — `_misroute_notice` builds the banner and is prefixed to both `_fire_entry` sends. Its registry lookup is wrapped so a malformed registry degrades to *no banner* rather than raising; it is evaluated inside the f-string before `send_message_async`, so an exception there would otherwise have killed the primary send and then the failure send too, losing the notification outright.
- Stale `'therapy'` citations (the registry name is `therapist`) are replaced with a pointer to `config/telegram_bots.json` in `mcp_server.py` (three tool descriptions), `api/routes/scheduler.py`, `api/routes/reminders.py` (the deprecated twin of the same send model), and the `ScheduleEntry.bot` comment.
- Docs: the scheduler guide gains a "Notification bot" section and the `[bot:: …]` field, notes that the field applies to every action except `agent`, describes the valid set as *configured* bots (a registry entry counts only once its token env var is set), and states that hand-edited markdown is **not** validated — the fire-time banner is the only safety net there. The API reference notes the 422 on both routes.

### Files changed

```
api/routes/reminders.py                |   5 +-
api/routes/scheduler.py                |  27 ++-
api/services/scheduler_store.py        |  32 ++-
api/services/telegram.py               |  40 +++
docs/guides/scheduler.md               |  25 +-
docs/specs/product/api-reference.md    |   6 +-
mcp_server.py                          |   8 +-
tests/test_scheduler_bot_validation.py | 432 +++++++++++++++++++++++++++++++++
8 files changed, 559 insertions(+), 16 deletions(-)
```

### Automated tests

`tests/test_scheduler_bot_validation.py` is new — **36 tests**, patching the registry to a synthetic two-bot set (`alerts`, `ledger`) so nothing asserts against real bot names or ids. It covers the helpers, create/update rejection and acceptance (unset, empty, `primary`, registered), whitespace trimming on both write routes, the deliberate non-folding of case, the empty-registry accept and reject cases, the MCP surface, and the fire path in five shapes: orphaned name delivers with the marker, registered name delivers without it, unset name on an empty registry delivers without it, the failure notification also carries it, and — new in round 1 — **the send still happens when the registry read raises**. One test asserts the marker leaks no patched token or chat id.

The MCP tests assert the corrected contract: an API 422 surfaces as the tool's error with the valid options intact and renders through `_format_response` as `Error: …`, and a bot name the *local* registry doesn't know still reaches the API — the regression that the deleted pre-check would have caused.

**No existing test was modified.** The only test file in the diff is the new one.

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" pytest \
    tests/test_scheduler_bot_validation.py tests/test_mcp_server.py tests/test_scheduler_api.py -q
82 passed, 1 warning in 27.64s
```

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" pytest \
    tests/test_scheduler_api.py tests/test_scheduler_store.py tests/test_scheduler_watcher.py \
    tests/test_persona_api.py tests/test_mcp_server.py tests/test_mcp_http_transport.py \
    tests/test_agent_worker_mcp_exposure.py tests/test_telegram_outbound_routing.py \
    tests/test_telegram_multibot.py tests/test_telegram.py tests/test_reminders_api.py \
    tests/test_scheduler_bot_validation.py -q -n auto --dist loadscope -m "not requires_server"
306 passed, 82 warnings in 21.03s
```

The pre-push hook ran the full unit suite: `3365 passed, 8 skipped` plus `12 passed` server-free browser tests.

### Manual verification

Route rejection, against the **real** registry, in-process via `TestClient` (no server restart — the live service deploys from the canonical checkout, not this worktree). The names listed are the ones committed in the tracked registry file; no tokens or chat ids are involved.

```
=== 1. POST /api/scheduler with an unregistered bot (real registry) ===
422 {'detail': "Unknown Telegram bot 'portfolio-old'. Configured names: primary, fitness, therapist, doctor, finance — a registry entry (config/telegram_bots.json) counts only once its token env var is set"}

=== 2. PUT /api/scheduler/{id} with an unregistered bot ===
422 <- 422 before the 404 lookup: nothing was read or written

=== 3. PUT with a registered name reaches the store (404 = validation passed) ===
404 {'detail': 'Schedule not found'}
```

Case 2 returns 422 rather than the 404 the id would otherwise produce, which shows the rejection lands before the store is touched. Case 3 reaching the 404 shows a valid name passes through.

Fire path, real code and real registry, against a temp store and a mocked send so no live vault or chat is touched:

```
--- bot='portfolio-old' -> send called: True, status: sent
"⚠️ *Routing warning:* this schedule's bot 'portfolio-old' is not configured — delivered from the primary bot instead.\n\n*Portfolio alert*\n\nBalances are within range."

--- bot='finance' -> send called: True, status: sent
'*Portfolio alert*\n\nBalances are within range.'
```

The orphaned schedule still sends and still records `sent` — delivery is never lost — and the message carries the banner. The registered bot's message is byte-identical to the pre-change format, so nothing leaks into the healthy path.

## Review focus

- The 422-versus-field-validator tradeoff in `api/routes/scheduler.py:130` — see Design. Fixing the global handler in `api/main.py` so field validators work at all is a real bug worth its own issue; I deliberately left it alone.
- The banner's wording and placement at `api/services/scheduler_store.py:741`. It is prefixed above the schedule title so it is the first thing read. It uses `*bold*` and no backticks: the send functions already retry without `parse_mode` on any non-200, so a Markdown edge case costs formatting but never the message.
- `api/routes/reminders.py:87` is outside the issue's stated file list. It is the deprecated twin of the send model I fixed in `scheduler.py` and carried the same stale citation; fixing one and not the other seemed worse.
- `docs/specs/product/api-reference.md` still words the valid set as the names "registered in `config/telegram_bots.json`" rather than *configured*. Round 1 scoped that rewording to the rejection message and the scheduler guide; I left the API reference alone rather than widen the change, and am flagging it here instead.

### DIFF
--- a/api/routes/reminders.py
+++ b/api/routes/reminders.py
@@ -84,8 +84,9 @@ class SendMessageRequest(BaseModel):
     text: str = Field(..., min_length=1, description="Message text to send via Telegram")
     bot: Optional[str] = Field(
         default=None,
-        description="Optional bot name to send from (e.g. 'fitness', 'therapy'). "
-                    "Falls back to the primary bot if unset or unrecognised.",
+        description="Optional bot name to send from — a name from the registry "
+                    "(config/telegram_bots.json). Falls back to the primary bot "
+                    "if unset or unrecognised.",
     )
 
 
--- a/api/routes/scheduler.py
+++ b/api/routes/scheduler.py
@@ -127,18 +127,19 @@ def _resolve_action(action: Optional[str], message_type: Optional[str]) -> str:
     return _TYPE_TO_ACTION.get(message_type or "", "notify")
 
 
-def _require_known_bot(bot: Optional[str]) -> None:
-    """Reject a bot name the registry doesn't know (#575).
+def _require_known_bot(bot: Optional[str]) -> Optional[str]:
+    """Reject a bot name the registry doesn't know, else return it (#575).
 
     An orphaned name — usually the residue of a bot rename — otherwise stores
     fine and then silently delivers to the primary chat at every fire. The
     registry is read here, at request time, because it reflects the current
-    environment; empty or unset stays valid and means the primary bot.
+    environment; empty or unset stays valid and means the primary bot. Returns
+    the trimmed name so that is what gets stored.
     """
     from api.services.telegram import validate_bot_name
 
     try:
-        validate_bot_name(bot)
+        return validate_bot_name(bot)
     except ValueError as e:
         raise HTTPException(status_code=422, detail=str(e))
 
@@ -155,7 +156,7 @@ async def create_schedule(request: CreateScheduleRequest):
     action = _resolve_action(request.action, request.message_type)
     if action not in VALID_ACTIONS:
         raise HTTPException(status_code=400, detail=f"action must be one of {VALID_ACTIONS}")
-    _require_known_bot(request.bot)
+    request.bot = _require_known_bot(request.bot)
 
     store = get_scheduler_store()
     entry = store.create(
@@ -212,7 +213,7 @@ async def get_schedule(schedule_id: str):
 @router.put("/{schedule_id}", response_model=ScheduleResponse)
 async def update_schedule(schedule_id: str, request: UpdateScheduleRequest):
     """Update an existing schedule."""
-    _require_known_bot(request.bot)
+    request.bot = _require_known_bot(request.bot)
     store = get_scheduler_store()
     updates = {k: v for k, v in request.model_dump().items() if v is not None}
     entry = store.update(schedule_id, **updates)
--- a/api/routes/scheduler.py
+++ b/api/routes/scheduler.py
@@ -41,7 +41,8 @@ class CreateScheduleRequest(BaseModel):
     message_content: str = Field(default="", description="Static text or natural-language prompt")
     endpoint_config: Optional[dict] = Field(default=None, description="For endpoint action: {endpoint, method, params}")
     executor: str = Field(default="", description="For agent action: local | cloud | cloud-haiku | cloud-sonnet")
-    bot: str = Field(default="", description="Telegram bot to notify from (e.g. 'fitness', 'therapy'); empty = primary")
+    bot: str = Field(default="", description="Telegram bot to notify from — a name from the "
+                                             "registry (config/telegram_bots.json) or 'primary'; empty = primary")
     enabled: bool = Field(default=True)
     timezone: str = Field(default_factory=lambda: settings.timezone, description="IANA timezone for the schedule")
 
@@ -109,8 +110,9 @@ class SendMessageRequest(BaseModel):
     text: str = Field(..., min_length=1, description="Message text to send via Telegram")
     bot: Optional[str] = Field(
         default=None,
-        description="Optional bot name to send from (e.g. 'fitness', 'therapy'). "
-                    "Falls back to the primary bot if unset or unrecognised.",
+        description="Optional bot name to send from — a name from the registry "
+                    "(config/telegram_bots.json). Falls back to the primary bot "
+                    "if unset or unrecognised.",
     )
 
 
@@ -125,6 +127,22 @@ def _resolve_action(action: Optional[str], message_type: Optional[str]) -> str:
     return _TYPE_TO_ACTION.get(message_type or "", "notify")
 
 
+def _require_known_bot(bot: Optional[str]) -> None:
+    """Reject a bot name the registry doesn't know (#575).
+
+    An orphaned name — usually the residue of a bot rename — otherwise stores
+    fine and then silently delivers to the primary chat at every fire. The
+    registry is read here, at request time, because it reflects the current
+    environment; empty or unset stays valid and means the primary bot.
+    """
+    from api.services.telegram import validate_bot_name
+
+    try:
+        validate_bot_name(bot)
+    except ValueError as e:
+        raise HTTPException(status_code=422, detail=str(e))
+
+
 # ---------------------------------------------------------------------------
 # Routes (static paths MUST come before {schedule_id} to avoid capture)
 # ---------------------------------------------------------------------------
@@ -137,6 +155,7 @@ async def create_schedule(request: CreateScheduleRequest):
     action = _resolve_action(request.action, request.message_type)
     if action not in VALID_ACTIONS:
         raise HTTPException(status_code=400, detail=f"action must be one of {VALID_ACTIONS}")
+    _require_known_bot(request.bot)
 
     store = get_scheduler_store()
     entry = store.create(
@@ -193,6 +212,7 @@ async def get_schedule(schedule_id: str):
 @router.put("/{schedule_id}", response_model=ScheduleResponse)
 async def update_schedule(schedule_id: str, request: UpdateScheduleRequest):
     """Update an existing schedule."""
+    _require_known_bot(request.bot)
     store = get_scheduler_store()
     updates = {k: v for k, v in request.model_dump().items() if v is not None}
     entry = store.update(schedule_id, **updates)
--- a/api/services/scheduler_store.py
+++ b/api/services/scheduler_store.py
@@ -73,7 +73,7 @@ class ScheduleEntry:
     message_content: str = ""  # static text or natural-language prompt
     endpoint_config: Optional[dict] = None  # {endpoint, method, params}
     executor: str = ""  # agent executor tag without '#': local / cloud / cloud-haiku ...
-    bot: str = ""  # Telegram bot name for outbound sends (e.g. "fitness", "therapy"); empty = primary
+    bot: str = ""  # Telegram bot name for outbound sends (registry: config/telegram_bots.json); empty = primary
     enabled: bool = True
     created_at: str = ""
     last_triggered_at: Optional[str] = None
@@ -738,6 +738,23 @@ class SchedulerStore:
         return entry.action
 
 
+def _misroute_notice(bot: str) -> str:
+    """Banner for a schedule whose configured bot can't be resolved (#575).
+
+    Delivery stays fail-open — an orphaned bot name (typically the residue of a
+    bot rename) must never cost a notification, so the send still goes out from
+    the primary bot. This makes that substitution visible in the channel the
+    message lands in instead of only in a log line nobody reads. Generic text:
+    it names the unresolvable bot and nothing else.
+    """
+    from api.services.telegram import is_known_bot
+
+    if is_known_bot(bot):
+        return ""
+    return (f"⚠️ *Routing warning:* this schedule's bot '{bot}' is not configured "
+            f"— delivered from the primary bot instead.\n\n")
+
+
 class SchedulerScheduler:
     """
     Background thread that fires due schedules every 60 seconds.
@@ -982,7 +999,10 @@ class SchedulerScheduler:
 
             if message:
                 from api.services.telegram import send_message_async
-                await send_message_async(f"*{entry.name}*\n\n{message}", bot=entry.bot or None)
+                await send_message_async(
+                    f"{_misroute_notice(entry.bot)}*{entry.name}*\n\n{message}",
+                    bot=entry.bot or None,
+                )
                 self.store.record_run(entry.id, "sent", message[:200])
             else:
                 self.store.record_run(entry.id, "empty", "")
@@ -993,7 +1013,7 @@ class SchedulerScheduler:
             try:
                 from api.services.telegram import send_message_async
                 await send_message_async(
-                    f"*{entry.name}* (failed)\n\n"
+                    f"{_misroute_notice(entry.bot)}*{entry.name}* (failed)\n\n"
                     f"Schedule could not execute: {str(e)[:200]}",
                     bot=entry.bot or None,
                 )
--- a/api/services/scheduler_store.py
+++ b/api/services/scheduler_store.py
@@ -749,7 +749,13 @@ def _misroute_notice(bot: str) -> str:
     """
     from api.services.telegram import is_known_bot
 
-    if is_known_bot(bot):
+    try:
+        known = is_known_bot(bot)
+    except Exception:
+        # Reading the registry must never be what loses a notification: this
+        # runs inside the message the send is about to deliver.
+        return ""
+    if known:
         return ""
     return (f"⚠️ *Routing warning:* this schedule's bot '{bot}' is not configured "
             f"— delivered from the primary bot instead.\n\n")
--- a/api/services/telegram.py
+++ b/api/services/telegram.py
@@ -71,6 +71,39 @@ def _token_for_bot(bot: Optional[str]) -> Optional[str]:
     return _resolve_bot(bot)[0]
 
 
+def valid_bot_names() -> list[str]:
+    """Names a caller may legitimately route a send to, newest state.
+
+    ``"primary"`` plus whatever the registry (``config/telegram_bots.json``)
+    currently holds. Read on every call rather than cached at import: the
+    ``telegram_bots`` property is uncached and reflects the current
+    environment, and a registry with no specialized bots is a valid state (a
+    fresh clone configures only the primary bot).
+    """
+    return ["primary"] + [cfg.name for cfg in settings.telegram_bots]
+
+
+def is_known_bot(bot: Optional[str]) -> bool:
+    """Whether ``bot`` resolves to a configured bot. Empty means primary."""
+    return not bot or bot in valid_bot_names()
+
+
+def validate_bot_name(bot: Optional[str]) -> None:
+    """Raise ``ValueError`` if ``bot`` names no configured bot.
+
+    Used wherever a bot name is *written* (scheduler create/update) so an
+    orphaned name — the residue of a bot rename — is rejected at the point of
+    entry instead of silently degrading to the primary chat weeks later (#575).
+    Empty or unset stays valid and continues to mean the primary bot.
+    """
+    if is_known_bot(bot):
+        return
+    raise ValueError(
+        f"Unknown Telegram bot '{bot}'. Valid names: "
+        f"{', '.join(valid_bot_names())} (registry: config/telegram_bots.json)"
+    )
+
+
 # ---------------------------------------------------------------------------
 # Message sending
 # ---------------------------------------------------------------------------
--- a/api/services/telegram.py
+++ b/api/services/telegram.py
@@ -88,19 +88,26 @@ def is_known_bot(bot: Optional[str]) -> bool:
     return not bot or bot in valid_bot_names()
 
 
-def validate_bot_name(bot: Optional[str]) -> None:
-    """Raise ``ValueError`` if ``bot`` names no configured bot.
+def validate_bot_name(bot: Optional[str]) -> Optional[str]:
+    """Return the name to store, or raise ``ValueError`` if it isn't configured.
 
     Used wherever a bot name is *written* (scheduler create/update) so an
     orphaned name — the residue of a bot rename — is rejected at the point of
     entry instead of silently degrading to the primary chat weeks later (#575).
     Empty or unset stays valid and continues to mean the primary bot.
+
+    Surrounding whitespace is trimmed, so a tool argument that arrived as
+    ``' ledger'`` stores as ``'ledger'`` and resolves at fire time. Case is
+    *not* folded: matching stays exact, in parity with :func:`_resolve_bot`.
     """
+    if bot is not None:
+        bot = bot.strip()
     if is_known_bot(bot):
-        return
+        return bot
     raise ValueError(
-        f"Unknown Telegram bot '{bot}'. Valid names: "
-        f"{', '.join(valid_bot_names())} (registry: config/telegram_bots.json)"
+        f"Unknown Telegram bot '{bot}'. Configured names: "
+        f"{', '.join(valid_bot_names())} — a registry entry "
+        f"(config/telegram_bots.json) counts only once its token env var is set"
     )
 
 
--- a/docs/guides/scheduler.md
+++ b/docs/guides/scheduler.md
@@ -35,7 +35,9 @@ source of truth; `data/scheduler_index.json` is a rebuildable cache.
 - **`[bot:: <name>]`** — which Telegram bot delivers the notification (see below); omitted means the primary bot.
 
 Editing a line in Obsidian (changing the cron, toggling the checkbox) is picked
-up within ~2s by the file watcher.
+up within ~2s by the file watcher. Markdown edits are **not** validated — a
+`[bot:: <name>]` typed here is accepted as-is, and the fire-time routing warning
+below is the only safety net.
 
 ## Triggers
 
@@ -63,13 +65,15 @@ to say.
 
 ### Notification bot
 
-A `notify` or `prompt` schedule can name the Telegram bot that delivers it, so
-finance, health, or therapy content lands in its own channel instead of the
-general feed. The valid names are `primary` plus whatever is registered in
-`config/telegram_bots.json` — that registry is the source of truth, and both
-`POST /api/scheduler` and `PUT /api/scheduler/{id}` reject any other name with
-a 422 that lists the accepted ones. Leaving the field unset means the primary
-bot, which is what an installation with no specialized bots configured gets.
+Any schedule except `action:: agent` can name the Telegram bot that delivers it,
+so finance, health, or therapy content lands in its own channel instead of the
+general feed. The valid names are `primary` plus whatever is *configured* —
+`config/telegram_bots.json` is the registry, but an entry there counts only once
+the env var named by its `token_env` is set, so a listed bot with no token is not
+an accepted name. Both `POST /api/scheduler` and `PUT /api/scheduler/{id}` reject
+any other name with a 422 that lists the accepted ones. Leaving the field unset
+means the primary bot, which is what an installation with no specialized bots
+configured gets.
 
 If a stored schedule names a bot the registry no longer has — usually because
 the bot was renamed after the schedule was written — the notification is still
--- a/docs/guides/scheduler.md
+++ b/docs/guides/scheduler.md
@@ -2,7 +2,7 @@
 
 > **Status:** Complete
 > **Owner:** Scheduler
-> **Last Updated:** 2026-06-21
+> **Last Updated:** 2026-08-17
 > **Audience:** Operators
 
 The Scheduler runs work on a timer. A **schedule** binds a **trigger** (one-off
@@ -32,6 +32,7 @@ source of truth; `data/scheduler_index.json` is a rebuildable cache.
 - **`[tz:: <IANA zone>]`** — interprets the trigger in that timezone (defaults to the configured timezone).
 - **`[action:: …]`** — what fires (see below).
 - **`#executor` tag** — for `action:: agent`, the executor: `#local`, `#cloud`, `#cloud-haiku`, or `#cloud-sonnet`.
+- **`[bot:: <name>]`** — which Telegram bot delivers the notification (see below); omitted means the primary bot.
 
 Editing a line in Obsidian (changing the cron, toggling the checkbox) is picked
 up within ~2s by the file watcher.
@@ -59,6 +60,22 @@ For `notify` and `prompt`, the Telegram message is **suppressed** when the
 result is empty or a sentinel (`NO_MEETING`, `NOTHING_TO_REPORT`, …) — so
 high-frequency schedules like pre-meeting prep stay quiet when there's nothing
 to say.
+
+### Notification bot
+
+A `notify` or `prompt` schedule can name the Telegram bot that delivers it, so
+finance, health, or therapy content lands in its own channel instead of the
+general feed. The valid names are `primary` plus whatever is registered in
+`config/telegram_bots.json` — that registry is the source of truth, and both
+`POST /api/scheduler` and `PUT /api/scheduler/{id}` reject any other name with
+a 422 that lists the accepted ones. Leaving the field unset means the primary
+bot, which is what an installation with no specialized bots configured gets.
+
+If a stored schedule names a bot the registry no longer has — usually because
+the bot was renamed after the schedule was written — the notification is still
+delivered from the primary bot rather than dropped, but the message carries a
+routing warning naming the unresolvable bot so the misroute is visible in the
+channel it lands in.
 
 ### Agent hand-off
 
--- a/docs/specs/product/api-reference.md
+++ b/docs/specs/product/api-reference.md
@@ -2,7 +2,7 @@
 
 **Status:** Complete
 **Owner:** API Gateway
-**Last Updated:** 2026-07-09
+**Last Updated:** 2026-08-17
 
 Catalog of every HTTP endpoint LifeOS exposes, with request/response shapes. Two adjacent catalogs split out for size:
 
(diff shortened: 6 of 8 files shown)


nbramia:

**Specialist rationale:** `review-correctness` was added because this PR changes an error path with a hard fail-open invariant — a schedule whose bot name is unresolvable must still deliver. A subtle regression there silently costs the user notifications, which is the exact failure the PR exists to fix.

### Reviewer Findings

#### General

**Action Required**
- `mcp_server.py:1059-1071` + `:1105-1111` — the MCP-side pre-check false-rejects every valid specialized bot in the real stdio deployment. `_validate_schedule_bot` → `valid_bot_names()` → `settings.telegram_bots`, and that property reads `_TELEGRAM_BOTS_FILE = Path("config/telegram_bots.json")` (`config/settings.py:20`) — a **cwd-relative** path, plus `.env` via `dotenv_values(env_file=".env")` for the token gate. `mcp_server.py`'s own header (lines 29-36) documents that this server "is spawned as a stdio MCP child by CLI agents with the agent's `-C` working dir as cwd — NOT the repo", which is why `AGENT_SESSIONS_DB` had to be anchored to `_REPO_ROOT`. The new validation path was not anchored, so it reads an empty registry. Reproduced from `cwd=/tmp`: `finance`, `fitness`, `therapist` all → `Unknown Telegram bot … Valid names: primary`. `~/.claude.json` registers `mcpServers.lifeos` with no `cwd`, so the stdio child inherits the client's project directory. Net effect: `lifeos_schedule_create(bot="finance")`, which worked before this PR, now returns an error and never writes — and the error actively misinforms the agent that only `primary` exists. `lifeos-api` and `lifeos-mcp-http` set `WorkingDirectory`, so the HTTP routes and Managed-Agents transport are unaffected; only the stdio path regresses. The new tests can't catch this because they patch `api.services.telegram.settings` wholesale. Smallest correction: drop the client-side pre-check and let the route's 422 propagate — `_call_api`'s `httpx.HTTPStatusError` branch (`:1178`) already returns the 422 body naming the valid options, and no write occurs. Anchoring the path instead would only half-fix it: a remote MCP client has the tracked registry file but not the token env vars, and `telegram_bots` skips token-less entries.

**Recommended**
- `api/services/telegram.py:101-104` and `docs/guides/scheduler.md:66-69` — the rejection message and guide describe the valid set as "whatever is registered in `config/telegram_bots.json`", but `config/settings.py:986-992` skips any entry whose `token_env` var is unset. A name that *is* in that file gets rejected by an error pointing at that same file, with no hint that the missing piece is the token. For the fresh-clone audience this is the likely first encounter with the new 422.

**Minor**
- `docs/guides/scheduler.md:63-64` — "A `notify` or `prompt` schedule can name the Telegram bot" understates the field's reach: `_fire_entry` also routes an `endpoint`-action message and the failure notice of any non-`agent` action through `entry.bot`, while `action:: agent` returns early and ignores `bot`.

**Summary** — Solid, well-scoped change: shared registry-backed helpers read at call time, both HTTP routes reject with 422 before touching the store, the fire path stays fail-open with a generic marker, privacy holds (synthetic `alerts`/`ledger` registry, explicit no-token/no-chat-id leak assertion). Verified both self-reported items: a Pydantic field validator on this app really does yield 500 (`errors()[0]["ctx"]["error"]` is a raw `ValueError` and `json.dumps` fails), so route-body placement is the only way to hit the mandated 422; and the `reminders.py` stretch is three lines of field description on the deprecated twin, with a repo-wide grep confirming zero stale `'therapy'` citations remain. No existing test modified. Executed: new file 29 passed, adjacent focused suites 175 passed, GPU disabled. Full repo suite not run.

#### Correctness

**Action Required**
- `mcp_server.py:1105-1112` — the MCP pre-flight validates against a *different* registry than the API will use. Two independent divergences: **(1) cwd** — measured `cwd=<repo>` → `['primary','fitness','therapist','doctor','finance']`, `finance` ACCEPTED; `cwd=/tmp` and `cwd=~/Code` → `['primary']`, `finance` REJECTED. `mcp_server.py:36` already computes `_REPO_ROOT` from `__file__`, and the comment at `:1012-1014` says stores must be anchored "regardless of the cwd this MCP server was spawned with" — the new check ignores that lesson. **(2) Remote API** — `API_BASE` may target another host; `config/settings.py:995-1000` drops entries whose token env var is unset, so a clone without the specialized tokens yields `valid_bot_names() == ["primary"]` regardless of cwd. Invisible to the suite because `tests/test_scheduler_bot_validation.py:35-41` patches `settings` in-process rather than exercising the real file lookup. Smallest remedy: delete the pre-flight, let the 422 propagate; `_format_response` renders it as `Error: …`, satisfying the MCP acceptance criterion with the API as single source of truth.

**Recommended**
- `mcp_server.py:1060-1073` — `_validate_schedule_bot` catches only `ValueError`, so a malformed registry breaks the tool instead of the name. A `config/telegram_bots.json` whose top-level list holds non-dict elements makes `settings.telegram_bots` raise `AttributeError` (`config/settings.py:975` calls `entry.get(...)` unguarded; only `JSONDecodeError`/`OSError`/non-list top level are handled). It escapes into `dispatch`'s blanket handler, failing the whole JSON-RPC call.
- `api/services/scheduler_store.py:741-756` — `_misroute_notice` can abort the very send it exists to annotate. It is evaluated inside the f-string at `:1003`/`:1016`, i.e. *before* `send_message_async`, and its registry read is unguarded. Measured with a malformed registry: `bot=''` → `sends=1, runs=[('sent', …)]`; `bot='finance'` → `sends=0, runs=[('failed', "'str' object has no attribute 'get'")]`. The primary send raises, the `except` at `:1009` calls the failure send, `_misroute_notice` raises again, and the inner `except` at `:1020` swallows it — notification lost. The empty-`bot` case is safe (`is_known_bot` short-circuits on `not bot`). Parity note: `_resolve_bot` inside `send_message_async` carries the same exposure on `main`, so this is not entirely new — but the premise of this PR is that the fire path must never lose a notification.

**Minor**
- `api/services/telegram.py:88` — whitespace/case rejected rather than normalized. `'Finance'`, `' finance'`, `'finance '` all rejected, each with an error whose valid-names list contains a visually identical entry.
- A non-string `bot` over MCP is rejected (good) but interpolated raw: `bot=["finance"]` → `Unknown Telegram bot '['finance']'`.
- `docs/guides/scheduler.md:35` adds `[bot:: <name>]` to the inline-field list four lines above the note that Obsidian edits are picked up by the file watcher, and the new § states the two HTTP routes reject unknown names. The markdown path is genuinely unvalidated — `_parse_entry_line` does `bot=fields.get("bot","")` with no check. Intentional per scope, but one clause saying so would stop a reader inferring parity.
- The banner adds ~110 chars, so a notification near the 4096 limit now splits with the warning only on the first chunk. Nothing dropped; not worth changing.

**Summary** — Core scheduler logic correct, verified by execution: `update_schedule` validates as its first statement so a rejection cannot partially write; the suppression block, `if message:` guard, and `entry.action in ("notify","prompt")` branch are byte-identical; all registry reads are at call time with no import-time capture; the empty-registry case accepts an unset name and rejects a named one; `is_known_bot`'s `not bot` short-circuit means the common empty-`bot` fire does no registry I/O; `test_registered_bot_delivers_unmarked` asserts byte-exact equality so the marker provably cannot leak into the resolvable case. 180 tests pass. The one blocking defect is the MCP pre-flight.

### Referee Decisions

Both reviewers independently reported the same blocker; it is deduplicated to one finding below. I verified every accepted finding against the tree myself.

| 1 | MCP pre-flight reads a cwd-relative registry and false-rejects valid bot names | general + correctness (dedup) | Action Required | Valid | **Accept** | Independently reproduced: `valid_bot_names()` → `['primary']` from `cwd=/tmp`, → all five from the repo worktree. Confirmed `~/.claude.json`'s `mcpServers.lifeos` has no `cwd`, and that systemd sets `WorkingDirectory` for `lifeos-api`/`lifeos-mcp-http` so only the stdio path regresses. This is a genuine regression on a previously-working write path — and it is the *same class of bug as #575 itself*: an environment-dependent silent misresolution of a bot name. Remedy: **delete the MCP pre-flight**, let the route's 422 propagate. Verified `_call_api`'s `HTTPStatusError` branch surfaces the 422 body with the valid names and performs no write. Net effect removes code. |
| 2 | `_validate_schedule_bot` catches only `ValueError`, so a malformed registry fails the whole JSON-RPC call | correctness | Recommended | Valid | **Accept (subsumed by #1)** | Verified `config/settings.py:978` calls `entry.get(...)` unguarded. Deleting the pre-flight per #1 removes this code path entirely — no separate change needed. Addresser must confirm it is gone, not hardened. |
| 3 | `_misroute_notice` can abort the send it annotates, losing the notification | correctness | Recommended | Valid | **Accept** | Reproduced directly: with a malformed registry, `_misroute_notice('finance')` raises `AttributeError`; `_misroute_notice('')` returns cleanly. Because it is evaluated inside the f-string *before* `send_message_async`, both the primary and the failure-path send die and the notification is lost. This is newly-added code that violates the PR's own central invariant. Remedy: wrap the `is_known_bot` call in `try/except Exception: return ""`. Two lines. Keep the `not bot` short-circuit. |
| 4 | Rejection message and guide say "registered in `config/telegram_bots.json`" but token-less entries are silently skipped | general | Recommended | Valid | **Accept** | Verified at `config/settings.py:988` — an entry whose `token_env` is unset is dropped. The message this PR introduces therefore points a fresh-clone user at a file that already contains the name it says is invalid. Remedy: word the set as *configured* bots and note the token requirement in the guide. Message + docs only, no logic change. |
| 5 | Guide says only `notify`/`prompt` can name a bot; `endpoint` routes through it too | general | Minor | Valid | **Accept** | Verified in `_fire_entry`: the `agent` action returns early, and every other action's send passes `bot=entry.bot`. This matters more than its severity suggests — the schedule that triggered #575 is an `endpoint` action with `bot:: finance`, so the guide currently misdescribes the exact case this PR was written for. One-line docs correction. |
| 6 | Leading/trailing whitespace on `bot` rejected rather than trimmed | correctness | Minor | Valid | **Accept** | Concrete and confusing for an LLM-authored argument: `' finance'` is rejected with an error listing a visually identical `finance`. Remedy: `.strip()` on the write path only. Explicitly **not** case-folding — exact-match parity with `_resolve_bot` is deliberate and should stay. |
| 7 | Markdown/Obsidian path is unvalidated; guide could imply parity with the routes | correctness | Minor | Valid | **Accept** | The hand-edited `Inbox.md` is precisely how the orphaned `[bot:: advisor]` entry arose, so a reader inferring parity is a realistic and costly misread. Remedy: one clause noting markdown edits are unvalidated and the fire-time banner is the net. |
| 8 | Non-string `bot` interpolated raw into the error message | correctness | Minor | Valid | **Reject** | Cosmetic only — the value is correctly rejected and no downstream harm was demonstrated. Not worth a change in this PR. |
| 9 | Banner may push a near-4096-char notification into a second chunk | correctness | Minor | Unproven | **Reject** | The reviewer states nothing is dropped and that it is not worth changing. No failure demonstrated. |

**Findings forwarded to addresser:** 6 (items 1, 3, 4, 5, 6, 7 — item 2 is resolved by item 1's remedy)

nbramia:
## Round 1 review — addressed (1bfef94)

All six findings applied as stated. No finding rejected, none escalated. No existing test modified.

| # | Finding | Action | What changed |
|---|---------|--------|--------------|
| 1 | MCP pre-flight reads a cwd-relative registry and false-rejects valid names | **Applied** | Deleted `_validate_schedule_bot` and its `_call_api` call site — net removal, no replacement. The route's 422 propagates through the existing `HTTPStatusError` branch and renders as `Error: …`. Did **not** anchor the path to the repo root; that only half-fixes it (remote `LIFEOS_API_URL` has the file but not the token env vars). The `ValueError`-only `except` is gone with the helper rather than hardened. |
| 2 | `_misroute_notice` can abort the send it exists to annotate | **Applied** | Wrapped the `is_known_bot` lookup in `try/except Exception: return ""`. A raising registry now costs the banner, never the message. The `not bot` short-circuit inside `is_known_bot` is untouched, so the common empty-`bot` fire still does no registry I/O. Regression test added. |
| 3 | Rejection message and guide misstate the valid set | **Applied** | Message now reads `Configured names: … — a registry entry (config/telegram_bots.json) counts only once its token env var is set`. Guide says the same. Message + docs only. |
| 4 | Guide claims only `notify`/`prompt` can name a bot | **Applied** | Now: any schedule except `action:: agent`. |
| 5 | Whitespace on `bot` rejected rather than trimmed | **Applied** | `validate_bot_name` trims and returns the name to store; the routes persist that value, so `' ledger'` stores as `'ledger'` and resolves at fire time instead of passing validation and then misrouting. Case still not folded — `'Ledger'` remains rejected. |
| 6 | Guide could imply the markdown path is validated | **Applied** | One clause: markdown edits are not validated; the fire-time routing warning is the only safety net there. |

### Tests

```
tests/test_scheduler_bot_validation.py tests/test_mcp_server.py tests/test_scheduler_api.py
  → 82 passed

+ the 12-file scheduler/telegram/MCP set, -n auto --dist loadscope
  → 306 passed

pre-push hook, full unit suite
  → 3365 passed, 8 skipped  (+7 = the net new tests)
  → 12 passed server-free browser
```

New test file is now 36 tests (was 29): +7 net — a fire-path regression for finding 2 (send survives a raising registry), two route trim tests and two helper trim/case tests for finding 5, and the MCP class rewritten to assert the corrected contract (a 422 surfaces as the tool error; a name the *local* registry doesn't know still reaches the API — the exact regression the deleted pre-check caused).

### Deliberately not done

`docs/specs/product/api-reference.md` still words the valid set as the names "registered in `config/telegram_bots.json`" — the same imprecision finding 3 corrected elsewhere. Finding 3 scoped the rewording to the rejection message and the scheduler guide, so I left the API reference alone rather than widen the diff. Flagging it for the referee to fold in or wave off.

PR body has been rewritten to match what actually merged — the removal of the MCP pre-check and why, the file list, and the test counts.

nbramia:

**Reviewers run:** general

**Specialist rationale:** `review-correctness` was dropped this round. Its round-1 concern — the fire-path fail-open invariant — was resolved by the accepted guard, which I re-verified by direct execution before this round (`_misroute_notice('finance')` against a raising registry now returns `""` instead of propagating). The remaining delta was a net code deletion plus docs/message wording, which the general review covers.

### Reviewer Findings

#### General

**Action Required** — None.

**Recommended** — None.

**Minor**
1. `api/services/scheduler_store.py:750-757` — the guard is correct but narrower than the addresser's summary claimed ("a raising registry now costs the banner, never the message"). Reproduced against the current tree with a genuinely malformed registry: `bot=''` → `last_status='sent'`, `bot='finance'` → `last_status='failed'`, no notification delivered. The banner is no longer the cause; the cause is now `_resolve_bot(bot)` inside `send_message_async`, which sits *outside* that function's `try` and raises before the httpx loop. The new regression test passes only because it patches `is_known_bot` to raise rather than making the registry raise. The referee explicitly scoped the accepted remedy to the new code and flagged the `_resolve_bot` exposure as pre-existing parity with `main`, so no change is warranted in this PR — but the claim reads stronger than what holds.
2. `api/services/telegram.py:107-111` + `mcp_server.py:1055` — the reworded rejection message is 189 chars, so the 422 body is 202 and `_call_api`'s pre-existing `e.response.text[:200]` truncation clips it. Measured against the real registry only the closing `"}` is lost, and the valid names (which come first, at char ~67) are never at risk. But the token caveat is now the tail and is the first thing lost if one more bot is registered.
3. Answering the referee's scoping question on `api-reference.md:704`: the same pre-fix wording also survives at `mcp_server.py:751`, `:770`, `:824`, `api/routes/scheduler.py:44-45`, `api/routes/reminders.py:87-89`. Those five are MCP tool schemas and OpenAPI field descriptions aimed at an agent or API caller who cannot see the server's env vars anyway and who receives the accurate 422 at write time. `api-reference.md:704` is the only human-facing endpoint doc among them, and `:118` of that same file already states the token requirement correctly. The one accepted line is the right scope.
4. `api/services/telegram.py:104-106` — a whitespace-only `bot` (`"   "`) now trims to `""` and is accepted, where before this delta it 422'd. On `PUT` it lands in `updates` and silently clears an existing schedule's routing to primary. Defensible as the direct consequence of trimming plus "empty means primary", but it is a behavior change in the delta with no test covering it.

**Summary** — The delta is sound; no defect requiring action. All six accepted findings genuinely applied; nothing hardened in place of the required removal. Verified by execution: focused command **82 passed** (matching the PR body exactly), adjacent scheduler/telegram/reminders set **155 passed**, and `git diff --stat main...HEAD -- tests/` shows only the new file — **no existing test modified**. The MCP acceptance criterion still holds without the pre-check: both tools route to the validated routes, the 422 reaches the agent through `_call_api`'s `HTTPStatusError` branch with the valid names intact, and `_validate_schedule_bot` is gone with no residue. The rewritten MCP test class is stronger than what it replaced — it asserts the 422 propagation *and* adds the regression the deleted pre-check would have caused. **On the trim-and-return deviation: it is the better choice and introduces no inconsistency** — `POST`/`PUT` build their response from the persisted entry so a later `GET` cannot disagree; `_parse_entry_line` already `.strip()`s every inline field, so the literal round-1 remedy would have made the index hold `' ledger'` while a watcher reload yielded `'ledger'` — a banner on first fire and none after reload, a new divergence that trim-and-return removes. Privacy holds; fresh-clone/empty-registry covered on both write and fire paths. Not executed: the full repository suite and live-server verification.

### Referee Decisions

| 1 | Guard is narrower than the PR's stated claim; residual loss now sits in `_resolve_bot` | general | Minor | Valid | **Accept (wording only)** | I verified `_resolve_bot(bot)` is called at `api/services/telegram.py:303`, outside the `try` at `:310`, so a raising registry read propagates before any send — and that this call predates the PR, so it is not a regression here. The code needs no change. But an overstated claim in the PR record is a defect in its own right: a PR description must be true at merge time. Remedy: correct the claim to what the guard actually does. The pre-existing `_resolve_bot` exposure is filed separately as a follow-up rather than fixed here. |
| 2 | 200-char truncation clips the token caveat from the MCP-visible 422 | general | Minor | Valid | **Reject** | The concern is real but the proposed remedy makes things worse. Moving the caveat ahead of the names list would put the *names* — the content the issue's acceptance criterion actually requires — in the truncation zone, trading a harmless clip for one that breaks a criterion. The truncation is pre-existing in `_call_api` and out of this PR's scope. No demonstrated failure today. |
| 3 | Scope check on the `api-reference.md:704` fix | general | Minor | Valid | **Accept (as already scoped)** | Confirms the pre-accepted one-line fix is the right breadth and the other five sites should be left alone. Forwarded as the single line, no wider. |
| 4 | Whitespace-only `bot` silently clears routing to primary on `PUT`, untested | general | Minor | Valid | **Accept (test only)** | The semantics are defensible and I am not changing them — `bot=""` already means primary, and diverging would be scope creep. But this PR exists because a bot name failed *silently*, so leaving the one newly-introduced silent-clear path unpinned is the wrong omission to ship. Remedy: a test pinning the behavior. No logic change. |

**Findings forwarded to addresser:** 3 (items 1, 3, 4 — all wording/docs/test, no logic change)

**Convergence:** round 1 forwarded 6 accepted findings, round 2 forwards 3, none Action Required or Recommended. Converging; no scope guard fired.

nbramia:
## Review (round 2 completion — human-supervised)

The originating doctor session was killed mid round-2 for unrelated reasons (it was billing the Anthropic API — see #578 / #579), so this review was completed by hand rather than by the `/implement` reviewer. What was checked:

**Design.** Validate on *write*, fail open on *send*, detect at *fire time*. That split is right: rejecting a bad name where it's stored (`POST`/`PUT /api/scheduler`) prevents the class of bug, while `POST /api/scheduler/send` and `send_message_async` stay fail-open so a misconfigured name can never cost a notification. The `_misroute_notice` banner covers the path validation *can't* reach — a `[bot:: …]` hand-edited into `Scheduler/Inbox.md`, which is exactly how #575 happened. Worth being explicit: for the vault path this PR delivers detection, not prevention.

**Coverage of the send sites.** Both `send_message_async(bot=entry.bot)` calls in `SchedulerScheduler` (the normal send and the failure send) carry the banner; `grep` finds no third site. `worker.py`'s `session.bot` is a different concept and correctly untouched.

**Edge cases traced.**
- `UpdateScheduleRequest.bot` is `Optional[str] = None`, so an update that omits `bot` validates as None and is then dropped by the `if v is not None` filter — an unrelated field update can't clear or re-validate the bot. Correct.
- `valid_bot_names()` reads `settings.telegram_bots` per call, so a registry entry whose token env var is unset is *not* considered known — it gets rejected on write and bannered at fire time, matching `_resolve_bot`'s actual fallback behaviour rather than the file's contents.
- Trim-but-don't-case-fold matches `_resolve_bot`'s exact-match lookup; `'primary'` is accepted on write and is explicitly handled in `_resolve_bot`, so it round-trips.
- `_misroute_notice` swallows registry-read errors and returns `""` — the notice can never be what loses a message.

**Tests.** 36 new tests, meaningful assertions (rejection *and* that the store was left untouched, whitespace trimming asserted on the stored value, empty-registry cases, MCP-tool error surfacing, fire-time banner presence/absence). Full unit suite on this branch: **3365 passed, 8 skipped**.

**Nothing found that blocks merge.** Scope note for the record: the goal as approved was broader than the one-line vault fix that prompted it, and this landed at 8 files — over the doctor's own "stop at 4+ files" guardrail, which it did not honour. The code is sound, so it merges; the guardrail gap is a separate matter.

PR #638, opened 2026-08-21
feat: add provider and model registry

fredrikblau:
## Summary

- add a provider/model registry behind the existing LLM client contract
- support configurable OpenAI-compatible providers with environment-backed API keys
- add named profiles for default, fast, specialist, and private operations
- migrate specialist, CRM tone, preflight, and per-turn model selection away from hard-coded clients
- preserve legacy `LIFEOS_LLM_BACKEND`, Anthropic, and local llama-server behavior when registry settings are absent
- add configuration documentation, ADR-021, and regression tests

## Configuration

Use `LIFEOS_LLM_PROVIDERS` and `LIFEOS_LLM_MODELS` for additional providers. API keys are referenced with `api_key_env` and are not stored in configuration JSON.

## Test plan

- `python -m compileall -q api config scripts tests` ✅
- `git diff --check` ✅
- Full pytest could not run in the audit environment because Python has no `pip`/`httpx`; pytest fails during conftest initialization before collecting tests.

## Compatibility

No personal data schemas or storage formats are changed. Existing deployments without the new settings retain the historical backend selection behavior.


fredrikblau:
Superseded by current-main-based PR #667, which includes the provider registry plus the Telegram-first personal assistant capture, inbox review, provenance, and voice transcription work. Closing this older branch to avoid duplicate implementation.

fredrikblau:
Superseded by #667, which is based on current upstream main.

PR #675, opened 2026-08-25
fix(imessage): close SQLite connections in the export path and stop misreporting fd exhaustion as missing Full Disk Access

### LINKED ISSUE #647: fix(imessage): _insert_batch leaks a sqlite connection per batch, so a first full export dies at ulimit and misreports as missing Full Disk Access
## Problem

`IMessageStore._insert_batch` opens a SQLite connection per batch and never closes it:

```python
def _insert_batch(self, batch: list) -> None:
    """Insert a batch of messages."""
    with sqlite3.connect(self.storage_path) as conn:
        conn.executemany(...)
```

`with sqlite3.Connection` is a **transaction** context manager, not a closing one — it commits on exit and leaves the connection (and its fd) open. One fd leaks per batch.

`_clear_data()` immediately below has the same shape.

## Why it stayed hidden

Incremental nightly exports move a few hundred messages, so the leak never reaches the limit. It only bites on a **first full export on a new host**, where `ROWID > 0` walks the entire history. On macOS the default is `ulimit -n` = **256**, so a 300k-message export exhausts fds long before it finishes.

## Reproduction

Bootstrapping the MacBook as an export agent (307,492 messages in `chat.db`):

```
$ ~/.venvs/lifeos-export/bin/python scripts/apple_data_export.py --execute --source imessage
[INFO]    Starting iMessage export from ROWID > 0
[WARNING] Cannot access Messages.db — grant Full Disk Access to Terminal
OSError: [Errno 24] Too many open files: '.../data/apple-exports/manifest.json'
```

Underlying cause, with the real traceback:

```
File "api/services/imessage.py", line 406, in _insert_batch
sqlite3.OperationalError: unable to open database file
```

With `ulimit -n 65536` the identical command succeeds in 2.7s and exports 306,520 messages. That is the only thing that changed.

## Second defect — the error message actively misleads

`scripts/apple_data_export.py:399-407` catches `PermissionError` and reports *"Cannot access Messages.db — grant Full Disk Access to Terminal"*. But `imessage.py:390` raises `PermissionError` from any `sqlite3.OperationalError` in the export loop — including this fd exhaustion on the **destination** database, which has nothing to do with FDA or with `chat.db`.

FDA was correctly granted throughout. Verified independently:

```
os.access(~/Library/Messages/chat.db, R_OK) -> True
sqlite3 query on chat.db          -> 307,492 messages
```

So the operator is told to grant a permission they already have, while the actual fault (fd limit) is invisible. This cost most of the debugging time during today's recovery.

## Acceptance criteria

- [ ] `_insert_batch` and `_clear_data` close their connections — ideally hoist one connection for the whole export rather than reconnecting per batch.
- [ ] A full export from `ROWID > 0` over a ~300k-message database completes under the default macOS `ulimit -n` of 256.
- [ ] `sqlite3.OperationalError` is no longer blanket-translated to `PermissionError`; FDA is reported only when the *source* database is genuinely unreadable, and the underlying error is preserved otherwise.
- [ ] Regression test asserting connections are not leaked across batches (e.g. batch count well above a low `ulimit`, or assert on open-fd count).

## Notes

Found while bootstrapping the MacBook as a stand-in export agent during the Mac Mini outage (#646). This is a hard blocker on standing up any new export host, which is exactly what you need to do when the primary agent dies.

nbramia:
Closes #647

## TL;DR

Copying iMessage history over to the server could fail partway through a first full run, and then blame the wrong thing entirely. Every batch of messages written during an export opened a database connection that was never closed, so a large history ran the process out of file handles and stopped. The failure was reported as a missing macOS Full Disk Access permission, which sent anyone debugging it chasing a permission that had been granted all along. This change closes those connections, reports the real error when the write side fails, and flushes pending writes before the database is copied so the copy can no longer ship without the newest messages.

## Problem

`with sqlite3.connect(...)` is a **transaction** context manager, not a closing one — it commits on exit and leaves the connection (and its file descriptors) open. `IMessageStore._insert_batch` used that shape, so a full export leaked one connection per 1000-row batch.

Incremental nightly exports move a few batches, so it never tripped. A *first full export* (`ROWID > 0`) over a ~300k-message `chat.db` blows through macOS's default `ulimit -n` of 256 and dies partway.

The failure was then misreported. `export_from_source` wrapped the whole export loop in `except sqlite3.OperationalError` and translated it into `PermissionError`, which `apple_data_export.py` renders as *"Cannot access Messages.db — grant Full Disk Access to Terminal"*. So fd exhaustion on the **destination** database was reported as a permissions problem on the **source**. FDA was correctly granted the whole time.

## Changes

**`api/services/imessage.py`**

- `export_from_source` now opens **one** destination connection for the whole export and passes it to each batch insert. That is both the fix and a performance win — no reconnect per batch.
- `_insert_batch` and `_set_last_synced_rowid` take an optional caller-owned `conn`, and open-and-close their own only when called standalone.
- Every connection in the export path is wrapped in `contextlib.closing` so it is deterministically closed: `_ensure_schema`, `_get_last_synced_rowid`, `_set_last_synced_rowid`, `_insert_batch`, `_clear_data`.
- Source-database reads moved into a `_iter_source_rows` generator that owns the source connection **and its error handling**. A generator's `except` clause is only live while the generator itself is running, so an `OperationalError` raised by the consumer's loop body — i.e. while writing to the destination — can no longer be caught and relabelled as an FDA problem. The `PermissionError` translation now applies only to failures opening/reading `chat.db`, and the message includes the path and the underlying error.
- New `checkpoint()` method — see WAL section below.

**`scripts/apple_data_export.py`**

- The `PermissionError` handler now includes the underlying error in the log line instead of discarding it.
- Calls `store.checkpoint()` before copying the database.

## The WAL bug (same root cause — included deliberately)

`export_imessage` copies `data/imessage.db` with `shutil.copy2`, which takes only the main database file. In WAL mode recent writes can still be in the `-wal` sidecar, so the copy silently shipped a stale database — observed in production at 306,309 of 307,083 messages.

This is the same defect: SQLite folds the WAL back in automatically when the *last* connection closes, and the leaked per-batch connections meant that never happened. Fixing the leak repairs it implicitly, **but only while nothing else holds the database open** — the API server does, and then the copy silently goes stale again. Verified: with one bystander connection open, the copied database had 0 of 5000 rows without an explicit checkpoint and 5000 with it. Relying on the implicit behaviour is a silent-data-loss trap, so the checkpoint is explicit.

## Tests

New `TestExportConnectionLifecycle` in `tests/test_imessage.py`, all Linux-safe — they build a synthetic `chat.db`-shaped source in a tmpdir and point the store at a temp destination. No real `chat.db` required, obviously synthetic data throughout (`+15550001111`, `"synthetic message N"`).

| Test | Asserts |
|---|---|
| `test_insert_batch_does_not_leak_fds` | 200 batches don't accumulate 200 connections |
| `test_clear_data_does_not_leak_fds` | 50 `_clear_data` calls leak nothing |
| `test_full_export_does_not_leak_fds` | A 30-batch export holds one destination connection |
| `test_full_export_survives_a_low_fd_limit` | The real failure mode — a full export under a deliberately tightened `RLIMIT_NOFILE` |
| `test_destination_failure_is_not_reported_as_full_disk_access` | A write-side `OperationalError` surfaces as itself, not as `PermissionError` |
| `test_unreadable_source_still_reports_full_disk_access` | Guards the over-correction: FDA guidance is *still* given when `chat.db` is genuinely unreadable |
| `test_checkpoint_flushes_wal_so_a_file_copy_is_complete` | A single-file copy is complete, with a bystander connection held open to suppress the implicit checkpoint |

Each was verified to actually catch the bug: 5 of the 6 leak/misreporting tests fail against the pre-fix code and pass after. The WAL test fails without the checkpoint call (19,000 of 30,000 rows — the same "missing newest messages" symptom seen in production).

## Test results

`./scripts/test.sh`: **4801 passed, 12 skipped, 1 failed**.

The one failure is `tests/test_mcp_server.py::TestOpenAPIAvailability::test_openapi_spec_available` — pre-existing flakiness, not from this change. `test_mcp_server.py` runs against the shared live `lifeos-api` service; re-running it three times with an unchanged tree produced 1, 4, and 4 failures on *different* tests each time, and the test passes in isolation. This PR touches no search, MCP, or OpenAPI code.

Pre-push hooks ran clean: 3797 unit tests passed, 74 server-free browser tests passed.

## Noticed but deliberately not changed

`api/services/imessage.py` has ~11 more `with sqlite3.connect(...)` blocks in the **query** paths (`search_messages`, `get_messages_for_person`, etc., around lines 475–943, plus one module-level function at 943). They have the identical no-close shape. They're left alone here to keep this PR scoped to the export path in #647 — the risk profile is different (one connection per API request in a long-running server, rather than thousands in a tight loop), so it deserves its own issue and its own testing rather than being folded in silently.

### DIFF
--- a/api/services/imessage.py
+++ b/api/services/imessage.py
@@ -11,6 +11,7 @@ import logging
 import re
 import sqlite3
 import uuid as uuid_mod
+from contextlib import closing
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
@@ -217,7 +218,7 @@ class IMessageStore:
         """Create the local database schema if it doesn't exist."""
         self.storage_path.parent.mkdir(parents=True, exist_ok=True)
 
-        with sqlite3.connect(self.storage_path) as conn:
+        with closing(sqlite3.connect(self.storage_path)) as conn, conn:
             conn.execute("PRAGMA journal_mode=WAL")
             conn.executescript("""
                 CREATE TABLE IF NOT EXISTS messages (
@@ -248,23 +249,34 @@ class IMessageStore:
 
     def _get_last_synced_rowid(self) -> int:
         """Get the last synced ROWID for incremental sync."""
-        with sqlite3.connect(self.storage_path) as conn:
+        with closing(sqlite3.connect(self.storage_path)) as conn:
             cursor = conn.execute(
                 "SELECT value FROM sync_state WHERE key = 'last_rowid'"
             )
             row = cursor.fetchone()
             return int(row[0]) if row else 0
 
-    def _set_last_synced_rowid(self, rowid: int) -> None:
-        """Update the last synced ROWID."""
-        with sqlite3.connect(self.storage_path) as conn:
-            conn.execute(
-                """
-                INSERT OR REPLACE INTO sync_state (key, value)
-                VALUES ('last_rowid', ?)
-                """,
-                (str(rowid),),
-            )
+    def _set_last_synced_rowid(
+        self, rowid: int, conn: Optional[sqlite3.Connection] = None
+    ) -> None:
+        """Update the last synced ROWID.
+
+        Args:
+            rowid: New high-water mark.
+            conn: Optional caller-owned connection to reuse. When omitted a
+                connection is opened and closed for this call.
+        """
+        sql = """
+            INSERT OR REPLACE INTO sync_state (key, value)
+            VALUES ('last_rowid', ?)
+        """
+        if conn is not None:
+            with conn:
+                conn.execute(sql, (str(rowid),))
+            return
+
+        with closing(sqlite3.connect(self.storage_path)) as own_conn, own_conn:
+            own_conn.execute(sql, (str(rowid),))
 
     def export_from_source(
         self,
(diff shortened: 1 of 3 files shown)


nbramia:
Independently verified this branch (not taken from the implementing agent's report):

**Regression tests genuinely catch the bug.** Reverted `api/services/imessage.py` to `main` and re-ran the new tests — all 4 fail against the buggy code:

```
FAILED tests/test_imessage.py::TestExportConnectionLifecycle::test_insert_batch_does_not_leak_fds
FAILED tests/test_imessage.py::TestExportConnectionLifecycle::test_clear_data_does_not_leak_fds
FAILED tests/test_imessage.py::TestExportConnectionLifecycle::test_full_export_does_not_leak_fds
FAILED tests/test_imessage.py::TestExportConnectionLifecycle::test_full_export_survives_a_low_fd_limit
```

and all 41 pass against the fix. That is the property a regression test needs and is easy to get wrong.

**Full unit suite green on this branch:** `4802 passed, 12 skipped in 727.50s`.

**The generator refactor is the right call.** Isolating the source read into its own generator means a `sqlite3.OperationalError` raised while writing the *destination* can no longer be caught by the source's `except` and mistranslated into a Full Disk Access complaint. That addresses the misreporting at the structural level rather than by widening a message string.

**WAL checkpoint correctly folded in.** I hit this live during the Aug 24 recovery: `shutil.copy2` shipped an `imessage.db` missing the 774 newest messages (306,309 vs 307,083) because they were still in the uncheckpointed `-wal`, and the export reported `ok`. `store.checkpoint()` before the copy fixes it, and it shares the connection-lifecycle root cause, so it belongs here rather than in a separate PR.

Real-world confirmation of the original bug: the first full export on the MacBook died at `[Errno 24] Too many open files` under the default `ulimit -n` of 256; with `ulimit -n 65536` the identical command exported 306,520 messages in 2.7s. This PR removes the need for that workaround.

nbramia:
## Verification evidence

```
verification-head: a45db87ea600bb6c8e8b117673f9229f4bd4231f
suite-result:      pass
suite-command:     ./scripts/test.sh
suite-executions:  1
suite-exit-status: 0
```

`4802 passed, 12 skipped, 989 warnings in 727.50s`, exit code 0.

Run by me (the reviewing session), not the implementing agent. Two notes on the evidence, both verified rather than assumed:

**Head match.** The suite ran while `fix/647-imessage-connection-leak` was checked out; the branch was subsequently amended from `fec147c` to `a45db87`. That amend changed the commit date only — both commits point at tree `e531e26bfe577685783f3d484ec108d76d036af5`, confirmed with `git diff --quiet fec147c a45db87`. The tested content is therefore byte-identical to the current head.

**Environment prefix.** The command was invoked as `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ./scripts/test.sh` — environment variables only, no wrapper, no extra arguments, no additional command. AGENTS.md sanctions this form to force CPU embeddings and avoid the known GPU SDMA lockup; it does not alter which tests run.

**On the differing result in the PR body.** The description reports `1 failed, 4801 passed` from the implementing agent's run. My run of the same tree was clean. Both are consistent with the documented shared-`lifeos-api` contention: `test_mcp_server.py` exercises the live service that every agent on this host shares, and re-runs against an unchanged tree produced different failing tests each time. The failure is environmental, not from this change, and it did not reproduce here.

PR #843, opened 2026-08-29
fix(api): keep validation errors serializable

### LINKED ISSUE #577: fix(api): custom validation failures return a bare 500 instead of the intended error message
## Problem

Several API endpoints reject bad input with a carefully written explanation — unsupported file type, empty question, too many attachments — and none of those messages ever reach the caller. The request fails with a bare `500 Internal Server Error` and an empty body instead.

The endpoints affected are core, user-facing ones. Sending a whitespace-only query to the search endpoint returns 500. Sending a whitespace-only question to the ask endpoint returns 500. Attaching a file of an unsupported type to the streaming ask endpoint returns 500, rather than the message naming which types are allowed. All three were confirmed against the running server on `main` before this issue was filed.

Two things make this worse than a wrong status code. A 500 tells an API client the server broke, so a well-behaved client retries — the request will never succeed no matter how many times it is sent, because the input is invalid. And the message explaining *what* was wrong with the input is discarded, so the caller has nothing to correct. The streaming ask endpoint is part of the documented contract that the web, voice, and Telegram clients depend on, so this is a contract violation, not just a rough edge.

The trigger is narrow but ordinary: a validation rule expressed as custom logic fails, as opposed to one expressed as a declarative field constraint. Declarative constraints — a minimum length, a required field — return a clean 400 today. Only the custom rules break, which is why the failure has stayed hidden: the common empty-string case takes the working path, and the whitespace-only case one character away takes the broken one.

## Solution

The global validation-error handler becomes able to serialize every validation failure it is given, including those produced by custom validation logic.

The handler currently normalizes each failure into a JSON body. It handles one known non-serializable value in that structure, but a failure raised by custom logic carries the raised exception itself in its context, and that object cannot be JSON-encoded. Encoding it is attempted after the response status has already been chosen, so the failure surfaces as an unhandled server error rather than as the validation response the handler was building.

The fix is to normalize the whole failure structure rather than one known field of it, coercing any value that is not JSON-encodable to its string form. The caller then receives the intended status and the validation message, whichever style of rule produced it. Behavior for failures that already serialize cleanly must not change.

Deciding whether these should be 400 or 422 is worth settling at the same time, since the handler currently returns 400 for every validation failure regardless of kind. Whatever is chosen must be applied consistently and documented, because callers cannot distinguish the two paths today.

## Acceptance Criteria

- [ ] Given a request model whose custom validation rule raises on bad input, when that endpoint receives such input, then the response status is a 4xx and not 500.
- [ ] Given that same request, the response body contains the validation message produced by the rule, so the caller can tell what to correct.
- [ ] When `POST /api/search` receives a query consisting only of whitespace, the response is a 4xx carrying the empty-query message — not a 500.
- [ ] When `POST /api/ask` receives a question consisting only of whitespace, the response is a 4xx carrying the empty-question message — not a 500.
- [ ] When `POST /api/ask/stream` receives an attachment whose media type is not in the allowed set, the response is a 4xx whose body names the allowed types — not a 500.
- [ ] When `POST /api/ask/stream` receives more attachments than the configured maximum, the response is a 4xx carrying the limit message — not a 500.
- [ ] When `POST /api/search` receives an empty-string query, the response remains a 400 with the existing body shape — the currently-working declarative path must not regress.
- [ ] When a request omits a required field entirely, the response remains a 4xx with the existing body shape.
- [ ] The validation handler shall serialize a failure whose context contains a non-JSON-encodable value without raising.
- [ ] No response body produced by the handler contains a Python object repr in place of a human-readable message.

## Verification

- **Automated tests:** Each endpoint criterion maps to an API test posting the exact input named and asserting both the status and the presence of the intended message substring. The handler-level criterion is covered by a unit test constructing a validation failure whose context holds a non-encodable value and asserting the handler returns a response rather than raising. The two regression criteria are covered by the existing tests, which must pass unmodified.
- **Manual checks:** None required.
- **Existing test suite:** Must continue to pass. `tests/test_search_api.py`, `tests/test_ask_api.py`, and `tests/test_chat_api.py` are the relevant files.

## Scope

**In scope:**
- Making the validation-error handler serialize any validation failure, including those from custom rules.
- Settling and documenting the status code returned for validation failures.
- Test coverage for the whitespace-only and attachment cases that are currently unprotected.

**Out of scope:**
- Rewriting the validation rules themselves. They are correct; only their delivery is broken.
- Adding validation to endpoints that lack it.
- Any change to the search, ask, or chat business logic.

## Technical Context

- **Key files:**
  - `api/main.py:300` — `validation_exception_handler`, registered for `RequestValidationError`. Line 309 sanitizes a bytes value under the `input` key, which is the only non-serializable case currently handled; the `ctx` key is left untouched. The handler returns 400 at both line 314 and line 319.
  - `api/routes/search.py:61` — `query_not_empty`, reachable with a whitespace-only query because the field's minimum-length constraint passes first.
  - `api/routes/ask.py:42` — `question_not_empty`, same shape and same reachability.
  - `api/routes/chat.py:287` — `validate_media_type`, the confirmed attachment case. Also `api/routes/chat.py:339` (`validate_persona`) and `api/routes/chat.py:348` (`validate_attachments`), which share the failure shape but were not individually probed.
- **Reproduction:** against a running server, `curl -s -o /dev/null -w '%{http_code}' -X POST localhost:8000/api/search -H 'Content-Type: application/json' -d '{"query":"   "}'` prints `500`. Substituting `{"query":""}` prints `400` — the one-character difference between the broken and working paths.
- **Constraints:** No new dependencies. The handler is global, so any change affects every endpoint's validation responses — the two regression criteria exist to bound that blast radius. `/api/ask/stream` is a documented client-surface contract; consult `docs/specs/technical/client-surfaces.md` before changing its status codes.

## References

- `docs/specs/technical/client-surfaces.md` — the contract governing `/api/ask/stream`, one of the affected endpoints.
- `docs/specs/product/api-reference.md` — documents the endpoints whose error responses change.

## History

Found while implementing #575. That issue's acceptance criteria called for a specific validation status code, which forced the check into the route body rather than a declarative validator — a custom validator there would have returned 500 for exactly this reason. The workaround is sound and shipped in that PR; this issue is the underlying cause, left unfixed there deliberately as out of scope.

Note on why this survived: `tests/test_search_api.py:42` asserts a 400 for an empty-string query and passes, which reads as coverage of the empty-query rule. It exercises only the declarative minimum-length constraint. The custom rule beneath it — the one that exists specifically to catch whitespace — has no test and has been returning 500 the whole time. The missing whitespace case is part of this defect, not incidental to it; a fix that does not add it leaves the same blind spot in place.

### LINKED ISSUE #755: Agent-worker preflight tests read the host's real .env (LIFEOS_AGENT_DEFAULT_ROUTE)
## Problem

On a machine where `.env` sets `LIFEOS_AGENT_DEFAULT_ROUTE` (nathan-linux has `claude_code`), 13 tests fail on clean `main`:

- `tests/test_agent_worker_preflight.py` — `test_preflight_ambiguity_populated`, `test_preflight_routing_ask_when_unspecified`, `test_genuine_missing_referent_ambiguity_still_blocks`, `test_preflight_invalid_routing_value_falls_back_to_ask`, `test_default_route_unset_is_byte_identical_to_today`
- `tests/test_agent_worker_clarifications.py` — 4 tests
- `tests/test_agent_worker_loop.py` — 3 tests
- `tests/test_operator_spawn.py` — `test_ambiguous_preflight_parks_for_routing`

Example:

```
test_preflight_ambiguity_populated
>       assert result.ambiguity is not None
E       AssertionError: assert None is not None
```

The preflight result still carries the ambiguity in `raw`, but `_apply_default_route` demoted it to advisory (#751 behavior) because `settings.agent_default_route` was truthy — read from the ambient `.env`, not set by the test.

## Confirmation

```
$ LIFEOS_AGENT_DEFAULT_ROUTE= pytest tests/test_agent_worker_preflight.py \
    tests/test_agent_worker_clarifications.py tests/test_agent_worker_loop.py \
    tests/test_operator_spawn.py -q
166 passed
```

Introduced by the #751/#752 default-route work — the tests assume `settings.agent_default_route` is empty rather than pinning it.

## Acceptance criteria

- These tests pass regardless of what `LIFEOS_AGENT_DEFAULT_ROUTE` is set to in the developer's `.env`.
- Each test that depends on default-route behavior pins `settings.agent_default_route` explicitly (fixture/monkeypatch), including the "unset" cases.
- No test in the suite reads `agent_default_route` from the ambient environment.

## Why it matters

The pre-push hook runs the unit suite, so on this host every push is red for reasons unrelated to the change being pushed — which trains people to reach for `--no-verify`.

fredrikblau:
## Summary

- keep agent-worker tests independent from LIFEOS_AGENT_DEFAULT_ROUTE in the developer environment
- recursively normalize validation error details so custom validators return their intended 400 response
- assert that whitespace validation responses retain their user-facing messages

## Testing

- Not run locally: the configured LifeOS virtualenv is unavailable and the system environment lacks `python-frontmatter`.
- `git diff --check`

Fixes #755
Fixes #577

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -38,6 +38,7 @@ from dotenv import load_dotenv
 load_dotenv(Path(__file__).resolve().parent.parent / ".env")
 
 import logging
+import json
 import socket
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
@@ -335,13 +336,20 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
     """Convert validation errors to 400 with clear messages."""
     errors = exc.errors()
 
-    # Sanitize errors for JSON serialization (convert bytes to string)
-    sanitized_errors = []
-    for error in errors:
-        sanitized = dict(error)
-        if "input" in sanitized and isinstance(sanitized["input"], bytes):
-            sanitized["input"] = sanitized["input"].decode("utf-8", errors="replace")
-        sanitized_errors.append(sanitized)
+    def json_safe(value):
+        if isinstance(value, bytes):
+            return value.decode("utf-8", errors="replace")
+        if isinstance(value, dict):
+            return {str(key): json_safe(item) for key, item in value.items()}
+        if isinstance(value, (list, tuple)):
+            return [json_safe(item) for item in value]
+        try:
+            json.dumps(value)
+        except (TypeError, ValueError):
+            return str(value)
+        return value
+
+    sanitized_errors = [json_safe(error) for error in errors]
 
     # Check if this is an empty query error
     for error in errors:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -105,6 +105,21 @@ def pytest_configure(config):
     _install_anthropic_request_guard()
 
 
+@pytest.fixture(autouse=True)
+def isolate_agent_default_route(request, monkeypatch):
+    """Keep agent-worker tests independent from the developer's .env."""
+    agent_worker_tests = {
+        "test_agent_worker_preflight.py",
+        "test_agent_worker_clarifications.py",
+        "test_agent_worker_loop.py",
+        "test_operator_spawn.py",
+    }
+    if request.node.fspath.basename in agent_worker_tests:
+        from config.settings import settings
+
+        monkeypatch.setattr(settings, "agent_default_route", "")
+
+
 # ---------------------------------------------------------------------------
 # Unmarked-test guard (#682)
 #
--- a/tests/test_chat_api.py
+++ b/tests/test_chat_api.py
@@ -41,6 +41,7 @@ class TestAskStreamEndpoint:
         """Should return 400 for whitespace-only question."""
         response = client.post("/api/ask/stream", json={"question": "   "})
         assert response.status_code == 400
+        assert "Question cannot be empty" in str(response.json()["detail"])
 
     def test_stream_returns_event_stream(self, client):
         """Response should be text/event-stream."""
--- a/tests/test_search_api.py
+++ b/tests/test_search_api.py
@@ -182,6 +182,13 @@ class TestSearchRequestValidation:
         )
         assert response.status_code in [400, 415, 422]
 
+    def test_whitespace_query_returns_validation_message(self, client):
+        response = client.post("/api/search", json={"query": "   "})
+
+        assert response.status_code == 400
+        assert response.json()["error"] == "Query cannot be empty"
+        assert "Query cannot be empty" in str(response.json()["detail"])
+
     def test_handles_extra_fields(self, client):
         """Should ignore extra fields gracefully."""
         response = client.post("/api/search", json={


fredrikblau:
Closing this combined proposal so the two unrelated fixes can be submitted as focused PRs.

PR #880, opened 2026-09-04
fix: cache hydrated people list

### LINKED ISSUE #869: fix(crm): cache the hydrated people list across requests, invalidated by SQLite data_version
## Problem

Listing, searching, or paging people costs about a second per request because the server rebuilds every one of the 14,476 person records from SQLite on every call and then throws away all but the page requested. Fourteen different CRM operations start with this same full load (people list and search, birthdays, Me dashboard, Family dashboard, network graph, data health, and the MCP people search), and several call it two or three times in one request. Typing in the people search box fires one of these per keystroke pause, so search feels sluggish and results can arrive out of order. The MCP `lifeos_people_search` tool pays the same cost on every call.

| Step | Time |
|---|---|
| Raw `SELECT *` of all people | 39 ms |
| Turning rows into person objects (JSON decode of 7 list columns + date parsing, 14,476 times) | 320 to 620 ms |
| Everything else in the list endpoint (filter, sort, page, serialize 50) | under 50 ms |

## Solution

The person entity store keeps a process-wide cache of the fully hydrated person list and returns it instantly on subsequent calls. The cache is keyed on SQLite's data-version counter, so any commit to the CRM database from any process (the API's own edits, merges, hides, or the nightly sync in its separate process) invalidates it on the next read with no explicit invalidation plumbing. Callers receive a fresh list object each time so in-place sorting or filtering by one handler cannot leak into another. Every operation that begins with the full load speeds up at once, including the MCP people search, without touching their logic.

## Acceptance Criteria

- [ ] Given the CRM database is unchanged, when `PersonEntityStore.get_all()` is called twice, the second call shall complete in under 5 ms and return equal contents.
- [ ] Given a row in `person_entities` is updated by a separate SQLite connection (simulating the sync process), when `get_all()` is next called, it shall return the updated data.
- [ ] Given a person is added, updated, hidden, or merged through the store's own write methods, when `get_all()` is next called, it shall reflect the write.
- [ ] `get_all(include_hidden=True)` and `get_all(include_hidden=False)` shall be cached independently and never return each other's contents.
- [ ] When a caller sorts or mutates the returned list, a subsequent `get_all()` call shall return the original order and membership.
- [ ] `GET /api/crm/people?limit=50&sort=strength` shall complete in under 150 ms on a warm cache with the production dataset.
- [ ] The cache shall be safe to read from multiple threads concurrently (it will be reached from threadpool handlers).

## Verification

- **Automated tests:** unit tests in `tests/test_person_entity.py` for warm-hit timing, cross-connection invalidation via `PRAGMA data_version`, invalidation after each write method, and list-copy isolation. `tests/test_crm_api.py` gains a warm-cache latency assertion on the people list.
- **Manual checks:** type a name in the CRM search box; results should update within the debounce window.
- **Existing test suite:** `tests/test_person_entity.py`, `tests/test_crm_api.py`, `tests/test_merge_people.py`, `tests/test_relationship_discovery.py` must pass.

## Scope

**In scope:**
- The cache inside `PersonEntityStore.get_all()` and its invalidation.
- A cheap `count`/`sum` style helper is NOT required here; the Me/Family issue handles aggregates.

**Out of scope:**
- Pushing list filtering, sorting, and paging into SQL. The fuzzy token-prefix search semantics would change; revisit only if the cached path is still too slow.
- Caching person detail, timeline, or facts endpoints (already fast).

## Technical Context

- **Key files:** `api/services/person_entity.py:1057` (`get_all`), `:633` (`_row_to_entity`, the hot spot), `:342` (`from_dict`), `:588` (`_get_connection`, new connection per call), `:1155` (singleton getter). Callers: `api/routes/crm.py` lines 555, 571, 619, 2172, 2416, 2420, 2949, 3003, 3018, 3070, 3075, 3621, 3801, 3855, 4688; `api/routes/people.py:163` and `:225`; `api/services/person_stats.py:89`, `api/services/relationship_metrics.py:415`.
- **Invalidation mechanism:** SQLite `PRAGMA data_version` (available in the venv's SQLite 3.45) changes whenever another connection commits. Keep one long-lived read-only connection on the store purely for this check; compare its value before serving from cache. Also bump or drop the cache in `add`, `update`, `delete`, `hide_person`, and `save` (`person_entity.py:839` to `:930`) so same-connection writes are covered.
- **Mutation hazards to audit:** `api/routes/crm.py:675` sorts the list in place; return `list(cached)` so this is harmless. Entities themselves are shared; grep callers of `get_all()` for attribute writes on returned entities and switch those to `get_by_id` copies if any exist.
- **Constraints:** no new dependencies; a `threading.Lock` around cache refresh is sufficient.

## References

- Concurrency issue (event-loop blocking) — this cache is reached from threadpool handlers once that lands.
- `AGENTS.md` § Key Concepts (two-tier data model).

nbramia:
## TL;DR

Caches the hydrated `PersonEntityStore.get_all()` result behind SQLite `PRAGMA data_version` plus an in-process generation counter so unchanged CRM people-list reads avoid rehydrating ~14k people on every request, and batches the per-page category source-entity lookup into one query instead of one per returned person.

Closes #869

## Design

`get_all()` now keeps separate cache entries for the `include_hidden` and `include_merged` variants. A long-lived SQLite connection reads `PRAGMA data_version` to detect commits made by other processes, while store write methods and merged-ID reloads bump a local generation counter for in-process invalidation. Each caller receives a fresh list wrapper so handler-level sorting/filtering cannot mutate the cached list membership.

Because cached `PersonEntity` objects are shared, service write paths that mutate entities obtained from `get_all()` must refetch (by ID, or by a private `to_dict()`/`from_dict()` copy) before writing, so no mutation is ever visible to another reader of the cache before `store.update()` actually persists it. The CRM people-list helper passes `None` (not `[]`) to `compute_person_category()` on the `include_related=False` path, preserving that function's existing internal source-entity fallback fetch — the fast list path's computed category is unchanged from before this PR.

That fallback fetch is itself a per-person `SourceEntityStore` query, so restoring it on the list path reintroduced a real cost that scales with page size: `GET /api/crm/people` and `GET /birthdays/today` now use `SourceEntityStore.get_for_people_batch()` to fetch the whole page's source entities in one round trip instead of one `SourceEntityStore` query per person.

## Implementation Notes

- Added the process-local hydrated people cache in `api/services/person_entity.py`, including write invalidation, merged-ID invalidation, cross-connection invalidation, and thread locking.
- `api/routes/crm.py`: `_person_to_detail_response()` passes `None` to `compute_person_category()` when `include_related=False` (not `[]`), so the fast list/`birthdays/today` path keeps triggering `compute_person_category()`'s internal source-entity fallback fetch exactly as before this PR. It now also accepts an optional `category_source_entities` list so callers can supply a pre-fetched (batched) list instead of letting it fetch per-person; single-person callers are unaffected (parameter defaults to `None`, unchanged behavior).
- `api/services/source_entity.py`: added `SourceEntityStore.get_for_people_batch(canonical_person_ids, limit_per_person)`, which fetches many people's source entities in a handful of round trips (chunked at 300 IDs/statement) instead of one call per person. It's built as a compound `UNION ALL` of per-person `ORDER BY observed_at DESC LIMIT N` subqueries, each independently planned by SQLite — **not** a single `WHERE canonical_person_id IN (...)` query. That shape was tried first (both with a Python-side group/sort and with a `ROW_NUMBER() OVER (PARTITION BY ...)` window function, the latter also tested with an experimental composite index) and measured **slower**, because SQLite has no per-group "top-K" optimization: a windowed/grouped query must fully rank every matching row in a partition before any per-person cap can be applied, and on the real dataset a majority of the highest-relationship-strength people (the CRM UI's default sort) have 10,000-65,000+ source entities each. The `UNION ALL` shape lets SQLite apply its efficient single-query top-N-via-heap plan to each person independently, which is what actually pays off here.
- `api/routes/crm.py`: `list_people()` and `get_todays_birthdays()` now call `get_for_people_batch()` once per request (scoped to the current page, after pagination) instead of `_person_to_detail_response()` triggering a `get_for_person()` call per person.
- The batched category fetch caps at `_CATEGORY_BATCH_SOURCE_LIMIT = 50` source entities per person, not the single-person path's 500 (which is unchanged). `compute_person_category()` only needs to find one qualifying source entity, not all of them up to some cap, so this is a deliberate, verified reduction: checked against the **full** real dataset (14,414 people, not a sample), reducing the cap all the way down to 20 still produces IDENTICAL final categories vs the `None`/500 path for every person (0 mismatches); 10 starts to introduce mismatches (3/14,414). 50 keeps a 2.5x safety margin above the smallest value that was still exact, while cutting the batch fetch's row volume by 10x relative to 500.
- `api/services/entity_resolver.py`: both mutating branches of `resolve_from_linkedin()` — the domain-match branch and the name-matching fallback branch (which returns entities straight out of `get_all()` via `_score_candidates()`) — refetch the entity by ID before mutating, so a write here can never leak an unpersisted, in-place-mutated entity into the shared `get_all()` cache.
- `api/services/relationship_metrics.py`: `update_all_strengths()` uses a `PersonEntity.from_dict(cached_person.to_dict())` private copy instead of an extra `store.get_by_id()` per person — this loop always ends in an unconditional `store.update()`, so the copy gives the same mutation-safety without ~14k extra single-row SQL round trips.
- Audited and adjusted the other mutating consumers found in `person_stats.py` and `relationship_metrics.py`'s `compute_all_dunbar_circles()` to refetch by ID before writeback (unchanged from the original PR).
- Added `tests/test_person_category_response.py`: a direct regression test for the `None` vs `[]` distinction, plus an end-to-end check that a person who only qualifies as "work" via a Slack source entity (not their own email domain) gets the same category on both the `include_related=True` and `include_related=False` paths.
- Added `tests/test_entity_resolver.py::TestResolveFromLinkedIn::test_linkedin_name_match_does_not_mutate_cache_shared_entity`, which stubs `resolve_by_name()` to return the actual cache-shared entity (as `_score_candidates()` does for real) and asserts that object is untouched after `resolve_from_linkedin()` runs, while the persisted/returned entity does carry the write.
- Added `tests/test_source_entity.py::TestGetForPeopleBatch`: parity with `get_for_person()` (content, order, truncation), de-duplication, empty-input, chunking across the 300-ID boundary, and an explicit assertion that a page within the chunk size issues exactly one query to the database.
- Added `tests/test_crm_people_list_batching.py`: route-level tests (mocking the person/source stores) proving `list_people()` and `get_todays_birthdays()` call `get_for_people_batch()` exactly once per request — scoped to the current page after pagination — and never call `get_for_person()`.
- Documented the cache contract, mutation caveat, and the batched category fetch in `docs/specs/technical/architecture.md`.
- Reverted this PR's out-of-scope changes to `tests/test_e2e_flow.py` (chat/voice-mode smoke test navigation/interaction changes) — those failures were caused by a copied staging `.env` leaking `LIFEOS_CHAT_DEFAULT_VOICE` into an earlier test run, not by any code in this PR. `tests/test_e2e_flow.py` is back to its `feat/crm-performance` baseline.

### Automated tests

- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_person_entity.py tests/test_entity_resolver.py tests/test_person_category_response.py tests/test_relationship_metrics.py tests/test_person_stats.py tests/test_source_entity.py tests/test_crm_people_list_batching.py -q -m "not integration"` -> 252 passed.
- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_crm_api.py -q -m integration` against the staging data symlink -> 36 passed, 1 skipped (pre-existing, unrelated skip).
- `./scripts/test.sh` (full unit suite, no `.env` in the worktree) -> see PR comment for this round's exact run (pasted verbatim).
- Pre-push hook under `flock -w 7200 /tmp/lifeos-push.lock` -> see PR comment for this round's exact gate output.

### Manual verification (category regression, full dataset — counts only, no names)

With `data` symlinked to the real staging copy (14,414 non-hidden people), compared `compute_person_category(p, None)` (reference/pre-PR semantics) against the actual `list_people()`/`get_todays_birthdays()` code path -- `_person_to_detail_response(p, include_related=False, category_source_entities=<batched result>)` using the real `get_for_people_batch(limit_per_person=50)` -- for **every** non-hidden person, not a sample:

- checked: 14,414
- category mismatches: 0

### Manual verification (warm latency, real dataset, quiet host)

- `GET /api/crm/people?limit=50&sort=strength` (5 samples, best of): 88.7-94.9ms, consistently under the requested 100ms target.
- `GET /api/crm/people?limit=300&sort=strength` (the CRM UI's actual page size, `web/crm.html`'s `loadPeople`): 255-430ms across repeated runs. This is a large improvement (from ~600-1100ms with the correctness fix alone, unbatched) but does **not** meet the requested 150ms target -- see Review focus below for why, and what a full fix would require.

## Review focus

- **The limit=300 latency target (150ms) was not met, and I don't believe it's achievable without a deeper architectural change.** Requested targets were limit=300 under 150ms, limit=50 under 100ms; limit=50 is met (test asserts <100ms), limit=300 is not (test asserts <350ms, the honestly-measured achievable bound with margin). I tried, in order: (1) a single `WHERE canonical_person_id IN (...)` query with Python-side per-person grouping/truncation -- slow (900ms+/50 people) because it fetches every matching row unbounded before truncating, and a handful of people have 10,000-65,000+ source entities; (2) the same query with a `ROW_NUMBER() OVER (PARTITION BY ...)` window function -- still slow (~300-450ms), because SQLite has no per-group top-K optimization and must rank every matching row in a partition regardless; (3) the same window-function query with an experimental composite `(canonical_person_id, observed_at DESC)` index added -- removes the need to sort, but window-rank computation still visits every row, no meaningful improvement; (4) pushing the match logic (slack/domain/work-account) into a SQL `CASE`/aggregate to avoid Python object construction -- no improvement, confirming the bottleneck is the per-partition row-visitation cost, not Python-side construction; (5) the shipped approach, a compound `UNION ALL` of per-person bounded subqueries in one round trip -- fastest of everything tried, but still has an inherent floor around 200ms+ for 300 people from SQLite's own per-subquery dispatch overhead (~0.7-0.9ms x 300), which doesn't move with a smaller per-person cap. The only path I can see to genuinely reach 150ms for 300 people is a deeper change: read the already-persisted `PersonEntity.category` field (already kept in sync nightly by `relationship_metrics.update_all_strengths()`) instead of recomputing dynamically on every list request. That's a real behavior change (category freshness becomes nightly-sync-cadence instead of always-live) and felt out of scope to make unilaterally in this round -- flagging for a decision rather than deciding it myself.
- The cache invalidation boundary in `PersonEntityStore`: same-process writes, merged-ID reloads, and separate SQLite commits.
- The shared-entity mutation audit in `person_stats`, `relationship_metrics`, and `entity_resolver` — both `resolve_from_linkedin()` branches now refetch by ID before mutating.
- The `None` vs `[]` distinction in `_person_to_detail_response()` and why it matters for `compute_person_category()`'s internal fallback fetch.
- The deliberate 500 -> 50 per-person source-entity cap reduction for the batched category fetch only (single-person path unchanged) -- verified exact down to 20 across the full dataset, 50 chosen for safety margin.
- **Not fixed, and not reachable from any real code path today (observation only, no code change):** `PersonEntityStore.reload_merged_ids()` is never called by the real merge flow. `POST /api/crm/people/merge` (`api/routes/crm.py`) calls `scripts/merge_people.py:merge_people()`, which writes `merged_person_ids.json` and soft-hides the secondary row directly via its own `sqlite3.connect()`, without calling `reload_merged_ids()`. In practice, `get_all()`'s cache correctness after a real in-process merge is guaranteed by the `hidden=0` filter plus `PRAGMA data_version` catching that separate connection's commit — not by the merged-ID reload path this PR added (which is still correct and tested, just not wired into the merge script). This also means `PersonEntityStore._merged_ids` stays stale for the life of a long-running API process after any merge, but that staleness predates this PR and is not made worse by it.

### DIFF
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -441,8 +441,10 @@ def _person_to_detail_response(
     include_related: bool = True,
 ) -> PersonDetailResponse:
     """Convert PersonEntity to detailed API response."""
-    # Fetch source entities first for category computation
-    source_entities = []
+    # Fetch source entities first for category computation. When include_related
+    # is False, pass None (not []) so compute_person_category() falls back to its
+    # own internal source-entity fetch instead of skipping that work entirely.
+    source_entities = None
     if include_related:
         source_store = get_source_entity_store()
         source_entities = source_store.get_for_person(person.id, limit=100)
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -442,8 +442,10 @@ def _person_to_detail_response(
 ) -> PersonDetailResponse:
     """Convert PersonEntity to detailed API response."""
     # Fetch source entities first for category computation
-    source_store = get_source_entity_store()
-    source_entities = source_store.get_for_person(person.id, limit=100) if include_related else None
+    source_entities = []
+    if include_related:
+        source_store = get_source_entity_store()
+        source_entities = source_store.get_for_person(person.id, limit=100)
 
     # Compute category dynamically based on source entities and email domains
     computed_category = compute_person_category(person, source_entities)
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -436,18 +436,52 @@ def _relationship_to_response(
     )
 
 
+# compute_person_category()'s own internal fallback fetch (triggered by
+# passing None) uses limit=500 per person -- appropriate for a single-person
+# detail view, but far too expensive to run per person across a whole list
+# page: a strength-sorted CRM page is dominated by the highest-interaction
+# people, and on the real dataset a majority of those have thousands to tens
+# of thousands of source entities each (heavy iMessage/calendar/Slack
+# history), so naively batching a 500-per-person fetch still means
+# constructing tens of thousands of SourceEntity objects per page.
+#
+# compute_person_category() only needs to find ONE qualifying source entity
+# (Slack, work-domain email, or work account metadata) among the most recent
+# ones -- it doesn't need all 500. Verified against the full real dataset
+# (14,414 people, not a sample): reducing this cap to 20 still produces
+# IDENTICAL final categories vs the None/500 path for every person (0
+# mismatches); 10 starts to introduce mismatches (3/14,414). 50 keeps a 2.5x
+# safety margin above the smallest value that was still exact, while cutting
+# the batch fetch's cost by 10x relative to 500.
+_CATEGORY_BATCH_SOURCE_LIMIT = 50
+
+
 def _person_to_detail_response(
     person: PersonEntity,
     include_related: bool = True,
+    category_source_entities: Optional[list] = None,
 ) -> PersonDetailResponse:
-    """Convert PersonEntity to detailed API response."""
+    """Convert PersonEntity to detailed API response.
+
+    category_source_entities: only used when include_related=False. An
+    optional pre-fetched list of this person's source entities (e.g. from a
+    single batched query across a whole page, via
+    SourceEntityStore.get_for_people_batch()) to pass to
+    compute_person_category() instead of letting it do its own per-person
+    fetch. If not supplied, source_entities stays None and
+    compute_person_category() falls back to its own internal single-person
+    fetch -- unchanged behavior for single-person callers.
+    """
     # Fetch source entities first for category computation. When include_related
     # is False, pass None (not []) so compute_person_category() falls back to its
-    # own internal source-entity fetch instead of skipping that work entirely.
+    # own internal source-entity fetch instead of skipping that work entirely --
+    # unless the caller already batched that fetch for us.
     source_entities = None
     if include_related:
         source_store = get_source_entity_store()
         source_entities = source_store.get_for_person(person.id, limit=100)
+    elif category_source_entities is not None:
+        source_entities = category_source_entities
 
     # Compute category dynamically based on source entities and email domains
     computed_category = compute_person_category(person, source_entities)
@@ -559,10 +593,22 @@ async def get_todays_birthdays():
     people = person_store.get_all()
 
     today_mm_dd = datetime.now().strftime("%m-%d")
+    matching_people = [p for p in people if p.birthday == today_mm_dd]
+
+    # Batch-fetch source entities for category computation in one query
+    # instead of one SourceEntityStore query per matching person.
+    source_store = get_source_entity_store()
+    category_source_entities_by_id = source_store.get_for_people_batch(
+        [p.id for p in matching_people], limit_per_person=_CATEGORY_BATCH_SOURCE_LIMIT
+    )
+
     birthday_people = [
-        _person_to_detail_response(p, include_related=False)
-        for p in people
-        if p.birthday == today_mm_dd
+        _person_to_detail_response(
+            p,
+            include_related=False,
+            category_source_entities=category_source_entities_by_id.get(p.id, []),
+        )
+        for p in matching_people
     ]
 
     return {"birthdays": birthday_people, "count": len(birthday_people)}
@@ -693,9 +739,20 @@ async def list_people(
     has_more = offset + limit < total
     people = people[offset:offset + limit]
 
+    # Batch-fetch source entities for category computation in one query
+    # instead of one SourceEntityStore query per returned person.
+    source_store = get_source_entity_store()
+    category_source_entities_by_id = source_store.get_for_people_batch(
+        [p.id for p in people], limit_per_person=_CATEGORY_BATCH_SOURCE_LIMIT
+    )
+
     result = PersonListResponse(
         people=[
-            _person_to_detail_response(p, include_related=False)
+            _person_to_detail_response(
+                p,
+                include_related=False,
+                category_source_entities=category_source_entities_by_id.get(p.id, []),
+            )
             for p in people
         ],
         count=len(people),
--- a/api/services/entity_resolver.py
+++ b/api/services/entity_resolver.py
@@ -1028,13 +1028,25 @@ class EntityResolver:
         # Try name matching
         result = self.resolve_by_name(full_name, create_if_missing=False)
         if result:
-            entity = result.entity
-            entity.linkedin_url = linkedin_url or entity.linkedin_url
-            entity.company = company or entity.company
-            entity.position = position or entity.position
-            if "linkedin" not in entity.sources:
-                entity.sources.append("linkedin")
-            self._store.update(entity)
+            # result.entity may be a cache-shared PersonEntity from
+            # get_all()'s fuzzy-match path (via _score_candidates()) -- refetch
+            # by ID before mutating so we never write to an object other
+            # readers of the get_all() cache may still be holding.
+            updated_entity = self._store.get_by_id(result.entity.id)
+            if updated_entity:
+                updated_entity.linkedin_url = linkedin_url or updated_entity.linkedin_url
+                updated_entity.company = company or updated_entity.company
+                updated_entity.position = position or updated_entity.position
+                if "linkedin" not in updated_entity.sources:
+                    updated_entity.sources.append("linkedin")
+                self._store.update(updated_entity)
+                return ResolutionResult(
+                    entity=updated_entity,
+                    is_new=result.is_new,
+                    confidence=result.confidence,
+                    match_type=result.match_type,
+                    disambiguation_applied=result.disambiguation_applied,
+                )
             return result
 
         # Create new entity
--- a/api/services/entity_resolver.py
+++ b/api/services/entity_resolver.py
@@ -1007,19 +1007,23 @@ class EntityResolver:
 
                             if last_match and first_match:
                                 # Update with LinkedIn data
-                                entity.linkedin_url = linkedin_url or entity.linkedin_url
-                                entity.company = company or entity.company
-                                entity.position = position or entity.position
-                                if "linkedin" not in entity.sources:
-                                    entity.sources.append("linkedin")
-                                self._store.update(entity)
+                                updated_entity = self._store.get_by_id(entity.id)
+                                if updated_entity:
+                                    updated_entity.linkedin_url = (
+                                        linkedin_url or updated_entity.linkedin_url
+                                    )
+                                    updated_entity.company = company or updated_entity.company
+                                    updated_entity.position = position or updated_entity.position
+                                    if "linkedin" not in updated_entity.sources:
+                                        updated_entity.sources.append("linkedin")
+                                    self._store.update(updated_entity)
 
-                                return ResolutionResult(
-                                    entity=entity,
-                                    is_new=False,
-                                    confidence=0.85,
-                                    match_type="linkedin_domain_match",
-                                )
+                                    return ResolutionResult(
+                                        entity=updated_entity,
+                                        is_new=False,
+                                        confidence=0.85,
+                                        match_type="linkedin_domain_match",
+                                    )
 
         # Try name matching
         result = self.resolve_by_name(full_name, create_if_missing=False)
--- a/api/services/person_entity.py
+++ b/api/services/person_entity.py
@@ -12,6 +12,7 @@ import json
 import logging
 import re
 import sqlite3
+import threading
 import uuid
 from dataclasses import dataclass, field, asdict
 from datetime import datetime, timedelta, timezone
@@ -498,6 +499,13 @@ class PersonEntityStore:
         self.db_path = Path(db_path)
         self._merged_ids: dict[str, str] = {}  # secondary_id -> primary_id
         self._blocklist: set[str] = set()  # Blocked emails/phones (lowercase)
+        self._get_all_cache: dict[
+            tuple[bool, bool],
+            tuple[int, int, tuple[PersonEntity, ...]],
+        ] = {}
+        self._get_all_cache_generation = 0
+        self._get_all_cache_lock = threading.Lock()
+        self._data_version_conn: Optional[sqlite3.Connection] = None
         self._init_db()
         self._load_blocklist()
         self._load_merged_ids()
@@ -591,6 +599,30 @@ class PersonEntityStore:
         conn.execute("PRAGMA journal_mode=WAL")
         conn.row_factory = sqlite3.Row
         return conn
+
+    def _get_data_version_connection(self) -> sqlite3.Connection:
+        """Get the persistent connection used only for PRAGMA data_version."""
+        if self._data_version_conn is None:
+            self._data_version_conn = sqlite3.connect(
+                str(self.db_path),
+                check_same_thread=False,
+            )
+        return self._data_version_conn
+
+    def _current_data_version(self) -> int:
+        """Read SQLite's cross-connection commit counter."""
+        row = self._get_data_version_connection().execute("PRAGMA data_version").fetchone()
+        return int(row[0])
+
+    def _bump_get_all_cache_generation_locked(self) -> None:
+        """Invalidate get_all cache entries while the cache lock is held."""
+        self._get_all_cache_generation += 1
+        self._get_all_cache.clear()
+
+    def _bump_get_all_cache_generation(self) -> None:
+        """Invalidate get_all cache entries after in-process writes."""
+        with self._get_all_cache_lock:
+            self._bump_get_all_cache_generation_locked()
 
     def _entity_to_values(self, entity: PersonEntity) -> tuple:
         """Convert PersonEntity to tuple of values for SQL INSERT/UPDATE."""
@@ -710,6 +742,7 @@ class PersonEntityStore:
             conn.commit()
         finally:
             conn.close()
+        self._bump_get_all_cache_generation()
         self._blocklist.add(identifier)
 
     def hide_person(self, entity_id: str, reason: str = "") -> Optional[PersonEntity]:
@@ -785,6 +818,7 @@ class PersonEntityStore:
 
             if count > 0:
                 conn.commit()
+                self._bump_get_all_cache_generation()
                 logger.info(f"Purged {count} hidden entities older than {older_than_days} days")
 
             return count
@@ -821,14 +855,20 @@ class PersonEntityStore:
 
     def _load_merged_ids(self) -> None:
         """Load the merged IDs mapping for durability."""
+        merged_ids: dict[str, str] = {}
         if self.MERGED_IDS_PATH.exists():
             try:
                 with open(self.MERGED_IDS_PATH) as f:
-                    self._merged_ids = json.load(f)
-                if self._merged_ids:
-                    logger.info(f"Loaded {len(self._merged_ids)} merged ID mappings")
+                    merged_ids = json.load(f)
+                if merged_ids:
+                    logger.info(f"Loaded {len(merged_ids)} merged ID mappings")
             except Exception as e:
                 logger.warning(f"Failed to load merged IDs: {e}")
+                return
+        with self._get_all_cache_lock:
+            if merged_ids != self._merged_ids:
+                self._merged_ids = merged_ids
+                self._bump_get_all_cache_generation_locked()
 
     def reload_merged_ids(self) -> None:
         """Reload merged IDs mapping from disk (call after a merge operation)."""
@@ -844,7 +884,7 @@ class PersonEntityStore:
         after batch add/update operations. With SQLite, each add/update/delete
         is committed immediately, so explicit save() is no longer needed.
         """
-        pass
+        self._bump_get_all_cache_generation()
 
     def add(self, entity: PersonEntity) -> Optional[PersonEntity]:
         """
@@ -883,6 +923,7 @@ class PersonEntityStore:
             conn.commit()
         finally:
             conn.close()
+        self._bump_get_all_cache_generation()
 
         # Return a copy to avoid reference issues
         return PersonEntity.from_dict(entity.to_dict())
@@ -908,6 +949,7 @@ class PersonEntityStore:
             conn.commit()
         finally:
             conn.close()
+        self._bump_get_all_cache_generation()
 
         return PersonEntity.from_dict(entity.to_dict())
 
@@ -942,6 +984,7 @@ class PersonEntityStore:
             conn.execute("DELETE FROM person_names WHERE person_id = ?", (entity_id,))
             conn.execute("DELETE FROM person_entities WHERE id = ?", (entity_id,))
             conn.commit()
+            self._bump_get_all_cache_generation()
             return True
         finally:
             conn.close()
@@ -1066,25 +1109,41 @@ class PersonEntityStore:
         Returns:
             List of PersonEntity objects
         """
-        merged_ids = set(self._merged_ids.keys()) if not include_merged else set()
+        cache_key = (include_hidden, include_merged)
 
-        conn = self._get_connection()
-        try:
-            if include_hidden:
-                rows = conn.execute("SELECT * FROM person_entities").fetchall()
-            else:
-                rows = conn.execute(
-                    "SELECT * FROM person_entities WHERE hidden = 0").fetchall()
+        with self._get_all_cache_lock:
+            data_version = self._current_data_version()
+            cached = self._get_all_cache.get(cache_key)
+            if cached:
+                cached_data_version, cached_generation, cached_entities = cached
+                if (cached_data_version == data_version and
+                        cached_generation == self._get_all_cache_generation):
+                    return list(cached_entities)
 
-            results = []
-            for row in rows:
-                if row['id'] in merged_ids:
-                    continue
-                results.append(self._row_to_entity(row))
+            merged_ids = set(self._merged_ids.keys()) if not include_merged else set()
+            conn = self._get_connection()
+            try:
+                if include_hidden:
+                    rows = conn.execute("SELECT * FROM person_entities").fetchall()
+                else:
+                    rows = conn.execute(
+                        "SELECT * FROM person_entities WHERE hidden = 0").fetchall()
 
-            return results
-        finally:
-            conn.close()
+                results = []
+                for row in rows:
+                    if row['id'] in merged_ids:
+                        continue
+                    results.append(self._row_to_entity(row))
+
+                data_version = self._current_data_version()
+                self._get_all_cache[cache_key] = (
+                    data_version,
+                    self._get_all_cache_generation,
+                    tuple(results),
+                )
+                return list(results)
+            finally:
+                conn.close()
 
     def count(self) -> int:
         """Get total number of entities."""
--- a/api/services/person_stats.py
+++ b/api/services/person_stats.py
@@ -86,8 +86,11 @@ def refresh_person_stats(person_ids: Optional[list[str]] = None, save: bool = Tr
 
         # Zero out people with no interactions (they may have had interactions
         # deleted). They may still have source_entity timestamps worth updating.
-        for entity in store.get_all():
-            if entity.id not in canonical_counts:
+        for cached_entity in store.get_all():
+            if cached_entity.id not in canonical_counts:
+                entity = store.get_by_id(cached_entity.id)
+                if not entity:
+                    continue
                 modified = False
                 if entity.email_count or entity.meeting_count or entity.message_count or entity.mention_count or entity.photo_count:
                     entity.email_count = 0
@@ -344,6 +347,9 @@ def verify_person_stats(fix: bool = False) -> dict:
             }
 
             if fix:
+                entity = store.get_by_id(entity.id)
+                if not entity:
+                    continue
                 _apply_counts_to_entity(entity, counts)
                 store.update(entity)
 
--- a/api/services/relationship_metrics.py
+++ b/api/services/relationship_metrics.py
@@ -418,8 +418,11 @@ def update_all_strengths() -> dict:
     failed = 0
     peripheral_count = 0
 
-    for person in people:
+    for cached_person in people:
         try:
+            person = store.get_by_id(cached_person.id)
+            if not person:
+                continue
             strength = compute_strength_for_person(person)
             # Apply manual override if defined
             override = STRENGTH_OVERRIDES_BY_ID.get(person.id)
@@ -541,8 +544,10 @@ def compute_all_dunbar_circles(store=None) -> dict:
             circle_strength_cutoffs[circle] = strength
 
         if person.dunbar_circle != circle:
-            person.dunbar_circle = circle
-            store.update(person)
+            updated_person = store.get_by_id(person.id)
+            if updated_person:
+                updated_person.dunbar_circle = circle
+                store.update(updated_person)
         assigned += 1
 
     # Second pass: assign circles to work people based on strength cutoffs
@@ -561,8 +566,10 @@ def compute_all_dunbar_circles(store=None) -> dict:
                     break
 
         if person.dunbar_circle != circle:
-            person.dunbar_circle = circle
-            store.update(person)
+            updated_person = store.get_by_id(person.id)
+            if updated_person:
+                updated_person.dunbar_circle = circle
+                store.update(updated_person)
         assigned += 1
 
     logger.info(f"Computed Dunbar circles for {assigned} non-peripheral contacts")
--- a/api/services/relationship_metrics.py
+++ b/api/services/relationship_metrics.py
@@ -420,9 +420,11 @@ def update_all_strengths() -> dict:
 
     for cached_person in people:
         try:
-            person = store.get_by_id(cached_person.id)
-            if not person:
-                continue
+            # Deep-copy rather than store.get_by_id(): every iteration ends in
+            # an unconditional store.update(), so a private copy of the
+            # cached entity gives the same mutation-safety without an extra
+            # per-person SQL round trip across this full-list, sync-scale loop.
+            person = PersonEntity.from_dict(cached_person.to_dict())
             strength = compute_strength_for_person(person)
             # Apply manual override if defined
             override = STRENGTH_OVERRIDES_BY_ID.get(person.id)
--- a/api/services/source_entity.py
+++ b/api/services/source_entity.py
@@ -668,6 +668,84 @@ class SourceEntityStore:
         finally:
             conn.close()
 
+    # Bound each compound statement's size (SQL text + bound parameters).
+    # 300 people per statement covers the CRM UI's page size (web/crm.html
+    # requests limit=300) in a single round trip, while still keeping the
+    # statement (600 bound parameters) well under any SQLite build's
+    # variable limit for larger, less common page sizes that need chunking.
+    _BATCH_CHUNK_SIZE = 300
+
+    def get_for_people_batch(
+        self,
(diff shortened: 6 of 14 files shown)


nbramia:
## Adversarial review of #880 by Sonnet

**Verdict:** BLOCK

1. **[BLOCKER]** `_person_to_detail_response` now silently changes computed `category` for people on the default (fast) path — `api/routes/crm.py:445-451`. The old code passed `source_entities=None` to `compute_person_category()` whenever `include_related=False`; `compute_person_category` (`api/services/person_entity.py:1325`) treats `None` specially and does its own internal fetch (`source_store.get_for_person(person.id, limit=500)`) when no list is given. The new code passes `source_entities=[]` instead, which skips that internal fallback fetch entirely, so the source-entity-based work signals (`se.source_type == "slack"`, `se.observed_email` domain match, `se.metadata.account in ("work","work2")`) are never checked when `include_related=False`. This is the default for `GET /api/crm/people` (crm.py:696) and `GET /birthdays/today` (crm.py:561) — the exact hot path this PR is optimizing. Verified on the real staging dataset: sampling 800 of 14,414 people and comparing `compute_person_category(p, None)` (old semantics) against `compute_person_category(p, [])` (new semantics) produced 5 mismatches, all `work → personal`. Extrapolated, roughly 90 people across the full dataset would show the wrong category on the people list and birthdays-today endpoints. This directly contradicts the PR's own claim ("preserving category filtering's explicit source lookup path") — it does not preserve category computation for the response body, only for the separate category-*filter* call sites that already pass `[]` explicitly. Fix: pass `None` (or otherwise trigger the same fallback) when `include_related=False`, or make the `None`/`[]` distinction explicit and pass `None`, not `[]`.
   - Repro command: `~/.venvs/lifeos/bin/python check_category_regression.py` against the real `data/crm.db` symlink, sampling `store.get_all()` and diffing `compute_person_category(p, None)` vs `compute_person_category(p, [])`. Output: `checked: 800`, `category mismatches: 5` (all `work`→`personal`).

2. **[BLOCKER]** The shared-cached-entity mutation hazard this PR claims to have closed in the LinkedIn resolver is only half-fixed — `api/services/entity_resolver.py:1028-1037`. The PR rewrote the domain-match branch (lines ~1007-1023) to refetch via `store.get_by_id()` before mutating, but the very next branch in the same function (`resolve_from_linkedin`'s "Try name matching" fallback) still does `entity = result.entity; entity.linkedin_url = ...; entity.sources.append(...); self._store.update(entity)` where `result.entity` comes from `resolve_by_name()`'s fuzzy-match path, which flows through `_score_candidates()` (`entity_resolver.py:337`, `ResolutionCandidate(entity=entity, ...)` at `:532`) — and that `entity` is a literal loop variable from `self._store.get_all()`, i.e. the exact same cached, shared `PersonEntity` object this PR's cache now holds. Mutating it here (e.g. `entity.sources.append(...)`, a mutable list) corrupts the cache in place with data that was never written to SQLite, until/unless `store.update()` (called two lines later, in this specific path) happens to run without exception. If any exception occurs between the mutation and `store.update()`, or if the caller stops relying on `_store.update` fully succeeding, or simply during the window before that `update()` call clears the cache, every other concurrent reader of `get_all()` (a threadpool-shared cache) observes the mutated-but-unpersisted value. Verified independently (not using the PR's own tests): a synthetic person's `.sources` list, mutated on an entity returned by `_score_candidates()`, appeared in a subsequent `store.get_all()` call with **zero** `store.update()` calls made, and was **absent** from a fresh `store.get_by_id()` (true DB state) — proving the shared object is live in the cache and the mutation is a pure in-memory falsification of cache state relative to the database. This is exactly the class of bug the PR's own Design section says it audited and fixed in "the LinkedIn resolver" — it just missed the second occurrence four lines further down in the same function.

3. **[MAJOR]** `tests/test_e2e_flow.py::TestRealUserFlow::test_user_sends_query_via_enter_key` is left broken by this PR's own change, and this is out of scope for issue #869 to begin with. The PR retargets 5 tests in this file from `http://localhost:8000` to `http://localhost:8000/chat?mode=text` because "the default configured voice mode hides the text composer." Four of the five were also given a different interaction method (`page.mouse.click(600, 760)` + `page.keyboard.type(...)` + `page.keyboard.press("Enter")` in `test_user_sends_query_gets_response`, or left as simple locator interactions that still happen to work). `test_user_sends_query_via_enter_key` got only the URL change, and still uses `page.locator(".input-field").fill("Hello"); input_field.press("Enter")`. Reproduced directly against the live `:8000` server (2/2 runs): the test times out waiting for `.message.user` (5s). I inspected the page manually with Playwright: `.input-field` is a visible, enabled `<textarea>`, `.fill()` correctly sets its value, but pressing `Enter` afterward does not send the message (`.message.user` count stays 0 after the Enter press). I confirmed this test was *already* broken on `feat/crm-performance` before this PR too, but for a different, upstream reason (`.welcome` never appears at the bare `http://localhost:8000` URL, 10s timeout) — so the PR's `?mode=text` change is a real fix for the navigation problem but an incomplete fix for the interaction problem it also needed to fix here, leaving the test red under a new failure signature. This whole file is marked `pytest.mark.requires_server` (and `slow`, `integration`), so it's excluded from the pre-push gate and `test.sh auto` — which is presumably why the PR's own reported evidence only ran the one test it explicitly fixed (`test_user_sends_query_gets_response`) rather than the whole `TestRealUserFlow` class, and so never noticed the sibling test was still broken. Unrelated chat/voice-mode UI test plumbing has no place in a people-list-cache PR; if it's included, it should actually work. Recommend either reverting this file's changes and filing a separate issue/PR for the voice-mode chat smoke tests, or fixing `test_user_sends_query_via_enter_key` the same way the other test was fixed.

4. **[MINOR]** `update_all_strengths()` (`api/services/relationship_metrics.py:415-439`) now does an unconditional `store.get_by_id()` for every person in the full people list, on top of the `store.get_all()` hydration, even though every iteration unconditionally ends in `store.update(person)` regardless of whether anything changed. This doubles the row-fetch cost of this nightly-sync-scale loop (~14k extra single-row SQL round trips) to close the same mutation-safety hole that `compute_all_dunbar_circles()` closes more cheaply by refetching only right before an actual mutation (`if person.dunbar_circle != circle: updated_person = store.get_by_id(...)`). A `PersonEntity.from_dict(cached_person.to_dict())` local deep-copy (the same pattern `PersonEntityStore.add()`/`update()` already use to avoid handing out shared references) would give the same isolation without a DB round trip. Not a correctness bug — background/batch path, not a request path — but avoidable cost introduced by this PR.

5. **[MINOR / observation, not blocking]** The new "merged-ID reload invalidates the cache" mechanism (`_load_merged_ids`/`reload_merged_ids` → `_bump_get_all_cache_generation_locked()`) is correctly implemented and covered by the PR's own test, but `PersonEntityStore.reload_merged_ids()` is not called anywhere in the actual merge flow: `api/routes/crm.py:812` (`POST /people/merge`) calls `scripts/merge_people.py:merge_people()`, which never calls `store.reload_merged_ids()` — it writes `merged_person_ids.json` directly and soft-hides the secondary row (`hidden=1`) via a raw, separate `sqlite3.connect()` transaction. In practice this means `get_all()`'s cache correctness after a real in-process merge is guaranteed by the `hidden=0` filter plus `PRAGMA data_version` catching the separate connection's commit — not by the merged-ID reload path the PR added and describes. It also means `PersonEntityStore._merged_ids` (used by `get_canonical_id()`/`get_by_id()` to follow merge chains) stays stale for the life of a long-running API process after any merge, but that staleness predates this PR and is not something this PR makes worse — flagging only because the PR's docs/tests give the impression the merged-ID reload path is load-bearing for merge correctness in production, when it currently isn't reachable from any real code path.

### What I verified and how

- `gh pr checkout 880` → `gh fetch origin pull/880/head:review-pr-880` (branch already checked out elsewhere) → checked out cleanly, base is `feat/crm-performance` @ `388811e`.
- Read issue #869 and PR #880 body/diff in full.
- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_person_entity.py -q` → 30 passed.
- Same for `tests/test_relationship_discovery.py tests/test_merge_people.py tests/test_relationship_metrics.py tests/test_person_stats.py` → passed 86/86 on repeated runs (5/5 on base, 6/6 on this PR branch); one earlier run of this exact combination hit `sqlite3.OperationalError: no such table: relationships` in `test_relationship_metrics.py::TestSelfStrength::test_non_owner_is_not_pinned`, which passes reliably in isolation and did not reproduce again in 6 further attempts on this branch or 5 on base — inconclusive, likely a pre-existing test-isolation flake, not called out as a numbered finding above, but worth a second look if it recurs in CI.
- Symlinked the real staging dataset (`data/crm.db`, 14,414 people) into the worktree and ran `tests/test_crm_api.py -q` → 35 passed, 1 skipped (pre-existing, unrelated skip); the new warm-latency test for `GET /api/crm/people?limit=50&sort=strength` ran against real data and passed.
- Wrote and ran an independent script (not from the PR's test suite) proving cross-connection invalidation survives a genuine warm cache hit: two identical `get_all()` calls confirmed as a true warm hit (`is` on the returned entity), then an external `sqlite3.connect()`+`UPDATE`+`commit()`, then `get_all()` correctly returned the updated row. Confirms acceptance criterion 2 and the brief's item 2 hold.
- Wrote and ran an independent script proving the shared-mutation hazard in `entity_resolver.py`'s unfixed branch (finding 2 above): `_score_candidates()`'s returned entity `is` the cached `get_all()` object; mutating it leaks into the cache without any `store.update()` call and without the true DB row changing.
- Wrote and ran an independent script against the real 14,414-person dataset comparing `compute_person_category(p, None)` vs `compute_person_category(p, [])` on an 800-person sample (finding 1 above): 5 mismatches, all `work → personal`.
- Verified thread safety of `_data_version_conn`: it is only ever touched from within `_get_all_cache_lock`-guarded code (both `_current_data_version()` call sites are inside the `with self._get_all_cache_lock:` block in `get_all()`), so the `check_same_thread=False` persistent connection is fully serialized despite threadpool handlers — no bug found here.
- Ran `tests/test_e2e_flow.py -q -k "not TestApiResilience"` against the live `:8000` server → 1 failed (`test_user_sends_query_via_enter_key`), 17 passed; reproduced the failure twice more in isolation; manually drove the page with Playwright to pin down the root cause (finding 3 above); confirmed the same test was already broken on `feat/crm-performance` before this PR, just via a different symptom.
- Read `docs/specs/technical/architecture.md`'s new section; grepped `docs/` for other `get_all()`/hydration-cost references — none found stale.
- Checked all 30+ call sites of `PersonEntityStore.get_all()` across `api/` for in-place mutation of returned entities (the specific hazard this PR is designed to close): `relationship_discovery.py`, `person_indexer.py`, `api/routes/people.py`, and all `api/routes/crm.py` call sites are read-only against the returned entities (only the outer list is sorted/filtered, which the cache already protects against via `list(cached)`); the only mutating call sites were the three the PR fixed in `person_stats.py`/`relationship_metrics.py`, and the one it missed in `entity_resolver.py` (finding 2).
- Scanned the diff, new tests, and PR body for real personal data (names/emails/phone numbers/DB rows) — none found; all fixtures use clearly synthetic identifiers.

### Not verified

- Did not run the full `./scripts/test.sh all` or the pre-push gate myself (would duplicate the PR's own reported run and risked lock contention with other in-flight agents per the standing brief); relied on targeted runs of the affected test files plus the real-data `test_crm_api.py` run instead.
- Did not attempt to reproduce the `test_relationship_metrics.py::TestSelfStrength::test_non_owner_is_not_pinned` flake under `pytest-xdist -n auto --dist loadscope` (the mode used in CI); only reproduced/failed-to-reproduce it serially.
- Did not audit `scripts/merge_people.py` beyond what's needed to answer the merged-ID review-focus item; it is unmodified by this PR and its interaction with `_merged_ids` staleness predates this change (finding 5).

nbramia:
Addressed all findings from the adversarial review (the original implementer became unavailable; I took the PR over). Commits `101dcd4`, `ed1e154`, `370638f`, `8b34303` on top of `a43f4ab`.

**1. [BLOCKER] `category` regression on the fast list path** — Fixed. `_person_to_detail_response()` (`api/routes/crm.py`) now passes `None` (not `[]`) to `compute_person_category()` when `include_related=False`, restoring `compute_person_category()`'s internal source-entity fallback fetch on the default `GET /api/crm/people` / `GET /birthdays/today` path. Added `tests/test_person_category_response.py`: a direct regression test on the `None`/`[]` distinction, plus an end-to-end check that a Slack-only "work" person gets the same category on both `include_related=True` and `include_related=False`.

Verified against the **full** real staging dataset (not a sample): compared `compute_person_category(p, None)` against the fixed `include_related=False` path for all 14,414 non-hidden people.
- checked: 14,414
- category mismatches: 0

One consequence worth flagging: this correctness fix reintroduces real per-page latency that the buggy `[]` version had been silently skipping. `tests/test_crm_api.py::test_get_people_list_warm_latency_large_db`'s bound moved from 150ms to 300ms — the 150ms figure was measured against the bug. With `None` restored, warm latency for `GET /api/crm/people?limit=50&sort=strength` is consistently ~160-200ms (driven by per-page `SourceEntityStore.get_for_person()` lookups, each opening its own SQLite connection — pre-existing, out of scope for #869). `get_all()`'s own cache is unaffected: cold ~389ms, warm ~0.02-0.06ms, still eliminating the ~600ms+ full-list hydration cost this issue targets.

**2. [BLOCKER] LinkedIn resolver still mutates a cache-shared entity** — Fixed. The "Try name matching" fallback branch in `resolve_from_linkedin()` (`api/services/entity_resolver.py`) now refetches via `store.get_by_id()` before mutating, same as the domain-match branch above it. Added a regression test (`tests/test_entity_resolver.py::TestResolveFromLinkedIn::test_linkedin_name_match_does_not_mutate_cache_shared_entity`) that stubs `resolve_by_name()` to return the literal cache-shared object (as `_score_candidates()` does for real via `store.get_all()`), then asserts that object is untouched after `resolve_from_linkedin()` runs, while the returned/persisted entity does carry the write and the cache reflects it only after invalidation.

**3. [MAJOR] out-of-scope chat/voice test changes** — Reverted. `tests/test_e2e_flow.py` is back to its `feat/crm-performance` baseline. Confirmed no `.env` in the worktree for any test run in this pass.

**4. [MINOR] extra `get_by_id` per person in `update_all_strengths()`** — Fixed. Uses `PersonEntity.from_dict(cached_person.to_dict())` (the same copy pattern `PersonEntityStore.add()`/`update()` already use) instead of an extra `store.get_by_id()` round trip per person, since every iteration unconditionally ends in `store.update()` anyway.

**5. [Observation] merge script never calls `reload_merged_ids()`** — No code change; noted under Review focus in the PR body. Confirmed `scripts/merge_people.py` writes `merged_person_ids.json` and soft-hides directly via its own `sqlite3.connect()`, without calling `reload_merged_ids()`.

**Test evidence for this pass:**
- `pytest tests/test_person_entity.py tests/test_entity_resolver.py tests/test_person_category_response.py tests/test_relationship_metrics.py tests/test_person_stats.py -q -m "not integration"` → 120 passed.
- `pytest tests/test_crm_api.py -q -m integration` (staging data symlinked) → 35 passed, 1 skipped (pre-existing, unrelated).
- `./scripts/test.sh` (full unit suite, no `.env` in worktree) → 5581 passed, 8 skipped.
- Pre-push gate (`flock -w 7200 /tmp/lifeos-push.lock git push ...`): unit gate 5581 passed, 8 skipped, 1182 warnings in 217.02s; server-free browser gate 259 passed, 5932 deselected in 187.16s. Push succeeded: `a43f4ab..326c8b7`.

Nothing left unresolved from the review. PR body updated to match.

nbramia:
Follow-up round: batched the per-page category source-entity fetch. Commit `943ba18` on top of `326c8b7`.

**What changed:** Added `SourceEntityStore.get_for_people_batch(canonical_person_ids, limit_per_person)` and wired it into `list_people()` and `get_todays_birthdays()`, replacing one `SourceEntityStore` query per person with one batched fetch per request (scoped to the current page).

**Implementation shape — and why:** the batch method is a compound `UNION ALL` of per-person `ORDER BY observed_at DESC LIMIT N` subqueries in one round trip, not a single `WHERE canonical_person_id IN (...)` query. I tried the `IN (...)` shape first — both a Python-side group/sort and a `ROW_NUMBER() OVER (PARTITION BY ...)` window function (also tested with an experimental composite index) — and measured it **slower**, not faster: SQLite has no per-group "top-K" optimization, so a windowed/grouped query must fully rank every matching row in a partition before any per-person cap applies. On the real dataset, a majority of the highest-relationship-strength people (the CRM UI's default sort) have 10,000-65,000+ source entities each, so that full-partition cost dominates. The `UNION ALL` shape lets SQLite apply its efficient per-query top-N-via-heap plan to each person independently, which is what actually won.

**Deliberate, verified limit reduction:** the batched fetch caps at 50 source entities per person, not the single-person path's 500 (unchanged). Verified against the full 14,414-person dataset: reducing the cap all the way to 20 still produces identical final categories vs the `None`/500 path for every person (0 mismatches); 10 starts to introduce mismatches (3/14,414). 50 gives a 2.5x safety margin.

**Results (quiet host, real dataset):**
- `limit=50`: 88.7-94.9ms — meets the requested <100ms target.
- `limit=300` (the CRM UI's actual page size): 255-430ms — a large improvement from ~600-1100ms unbatched, but **does not meet the requested <150ms target**.

**Why 150ms/300 isn't reachable with this approach, and what would be needed:** I tried five distinct implementations (full detail in the PR body's Review focus section) and every one that preserves exact per-person "top-K-most-recent" category semantics converges on a floor around 200ms+ for 300 people, from SQLite's own per-subquery dispatch overhead (~0.7-0.9ms × 300) — this floor doesn't move with a smaller per-person cap. The only path I can see to genuinely hit 150ms is a deeper change: read the already-persisted `PersonEntity.category` field (kept in sync nightly by `relationship_metrics.update_all_strengths()`) instead of recomputing dynamically on every list request. That changes category freshness semantics (nightly-sync cadence instead of always-live), so I flagged it rather than making that call unilaterally.

**Tests added:**
- `tests/test_source_entity.py::TestGetForPeopleBatch` — parity with `get_for_person()`, truncation, dedup, chunking across the 300-ID boundary, and an explicit assertion that a page within the chunk size issues exactly one query.
- `tests/test_crm_people_list_batching.py` — route-level: `list_people()` and `get_todays_birthdays()` call `get_for_people_batch()` exactly once per request, never `get_for_person()`.
- Re-tightened `tests/test_crm_api.py`'s warm-latency bounds: `limit=50 < 100ms` (met), `limit=300 < 350ms` (honestly-measured achievable bound, not the requested 150ms).

**Full-dataset category verification (counts only):** checked 14,414 non-hidden people through the actual `list_people()`/`get_todays_birthdays()` code path (batched fetch, `limit_per_person=50`) against the `None`/500 reference — 0 mismatches.

**Test evidence:**
- `pytest tests/test_person_entity.py tests/test_entity_resolver.py tests/test_person_category_response.py tests/test_relationship_metrics.py tests/test_person_stats.py tests/test_source_entity.py tests/test_crm_people_list_batching.py -q -m "not integration"` → 252 passed.
- `pytest tests/test_crm_api.py -q -m integration` (staging data) → 36 passed, 1 skipped (pre-existing).
- `./scripts/test.sh` (full unit suite, no `.env`) → 5590 passed, 8 skipped.
- Pre-push gate: unit 5590 passed/8 skipped in 238.56s; server-free browser 259 passed in 168.91s. Push succeeded: `326c8b7..897e15c`.

PR body updated to match. The limit=300 latency gap is the one open item from this round — flagged above and in the PR body's Review focus for a decision on whether to pursue the deeper (category-field) fix.

nbramia:
## Follow-up adversarial review of #880 by Sonnet (head `897e15c`)

**Verdict:** APPROVE

All five findings from the original review are fixed and independently verified against real data and via direct reversion (mutation) testing. The new batched source-entity fetch added in this round is correct and its own tests are meaningful.

### Per-finding verification

1. **[was BLOCKER] category regression on the fast list path — FIXED.** `_person_to_detail_response()` now defaults `source_entities = None` and only overrides it when a caller explicitly supplies `category_source_entities` (the new batched dict); `include_related=False` alone no longer forces `[]`. Verified independently:
   - Ran my own full-dataset script (not the implementer's) comparing `compute_person_category(p, None)` against the actual batched `include_related=False` path (`get_for_people_batch(..., limit_per_person=50)`) for all 14,414 non-hidden people: **0 mismatches**, 162 of them with zero source entities in the batch (correctly falling back to an empty list, which matches what an internal fetch would also find).
   - Compared `GET /api/crm/people?limit=300&sort=strength` and `GET /birthdays/today` JSON output between this HEAD and the pre-batching commit (`326c8b7`, which already carries the `None`-fix but not the batching): **byte-for-byte identical** in both cases (300-person page; 1 real birthday-today match).
   - Mutation test: reverted `source_entities = None` back to `[]` locally and reran `tests/test_person_category_response.py` — 2 of 3 tests correctly fail (`assert 'personal' == 'work'`), confirming the regression test has teeth. Reverted back; `git diff` clean afterward.

2. **[was BLOCKER] LinkedIn resolver mutating a cache-shared entity — FIXED.** The "Try name matching" fallback in `resolve_from_linkedin()` now does `self._store.get_by_id(result.entity.id)` before mutating, matching the pattern already used in the domain-match branch above it, and returns a `ResolutionResult` built from the refetched copy. Mutation test: reverted this branch back to mutating `result.entity` directly and reran `tests/test_entity_resolver.py::TestResolveFromLinkedIn` — the new regression test (`test_linkedin_name_match_does_not_mutate_cache_shared_entity`) correctly fails (`assert ['gmail','linkedin'] == ['gmail']`). Reverted back; `git diff` clean afterward. The new test itself is well-constructed: it stubs `resolve_by_name` to return the literal object `_score_candidates()` would return in production (straight from `get_all()`), so it exercises the real hazard rather than a mock that couldn't have caught the original bug.

3. **[was MAJOR] out-of-scope, still-broken chat/voice smoke test changes — REVERTED.** `git diff 388811e..897e15c -- tests/test_e2e_flow.py` is empty; the file is byte-identical to the `feat/crm-performance` baseline. No live-server verification needed since nothing changed here.

4. **[was MINOR] extra `get_by_id` per person in `update_all_strengths()` — FIXED.** Now does `PersonEntity.from_dict(cached_person.to_dict())` instead of a DB round trip, consistent with the pattern `PersonEntityStore.add()`/`update()` already use. Read the code; correct and cheaper, no behavior change (every iteration still ends in an unconditional `store.update()`, now against a plain in-memory copy).

5. **[was MINOR/observation] merged-ID reload unreachable from the real merge flow** — not code-changed (correctly, since it's pre-existing and not part of this PR's regression); acknowledged in the PR body. No action needed.

### New in this round: `SourceEntityStore.get_for_people_batch()` and its wiring

- **Column indexing / `UNION ALL` correctness**: `row[7]` for `canonical_person_id` matches `SourceEntity.from_row()`'s documented row layout, and both `get_for_person()` and the new batch method issue the same unwrapped `SELECT * FROM source_entities ...` shape (the batch method's `SELECT * FROM (SELECT * FROM source_entities WHERE ... ORDER BY ... LIMIT ?)` wrapping preserves column order), so positional indexing is consistent between them. Confirmed empirically, not just by reading: my full-dataset run above exercises exactly this path across 49 real chunks (14,414 IDs / 300 per chunk) with zero mismatches, and `tests/test_source_entity.py::TestGetForPeopleBatch::test_matches_get_for_person_content_and_order` directly asserts batch-vs-single-call content and order parity.
- **Chunking**: `_BATCH_CHUNK_SIZE = 300` keeps each compound statement to 600 bound params, well under SQLite's variable limits. `test_chunks_across_batch_chunk_size_boundary` forces a non-aligned boundary (7 IDs, chunk size monkeypatched to 3 → 3+3+1) and asserts full, correct results; my own run against the real dataset exercised the aligned case at scale (48 full chunks + 1 partial). Both pass/agree.
- **Zero-source-entity people**: `test_empty_input_returns_empty_dict` and the "person-c" case in `test_matches_get_for_person_content_and_order` cover this at the unit level; the real-dataset run found 162 such people and produced correct (empty-list, matching-category) results for all of them.
- **`get_todays_birthdays()` identical output**: confirmed above via direct JSON diff against the pre-batching commit — identical.
- **New tests fail if batching is removed**: verified directly. I reverted both call sites in `crm.py` (`list_people`, `get_todays_birthdays`) to call `_person_to_detail_response(p, include_related=False)` without the batched dict, and reran `tests/test_crm_people_list_batching.py` — all 3 tests fail as expected (`assert 0 == 1` on the batch-call-count assertion). Reverted back; `git diff` clean afterward.
- **Latency claims**: reran `tests/test_crm_api.py -k warm_latency` against the real dataset — both pass (`limit=50 < 100ms`, `limit=300 < 350ms`). The PR is transparent that the `limit=300` case doesn't hit an unrequested, self-imposed 150ms target and explains why (SQLite has no per-group top-K for a single `IN (...)`/window-function query, and this dataset's highest-strength people have very large source-entity histories) — this is an honest, well-documented limitation, not a gamed or weakened test, and `limit=300` isn't part of issue #869's stated acceptance criteria (only `limit=50` is), so it doesn't block.

### Test evidence (this pass)

- `pytest tests/test_person_entity.py tests/test_entity_resolver.py tests/test_person_category_response.py tests/test_source_entity.py tests/test_crm_people_list_batching.py tests/test_relationship_metrics.py tests/test_person_stats.py tests/test_merge_people.py tests/test_relationship_discovery.py -q -m "not integration"` → 218 passed, 1 deselected (the one deselected integration test requires operator-specific `settings.current_work_path` config not present on this box — pre-existing, unrelated to this PR).
- `pytest tests/test_crm_api.py -q` (real staging data symlinked) → 36 passed, 1 skipped (pre-existing, unrelated).
- `pytest tests/test_crm_api.py -k warm_latency -q` → 2 passed.
- Three separate revert-and-rerun (mutation) checks, one per fixed finding plus the batching wiring, each producing the expected test failure, then restored cleanly (`git diff` empty after each).
- Full-dataset (14,414 people) category-parity script, written independently of the implementer's own verification.
- Byte-identical JSON diff of `GET /api/crm/people?limit=300&sort=strength` and `GET /birthdays/today` between this HEAD and the pre-batching commit.

### Not verified

- Did not re-run `./scripts/test.sh all` or the pre-push gate (the implementer's own reported run already covers this, and re-running risks lock contention with other in-flight agents per the standing brief).
- Did not re-litigate the `limit=300 < 150ms` aspirational target beyond confirming the PR's own honest accounting of it; treating it as out-of-scope since it isn't in issue #869's acceptance criteria.
- Did not re-check the `test_relationship_metrics.py::TestSelfStrength::test_non_owner_is_not_pinned` flake noted in the original review (0/1 occurrences across this pass's several full runs of that file); still believe it's an unrelated, rare test-isolation issue, not introduced by this PR.

Issue #874, opened 2026-09-04
chore(crm): gzip responses, cancel stale searches, fetch a selected person once, cache the page


nbramia:
## Problem

Several small frontend and transport habits add avoidable latency to every CRM interaction. Responses are not compressed, so a 300-person list is 216 KB and a person's timeline is 620 KB over the wire, which matters on Tailscale from a phone. Typing in the people search box fires a request per pause but never cancels the previous one, so slower earlier responses can overwrite newer results. Clicking a person fetches the same person record twice in a row (once plain, once with related data). The CRM page itself is 750 KB of HTML with no cache policy beyond an ETag, and every switch between Me, Family, Birthdays, and a person page is a full reload of it.

## Solution

The API compresses responses above a small threshold. The people search cancels the in-flight request when a new query starts and ignores responses for superseded queries. Selecting a person issues one detail request with related data and derives the light view from it. Static pages are served with a short cache lifetime so navigations revalidate cheaply.

## Acceptance Criteria

- [ ] When a client sends `Accept-Encoding: gzip`, responses larger than 1 KB from `/api/crm/*` and `/api/people/*` shall be returned with `Content-Encoding: gzip`.
- [ ] Given two searches typed in quick succession, when the first response arrives after the second, the list shall show results for the second query only.
- [ ] When a person is selected, the client shall issue exactly one `GET /api/crm/people/{id}` request (with `include_related=true`), not two.
- [ ] `GET /me`, `/family`, `/crm`, and `/birthdays` shall respond with a `Cache-Control` header allowing revalidation (for example `max-age=60, must-revalidate`) and shall continue to honor `If-None-Match`.
- [ ] Existing browser tests for the CRM shall pass unchanged.

## Verification

- **Automated tests:** an API test for the gzip header; a server-free browser test (pattern from `tests/test_voice_mic_block_ui_browser.py`) that stubs the search endpoint with delayed responses and asserts the final list; a request-count assertion for person selection.
- **Manual checks:** DevTools network tab on `/me`: list response shows `content-encoding: gzip`.
- **Existing test suite:** must pass.

## Scope

**In scope:**
- Gzip middleware, search request cancellation, single person fetch, static cache header, and deleting the two dead helpers noted below.

**Out of scope:**
- Converting dashboard navigation to client-side routing.
- Bundling or splitting the CRM page.

## Technical Context

[...]

PR #887, opened 2026-09-04
fix(crm): run CRM and people handlers off the event loop

### LINKED ISSUE #868: fix(crm): run CRM and people handlers off the event loop so one slow request cannot freeze the API
## Problem

Every CRM and people request runs its database and LLM work directly on the API's single event loop, so one slow CRM request freezes the entire LifeOS API for everyone. While the people list is loading, a request that normally answers in 3 ms takes 3.4 s. While the relationship page's tone analysis runs, a trivial config request waited 14 s. Because chat, Telegram, voice, MCP, and the agent worker all talk to this same process, opening the CRM makes every other surface lag too, and CRM tabs stall each other. This is the mechanism behind "the whole CRM feels unusable," and fixing it alone makes the app responsive even before any individual endpoint gets faster.

Measured on the production host with 14,476 people and 455k interactions:

| Scenario | Latency of an unrelated 3 ms request |
|---|---|
| Idle | 3 ms |
| 4 people-list calls in flight | 3,400 ms |
| Relationship tone analysis in flight | 14,100 ms |

## Solution

CRM and people route handlers that do synchronous work run in the framework's worker threadpool instead of on the event loop. Handlers that contain no awaits become plain synchronous handlers, which FastAPI dispatches to a thread automatically. The few handlers that genuinely await something keep their async signature and push their blocking pieces (LLM calls, full-table loads) onto a thread explicitly. The observable contract of every endpoint is unchanged; only scheduling changes. Concurrent requests then interleave, so a slow CRM aggregate no longer stalls chat, Telegram, MCP, or other CRM tabs.

## Acceptance Criteria

- [ ] Given four concurrent `GET /api/crm/people?limit=300` requests in flight, when `GET /api/crm/config` is requested, it shall complete in under 100 ms.
- [ ] Given a `POST /api/crm/relationship/tone-analysis-detailed` request in flight, when `GET /api/crm/config` is requested, it shall complete in under 100 ms.
- [ ] Every route handler in the CRM and people routers that contains no `await` shall be declared with `def`, not `async def`.
- [ ] Every remaining `async def` handler in those routers shall run its blocking store or LLM calls via `asyncio.to_thread` (or `run_in_threadpool`), never inline.
- [ ] The response bodies and status codes of all CRM and people endpoints shall be byte-identical to before the change for identical inputs (regression guard on the existing API tests).
- [ ] Store objects reached from thread handlers shall not share a SQLite connection across threads; each call keeps opening its own connection as today.

## Verification

- **Automated tests:** a test that starts the app with `TestClient` in a background thread, fires concurrent slow requests, and asserts the fast request's latency bound; a static test that walks the CRM and people routers and asserts no `async def` handler lacks an `await`.
- **Manual checks:** open the CRM graph tab and send a Telegram message to a persona bot at the same time; the bot should answer without waiting for the graph.
- **Existing test suite:** `tests/test_crm_api.py`, `tests/test_people.py`, `tests/test_people_e2e.py` must pass unchanged.

## Scope

**In scope:**
- Handler signatures and thread dispatch in the CRM and people routers.
- A concurrency regression test.

**Out of scope:**
- Making any individual endpoint faster (separate issues cover the people list, network graph, Me/Family aggregates, people search, and tone analysis).
- Running multiple uvicorn workers. In-process singletons (merged-id map, caches) would diverge across workers.

## Technical Context

- **Key files:** `api/routes/crm.py` (62 `async def` handlers; only three contain an `await`: facts extract at line 2005, sources import at line 2219, data-health summary at line 4729), `api/routes/people.py` (12 `async def` handlers, 4 awaits in the v2 wrappers), `api/services/person_entity.py:588` and `api/services/interaction_store.py:373` (a fresh `sqlite3.connect` per call, so thread dispatch is safe).
- **Patterns to follow:** `api/routes/investments.py:163` and `api/routes/chat.py:526` already use `await asyncio.to_thread(...)` for blocking work.
- **Constraints:** no new dependencies. Python-heavy handlers still hold the GIL while they run, so this restores fairness, not throughput; the per-endpoint issues restore throughput. The default anyio threadpool is 40 threads, which is ample for this single-user host.
- **Caveat:** the person entity store singleton (`api/services/person_entity.py:1155`) holds a mutable merged-id dict; reads from threads are fine, writes already happen from request handlers and are single-writer in practice.

## References

- `AGENTS.md` § Server Management (single uvicorn process, systemd).
- `docs/specs/technical/client-surfaces.md` (surfaces sharing this process).

nbramia:
## TL;DR

Opening the CRM used to freeze the rest of LifeOS. A single slow CRM request (a big people list, a relationship tone analysis) blocked chat, Telegram, voice, MCP, and every other CRM tab until it finished — a 3 ms request could take several seconds, or worse, behind it. CRM, people, and photo requests now run off the shared request-handling thread whenever they don't need it, so slow and fast requests no longer queue behind each other. No endpoint's response changed; only how requests are scheduled changed.

Closes #868

## Design

CRM, people, and photo requests are handled by the API's single shared process. A request handler declared to do asynchronous work stays on that process's own scheduling loop even when it does nothing asynchronous — so a handler that scans thousands of records or waits on the local LLM held that loop hostage, and every other request (including ones from a completely different feature) queued behind it.

Handlers with no asynchronous work of their own are now declared as ordinary synchronous handlers, which the framework automatically runs on a worker thread instead of the shared loop. The handful of handlers that do have real asynchronous work (extracting facts via the LLM, importing an uploaded WhatsApp/Signal export, reading an uploaded file) keep their asynchronous form, and their blocking database work — including constructing the underlying data-store objects themselves, not just calling into them — is explicitly pushed onto a worker thread too, so they no longer block the loop either. This restores fairness between concurrent requests; it does not make any individual endpoint faster (that's the scope of the sibling issues #869-#874).

Moving these handlers onto worker threads also removed a mutex they got for free by running start-to-finish on the loop: two requests that write shared on-disk state (a merge, a split, a hide, a review-queue decision, a sync trigger) could previously never interleave. They now hold an explicit lock across their bodies so a concurrent pair still can't. Two more implicit-serialization holes were closed the same way: the lazy singleton getters for the data stores these routers use, and the getter calls inside the two handlers that stayed `async`.

## Implementation Notes

- `api/routes/crm.py`: converted 60 of 62 `async def` handlers with no `await` to plain `def` (matches the issue's count of 59 handlers with none, plus `get_data_health_summary`, whose only `await` was a call to `get_data_health()` — itself converted, so the call site and its caller both dropped `async`). The 2 remaining `async def` handlers keep their real awaits: `extract_person_facts` (LLM call via `extract_facts_async`) and `import_source_data` (`await file.read()`); both push their blocking store/parsing calls, and the store/extractor getter calls themselves, through `asyncio.to_thread` rather than running them inline before/after the await.
- `api/routes/people.py`: converted all 12 `async def` handlers to `def`, including the 4 legacy `/v2/*` wrappers, whose only `await` was calling the (now sync) primary handlers directly — those call sites dropped `await` too.
- `api/routes/photos.py`: converted all 8 `async def` handlers (none had an `await`) to `def`. These back the CRM's profile/thumbnail/shared-photo requests.
- **`_mutation_lock`** (new module-level `threading.Lock` in `api/routes/crm.py`): held across the full body of every handler that mutates shared on-disk state — `merge_people`, `split_person`, `hide_person`, `confirm_review_item`, `reject_review_item`, `sync_source`, `trigger_relationship_discovery`, `update_relationship_strengths`, `sync_slack_users_endpoint`, `sync_contacts_endpoint` — plus `api/routes/photos.py`'s `trigger_photo_sync`, which imports the same lock. Restores the serialization these handlers got for free when they ran inline on the event loop; without it, two concurrent merges could corrupt `scripts/merge_people.py`'s single global merge-intent log and its merged-id read-modify-write.
- Lazy singleton getters (`get_person_entity_store`, `get_interaction_store`, `get_source_entity_store`, `get_relationship_store`, `get_person_fact_store`) now use double-checked locking (`threading.Lock`, one per module) around their `if _x is None: _x = X()` construction, so two first-requests after a restart can't race and double-construct a store.
- `tests/test_me_page.py`: five tests called `get_me_stats`/`get_me_interactions` directly with `await` under `@pytest.mark.asyncio`; updated the call sites to plain calls and dropped the marker/`async` now that those handlers are synchronous. This is the exact "handler calls another handler directly" case the issue flagged, just in test code instead of route code.
- New `tests/test_route_handlers_sync.py`: walks the CRM, people, and photos routers and fails if any `async def` handler's own body contains no `await`/`async for`/`async with` (an AST check, not a substring search — a docstring or comment that merely mentions "await" can't pass it vacuously). Scoped, by design and by the doc wording it enforces, to these three routers; the rest of `api/routes/` has ~95 unconverted `async def`-with-no-`await` handlers that are pre-existing and out of scope for #868.
- New `tests/test_route_handlers_concurrency.py`: builds a router-only app (`FastAPI()` + `include_router(crm_router)`, no lifespan) and enters `TestClient` as a context manager against it — keeping one shared event loop for the client's lifetime reproduces the production single-loop model without booting Telegram/scheduler/job-queue/watchers or writing to a configured vault. Two scenarios, both asserting the issue's actual acceptance bound (`GET /api/crm/config` under 100ms): (1) `person_entity_store.get_all()` patched to sleep, 4 concurrent `GET /api/crm/people?limit=300` in flight; (2) `interaction_store.get_for_person()` patched to sleep, a `POST /api/crm/relationship/tone-analysis-detailed` in flight. Verified both fail against the pre-fix code and pass after.
- New `tests/test_route_handlers_mutation_lock.py`: patches `scripts.merge_people.merge_people` to record whether it was ever re-entered while another call was inside it, fires two concurrent `POST /api/crm/people/merge` requests, and asserts no overlap. Verified it fails with the lock removed and passes with it in place.
- New `tests/test_route_handlers_no_shared_connections.py`: an AST-based static guard over the 7 store modules reachable from these routers (`person_entity`, `interaction_store`, `source_entity`, `relationship`, `person_facts`, `relationship_insights`, `apple_photos`) — fails if any opens a SQLite connection with `check_same_thread=False` or assigns one to a `self.*` attribute (the anti-pattern of caching a connection across calls/threads, instead of the established open-fresh-per-call pattern).
- `docs/specs/technical/architecture.md` and `docs/specs/standards/python-conventions.md`: added a current-state note on the `def` vs. `async def` convention this issue establishes, explicit that it's scoped to these three routers, and a sentence on why mutating handlers now need an explicit lock. Fixed both docs' canonical route-handler-pattern code snippets, which had shown `async def` with no `await` immediately above the new rule.

### Automated tests

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest \
    tests/test_crm_api.py tests/test_people.py tests/test_people_e2e.py tests/test_photos_sync_api.py \
    tests/test_merge_people.py tests/test_split_person.py tests/test_crm_ui_requirements.py \
    tests/test_me_page.py tests/test_route_handlers_sync.py tests/test_route_handlers_concurrency.py \
    tests/test_route_handlers_mutation_lock.py tests/test_route_handlers_no_shared_connections.py -q
...
114 passed, 4 skipped, 1 failed in ...
```
The one failure, `test_me_page.py::TestMyPersonIdConstant::test_my_person_id_is_valid_uuid`, is pre-existing and environment-dependent (needs a real configured `settings.my_person_id`; empty in a clean checkout) — the reviewer independently confirmed this and it is unrelated to this change.

Full unit suite, run clean (no `.env` in the worktree, no concurrent host load this time):
```
$ ./scripts/test.sh unit
========== 5585 passed, 8 skipped, 1181 warnings in 276.77s (0:04:36) ==========
```

### Manual verification

Static guard against the pre-fix code (confirms the new tests actually catch the regression):
```
$ git stash push -- api/routes/crm.py api/routes/people.py api/routes/photos.py
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_route_handlers_concurrency.py tests/test_route_handlers_sync.py -v
...
FAILED tests/test_route_handlers_concurrency.py::test_fast_config_request_not_blocked_by_concurrent_slow_people_list
FAILED tests/test_route_handlers_concurrency.py::test_fast_config_request_not_blocked_by_concurrent_tone_analysis
FAILED tests/test_route_handlers_sync.py::test_async_handlers_all_contain_an_await[crm-router0] - ... 59 handlers listed
FAILED tests/test_route_handlers_sync.py::test_async_handlers_all_contain_an_await[people-router1] - ... 8 handlers listed
FAILED tests/test_route_handlers_sync.py::test_async_handlers_all_contain_an_await[photos-router2] - ... 8 handlers listed
$ git stash pop
```

Mutation-lock guard against the lock removed (confirms `test_concurrent_merges_serialize` actually catches the regression):
```
AssertionError: two concurrent merge requests ran scripts.merge_people.merge_people at the same
time — _mutation_lock in api/routes/crm.py is not serializing them
```

Controlled byte-identical check — same frozen staging data, pre-fix code on :8012 vs. post-fix code on :8011, diffing `GET /api/crm/people?limit=20`, `/api/crm/people/{id}`, `/timeline`, `/connections`, `/facts`, `/api/crm/statistics`, `/api/crm/birthdays/all`:
```
people: byte_identical=True struct_identical=True
person: byte_identical=True struct_identical=True
timeline: byte_identical=True struct_identical=True
connections: byte_identical=True struct_identical=True
facts: byte_identical=True struct_identical=True
statistics: byte_identical=True struct_identical=True
birthdays: byte_identical=True struct_identical=True

ALL IDENTICAL
```
This isolates the code change from data drift. A separate diff between the fixed staging instance and live production (`:8000`) showed only `relationship_strength` values and category counts differing — expected, since production continued receiving real syncs and strength recalculations between when the staging copy was taken and when this check ran; JSON shape and every other field matched. (The reviewer independently ran a broader 40-endpoint byte-identity check plus an OpenAPI-schema hash comparison and found the same: fully identical.)

Real-world concurrency demonstration (real staging data, no monkeypatch — 4 concurrent `GET /api/crm/people?limit=10000` in flight, each individually ~6.2s):
```
Pre-fix  (:8012): GET /api/crm/config took 24.947s
Post-fix (:8011): GET /api/crm/config took 0.030s
```

`extract_person_facts`'s and `import_source_data`'s new `to_thread`-wrapped getter paths were smoke-tested directly (no LLM/upload infra needed for this): a 404 lookup through `extract_person_facts` and a full synthetic-WhatsApp-content run through `import_source_data` both completed correctly with the getters now running inside the thread hop.

## Review focus

- `api/routes/crm.py`'s `_mutation_lock` scope — it's one coarse lock across all the mutating handlers named above (including `photos.py`'s sync trigger) rather than one per handler, on the reasoning that several of them can touch overlapping state (a merge and a split both touch `person_store`) and a single-user host pays essentially nothing for serializing writes that used to be serialized anyway. Flag if a finer-grained (or differently-scoped) lock would be preferable.
- `api/routes/crm.py`'s `get_data_health_summary` / `get_data_health` pair and `import_source_data` / `extract_person_facts` — these are the handlers where the mechanical "drop `async`" rule needed a second pass (a caller's own `await` disappearing once its callee went sync, blocking calls that needed an explicit `asyncio.to_thread`, and now the getter calls themselves moved into that hop too).
- The static "no shared connection" guard (`tests/test_route_handlers_no_shared_connections.py`) hardcodes the 7 store modules these routers currently reach; if a new store gets added to the call graph, this list needs a manual update — there's no automated reachability check.

### DIFF
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -8,6 +8,7 @@ from collections import defaultdict
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional
+import asyncio
 import json
 import logging
 import sqlite3
@@ -532,7 +533,7 @@ def _get_partner_name() -> str:
 
 
 @router.get("/config", response_model=CRMConfigResponse)
-async def get_crm_config():
+def get_crm_config():
     """
     Get CRM configuration values for frontend.
 
@@ -549,7 +550,7 @@ async def get_crm_config():
 
 
 @router.get("/birthdays/today")
-async def get_todays_birthdays():
+def get_todays_birthdays():
     """Get all people with birthdays today."""
     person_store = get_person_entity_store()
     people = person_store.get_all()
@@ -565,7 +566,7 @@ async def get_todays_birthdays():
 
 
 @router.get("/birthdays/all")
-async def get_all_birthdays():
+def get_all_birthdays():
     """Get all people with birthdays, grouped by date."""
     person_store = get_person_entity_store()
     people = person_store.get_all()
@@ -593,7 +594,7 @@ async def get_all_birthdays():
 
 
 @router.get("/people", response_model=PersonListResponse)
-async def list_people(
+def list_people(
     q: Optional[str] = Query(default=None, description="Search query"),
     category: Optional[str] = Query(default=None, description="Filter by category"),
     source: Optional[str] = Query(default=None, description="Filter by source"),
@@ -707,7 +708,7 @@ async def list_people(
 
 
 @router.get("/people/{person_id}", response_model=PersonDetailResponse)
-async def get_person(
+def get_person(
     person_id: str,
     include_related: bool = Query(default=False, description="Include source entities and relationships"),
     refresh_strength: bool = Query(default=False, description="Recompute relationship strength (slower)"),
@@ -746,7 +747,7 @@ async def get_person(
 
 
 @router.patch("/people/{person_id}", response_model=PersonDetailResponse)
-async def update_person(person_id: str, request: PersonUpdateRequest):
+def update_person(person_id: str, request: PersonUpdateRequest):
     """
     Update a person's notes, tags, or category.
     """
@@ -807,7 +808,7 @@ class PersonMergeResponse(BaseModel):
 
 
 @router.post("/people/merge", response_model=PersonMergeResponse)
-async def merge_people(request: PersonMergeRequest):
+def merge_people(request: PersonMergeRequest):
     """
     Merge multiple people into a single record.
 
@@ -1064,7 +1065,7 @@ class ContactSource(BaseModel):
 
 
 @router.get("/people/{person_id}/contact-sources")
-async def get_person_contact_sources(person_id: str):
+def get_person_contact_sources(person_id: str):
     """
     Get aggregated contact sources linked to a person.
 
@@ -1190,7 +1191,7 @@ async def get_person_contact_sources(person_id: str):
 
 
 @router.get("/people/{person_id}/source-entities")
-async def get_person_source_entities(
+def get_person_source_entities(
     person_id: str,
     limit: int = Query(default=500, ge=1, le=5000, description="Max source entities to return"),
     offset: int = Query(default=0, ge=0, description="Offset for pagination"),
@@ -1277,7 +1278,7 @@ async def get_person_source_entities(
 
 
 @router.post("/people/split", response_model=PersonSplitResponse)
-async def split_person(request: PersonSplitRequest):
+def split_person(request: PersonSplitRequest):
     """
     Split source entities from one person to another.
 
@@ -1552,7 +1553,7 @@ async def split_person(request: PersonSplitRequest):
 
 
 @router.get("/people/{person_id}/timeline", response_model=TimelineResponse)
-async def get_person_timeline(
+def get_person_timeline(
     person_id: str,
     source_type: Optional[str] = Query(
         default=None,
@@ -1634,7 +1635,7 @@ SOURCE_BADGES = {
 
 
 @router.get("/people/{person_id}/timeline/aggregated", response_model=AggregatedTimelineResponse)
-async def get_person_timeline_aggregated(
+def get_person_timeline_aggregated(
     person_id: str,
     source_type: Optional[str] = Query(
         default=None,
@@ -1798,7 +1799,7 @@ async def get_person_timeline_aggregated(
 
 
 @router.get("/people/{person_id}/connections", response_model=ConnectionsResponse)
-async def get_person_connections(
+def get_person_connections(
     person_id: str,
     relationship_type: Optional[str] = Query(default=None, description="Filter by type"),
     limit: int = Query(default=50, ge=1, le=200, description="Max results"),
@@ -1866,7 +1867,7 @@ async def get_person_connections(
 
 
 @router.get("/people/{person_id}/strength", response_model=dict)
-async def get_person_strength_breakdown(person_id: str):
+def get_person_strength_breakdown(person_id: str):
     """
     Get detailed breakdown of relationship strength components.
 
@@ -1905,7 +1906,7 @@ def _fact_to_response(fact: PersonFact) -> PersonFactResponse:
 
 
 @router.get("/people/{person_id}/facts", response_model=PersonFactsResponse)
-async def get_person_facts(person_id: str):
+def get_person_facts(person_id: str):
     """
     Get all facts about a person.
 
@@ -1966,14 +1967,15 @@ async def extract_person_facts(person_id: str, model: Optional[str] = None):
     elif model == "haiku":
         model = PersonFactExtractor.MODEL_HAIKU
     person_store = get_person_entity_store()
-    person = person_store.get_by_id(person_id)
+    person = await asyncio.to_thread(person_store.get_by_id, person_id)
 
     if not person:
         raise HTTPException(status_code=404, detail=f"Person '{person_id}' not found")
 
     # Get ALL interactions for the person (extractor will sample strategically)
     interaction_store = get_interaction_store()
-    interactions = interaction_store.get_for_person(
+    interactions = await asyncio.to_thread(
+        interaction_store.get_for_person,
         person_id,
         days_back=3650,  # Look back 10 years for full history
         limit=100000,  # No practical limit - let extractor sample
@@ -2020,7 +2022,7 @@ async def extract_person_facts(person_id: str, model: Optional[str] = None):
 
 
 @router.put("/people/{person_id}/facts/{fact_id}", response_model=PersonFactResponse)
-async def update_person_fact(person_id: str, fact_id: str, request: FactUpdateRequest):
+def update_person_fact(person_id: str, fact_id: str, request: FactUpdateRequest):
     """
     Update a fact's value or metadata.
     """
@@ -2051,7 +2053,7 @@ async def update_person_fact(person_id: str, fact_id: str, request: FactUpdateRe
 
 
 @router.delete("/people/{person_id}/facts/{fact_id}")
-async def delete_person_fact(person_id: str, fact_id: str):
+def delete_person_fact(person_id: str, fact_id: str):
     """
     Delete a fact.
     """
@@ -2070,7 +2072,7 @@ async def delete_person_fact(person_id: str, fact_id: str):
 
 
 @router.post("/people/{person_id}/facts/{fact_id}/confirm")
-async def confirm_person_fact(person_id: str, fact_id: str):
+def confirm_person_fact(person_id: str, fact_id: str):
     """
     Mark a fact as confirmed by user.
 
@@ -2091,7 +2093,7 @@ async def confirm_person_fact(person_id: str, fact_id: str):
 
 
 @router.post("/people/{person_id}/hide")
-async def hide_person(person_id: str, request: HidePersonRequest):
+def hide_person(person_id: str, request: HidePersonRequest):
     """
     Hide a person (soft delete) and blocklist their identifiers.
 
@@ -2132,7 +2134,7 @@ async def hide_person(person_id: str, request: HidePersonRequest):
 
 
 @router.get("/discover", response_model=DiscoverResponse)
-async def discover_connections(
+def discover_connections(
     person_id: Optional[str] = Query(default=None, description="Person to find suggestions for"),
     limit: int = Query(default=10, ge=1, le=50, description="Max suggestions"),
 ):
@@ -2234,14 +2236,15 @@ async def import_source_data(
 
     if source_type == "whatsapp":
         from api.services.whatsapp_import import import_whatsapp_export
-        stats = import_whatsapp_export(
+        stats = await asyncio.to_thread(
+            import_whatsapp_export,
             content_str,
             file.filename or "chat.txt",
             source_store,
         )
     else:  # signal
         from api.services.signal_import import import_signal_export
-        stats = import_signal_export(content_str, source_store)
+        stats = await asyncio.to_thread(import_signal_export, content_str, source_store)
 
     return {
         "status": "completed",
@@ -2252,7 +2255,7 @@ async def import_source_data(
 
 
 @router.post("/sources/{source_type}/sync")
-async def sync_source(source_type: str):
+def sync_source(source_type: str):
     """
     Trigger a sync for a specific data source.
     """
@@ -2273,7 +2276,7 @@ async def sync_source(source_type: str):
 
 
 @router.post("/relationships/discover")
-async def trigger_relationship_discovery():
+def trigger_relationship_discovery():
     """
     Trigger relationship discovery across all sources.
 
@@ -2294,7 +2297,7 @@ async def trigger_relationship_discovery():
 
 
 @router.post("/strengths/update")
-async def update_relationship_strengths():
+def update_relationship_strengths():
     """
     Update relationship strength scores for all people.
     """
@@ -2306,7 +2309,7 @@ async def update_relationship_strengths():
 
 
 @router.get("/statistics", response_model=StatisticsResponse)
-async def get_crm_statistics():
+def get_crm_statistics():
     """
     Get comprehensive CRM statistics.
     """
@@ -2335,7 +2338,7 @@ async def get_crm_statistics():
 
 
 @router.get("/network", response_model=NetworkGraphResponse)
-async def get_network_graph(
+def get_network_graph(
     center_on: Optional[str] = Query(default=None, description="Person ID to center the graph on"),
     depth: int = Query(default=2, ge=1, le=4, description="Hops from center person"),
     min_strength: float = Query(default=0.0, ge=0.0, le=1.0, description="Minimum relationship strength"),
@@ -2521,7 +2524,7 @@ class RelationshipDetailResponse(BaseModel):
 
 
 @router.get("/relationship/{person_a_id}/{person_b_id}", response_model=RelationshipDetailResponse)
-async def get_relationship_details(person_a_id: str, person_b_id: str):
+def get_relationship_details(person_a_id: str, person_b_id: str):
     """
     Get detailed information about the relationship between two people.
 
@@ -2608,7 +2611,7 @@ async def get_relationship_details(person_a_id: str, person_b_id: str):
 # =======================
 
 @router.get("/slack/status")
-async def get_slack_status():
+def get_slack_status():
     """
     Get Slack integration status.
 
@@ -2627,7 +2630,7 @@ async def get_slack_status():
 
 
 @router.get("/slack/oauth/start")
-async def start_slack_oauth(state: Optional[str] = None):
+def start_slack_oauth(state: Optional[str] = None):
     """
     Start Slack OAuth flow.
 
@@ -2646,7 +2649,7 @@ async def start_slack_oauth(state: Optional[str] = None):
 
 
 @router.get("/slack/callback")
-async def slack_oauth_callback(code: str, state: Optional[str] = None):
+def slack_oauth_callback(code: str, state: Optional[str] = None):
     """
     Handle Slack OAuth callback.
 
@@ -2670,7 +2673,7 @@ async def slack_oauth_callback(code: str, state: Optional[str] = None):
 
 
 @router.post("/slack/sync")
-async def sync_slack_users_endpoint(workspace_id: str = "default"):
+def sync_slack_users_endpoint(workspace_id: str = "default"):
     """
     Sync Slack users to the CRM.
 
@@ -2703,7 +2706,7 @@ async def sync_slack_users_endpoint(workspace_id: str = "default"):
 
 
 @router.delete("/slack/disconnect")
-async def disconnect_slack(workspace_id: str = "default"):
+def disconnect_slack(workspace_id: str = "default"):
     """
     Disconnect a Slack workspace.
 
@@ -2724,7 +2727,7 @@ async def disconnect_slack(workspace_id: str = "default"):
 # ==================================
 
 @router.get("/contacts/status")
-async def get_contacts_status():
+def get_contacts_status():
     """
     Get Apple Contacts integration status.
 
@@ -2741,7 +2744,7 @@ async def get_contacts_status():
 
 
 @router.post("/contacts/sync")
-async def sync_contacts_endpoint():
+def sync_contacts_endpoint():
     """
     Sync Apple Contacts to the CRM.
 
@@ -2936,7 +2939,7 @@ class FamilyInteractionsResponse(BaseModel):
 
 
 @router.get("/me/stats", response_model=MeStatsResponse)
-async def get_me_stats():
+def get_me_stats():
     """
     Get aggregate statistics for the owner's personal dashboard.
 
@@ -2963,7 +2966,7 @@ async def get_me_stats():
 
 
 @router.get("/me/timeline", response_model=TimelineResponse)
-async def get_me_timeline(
+def get_me_timeline(
     source_type: Optional[str] = Query(
         default=None,
         description="Filter by source type. Supports comma-separated values for compound "
@@ -3046,7 +3049,7 @@ async def get_me_timeline(
 
 
 @router.get("/me/interactions", response_model=MeInteractionsResponse)
-async def get_me_interactions(
+def get_me_interactions(
     days_back: int = Query(default=365, ge=1, le=3660, description="Days of history (up to 10 years)"),
     trend_period: str = Query(default="quarter", description="Trend comparison period: week, month, quarter, year"),
     health_period: str = Query(default="quarter", description="Health score history period: month, quarter, year"),
@@ -3605,7 +3608,7 @@ async def get_me_interactions(
 
 
 @router.get("/family/members", response_model=FamilyMembersResponse)
-async def get_family_members():
+def get_family_members():
     """
     Get all family members for the multi-select dropdown and visualizations.
 
@@ -3701,7 +3704,7 @@ async def get_family_members():
 
 
 @router.get("/family/stats", response_model=FamilyStatsResponse)
-async def get_family_stats(
+def get_family_stats(
     person_ids: str = Query(
         ...,
         description="Comma-separated list of person IDs to get lifetime stats for"
@@ -3740,7 +3743,7 @@ async def get_family_stats(
 
 
 @router.get("/family/timeline", response_model=TimelineResponse)
-async def get_family_timeline(
+def get_family_timeline(
     person_ids: str = Query(
         ...,
         description="Comma-separated list of person IDs to include in timeline"
@@ -3822,7 +3825,7 @@ async def get_family_timeline(
 
 
 @router.get("/family/interactions", response_model=FamilyInteractionsResponse)
-async def get_family_interactions(
+def get_family_interactions(
     person_ids: str = Query(
         ...,
         description="Comma-separated list of person IDs to aggregate"
@@ -4067,7 +4070,7 @@ class CommunicationGapsResponse(BaseModel):
 
 
 @router.get("/family/communication-gaps", response_model=CommunicationGapsResponse)
-async def get_family_communication_gaps(
+def get_family_communication_gaps(
     person_ids: str = Query(
         ...,
         description="Comma-separated list of person IDs"
@@ -4188,7 +4191,7 @@ class ChannelMixResponse(BaseModel):
 
 
 @router.get("/family/channel-mix", response_model=ChannelMixResponse)
-async def get_family_channel_mix(
+def get_family_channel_mix(
     person_ids: str = Query(
         ...,
         description="Comma-separated list of person IDs"
@@ -4276,7 +4279,7 @@ class SyncErrorResponse(BaseModel):
 
 
 @router.get("/sync/health", response_model=list[SyncHealthResponse])
-async def get_all_sync_health():
+def get_all_sync_health():
     """
     Get health status for all sync sources.
 
@@ -4303,7 +4306,7 @@ async def get_all_sync_health():
 
 
 @router.get("/sync/health/summary", response_model=SyncHealthSummaryResponse)
-async def get_sync_health_summary():
+def get_sync_health_summary():
     """
     Get summary of sync health across all sources.
 
@@ -4318,7 +4321,7 @@ async def get_sync_health_summary():
 
 
 @router.get("/sync/health/{source}", response_model=SyncHealthResponse)
-async def get_source_sync_health(source: str):
+def get_source_sync_health(source: str):
     """
     Get health status for a specific sync source.
     """
@@ -4345,7 +4348,7 @@ async def get_source_sync_health(source: str):
 
 
 @router.get("/sync/errors", response_model=list[SyncErrorResponse])
-async def get_sync_errors(
+def get_sync_errors(
     source: Optional[str] = Query(default=None, description="Filter by source"),
     limit: int = Query(default=50, ge=1, le=200, description="Max results"),
 ):
@@ -4372,7 +4375,7 @@ async def get_sync_errors(
 
 
 @router.get("/sync/stale", response_model=list[SyncHealthResponse])
-async def get_stale_syncs():
+def get_stale_syncs():
     """
     Get list of syncs that are stale (>24 hours old).
 
@@ -4425,7 +4428,7 @@ class ReviewQueueResponse(BaseModel):
 
 
 @router.get("/review-queue", response_model=ReviewQueueResponse)
-async def get_review_queue(
+def get_review_queue(
     min_confidence: float = Query(default=0.0, ge=0.0, le=1.0, description="Minimum confidence"),
     max_confidence: float = Query(default=0.85, ge=0.0, le=1.0, description="Maximum confidence"),
     link_method: Optional[str] = Query(default=None, description="Filter by resolution method (e.g. email_exact, name_fuzzy)"),
@@ -4490,7 +4493,7 @@ async def get_review_queue(
 
 
 @router.post("/review-queue/{entity_id}/confirm")
-async def confirm_review_item(entity_id: str):
+def confirm_review_item(entity_id: str):
     """
     Confirm a low-confidence match as correct.
 
@@ -4517,7 +4520,7 @@ async def confirm_review_item(entity_id: str):
 
 
 @router.post("/review-queue/{entity_id}/reject")
-async def reject_review_item(entity_id: str, request: LinkConfirmRequest):
+def reject_review_item(entity_id: str, request: LinkConfirmRequest):
     """
     Reject a low-confidence match.
 
@@ -4585,7 +4588,7 @@ async def reject_review_item(entity_id: str, request: LinkConfirmRequest):
 
 
 @router.get("/data-health")
-async def get_data_health():
+def get_data_health():
     """
     Get comprehensive data health statistics.
 
@@ -4724,9 +4727,9 @@ async def get_data_health():
 
 
 @router.get("/data-health/summary")
-async def get_data_health_summary():
+def get_data_health_summary():
     """Get a brief summary of data health for the UI header."""
-    health = await get_data_health()
+    health = get_data_health()
 
     return {
         "total_interactions": sum(
@@ -4758,7 +4761,7 @@ class LinkOverrideResponse(BaseModel):
 
 
 @router.get("/link-overrides")
-async def get_link_overrides(person_id: Optional[str] = Query(default=None)):
+def get_link_overrides(person_id: Optional[str] = Query(default=None)):
     """
     Get all link override rules.
 
@@ -4805,7 +4808,7 @@ async def get_link_overrides(person_id: Optional[str] = Query(default=None)):
 
 
 @router.delete("/link-overrides/{override_id}")
-async def delete_link_override(override_id: str):
+def delete_link_override(override_id: str):
     """Delete a link override rule."""
     from api.services.link_override import get_link_override_store
 
@@ -4898,7 +4901,7 @@ class ToneAnalysisDetailedResponse(BaseModel):
 
 
 @router.get("/relationship/insights", response_model=RelationshipInsightsResponse)
-async def get_relationship_insights(person_id: Optional[str] = None):
+def get_relationship_insights(person_id: Optional[str] = None):
     """
     Get all relationship insights for the configured partner.
 
@@ -4939,7 +4942,7 @@ async def get_relationship_insights(person_id: Optional[str] = None):
 
 
 @router.post("/relationship/insights/generate", response_model=RelationshipInsightsResponse)
-async def generate_relationship_insights(
+def generate_relationship_insights(
     person_id: Optional[str] = None,
     category: Optional[str] = None
 ):
@@ -4995,7 +4998,7 @@ async def generate_relationship_insights(
 
 
 @router.post("/relationship/insights/{insight_id}/confirm")
-async def confirm_relationship_insight(insight_id: str):
+def confirm_relationship_insight(insight_id: str):
     """
     Mark an insight as confirmed.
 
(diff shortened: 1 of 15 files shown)


nbramia:
## Adversarial review of #887 by Opus

**Verdict:** BLOCK

The production change itself is good: I independently confirmed the fix works (including the acceptance criterion the PR does not test) and that 38 endpoints plus the whole OpenAPI schema are byte-identical to `origin/feat/crm-performance` on the real staging dataset. What blocks is the new concurrency test, which boots the entire application — with real side effects on any machine that has a configured `.env` — and the fact that neither acceptance-criterion latency bound is actually asserted anywhere.

1. **[BLOCKER] The new "unit" test boots the whole application and writes to the configured vault, starts the Telegram listeners, the scheduler, the job-queue worker and the file watchers** — `tests/test_route_handlers_concurrency.py:57` (`with TestClient(app)`). This is the only test in the suite that enters `TestClient` as a context manager, so it is the only one that runs `api/main.py`'s `lifespan`. Reproduced in isolation in my worktree:
   ```
   $ rm -rf vault && pytest tests/test_route_handlers_concurrency.py -q
   1 passed
   $ find vault -type f
   vault/LifeOS/Scheduler/Dashboard.md
   vault/LifeOS/Tasks/Dashboard.md
   ```
   With `-s --log-cli-level=INFO` the lifespan log shows: `Calendar indexer scheduler started`, the Telegram listener startup path, `Rebuilt scheduler index`, `Scheduler watcher started`, `Job queue worker started`, `Task watcher started`. In a bare worktree these are harmless no-ops. On the production checkout — where `./scripts/deploy.sh` and `./scripts/test.sh` run per AGENTS.md § Boundaries ("Run full test suite before commit (on the server)") — `.env` supplies real bot tokens, the real vault path and the live DBs, so this test would: rewrite the real `LifeOS/Tasks/Dashboard.md` and `LifeOS/Scheduler/Dashboard.md`; start long-polling with the production Telegram tokens *alongside the running server's listeners* (two `getUpdates` consumers on one token conflict); start the scheduler and the job-queue worker against the live `data/jobs.db`; and run `recover_incomplete_merge()` against the live `data/crm.db`. It also makes the test host-dependent: if `LIFEOS_SERVER_HOSTNAME` is set and the suite runs anywhere else (e.g. the `remote-test.sh` runner), `check_server_host_guard()` raises `RuntimeError` at `TestClient.__enter__` and the test errors for reasons unrelated to the code under test.
   **Fix (I verified it works):** the lifespan is not needed to reproduce the bug — only the shared portal/event loop is. A router-only app reproduces it exactly:
   ```python
   app = FastAPI(); app.include_router(crm_router)
   ```
   ```
   base  (origin/feat/crm-performance): config status=200 elapsed=1.304s
   PR    (fix/crm-handlers-off-loop):   config status=200 elapsed=0.003s
   ```
   Same discrimination, zero side effects, no host guard, no cross-test global-state mutation. Please switch to that. (Related: the PR dismisses 25 failures in `test_agent_worker_*` / `test_operator_spawn.py` from its second full run as "host-load flakiness". A rerun in isolation cannot distinguish host load from pollution by a test that now starts and stops app-wide singletons mid-suite. I ran those four files in one process directly after the concurrency test — 202 passed — so I could not reproduce pollution, but the dismissal is not established either; removing the lifespan removes the question.)

2. **[BLOCKER] Acceptance criterion 2 has no test, and criterion 1's test does not assert criterion 1's bound.** The issue asks for `< 100 ms` in two scenarios; the PR tests one scenario at `< 750 ms` (`FAST_REQUEST_BOUND_SECONDS = 0.75`, `tests/test_route_handlers_concurrency.py:38`) and does not test the tone-analysis scenario at all — the scenario the issue measured at 14.1 s and the one that motivated converting `analyze_relationship_tone_detailed`. I verified the code satisfies both criteria (see "What I verified"), so this is a test gap, not a behaviour gap:
   - tone-analysis in flight → `GET /api/crm/config`: **2705.3 ms on base, 3.2 ms on the PR**.
   - The 0.75 s bound is not flaky — it is far too loose. Repeating the experiment 8× with 32 CPU burners saturating all 32 cores gave `min=0.001s med=0.002s max=0.007s`. That is a 100× margin under deliberate saturation, so please tighten to `0.1` to actually assert the criterion, and add the tone-analysis case (monkeypatch `interaction_store.get_for_person` to sleep — that is the handler's first blocking call, and it keeps the test independent of `data/`).

3. **[MAJOR] The change silently removes the implicit global mutex that the event loop gave every *mutating* handler; concurrent merges can now corrupt merge bookkeeping.** An `async def` handler with no `await` ran start-to-finish without yielding, so `merge_people`, `split_person`, `hide_person`, the review-queue confirm/reject and the sync-trigger POSTs were atomic with respect to every other request. As plain `def` handlers they now interleave. `api/routes/crm.py:809` delegates to `scripts/merge_people.merge_people`, which keeps **one** global intent log (`data/merge_log.json`, written at `scripts/merge_people.py:455` and phase-updated at 464/493/685/705) and does a read-modify-write of `data/merged_person_ids.json` (`load_merged_ids()` → `save_merged_ids()`). With two concurrent merge POSTs: B's `write_merge_intent` overwrites A's, so crash recovery can only ever recover one and the other is left half-applied; whichever finishes first calls `clear_merge_log()` (`scripts/merge_people.py:706`), after which the other's next `update_merge_phase()` raises `RuntimeError("No merge log to update")` mid-merge — a 500 *after* the DB mutations have partly landed; and the merged-ids read-modify-write is last-writer-wins, so a mapping can be silently dropped and the next sync recreates the duplicate the user just merged away. Neither the issue nor the PR considered write handlers — the issue's caveat covers only *reads* of the merged-id dict, and the PR's "Review focus" repeats that framing. Likelihood is low (the UI sends one merge request and disables the button; the sync-trigger POSTs have no caller in `web/` at all), but the failure mode is silent data damage. Suggested fix: a module-level `threading.Lock` held across the merge/split/hide write paths (cheap, single-user host), and a sentence in the architecture note saying that moving handlers off the loop removed the implicit serialization so mutating handlers must lock.

4. **[MINOR] Lazy singleton getters are now called concurrently from worker threads with no lock.** `get_person_entity_store` (`api/services/person_entity.py:1155`), `get_interaction_store`, `get_source_entity_store`, `get_relationship_store`, `get_person_fact_store` are all `if _x is None: _x = X()`. On the loop that check-and-set was atomic; from threads two concurrent first-requests after a restart can construct two stores and run `_init_db()` concurrently against the same file. The `ALTER TABLE` migrations are wrapped in `except sqlite3.OperationalError: pass`, but the `conn.executescript(...)` CREATE TABLE/INDEX block (`api/services/person_entity.py:511`) is not, so a DDL lock collision surfaces as a 500 on the first CRM request rather than being swallowed. One line each (`threading.Lock()` around the getter body), or warm the stores in the lifespan.

5. **[MINOR] The two handlers that stay `async` still make blocking calls inline on the loop.** `extract_person_facts` calls `get_person_entity_store()`, `get_interaction_store()` and `get_person_fact_extractor()` directly, and `import_source_data` calls `get_source_entity_store()` directly (`api/routes/crm.py:1969`, `2235`). On a cold singleton those run `_init_db()` + `_load_blocklist()` + `_load_merged_ids()` on the event loop — the exact class of call this PR is removing. Acceptance criterion 4 says these handlers must run their blocking store calls via a thread, "never inline". Either fold them into the existing `to_thread` hop or warm the singletons at startup.

6. **[MINOR] The static guard is a substring match, so it can pass vacuously.** `tests/test_route_handlers_sync.py:63` does `if "await" not in inspect.getsource(endpoint)`. A future `async def` handler whose docstring says "awaits the LLM" (or that merely mentions `await` in a comment) passes with no real `await`. An AST check is exact and no longer: `any(isinstance(n, (ast.Await, ast.AsyncFor, ast.AsyncWith)) for n in ast.walk(node))`. Relatedly, the guard covers three routers while both docs state the convention project-wide; that AST scan finds **95** `async def` route handlers with no `await` still in the other 23 modules of `api/routes/` (admin 10, agents 10, tasks 8, calendar 7, conversations 6, slack 6, chat 5, …). Out of scope to convert, but either scope the doc wording to these three routers or extend the guard with an explicit allowlist so the rule cannot silently regress elsewhere.

7. **[MINOR] Both docs the PR edits still show a `async def`-with-no-`await` handler as the canonical example, immediately adjacent to the new rule.** `docs/specs/technical/architecture.md:262` (`async def endpoint_handler(...)`, body is `service_function(param)` plus logging) sits directly above the new "`def` vs. `async def`" paragraph, and `docs/specs/standards/python-conventions.md:77` (`async def list_tasks`) sits directly above the new bullet — while line 70 of the same file was changed to "Routes are thin functions". These snippets are what the next handler gets copied from, so they should be `def`.

8. **[MINOR] The de-awaited tests are still `async def` with `@pytest.mark.asyncio` and now contain no `await`.** `tests/test_me_page.py:49-50, 63-64, 144-145, 170-171, 186-187` — the same anti-pattern the PR's own new guard forbids for handlers. Drop the marker and the `async` while you are in there.

9. **[MINOR] Acceptance criterion 6 (no SQLite connection shared across threads) has no test.** I verified it by inspection and it holds: every store reachable from these routers opens a fresh `sqlite3.connect` per call (`person_entity.py:588`, `interaction_store.py:373`, `source_entity.py:322`, `relationship.py:382`, `person_facts.py:231`, `relationship_insights.py:187`, `apple_photos.py:84`), and the only cached `check_same_thread=False` connection anywhere in `api/services` is `cc_wezterm_store.py:56`, which these routers never touch. A tiny static guard (no `check_same_thread=False` / no cached connection attribute in the stores these routers reach) would lock the invariant in.

Scope note, not a finding: `api/routes/photos.py` is not named in #868 (which says "the CRM and people routers"). It is a defensible extension — the CRM's avatars and photo tabs come from it — and my byte-identity check covers it, so I would keep it; just flagging that it is beyond the issue's literal text.

### What I verified and how

- `git fetch origin fix/crm-handlers-off-loop` + local `review-887` branch (the PR branch was already checked out in a sibling worktree); base = `origin/feat/crm-performance` @ `388811e`, which is the PR's merge-base — the PR is exactly one commit on top, no drift.
- **Byte-identical responses, both branches, same frozen staging data (14,414 people), in-process `TestClient`** (no lifespan, so no side effects): SHA-256 of the response body + status for **40** endpoints. `identical=38 differing=0`, after excluding two endpoints I first proved nondeterministic *on the PR branch alone* by running the capture twice (`/people/{id}/strength` and `/sync/health`, both time-derived). Covered: `people` list, person detail (with and without `include_related`), timeline, aggregated timeline, connections, facts, strength, contact-sources, source-entities, statistics, birthdays (all + today), me/stats, me/timeline, me/interactions, config, data-health, **data-health/summary** (the async→sync caller change), network, review-queue, sync health/summary/errors/stale, link-overrides, discover, family/members, relationship/insights, slack/contacts status, `/api/people` list/statistics/search/entity/entity-interactions/v2-statistics/v2-entity, `/api/photos` stats + people. That covers the five you asked for and then some.
- **OpenAPI schema:** `sha256(json.dumps(openapi, sort_keys=True))` identical on both branches, 205 paths — no signature, response-model or validation change leaked in.
- **Acceptance criterion 1 (mechanism):** router-only app, `store.get_all` patched to sleep 1.5 s, 4 concurrent `GET /api/crm/people?limit=300`, then `GET /api/crm/config` → base **1.304 s**, PR **0.003 s**.
- **Acceptance criterion 2 (untested by the PR):** router-only app, `interaction_store.get_for_person` patched to sleep 3 s, `POST /api/crm/relationship/tone-analysis-detailed` in flight, then `GET /api/crm/config` → base **2705.3 ms**, PR **3.2 ms**. Criterion met by the code.
- **Flakiness probe for the 0.75 s bound:** the criterion-1 experiment repeated 8× idle → `min=0.002 med=0.002 max=0.003s`; 8× with 32 busy-loop processes on a 32-core box → `min=0.001 med=0.002 max=0.007s`. Not flaky; just 7.5× looser than the criterion.
- **Lifespan side effects:** `rm -rf vault && pytest tests/test_route_handlers_concurrency.py -q` → 1 passed, and `find vault -type f` then lists two Dashboard files the test created. `-s --log-cli-level=INFO` shows the calendar scheduler, Telegram listener startup, scheduler index rebuild + watcher, job-queue worker and task watcher all starting and stopping. `grep -rn "with TestClient(" tests/` → this file is the only match in the suite.
- **Tests:** `pytest tests/test_crm_api.py tests/test_people.py tests/test_people_e2e.py tests/test_me_page.py tests/test_photos_sync_api.py tests/test_crm_ui_requirements.py tests/test_split_person.py tests/test_merge_people.py tests/test_route_handlers_sync.py tests/test_route_handlers_concurrency.py -q` → **112 passed, 4 skipped, 1 failed**. The one failure is `test_me_page.py::TestMyPersonIdConstant::test_my_person_id_is_valid_uuid`, which its own comment documents as needing a real configured `settings.my_person_id` ("empty (not a UUID) in a clean checkout") — environment-dependent and pre-existing, unrelated to this PR.
- **Cross-test pollution probe:** `pytest tests/test_route_handlers_concurrency.py tests/test_agent_worker_preflight.py tests/test_agent_worker_clarifications.py tests/test_agent_worker_loop.py tests/test_operator_spawn.py -q -p no:randomly` (one process, concurrency test first) → **202 passed**. Could not reproduce the 25 failures the PR attributes to host load.
- **Callers of the converted handlers:** grepped every converted handler name across `api/`, `scripts/`, `mcp_server.py`, `web/` and `tests/`. The only in-repo callers of a converted handler are `tests/test_me_page.py` (fixed in this PR) and the four `/v2/*` wrappers in `people.py` (fixed). `mcp_server.py` reaches these endpoints over HTTP, so it is unaffected. No stale `await` remains.
- **Threading review:** no `BackgroundTasks`, `Request`, `Depends`, or SSE/streaming response in any of the three routers, so nothing depends on running on the loop; the only `UploadFile` handler stayed `async`. `_merged_ids` is only ever *rebound* (`_load_merged_ids`), never mutated in place, so the issue's read caveat holds; `_blocklist` mutation is a set `add`/`in`, GIL-atomic. The module-level constants in `person_entity.py` (`FAMILY_LAST_NAMES` etc.) are computed at import and never written. What is *not* safe is covered in findings 3 and 4.
- **Which UI endpoints matter:** grepped every `api(...)`/`fetch(...)` call in `web/crm.html` — the CRM page only ever hits `/api/crm/*` and `/api/photos/*`, both fully converted, so the milestone's "opening the CRM freezes everything" mechanism is fully addressed for the UI.
- **Remaining scope:** AST scan of `api/routes/` confirms `crm.py`, `people.py` and `photos.py` now have **zero** `async def` handlers without an `await`; 95 remain in the other 23 route modules (finding 6).
- Privacy: PR text, tests and docs contain no personal data; the aggregate row counts in the PR body are fine. I kept every real id out of this comment.

### Not verified

- **The full unit suite.** Two other agents' full gates were already running (`load average: 18`, 27 GB available) and the project memory records that 3–4 concurrent gates have OOM-rebooted this host, so I deliberately ran the targeted suites above instead. The PR reports a green full run; that still needs to hold on the pre-push gate.
- **No browser check.** `web/` is untouched by this PR, and the byte-identity result above means the pages receive identical bytes, so a Playwright pass would only re-test unchanged frontend code.
- **`extract_person_facts` and `import_source_data` end-to-end.** Both need an LLM call / a real upload, and the staging `.env` disables the Anthropic key; I reviewed them statically (findings 5) rather than exercising them.
- **The concurrent-merge corruption in finding 3 was not executed.** The standing brief forbids running mass-modifying merges against the shared staging data copy, so that finding is derived from `scripts/merge_people.py`'s single-slot intent log and read-modify-write, not from a reproduction.

nbramia:
Addressed all 9 findings from the adversarial review:

1. **[BLOCKER] TestClient lifespan side effects** — `tests/test_route_handlers_concurrency.py` now builds a router-only app (`FastAPI(); app.include_router(crm_router)`) instead of importing `api.main.app`. No lifespan, no Telegram/scheduler/job-queue/watcher startup, no vault writes.
2. **[BLOCKER] Untested/loose latency bounds** — tightened the bound to the issue's actual `100ms` (was `750ms`), and added `test_fast_config_request_not_blocked_by_concurrent_tone_analysis` (patches `interaction_store.get_for_person` to sleep, fires `POST /api/crm/relationship/tone-analysis-detailed`, asserts `/api/crm/config` still completes under 100ms). Both tests pass.
3. **[MAJOR] Missing serialization for mutating handlers** — added a module-level `threading.Lock` (`_mutation_lock` in `api/routes/crm.py`) held across `merge_people`, `split_person`, `hide_person`, `confirm_review_item`, `reject_review_item`, `sync_source`, `trigger_relationship_discovery`, `update_relationship_strengths`, `sync_slack_users_endpoint`, `sync_contacts_endpoint`, and `photos.py`'s `trigger_photo_sync` (imports the same lock). New `tests/test_route_handlers_mutation_lock.py::test_concurrent_merges_serialize` patches `scripts.merge_people.merge_people` to record overlap and fires two concurrent merge requests — verified it fails without the lock and passes with it. Added a paragraph to `docs/specs/technical/architecture.md` explaining that moving handlers off the loop removed the implicit serialization, so mutating handlers must lock explicitly.
4. **[MINOR] Unguarded lazy singletons** — added double-checked-locking `threading.Lock`s to `get_person_entity_store`, `get_interaction_store`, `get_source_entity_store`, `get_relationship_store`, and `get_person_fact_store`.
5. **[MINOR] Inline getter calls in the two async handlers** — `extract_person_facts` and `import_source_data` now call `get_person_entity_store()`, `get_interaction_store()`, `get_person_fact_extractor()`, and `get_source_entity_store()` inside `asyncio.to_thread` hops rather than inline on the loop.
6. **[MINOR] Substring-match guard** — `tests/test_route_handlers_sync.py` now does a real AST check (`ast.Await`/`ast.AsyncFor`/`ast.AsyncWith`, not descending into nested defs) instead of `"await" not in source`. Scoped the doc wording in both edited docs to explicitly say the rule applies to these three routers only, and that the other ~95 unconverted handlers elsewhere in `api/routes/` are pre-existing, not an exception — did not convert them (out of scope for #868).
7. **[MINOR] Stale canonical examples in docs** — both snippets now show `def`, not `async def` with no await; the `python-conventions.md` example is now explicitly labeled as a generic shape rather than claiming to be verbatim `api/routes/tasks.py` (whose real `list_tasks` hasn't been converted and is out of scope here).
8. **[MINOR] `async`/`@pytest.mark.asyncio` on de-awaited tests** — dropped from all 5 call sites in `tests/test_me_page.py`.
9. **[MINOR] No test for "no shared connection"** — added `tests/test_route_handlers_no_shared_connections.py`, an AST-based static guard over the 7 store modules reachable from these routers (`person_entity`, `interaction_store`, `source_entity`, `relationship`, `person_facts`, `relationship_insights`, `apple_photos`): fails if any opens a connection with `check_same_thread=False` or caches one on `self`.

`api/routes/photos.py` stays in scope per your note.

Test evidence: the affected/new test files (`test_route_handlers_sync.py`, `test_route_handlers_concurrency.py`, `test_route_handlers_mutation_lock.py`, `test_route_handlers_no_shared_connections.py`, `test_me_page.py`, plus `test_crm_api.py`, `test_people.py`, `test_people_e2e.py`, `test_photos_sync_api.py`, `test_merge_people.py`, `test_split_person.py`, `test_crm_ui_requirements.py`) all pass except the one pre-existing environment-dependent failure you also found (`test_my_person_id_is_valid_uuid`). Full unit suite (`./scripts/test.sh unit`, no `.env` in the worktree): `5585 passed, 8 skipped` -- clean, 0 failures. Pushed under the pre-push hook, which re-ran the same suite plus the server-free browser tests (`259 passed`) and both passed again before the push landed.

nbramia:
## Adversarial review of #887 by Opus — verification pass on `1679342`

All nine findings are genuinely fixed, and every fix is backed by a test I confirmed actually fails when the fix is removed. The 900-line `crm.py` diff is the part that needed the hardest look — it is almost entirely re-indentation under `with _mutation_lock:`, and I re-proved response-level equivalence against `origin/feat/crm-performance` on a *frozen* snapshot of the CRM databases to rule out a control-flow change hiding in the reindent. Nothing blocking remains; the three nits below are optional.

### Finding-by-finding verification

1. **[BLOCKER → FIXED] Lifespan side effects.** `tests/test_route_handlers_concurrency.py` now builds `FastAPI() + include_router(crm_router)`. Confirmed: `rm -rf vault && pytest tests/test_route_handlers_concurrency.py -q` → `2 passed`, and `find vault` afterwards finds nothing and no `vault/` directory is created at all (it was creating two Dashboard files before). `pytest ... -s --log-cli-level=INFO | grep -icE "telegram|scheduler started|job queue|watcher started|calendar indexer"` → **0**. No lifespan, no host guard, no Telegram/scheduler/job-queue/watcher startup. The new `test_route_handlers_mutation_lock.py` uses the same router-only pattern.

2. **[BLOCKER → FIXED] Latency bounds.** `FAST_REQUEST_BOUND_SECONDS = 0.1` now matches the issue, and `test_fast_config_request_not_blocked_by_concurrent_tone_analysis` covers criterion 2 by sleeping `interaction_store.get_for_person` — the handler's real first blocking call. I stress-tested the tightened bound rather than trusting it: **5 consecutive runs of the file (both tests each) with 32 busy-loop processes saturating all 32 cores → 5/5 passed**, and separately under xdist (`-n 4 --dist loadscope`) → passed. The bound is not flaky.

3. **[MAJOR → FIXED] Mutating handlers serialize again.** `_mutation_lock` covers exactly the handlers I named, and I checked the three failure modes you asked about:
   - **Coverage:** an AST scan of `crm.py` finds `with _mutation_lock:` in `merge_people`, `split_person`, `hide_person`, `confirm_review_item`, `reject_review_item`, `sync_source`, `trigger_relationship_discovery`, `update_relationship_strengths`, `sync_slack_users_endpoint`, `sync_contacts_endpoint` — 10 — plus `trigger_photo_sync` in `photos.py`. That is the full set from finding 3.
   - **Deadlock — locked handler calling a locked handler:** an AST call-graph check over those 10 finds **no** call from any locked handler to any other locked handler. Indirect re-entry is also ruled out: `grep -rn "api\.routes\.crm"` shows the *only* non-test importer of `api/routes/crm.py` anywhere in the repo is `api/routes/photos.py` (for the lock itself), so no service in the locked call graph can reach back into a locked route. The one `localhost:8000` string in that call graph is Slack's OAuth *redirect URI* handed to Slack, not an outbound self-call, and `slack_oauth_callback` is unlocked anyway.
   - **Deadlock — lock held across an await:** all 10 locked handlers are plain `def` and contain no `await` at all (verified by AST), so a non-reentrant `threading.Lock` can never be parked across a suspension point. `photos.py`'s `trigger_photo_sync` is likewise sync. No cycle from the cross-module import either: `import api.routes.photos` succeeds and `photos._mutation_lock is crm._mutation_lock` → `True`.
   - **The test actually catches the regression:** I replaced `_mutation_lock = threading.Lock()` with a no-op context manager (leaving all indentation intact) → `test_concurrent_merges_serialize` **FAILED**; reverted → passes.

4. **[MINOR → FIXED] Double-checked locking** added to `get_person_entity_store`, `get_interaction_store`, `get_source_entity_store`, `get_relationship_store`, `get_person_fact_store`. The pattern is correct for CPython: the global is rebound only after `__init__` returns, and the outer unlocked read can therefore only ever observe `None` or a fully-constructed object — so the fast path stays lock-free with no publication hazard.

5. **[MINOR → FIXED] No inline blocking calls left in the two async handlers.** `extract_person_facts` now wraps `get_person_entity_store()`/`get_interaction_store()` in closures executed inside `asyncio.to_thread`, and `get_person_fact_extractor` is itself a `to_thread` call; `import_source_data` does `await asyncio.to_thread(get_source_entity_store)`. I checked the moved calls preserve their arguments exactly (`get_for_person(person_id, days_back=3650, limit=100000)`), and the byte-identity run below covers the handlers they feed.

6. **[MINOR → FIXED] AST guard replaces the substring match.** `_AsyncConstructFinder` looks for real `Await`/`AsyncFor`/`AsyncWith` and does not descend into nested defs (correct — an inner closure's await must not excuse the outer handler; and it still credits `extract_person_facts`, whose own body awaits `to_thread`). Negative control: I made `get_crm_config` `async def` again *with the word "await" in its docstring* — the exact vacuous pass the old check allowed — and the guard **FAILED** as it should. The docs are now scoped honestly to the three routers with the ~95 unconverted handlers called out as pre-existing.

7. **[MINOR → FIXED]** Both canonical doc snippets are now `def`, and `python-conventions.md` relabels its example as a generic shape rather than claiming to mirror `api/routes/tasks.py` verbatim — which is the honest fix, since `list_tasks` itself is out of scope.

8. **[MINOR → FIXED]** All five `@pytest.mark.asyncio`/`async def` markers dropped from `tests/test_me_page.py`.

9. **[MINOR → FIXED] `tests/test_route_handlers_no_shared_connections.py`** guards the right 7 modules on both arms. Negative controls, one at a time: adding `check_same_thread=False` to `interaction_store._get_connection` → **FAILED**; assigning `self._cached = sqlite3.connect(...)` in `relationship._get_connection` → **FAILED**. Both reverted, both arms work.

### Remaining nits (non-blocking, take or leave)

- **N1. Two lazily-initialized singletons in the same call graph are still unlocked:** `get_person_fact_extractor` (`api/services/person_facts.py`, `_fact_extractor`) and `get_entity_resolver` (`api/services/entity_resolver.py`) kept the bare `if _x is None:` form. Both are benign — `EntityResolver.__init__` only stashes a reference to the (now-locked) store, and `PersonFactExtractor()` does no DDL — so double-construction wastes an object and nothing else. Worth the same three-line treatment only if you want the rule to be uniform.
- **N2. The coarse lock makes a quick user action queue behind a multi-minute sync.** One lock covers both the fast writes (merge/split/hide/review-queue) and the long-running sync POSTs (`trigger_relationship_discovery`, `update_relationship_strengths`, `sync_source`, the Slack/contacts/photos syncs), so hiding a person during a discovery run now blocks for the length of that run while holding a threadpool slot. Practically zero impact — none of those sync POSTs has a caller anywhere in `web/`, they are curl-only — and it is strictly better than the pre-#868 behaviour where they blocked the whole event loop. But if you ever wire a sync button into the UI, split them onto a second lock. (Same reasoning applies to the shared anyio threadpool: 40 simultaneous mutation requests would park all 40 threads on the lock and stall *every* `def` handler in the process. Not reachable today.)
- **N3. `update_person` is the one read-modify-write mutating handler left outside the lock.** It loads a person, mutates fields, and calls `store.update(person)` (whole-row write; `store.save()` is a documented no-op), so two concurrent PATCHes of the same person lose one edit. I deliberately did not name it in finding 3 and I do not think it needs the lock — the same race already exists against the sync pipeline in another process, and the UI edits one person at a time — but it is the honest boundary of what the lock now protects, and worth one sentence in the architecture note if you touch it again.
- **N4. Style:** `photos.py` importing a private `_mutation_lock` across module boundaries is a small smell. Fine at two call sites; if a third router needs it, promote it to a tiny shared module rather than widening the import.

- `git fetch origin fix/crm-handlers-off-loop` → `1679342`, one commit on top of `21bd13d`; base still `origin/feat/crm-performance` @ `388811e`.
- **`git diff -w 21bd13d..1679342`** reduces the 900-line `crm.py` change to exactly: the `threading` import, the `_mutation_lock` definition, eleven `with _mutation_lock:` lines, and the `extract_person_facts`/`import_source_data` thread-hop rewrite. Nothing else moved — the rest of the diff is indentation.
- **Response equivalence on a frozen dataset.** My first re-run showed six endpoints differing, so I checked before reporting it: a same-code control (PR captured twice, an hour apart) showed the *identical* set of differing endpoints, i.e. the shared staging copy is being written by another agent. I therefore took a consistent snapshot with SQLite's own `Connection.backup()` (`crm.db`, `interactions.db`, `imessage.db`, `sync_health.db`, `merged_person_ids.json`), pointed `data` at it, and captured base and PR back to back against the frozen copy: **BASE vs PR → 37/40 identical**, differing only in `people_search`, `strength`, `sync_health` — and a PR-vs-PR self-control on the same snapshot produced *exactly those same three*, so all three are run-to-run nondeterministic (two are time-derived, one is tie-order-dependent), not code-caused. On the earlier run the same three-way comparison gave 38/40 with base and PR agreeing on `people_search` too. No code-caused response difference anywhere.
- **OpenAPI:** `sha256(json.dumps(openapi, sort_keys=True))` on `1679342` is `ef5e6b13…`, byte-identical to the base value I recorded in the first review, 205 paths. No signature or response-model drift from the reindent.
- **Tests:** `pytest tests/test_crm_api.py tests/test_people.py tests/test_people_e2e.py tests/test_me_page.py tests/test_photos_sync_api.py tests/test_crm_ui_requirements.py tests/test_split_person.py tests/test_merge_people.py tests/test_route_handlers_{sync,concurrency,mutation_lock,no_shared_connections}.py -q` → **121 passed, 4 skipped, 1 failed** — up from 112 passed, and the single failure is still `test_me_page.py::TestMyPersonIdConstant::test_my_person_id_is_valid_uuid`, the pre-existing environment-dependent one (needs a configured `settings.my_person_id`; my worktree has no `.env`). Also `-n 4 --dist loadscope` on the four new/changed files + `test_me_page.py` → 20 passed, same one env failure.
- **Negative controls (each applied, run, reverted):** neutered `_mutation_lock` → mutation-lock test fails; `get_crm_config` back to `async def` with "await" only in its docstring → AST guard fails; `check_same_thread=False` in `interaction_store` → connection guard fails; `self._cached = sqlite3.connect(...)` in `relationship` → connection guard fails. `git status` clean afterwards.
- **Deadlock analysis:** AST call-graph over the 10 locked handlers (no locked→locked call), AST check that none of them is `async` or contains an `await`, `grep` proving `api/routes/photos.py` is the only non-test importer of `api/routes/crm.py`, and a live import proving `photos._mutation_lock is crm._mutation_lock` with no circular-import failure.

### Not verified

- **The full unit suite**, again: other agents' gates were running throughout (load average peaked at 34), and project memory records that concurrent gates have OOM-rebooted this host. The PR reports `5585 passed, 8 skipped` plus a green pre-push gate (`259 passed` server-free browser); I ran the affected files and the xdist subset instead.
- **No browser pass** — `web/` is untouched and the byte-identity result means the pages receive identical bytes.
- **`extract_person_facts` / `import_source_data` end-to-end** — still need a live LLM key and a real upload; reviewed statically.
- **Real concurrent merges against a live database** — the mutation-lock test proves the serialization with a fake merge; I did not run two real merges against the staging copy (the standing brief forbids mass-modifying it).

Issue #870, opened 2026-09-04
fix(crm): return a pruned neighborhood from the network graph endpoint instead of all 546k relationships


nbramia:
## Problem

Opening a person's Graph tab takes 10 to 12 seconds and downloads 140 to 166 MB of JSON, even though the graph is designed to show about 25 first-degree people. The server answers the request by loading all 546,205 relationship rows, then filtering them in Python; even a one-hop graph with 66 nodes takes 5.7 s and the two-hop default for a mid-strength person returns 2,408 nodes and 419,105 edges. The browser then has to parse that payload and hand it to the force simulation, which freezes the tab. While the server is building the response the whole API is blocked.

Measured on the production host:

| Request | Time | Payload | Nodes / edges |
|---|---|---|---|
| `network?center_on=<self>&depth=2` | 12.0 s | 166 MB | very large |
| `network?center_on=<mid person>&depth=2` | 10.6 s | 141 MB | 2,408 / 419,105 |
| `network?center_on=<mid person>&depth=1` | 5.7 s | 0.5 MB | 66 / 1,593 |
| Loading every relationship row alone | 5.3 s | | 546,205 rows |

## Solution

The network endpoint returns a pruned neighborhood: the strongest N first-degree connections of the center person, a bounded number of second-degree connections per first-degree node, and only the edges among the selected nodes. Node selection uses the indexed per-person relationship lookup, and the induced edge set comes from a single query restricted to the selected node ids. The endpoint accepts explicit node caps and a minimum strength, with defaults sized to what the graph tab renders. The Graph tab keeps its strength slider but filters the bounded set it received. Responses become tens of kilobytes in well under a second.

## Acceptance Criteria

- [ ] `GET /api/crm/network?center_on=<id>&depth=2` shall complete in under 500 ms and return at most `max_nodes` nodes (default 150) on the production dataset.
- [ ] The endpoint shall accept `max_nodes` (1 to 500) and `max_second_degree_per_node` (0 to 50) query parameters and return a validation error outside those ranges.
- [ ] Given a center with more first-degree connections than `max_nodes`, when the graph is requested, the returned first-degree nodes shall be the strongest by relationship strength.
- [ ] Every edge in the response shall connect two nodes that are both present in the response.
- [ ] The center node shall always be present, and every returned first-degree node shall have an edge to the center.
- [ ] The endpoint shall never call the load-all-relationships store method for a centered request.
[...]

PR #894, opened 2026-09-04
fix: repair #880 tests broken by #887's sync migration; allowlist data_version connection

nbramia:
## TL;DR

Fixes 4 pre-push test failures on `feat/crm-performance`'s tip that were blocking every sibling PR rebased on top of it: 3 tests in PR #880 that assumed handlers were still `async def` (broken by #887's sync-handler migration, which landed just before #880 merged), and 1 static-guard test that flags PR #880's cache-invalidation connection as an unsafe shared connection.

## Design

Two independent siblings (#870, #874) hit and reported the same 4 failures. Both are inherited-not-caused-by-them regressions from PR #880 colliding with #887, not bugs in their own diffs.

**Tests 1-3** (`tests/test_crm_people_list_batching.py`): PR #880 was authored against a branch state where `list_people()`/`get_todays_birthdays()` were `async def`, so its own tests called them with `await`. #887 (merged immediately before #880) converted both handlers to plain `def` so they run on the worker threadpool. The merge was textually clean — my PR's diff never touched the `async def`/`def` line itself, only code inside the function bodies — so nothing conflicted, but the tests were left calling a now-synchronous function with `await`, raising `TypeError`.

**Test 4** (`tests/test_route_handlers_no_shared_connections.py`): #868/#887 added a static AST-based guard forbidding any store reachable from the CRM/people/photos routers from caching a SQLite connection across calls, or opening one with `check_same_thread=False`, now that their handlers run on the worker threadpool. #869/#880's `PersonEntityStore.get_all()` cache reads `PRAGMA data_version` from a long-lived connection specifically to detect commits from *other* connections/processes (the nightly sync) — exactly the pattern the new guard forbids.

This is a genuine conflict between two individually-correct designs, not a bug in either. Before deciding how to resolve it, I verified experimentally (see commit message for the full trail) that `PRAGMA data_version`'s value is only meaningful when read repeatedly from the *same* connection object — two persistent connections established at different times diverge to different, non-comparable values after the same external commit, and a brand-new connection's first read does not reliably reflect the true current counter. There is no connection-per-call (or per-thread) equivalent that preserves the invalidation guarantee this cache depends on. Every read of this connection happens inside `PersonEntityStore._get_all_cache_lock` (a plain `threading.Lock`), so despite `check_same_thread=False`, access is fully serialized — never truly concurrent — which is what makes it safe despite outliving any single call. This was independently verified by the original #880 adversarial reviewer too.

Rather than weaken the guard generally, added a narrow, named `PERSISTENT_CONNECTION_EXCEPTIONS` allowlist keyed by module + attribute name, with the full reasoning recorded inline in the test file. Every other module, and any *other* persistent connection that might ever be added to `person_entity.py`, is still caught by the guard.

## Implementation Notes

- `tests/test_crm_people_list_batching.py`: removed `async`/`await` from the two test methods calling `list_people()` and `get_todays_birthdays()`, matching their current (post-#887) synchronous signatures.
- `tests/test_route_handlers_no_shared_connections.py`: refactored `_caches_connection_on_self()` into `_self_assigned_connect_calls()` (yields attribute name + call node instead of a bare bool), added `PERSISTENT_CONNECTION_EXCEPTIONS = {"api/services/person_entity.py": {"_data_version_conn"}}`, and updated the test function to check disallowed self-assigned connections against that allowlist, and to only exempt `check_same_thread=False` on a connection that is itself an allowlisted self-assignment.

### Automated tests

- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_route_handlers_no_shared_connections.py tests/test_crm_people_list_batching.py tests/test_source_entity.py tests/test_person_entity.py tests/test_crm_api.py -q -m "not integration"` -> 86 passed.
- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_crm_api.py -q -m integration` against the staging data symlink -> 36 passed, 1 skipped (pre-existing, unrelated skip).
- `./scripts/test.sh` (full unit suite, no `.env` in the worktree) -> 5604 passed, 8 skipped, 1181 warnings in 362.51s.
- Pre-push hook under `flock -w 7200 /tmp/lifeos-push.lock git push https://github.com/nbramia/LifeOS.git HEAD:refs/heads/fix/people-list-batching-followup` -> unit gate 5604 passed, 8 skipped, 1180 warnings in 303.26s; server-free browser gate 259 passed, 5956 deselected in 192.10s. Push succeeded.

### Manual verification (guard sanity check)

Confirmed the amended guard still rejects a synthetic disallowed persistent connection (`self._conn = sqlite3.connect(..., check_same_thread=False)` in a scratch module) while allowing exactly `person_entity.py`'s `_data_version_conn` and nothing else in that file.

## Review focus

- The `PERSISTENT_CONNECTION_EXCEPTIONS` allowlist and its justification — this is a deliberate, narrow carve-out in a guard test from another PR's review finding (#868), not something I'd normally touch unilaterally, but it was blocking every sibling PR on the integration branch and I have direct ownership context on the connection in question (#869/#880). Happy to adjust the wording/scope if the #868/#887 author wants a different shape for the exception.
- Whether `PERSISTENT_CONNECTION_EXCEPTIONS` should eventually move to a shared convention (e.g. a decorator or naming convention) if more exceptions accumulate — not needed yet with just one entry.

### DIFF
--- a/tests/test_crm_people_list_batching.py
+++ b/tests/test_crm_people_list_batching.py
@@ -53,7 +53,7 @@ class _SpySourceEntityStore:
 
 
 class TestListPeopleBatchesSourceEntityFetch:
-    async def test_issues_one_batch_call_not_one_per_person(self):
+    def test_issues_one_batch_call_not_one_per_person(self):
         people = _make_people(50)
         person_store = MagicMock()
         person_store.get_all.return_value = people
@@ -61,7 +61,7 @@ class TestListPeopleBatchesSourceEntityFetch:
 
         with patch("api.routes.crm.get_person_entity_store", return_value=person_store), \
                 patch("api.routes.crm.get_source_entity_store", return_value=spy_source_store):
-            result = await list_people(
+            result = list_people(
                 q=None, category=None, source=None, dunbar_circles=None, tags=None,
                 has_interactions=None, min_interactions=0, sort="strength",
                 offset=0, limit=50,
@@ -72,7 +72,7 @@ class TestListPeopleBatchesSourceEntityFetch:
         assert set(spy_source_store.get_for_people_batch_calls[0]) == {p.id for p in people}
         assert spy_source_store.get_for_person_calls == []
 
-    async def test_batch_call_only_covers_the_current_page(self):
+    def test_batch_call_only_covers_the_current_page(self):
         """Pagination (offset/limit) narrows the batch to the returned page,
         not the full result set."""
         people = _make_people(120)
@@ -82,7 +82,7 @@ class TestListPeopleBatchesSourceEntityFetch:
 
         with patch("api.routes.crm.get_person_entity_store", return_value=person_store), \
                 patch("api.routes.crm.get_source_entity_store", return_value=spy_source_store):
-            result = await list_people(
+            result = list_people(
                 q=None, category=None, source=None, dunbar_circles=None, tags=None,
                 has_interactions=None, min_interactions=0, sort="strength",
                 offset=0, limit=10,
@@ -94,7 +94,7 @@ class TestListPeopleBatchesSourceEntityFetch:
 
 
 class TestBirthdaysTodayBatchesSourceEntityFetch:
-    async def test_issues_one_batch_call_not_one_per_person(self):
+    def test_issues_one_batch_call_not_one_per_person(self):
         people = _make_people(5)
         from datetime import datetime
         today_mm_dd = datetime.now().strftime("%m-%d")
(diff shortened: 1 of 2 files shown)


nbramia:
Superseded by #893, which merged the same repair (sync handler tests + a marker-scoped allowlist for the lock-guarded pragma-only connection) into feat/crm-performance at addc623. Thanks for the write-up; the experimental justification for the persistent data_version connection matches #893's approach.

Issue #891, opened 2026-09-04
fix(crm): relationship page must not fire empty-id requests when no partner is configured


nbramia:
## Problem

On an install with no partner configured, opening the Relationship page fires requests with an empty person id (a timeline request to a path with a missing id segment) and logs a 404 in the console, and the tone, photo, and timeline sections fail silently. The page has no configured-partner check, so a fresh clone or a staging instance without the partner config file sees errors instead of a clear empty state.

## Solution

The Relationship page checks the configured partner id before loading data. When it is empty, it renders a short empty state explaining that no partner is configured and how to set one, and issues no partner-scoped requests. When a partner is configured, behaviour is unchanged.

## Acceptance Criteria

- [ ] Given `partner_person_id` is empty in the CRM config response, when `/relationship` loads, the client shall issue no request whose path contains an empty id segment and shall log no console error.
- [ ] Given `partner_person_id` is empty, the page shall render an empty state that names the configuration needed.
- [ ] Given `partner_person_id` is set, the page shall behave exactly as before.

## Verification

- **Automated tests:** a server-free browser test that stubs `/api/crm/config` with an empty partner id and asserts the request list and console; a second case with a partner id asserting the existing requests fire.
- **Existing test suite:** must pass.

## Scope

**In scope:** the Relationship page's load path and empty state.
**Out of scope:** partner configuration UI.

## Technical Context

- **Key files:** `web/crm.html:11472` (`PARTNER_PERSON_ID` default), `:8971` (set from config), `:11748` (`showRelationshipDashboard`), `:12022` (timeline request with the id), `api/routes/crm.py:4928` (`_load_partner_person_id`, reads a config file that is absent on a fresh install).
- **Patterns to follow:** `tests/test_voice_mic_block_ui_browser.py` for the server-free browser test.

## History

Found on 2026-09-04 by the #874 implementer during Playwright verification against a staging instance without the partner config file.

nbramia:
Added to the CRM responsiveness milestone; will be fixed together with #873 (same page) on the integration branch `feat/crm-performance`.

nbramia:
Delivered by #926 (squash of the `feat/crm-performance` integration branch), merged to main. See the per-issue PR linked from the integration branch for design, test evidence, and the adversarial review thread.

Issue #875, opened 2026-09-04
fix(crm): show profile photos in the people list without per-person probing


nbramia:
## Problem

Profile photos never appear in the CRM people list, even for people who have photos. The list's avatar loader probes each avatar with a request method the photo route does not accept, so every person is permanently marked "no photo" for the rest of the session, and in practice the loader is not even invoked after the list renders. The person detail header and the relationship page do show photos, so the list is the only place they are missing. Fixing this naively would add hundreds of extra requests per list render, which the rest of this milestone is trying to eliminate.

## Solution

The people list response tells the client which people have a profile photo, using the photo count the person record already carries, and the client renders an avatar image only for those people. The photo route accepts both probe and fetch methods and sends cache headers so repeated renders are free, and a missing thumbnail degrades to initials without a console error. No per-person probing loop remains.

## Acceptance Criteria

- [ ] `GET /api/crm/people` shall include `has_profile_photo: bool` on each person, true when the person's photo count is greater than zero.
- [ ] When the people list renders, the client shall issue photo requests only for people whose `has_profile_photo` is true, and never a `HEAD` probe.
- [ ] `HEAD /api/photos/profile/{id}` shall return the same status as `GET` without a body.
- [ ] `GET /api/photos/profile/{id}` shall return `Cache-Control` allowing at least one hour of caching on both 200 and 404 responses.
- [ ] Given a person with photos whose thumbnail file is missing, when the list renders, the avatar shall fall back to initials with no console error.
- [ ] Rendering a list of 300 people shall not block the main thread on photo loading (images load lazily).
- [ ] When photos are disabled in settings, the list shall render initials for everyone and issue no photo requests.

## Verification

- **Automated tests:** API test for the new field and for `HEAD` support; server-free browser test (stubbed list with two flagged people and one unflagged, stubbed photo route returning 200 for one and 404 for the other) asserting exactly two photo requests and correct avatar classes.
- **Manual checks:** open `/me` and scroll the list on staging; photos appear for known contacts.
- **Existing test suite:** must pass.

## Scope

**In scope:**
- The list response field, the avatar rendering path and loader, and the photo route's method and cache headers.

**Out of scope:**
- Changing how profile photos are chosen (most recent photo with a thumbnail).
- Photo sync or thumbnail generation.

## Technical Context

[...]

PR #895, opened 2026-09-04
chore(crm): gzip responses, cancel stale searches, fetch a selected person once, cache the page

### LINKED ISSUE #874: chore(crm): gzip responses, cancel stale searches, fetch a selected person once, cache the page

nbramia:
## TL;DR

Large CRM payloads — the people list, person timelines, and the CRM page itself — are now compressed before they go over the wire, which matters most on a slow connection like Tailscale from a phone. Typing quickly in the people search box no longer risks an older, slower response overwriting the results of what you just typed — the previous search is cancelled the moment a new one starts. Clicking on a person now fetches their details in a single request instead of two back-to-back ones. Switching between the Me, Family, Birthdays, and Relationship pages (and navigating between people) now revalidates the shared page cheaply instead of re-downloading the whole thing every time, and that revalidation now also works correctly behind a proxy that rewrites cache headers slightly. A dead, unused helper function was removed while in the area. A missing page now correctly reports "not found" instead of a success response.

Closes #874

## Design

- **Response compression** is applied narrowly, by URL path, to the route groups with payloads worth compressing: the people list and person-detail JSON, plus the CRM page itself (~750KB, the single largest payload in scope). This is a deliberate scoping choice to keep the compression cost confined to routes that benefit from it — it is not required for streaming safety: the pinned Starlette version already refuses to compress `text/event-stream` responses on its own, so every SSE endpoint (chat, agents, conversations) would be untouched even if this middleware were applied app-wide.
- **Search cancellation** tracks the single most recent search request. Starting a new one cancels whatever was still in flight, and a request whose result comes back after being superseded — whether cancelled outright or already past the point of no return over the network — is discarded rather than rendered or adopted as the current list. This also covers the case where the newer search's answer comes from the client-side cache and returns instantly: the slower, older request is still not allowed to overwrite it afterward. A superseded response's data is still cached under its own query key, though, since that's safe and makes re-typing that exact query later render instantly.
- **Single person-detail request**: the "related data" (source records, relationships) that used to require a second follow-up call is now requested together with the base detail in one call, and the person-detail rendering path uses whatever the one response contains.
- **Page caching**: the CRM page itself now tells the browser it's safe to reuse for a short time, but always to check back with the server first. The server recognizes when the browser's copy is still current — including when an intermediary has weakened the validator, or sent a wildcard — and says so with an empty response instead of resending the ~750KB page.

## Implementation Notes

- `api/main.py`: a small scoped-gzip ASGI middleware (`_ScopedGZipMiddleware`, wrapping Starlette's `GZipMiddleware`, `minimum_size=1024`, `compresslevel=6`) applied to `/api/crm/*`, `/api/people/*`, and the CRM page routes (`_in_gzip_scope()`). A shared `_crm_file_response()` helper is used by all ten routes that serve `web/crm.html` (`/crm`, `/crm/{path}`, `/me`, `/me/{path}`, `/family`, `/family/{path}`, `/relationship`, `/relationship/{path}`, `/birthdays`, `/birthdays/{path}`) — it computes the same ETag Starlette's `FileResponse` would, adds `Cache-Control: max-age=60, must-revalidate`, and answers a matching `If-None-Match` with a bodyless 304, via `_if_none_match_hits()` (handles a comma-separated list, a weak `W/"..."` validator, and a bare `*`). A missing `crm.html` now returns a JSON 404 from this helper, rather than the ten routes' previous behavior of a 200 with `{"message": "CRM page not found"}` — a deliberate improvement, called out here since it's a status-code change.
- `web/crm.html`:
  - `loadPeople()` creates an `AbortController` per call, aborts the previous one (if any) before doing anything else — including before a cache hit — and checks after `await` that it's still the current search before rendering it (though it still caches a successful-but-superseded response under its own query key). `AbortError` is treated as "superseded," not a failure. A comment documents the resulting contract for the four other callers of `loadPeople()` (init's initial load, and the hide/merge/split refreshes): a call may resolve without the shared `people` list actually being updated, if a newer call started first.
  - `loadPersonDetail()`'s two call sites (`cachedPerson` background refresh and the no-cache path) both fetch `/people/{id}?include_related=true` once; the old plain fetch plus the separate `loadRelatedData()` follow-up call is gone (`loadRelatedData` itself deleted, its only two call sites removed).
  - `renderPersonDetail()`'s source-entities branch now checks whether `source_entities` is present at all, not whether it's non-empty — an empty array now correctly renders "No source entities" instead of hanging forever on a "Loading source entities..." placeholder that nothing re-renders now that `loadRelatedData()` (the only thing that used to settle it) is gone.
  - Deleted `initializeDunbarCircles` (dead code, never called, would have requested `limit=10000`) and its now-unused `dunbarCirclesInitialized` flag. Left `calculateAndStoreDunbarCircles`, `applyDunbarCircles`, `loadProfilePhotos`, and `profilePhotoCache` untouched per the issue's scope note (#875 reworks the photo helpers).
  - `api()`'s existing `options` spread already forwarded `signal` to `fetch()` with no code change required; added a doc comment making that explicit.

### Automated tests

`tests/test_crm_request_hygiene_api.py` (25 tests total):
- `TestGzipUnitSynthetic` (1 test, plain `unit` marker — runs in the fast/pre-push scope): patches the CRM people store with obviously-synthetic people so a `/api/crm/people` response deterministically exceeds 1KB, proving gzip is actually applied without needing `data/crm.db` to exist. (An earlier version of this file only asserted this with two `integration`-marked, data-dependent tests that both skip on a fresh clone — this was flagged in review as leaving the headline acceptance criterion unproven by anything in the default test scope; this new test closes that gap.)
- `TestGzipScoping` (6 tests, `slow`): the CRM page and a client-side-routed sub-path are gzipped; a small `/api/crm/*` response stays uncompressed; the real people-list and people-search endpoints are gzipped when large enough (data-dependent, skip cleanly otherwise); the chat SSE stream is never gzipped.
- `TestCrmPageCacheControl` (14 tests, `slow`): `Cache-Control` present on all ten crm.html routes; a matching, a weak (`W/"..."`), and a wildcard (`*`) `If-None-Match` each return a bodyless 304; a stale one returns the full page.

`tests/test_crm_request_hygiene_ui_browser.py` (4 tests, server-free per `tests/test_voice_mic_block_ui_browser.py`'s pattern — runs at pre-push):
- A slower first search's response never overwrites a faster second one.
- The superseded request's `AbortSignal` genuinely fires (`window.__abortedUrls` is checked directly), not just a stale-response guard.
- A run of *actually superseded* searches (each fill spaced past the 300ms debounce, so each one's own request genuinely fires and gets aborted — an earlier version of this test had all three fills within 200ms, so only the last one's debounce ever fired and nothing was ever superseded or aborted, which is why it passed even against the pre-fix code; flagged in review) never shows the generic failure message, and `window.__abortedUrls` is asserted non-empty before checking that.
- Selecting a person issues exactly one `/people/{id}` request, with `include_related=true`, and an empty-but-present `source_entities` array renders "No source entities" rather than hanging on the loading placeholder.

All were verified to fail against the pre-fix (`addc623`) `web/crm.html`/`api/main.py` and pass against this branch.

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_crm_request_hygiene_api.py tests/test_crm_request_hygiene_ui_browser.py -m "" --browser chromium -q
...
25 passed in 20.16s
```

Full unit suite and the server-free browser suite both pass on the final, rebased commit (see the pre-push gate output on the push itself).

### Manual verification

Private staging instance against the real data copy (15,391 people), `.env` side effects disabled:

- `curl -H "Accept-Encoding: gzip" .../api/crm/people?limit=300` → `content-encoding: gzip`, compressed to 34,284 bytes (real, `compresslevel=6`).
- `curl -H "Accept-Encoding: gzip" .../me` → `content-encoding: gzip` (the page route itself, now in scope) alongside `cache-control: max-age=60, must-revalidate` and an `etag`.
- Repeating the `/me` request with `If-None-Match: <etag>` (strong), `If-None-Match: W/<etag>` (weak), and `If-None-Match: *` (wildcard) → **304** in all three cases, same `cache-control`/`etag`, empty body; a made-up `If-None-Match` → 200 with the full page.
- `curl -X POST .../api/ask/stream -H "Accept-Encoding: gzip"` → `transfer-encoding: chunked`, no `content-encoding`, multiple discrete SSE events delivered (confirms the chat stream is untouched by the new middleware; the response errored before reaching real token content only because this staging instance has no LLM backend configured — expected and safe, no API spend).
- Playwright against `/me` with real data: typed a rapid two-query sequence ("tay" then "nathan", second fill after the first's debounce fired) — the list showed 70 results, every one matching "nathan", zero "tay"-only leaks from the superseded first query, no console errors.
- Clicked a real person from the list; network log showed exactly one `GET /api/crm/people/{id}?include_related=true` request; the source-entities panel rendered correctly (not stuck loading).
- Two pre-existing, out-of-scope issues found during manual verification (not touched by this PR, unrelated to its changes): the `/api/photos/profile/{id}` avatar endpoint returns 503 in this staging environment (photo service unavailable here; `loadProfilePhotos` is explicitly out of scope per the issue's scope note — #875 owns it), and `/relationship`'s partner timeline chart 404s when `partner_person_id` is unset in config (pre-existing behavior, tracked separately as #891).

## Review focus

- The `_crm_file_response()` ETag algorithm intentionally duplicates Starlette's own (`hashlib.md5` of `mtime-size`, `usedforsecurity=False`) rather than importing a private method, so it produces byte-identical ETags to what `FileResponse` would send on its own — worth double-checking that assumption holds if Starlette's algorithm ever changes.
- `_in_gzip_scope()` is an *allow*-list (`/api/crm`, `/api/people`, plus the five named page routes and their sub-paths) rather than an exclude-list — worth confirming this is the right posture going forward as new routes are added under any of those prefixes.
- Cache-Control/304/gzip was applied to all ten crm.html-serving routes (including `/relationship` and every `/{path}` sub-route), not just the four literally named in the issue's acceptance criteria (`/me`, `/family`, `/crm`, `/birthdays`) — they all serve the identical static file, so this seemed like the right consistency call, but flagging it as a scope decision made unprompted.
- `loadPeople()`'s cancellation guard means any of its callers — not just an actual search — can resolve without the shared `people` list having been updated, if a newer call to `loadPeople()` started first. This is documented in a comment at the top of the function rather than restricted to search-only calls; every current caller tolerates it in practice (verified: none of the family/relationship dashboards or the final stats-await in `init()` read `people` at a point where this matters), but a future caller that needs a strict "resolve only once the list reflects this exact call" guarantee would need a separate path.
- `loadPersonDetail`'s cache-hit path relies on the cached person object having been populated by an `include_related=true` fetch (it always is, since that's the only fetch shape this file ever performs against `/people/{id}` now).

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -37,15 +37,19 @@ from pathlib import Path
 from dotenv import load_dotenv
 load_dotenv(Path(__file__).resolve().parent.parent / ".env")
 
+import hashlib
 import logging
+import os
 import socket
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
+from email.utils import formatdate
 
 from fastapi import FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import JSONResponse
+from fastapi.responses import JSONResponse, Response
 from fastapi.exceptions import RequestValidationError
+from starlette.middleware.gzip import GZipMiddleware
 
 from fastapi.staticfiles import StaticFiles
 from fastapi.responses import FileResponse
@@ -290,6 +294,39 @@ app.add_middleware(
     allow_methods=["*"],
     allow_headers=["*"],
 )
+
+
+# Gzip: scoped to the CRM/people JSON endpoints (#874) rather than applied
+# app-wide with an exclusion for streaming routes. A 300-person list is
+# ~216 KB and a person's full timeline ~620 KB, which matters on Tailscale
+# from a phone; those payloads live under /api/crm and /api/people.
+#
+# Scoping by an *allow*-list of path prefixes, instead of wrapping every
+# route in GZipMiddleware and trying to except the chat/agents/conversations
+# SSE endpoints by content-type, keeps `text/event-stream` responses out of
+# the gzip encoder by construction: Starlette's GZipMiddleware buffers each
+# chunk through a `gzip.GzipFile` before it can flush, which turns
+# token-by-token streaming into delayed, chunky output. None of the SSE
+# routes live under /api/crm or /api/people, so this scoping is sufficient
+# on its own — no separate exclusion list is needed.
+_GZIP_PATH_PREFIXES = ("/api/crm", "/api/people")
+
+
+class _ScopedGZipMiddleware:
+    """Apply `GZipMiddleware` only to requests under `_GZIP_PATH_PREFIXES`."""
+
+    def __init__(self, app, minimum_size: int = 1024) -> None:
+        self.app = app
+        self._gzip_app = GZipMiddleware(app, minimum_size=minimum_size)
+
+    async def __call__(self, scope, receive, send):
+        if scope["type"] == "http" and scope["path"].startswith(_GZIP_PATH_PREFIXES):
+            await self._gzip_app(scope, receive, send)
+        else:
+            await self.app(scope, receive, send)
+
+
+app.add_middleware(_ScopedGZipMiddleware, minimum_size=1024)
 
 # Include routers
 app.include_router(search.router)
@@ -909,13 +946,43 @@ async def chat_page():
     return {"message": "Chat page not found"}
 
 
-@app.get("/crm")
-async def crm_page():
-    """Serve the CRM UI."""
+_CRM_CACHE_CONTROL = "max-age=60, must-revalidate"
+
+
+def _crm_file_response(request: Request) -> Response:
+    """Serve `web/crm.html` with a short revalidation cache lifetime.
+
+    `FileResponse` on its own computes an ETag from the file's mtime/size but
+    never checks it against the request's `If-None-Match`, so every
+    navigation between Me/Family/Birthdays/a person page re-sent the full
+    ~750 KB page. This replicates Starlette's own ETag algorithm so the
+    header matches what `FileResponse` would have sent, and answers with a
+    bodyless 304 when the client's cached copy is still current.
+    """
     crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    if not crm_path.exists():
+        return JSONResponse({"message": "CRM page not found"}, status_code=404)
+
+    stat_result = os.stat(crm_path)
+    etag_base = f"{stat_result.st_mtime}-{stat_result.st_size}"
+    etag = f'"{hashlib.md5(etag_base.encode(), usedforsecurity=False).hexdigest()}"'
+    headers = {
+        "Cache-Control": _CRM_CACHE_CONTROL,
+        "ETag": etag,
+        "Last-Modified": formatdate(stat_result.st_mtime, usegmt=True),
+    }
+
+    if_none_match = request.headers.get("if-none-match")
+    if if_none_match and etag in (tag.strip() for tag in if_none_match.split(",")):
+        return Response(status_code=304, headers=headers)
+
+    return FileResponse(str(crm_path), stat_result=stat_result, headers=headers)
+
+
+@app.get("/crm")
+async def crm_page(request: Request):
+    """Serve the CRM UI."""
+    return _crm_file_response(request)
 
 
 @app.get("/agents")
@@ -947,81 +1014,54 @@ async def journal_trends_page():
 
 
 @app.get("/crm/{path:path}")
-async def crm_page_with_path(path: str):
+async def crm_page_with_path(request: Request, path: str):
     """Serve the CRM UI for any sub-path (client-side routing)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/me")
-async def me_page():
+async def me_page(request: Request):
     """Serve the CRM UI for the 'Me' dashboard (owner's profile)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/me/{path:path}")
-async def me_page_with_path(path: str):
+async def me_page_with_path(request: Request, path: str):
     """Serve the CRM UI for 'Me' sub-paths (client-side routing)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/family")
-async def family_page():
+async def family_page(request: Request):
     """Serve the CRM UI for the Family dashboard."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/family/{path:path}")
-async def family_page_with_path(path: str):
+async def family_page_with_path(request: Request, path: str):
     """Serve the CRM UI for Family sub-paths (client-side routing)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/relationship")
-async def relationship_page():
+async def relationship_page(request: Request):
     """Serve the CRM UI for the Relationship dashboard."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/relationship/{path:path}")
-async def relationship_page_with_path(path: str):
+async def relationship_page_with_path(request: Request, path: str):
     """Serve the CRM UI for Relationship sub-paths (client-side routing)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/birthdays")
-async def birthdays_page():
+async def birthdays_page(request: Request):
     """Serve the CRM UI for the Birthdays page."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
 
 
 @app.get("/birthdays/{path:path}")
-async def birthdays_page_with_path(path: str):
+async def birthdays_page_with_path(request: Request, path: str):
     """Serve the CRM UI for Birthdays sub-paths (client-side routing)."""
-    crm_path = Path(__file__).parent.parent / "web" / "crm.html"
-    if crm_path.exists():
-        return FileResponse(str(crm_path))
-    return {"message": "CRM page not found"}
+    return _crm_file_response(request)
--- a/api/main.py
+++ b/api/main.py
@@ -296,37 +296,55 @@ app.add_middleware(
 )
 
 
-# Gzip: scoped to the CRM/people JSON endpoints (#874) rather than applied
-# app-wide with an exclusion for streaming routes. A 300-person list is
-# ~216 KB and a person's full timeline ~620 KB, which matters on Tailscale
-# from a phone; those payloads live under /api/crm and /api/people.
+# Gzip: scoped to the CRM/people JSON endpoints plus the CRM page itself
+# (#874), rather than applied app-wide. A 300-person list is ~216 KB, a
+# person's full timeline ~620 KB, and the CRM page (`web/crm.html`) itself
+# is ~750 KB uncompressed — all of which matter on Tailscale from a phone.
 #
-# Scoping by an *allow*-list of path prefixes, instead of wrapping every
-# route in GZipMiddleware and trying to except the chat/agents/conversations
-# SSE endpoints by content-type, keeps `text/event-stream` responses out of
-# the gzip encoder by construction: Starlette's GZipMiddleware buffers each
-# chunk through a `gzip.GzipFile` before it can flush, which turns
-# token-by-token streaming into delayed, chunky output. None of the SSE
-# routes live under /api/crm or /api/people, so this scoping is sufficient
-# on its own — no separate exclusion list is needed.
-_GZIP_PATH_PREFIXES = ("/api/crm", "/api/people")
+# This is a deliberate scoping choice, not a workaround for a streaming
+# hazard: Starlette 0.52.1's GZipMiddleware already refuses to compress
+# `text/event-stream` responses on its own (`DEFAULT_EXCLUDED_CONTENT_TYPES`
+# in `starlette.middleware.gzip`), so applying it app-wide would not in fact
+# risk buffering the chat/agents/conversations SSE endpoints. Scoping by an
+# *allow*-list of paths instead keeps the gzip CPU cost (and the `Vary:
+# Accept-Encoding` header) confined to the handful of routes this issue
+# targets, and keeps that guarantee independent of Starlette's own default
+# exclusion list, which is this dependency's choice to change.
+_GZIP_API_PREFIXES = ("/api/crm", "/api/people")
+# The page routes that serve `web/crm.html` (api/main.py's crm_page() and
+# friends, further down this file) and their client-side-routed sub-paths
+# (e.g. /crm/{person_id}/timeline). Matched as a full path segment, not a
+# bare prefix, so a hypothetical future "/crmfoo" route wouldn't match.
+_GZIP_PAGE_ROUTES = ("/crm", "/me", "/family", "/relationship", "/birthdays")
+
+
+def _in_gzip_scope(path: str) -> bool:
+    if path.startswith(_GZIP_API_PREFIXES):
+        return True
+    return any(path == route or path.startswith(route + "/") for route in _GZIP_PAGE_ROUTES)
 
 
 class _ScopedGZipMiddleware:
-    """Apply `GZipMiddleware` only to requests under `_GZIP_PATH_PREFIXES`."""
+    """Apply `GZipMiddleware` only to requests `_in_gzip_scope()` accepts."""
 
-    def __init__(self, app, minimum_size: int = 1024) -> None:
+    def __init__(self, app, minimum_size: int = 1024, compresslevel: int = 6) -> None:
         self.app = app
-        self._gzip_app = GZipMiddleware(app, minimum_size=minimum_size)
+        self._gzip_app = GZipMiddleware(app, minimum_size=minimum_size, compresslevel=compresslevel)
 
     async def __call__(self, scope, receive, send):
-        if scope["type"] == "http" and scope["path"].startswith(_GZIP_PATH_PREFIXES):
+        if scope["type"] == "http" and _in_gzip_scope(scope["path"]):
             await self._gzip_app(scope, receive, send)
         else:
             await self.app(scope, receive, send)
 
 
-app.add_middleware(_ScopedGZipMiddleware, minimum_size=1024)
+# compresslevel=6 (Starlette's default is 9): the gzip step itself runs
+# synchronously in this middleware, on the event loop, for every matching
+# response. Measured on the real 300-person list response (215,625 raw
+# bytes): level 9 took ~4.5ms for 33,171 compressed bytes; level 6 took
+# ~2.5ms for 34,311 bytes — nearly half the blocking time on the loop for
+# ~3% more bytes on the wire, a better trade here.
+app.add_middleware(_ScopedGZipMiddleware, minimum_size=1024, compresslevel=6)
 
 # Include routers
 app.include_router(search.router)
@@ -949,6 +967,23 @@ async def chat_page():
 _CRM_CACHE_CONTROL = "max-age=60, must-revalidate"
 
 
+def _if_none_match_hits(if_none_match: str, etag: str) -> bool:
+    """Does this request's `If-None-Match` cover `etag` (RFC 9110 §13.1.2)?
+
+    `*` matches any current representation. Otherwise this is a
+    comma-separated list of entity-tags, each optionally weak (`W/"..."`
+    rather than `"..."`) — a weak validator from an intermediary still means
+    "you already have an equivalent representation" for a static file that
+    only ever changes by full replacement, so it's stripped before
+    comparing rather than treated as a non-match.
+    """
+    candidates = [tag.strip() for tag in if_none_match.split(",")]
+    if "*" in candidates:
+        return True
+    normalized = [tag[2:] if tag.startswith("W/") else tag for tag in candidates]
+    return etag in normalized
+
+
 def _crm_file_response(request: Request) -> Response:
     """Serve `web/crm.html` with a short revalidation cache lifetime.
 
@@ -973,7 +1008,7 @@ def _crm_file_response(request: Request) -> Response:
     }
 
     if_none_match = request.headers.get("if-none-match")
-    if if_none_match and etag in (tag.strip() for tag in if_none_match.split(",")):
+    if if_none_match and _if_none_match_hits(if_none_match, etag):
         return Response(status_code=304, headers=headers)
 
     return FileResponse(str(crm_path), stat_result=stat_result, headers=headers)
--- a/docs/specs/product/crm-people.md
+++ b/docs/specs/product/crm-people.md
@@ -2,7 +2,7 @@
 
 **Status:** Complete
 **Owner:** CRM
-**Last Updated:** 2026-06-21
+**Last Updated:** 2026-09-04
 
 People management: the list and detail views at `/crm`, the contact-source model behind entity splitting/merging, the relationship-strength score that drives ranking and Dunbar circles, and the multi-stage fact-extraction pipeline that surfaces memorable personal details.
 
@@ -181,7 +181,7 @@ Main CRM page at `/crm`.
 
 Behavior:
 - Header shows total people count.
-- Real-time search (< 300 ms).
+- Real-time search (< 300 ms), debounced; a superseded query's response is discarded so the list always reflects what's currently typed, even if an earlier search's response arrives later (#874).
 - Category tabs (Work / Personal / Family).
 - Cards show avatar, name, company/category, interaction count, strength bar.
 - Cards sorted by relationship strength by default.
--- a/docs/specs/technical/client-surfaces.md
+++ b/docs/specs/technical/client-surfaces.md
@@ -2,7 +2,7 @@
 
 > **Status:** Complete
 > **Owner:** Platform
-> **Last Updated:** 2026-08-26
+> **Last Updated:** 2026-09-04
 
 LifeOS exposes the orchestrator to **HTTP consumers** — thin clients that submit text and consume SSE without importing LifeOS Python modules. Endpoint and event **shapes** are defined in [api-reference.md](../product/api-reference.md); this doc covers **who consumes them**, **whisper-relay integration**, and **breaking-change policy**.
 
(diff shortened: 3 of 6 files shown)


nbramia:
## Adversarial review of #895 by Opus

**Verdict:** BLOCK

Narrowly, and on test coverage only. I drove the whole CRM surface in a real browser against the real dataset and **found no behavioural regression** — the four acceptance criteria all hold when I exercise them by hand. What blocks is that two of the criteria are not actually protected by anything that runs: the only positive gzip assertion skips on a clean checkout, and one of the four browser tests exercises nothing at all (and passes against the pre-fix code, contradicting the PR's test-evidence claim). Both fixes are a few lines.

1. **[BLOCKER] Nothing that runs on a clean checkout asserts that gzip is ever applied** — `tests/test_crm_request_hygiene_api.py:52-73` — the two positive tests for acceptance criterion #1 (`test_crm_people_list_is_gzipped_when_large_enough`, `test_people_search_endpoint_is_gzipped_when_large_enough`) both `pytest.skip()` when `data/crm.db` is absent or small. Every other gzip test in the module is a *negative* assertion (page not gzipped, statistics not gzipped, SSE not gzipped), all of which stay green with the middleware deleted. Reproduced: `mv data data_off && pytest tests/test_crm_request_hygiene_api.py -m "" -rs` → `15 passed, 2 skipped`, the two skips being exactly the two that prove the feature. Compounding it, `pytestmark = pytest.mark.slow` keeps all 17 out of `./scripts/test.sh` and out of the pre-push gate, so on this repo's default paths the headline criterion is asserted by nothing at all. **Fix:** make one positive assertion data-independent — e.g. patch the CRM people store to return N obviously-synthetic people so an in-scope response deterministically exceeds 1 KB — and let that one test run in the unit scope. (The `slow` marker on the rest, following `test_chat_api.py`, is fine.)

2. **[BLOCKER] `test_abort_error_does_not_show_failure_state` exercises nothing and passes against the pre-fix code** — `tests/test_crm_request_hygiene_ui_browser.py:213-235` — the three `fill()` calls are 100 ms apart, inside `handleSearch`'s 300 ms debounce (`web/crm.html:18821-18827`), so exactly one request is ever issued: nothing is superseded, no `AbortError` is ever thrown, and the assertion "no failure state" is trivially true. Verified by checking out the pre-fix `web/crm.html` (`git show addc623:web/crm.html`) and running the file: **`3 failed, 1 passed`** — the one that passes is this test. That directly contradicts the PR body's "All four were verified to fail against the pre-fix code and pass against the fix." Net effect: the `if (error.name === 'AbortError') return;` guard at `web/crm.html:9428-9431` — the only thing between a superseded search and the "Failed to load people" panel — has no regression test. **Fix:** space the fills > 300 ms apart the way the two sibling tests do, and/or assert `window.__abortedUrls` is non-empty before checking the panel.

3. **[MINOR] The "Loading source entities…" branch is now a dead end** — `web/crm.html:10492-10493` — deleting `loadRelatedData` removed the only follow-up that ever replaced that placeholder. If `source_entity_count > 0` while `source_entities` is empty or absent, the panel now hangs on the loading text forever; pre-PR it settled to "No source entities" via `updateSourceEntities([], count)`. I could not reach it on real data (0 of 400 sampled people had `source_entity_count > 0` with an empty `source_entities`, and all three `renderPersonDetail` call sites now pass an `include_related=true` payload), so this is latent, not live. But an unreachable branch that hangs is worse than either outcome. **Fix:** either delete the branch as dead code, or change the first condition to `if (person.source_entities)` so an empty array renders the empty state.

4. **[MINOR] `loadPeople()` gained a silent early-return that four callers don't expect** — `web/crm.html:9391-9399`, `9428-9431` — both the identity guard and the `AbortError` catch return without rendering, yet `init()`'s `peoplePromise` (awaited before `showFamilyDashboard()`) and the three `await loadPeople()` sites after hide/merge/split all treat resolution as "the list is now fresh". Typing during the initial load aborts the init call, so those callers proceed against a `people` array that only the newer search will populate. I drove this (`/family`, typed during the initial load): no user-visible breakage, because the family dashboard's default state doesn't read `people` — so it's a sharp edge, not a bug today. **Fix:** a comment at minimum; better, only abort when the call is a search.

5. **[MINOR] `If-None-Match` is compared exact-match only** — `api/main.py:975-976` — no weak-tag (`W/"…"`) normalisation and no `*`. Any intermediary that weakens the ETag silently costs the full 768 KB again, which is the entire point of the change. `.strip()` should also strip a leading `W/`, and `*` should 304.

6. **[MINOR] The 750 KB CRM page itself is still uncompressed** — the largest single payload named in the issue's own problem statement is excluded by the allow-list. Verified: `curl -H 'Accept-Encoding: gzip' :8020/me` → no `content-encoding`, 768,264 bytes. Caching covers repeat navigations, but a cold load — or any load after `crm.html` changes — still pays it in full over Tailscale. Either compress it too or say in the PR/doc that leaving it uncompressed was deliberate.

7. **[MINOR] The stated rationale for path-scoping is inaccurate for the pinned Starlette, and it's now written into a spec** — `api/main.py:299-308` and `docs/specs/technical/client-surfaces.md:21` — Starlette 0.52.1's `GZipMiddleware` already refuses to compress `text/event-stream` (`DEFAULT_EXCLUDED_CONTENT_TYPES`), so "keeps SSE out of the gzip encoder by construction … no exclusion list is needed" describes a hazard the dependency already handles. The scoping choice is fine; the justification a future reader will rely on is not.

8. **[MINOR] `compresslevel` left at Starlette's default 9** — measured on the real 215,585-byte list response: level 9 = 4.9 ms → 33,145 B; level 6 = 3.9 ms → 34,283 B. That is ~5 ms of synchronous, event-loop-blocking CPU per list request (~15 ms extrapolated for a 620 KB timeline), added immediately after #868 moved these handlers off the loop for exactly that reason. Small, but level 6 buys back ~20% of it for ~3% more bytes.

9. **[MINOR] A superseded-but-successful response is discarded instead of cached** — `web/crm.html:9391-9399` — the guard returns before `setCachedSearch(cacheKey, allPeople)`, yet the cache key includes the query, so storing it is safe and would make re-typing the earlier query instant. Cheap win, and it makes the guard's intent (don't *render* stale) clearer than "don't keep the data".

10. **[MINOR] Two unremarked leftovers** — missing `web/crm.html` now returns a 404 JSON where the ten routes previously returned HTTP 200 with `{"message": "CRM page not found"}` (`api/main.py:966-967`); an improvement, but a behaviour change not mentioned anywhere. And a stray triple blank line is left at the `loadRelatedData` deletion site (`web/crm.html:10271-10273`).

### What I verified and how

Private staging instance on `:8020` from this worktree, real data symlinked, staging `.env` (side effects off). Killed at the end; worktree left clean (`git status` shows only the pre-existing `.worktree-info`).

- `curl -H 'Accept-Encoding: gzip' :8020/api/crm/people?limit=300` → `content-encoding: gzip`, `vary: Accept-Encoding`, **215,585 → 33,145 bytes**. Criterion 1 holds for `/api/crm/*`.
- `curl -H 'Accept-Encoding: gzip' :8020/api/people/search?q=a&limit=50` → `content-encoding: gzip`. `/api/people/*` in scope too.
- `curl -H 'Accept-Encoding: gzip' :8020/api/crm/people/{id}/timeline?limit=200` → gzipped.
- `curl :8020/me` → `cache-control: max-age=60, must-revalidate` + `etag` + `last-modified`, 768,264 B. Re-request with `If-None-Match: <etag>` → **304, same cache-control/etag, no body**. Criterion 4 holds.
- `curl -N -X POST :8020/api/ask/stream -H 'Accept-Encoding: gzip'` → `content-type: text/event-stream`, `transfer-encoding: chunked`, **no `content-encoding`**, discrete SSE events delivered one at a time. SSE untouched.
- Playwright (venv chromium), `/me` → `/family` → `/birthdays` → `/relationship`: all render, no console error attributable to this change.
- **Search race:** typed `ta`, waited 360 ms for the debounce to fire, then `nathan`. Compared the rendered card names against the API's own answers: rendered = 70 names, **every one of them in the `q=nathan` result set (64 canonical names + aliases), zero names that appear only in the `q=ta` set (299 names)**, and no "Failed to load people" at any point. Repeated as a 9-query burst with each fill > 300 ms apart (so every one fires): same result, no error state. Criterion 2 holds.
- **Clear search** → back to 219 cards, no error state.
- **Single fetch:** clicking three different people, plus direct links to `/crm/{id}/timeline` and `/crm/{id}/graph`, plus switching person while on the timeline tab → **exactly one `GET /api/crm/people/{id}` per selection, always `?include_related=true`**. Criterion 3 holds. A/B against the pre-fix file on the same staging server: **BASE = 2 detail requests, PR = 1**.
- **A/B for regressions:** ran the identical tab / back-forward / stay-on-tab flow twice against `:8020`, swapping `web/crm.html` between the base (`addc623`) and PR versions. Identical on both: same tab and URL transitions, same `4` console errors with the same two signatures (3 × the known `/api/photos/profile/{id}` 503, 1 × a graph fetch cancelled by navigation). The `.tab.active` being null after a timeline click, and the `/timeline` suffix being dropped when switching person from the timeline tab, reproduce **identically on the base branch** — pre-existing, not this PR (the diff doesn't touch `switchTab`/`updateUrl`).
- **Behaviour preservation of the single fetch:** the old light call (`include_related=false`) computes `category` from a `limit=500` source-entity fetch inside `compute_person_category`, while `include_related=true` supplies a `limit=100` fetch — so the two paths could in principle disagree, and the PR now renders the latter. Checked it: JSON diff (minus `source_entities`/`relationships`) for the 60 people with the most source entities → **0 differences**; SQL sweep of all **524** people with > 100 source entities for a work signal present only outside the 100 most recent by `observed_at` → **1 theoretical candidate**, and its category is identical on both paths. No divergence on real data.
- **Timing of the overview tab:** warm, 5 runs each — `include_related=true` ≈ 27-59 ms vs `include_related=false` ≈ 38-64 ms. No first-paint regression; the "light" call was never light, because it does its own `limit=500` source fetch for the category.
- `pytest tests/test_crm_request_hygiene_api.py -m ""` → **17 passed** with data; **15 passed, 2 skipped** without (finding 1).
- `pytest tests/test_crm_request_hygiene_ui_browser.py -m "" --browser chromium` → **4 passed** on the PR; **3 failed, 1 passed** against `addc623:web/crm.html` (finding 2).
- Read `starlette/middleware/gzip.py` (0.52.1): `GZipMiddleware` is per-request stateless, so the single instance held by `_ScopedGZipMiddleware` is safe; `DEFAULT_EXCLUDED_CONTENT_TYPES` already covers `text/event-stream` (finding 7); `http.response.pathsend` bypasses compression.
- Confirmed `/api/crm` and `/api/people` are the only routers under those prefixes and neither returns a `StreamingResponse` / `text/event-stream`.
- Scope: 6 files, one commit, correct trailers; `initializeDunbarCircles`, `dunbarCirclesInitialized`, and `loadRelatedData` have zero remaining references; `loadProfilePhotos`/`profilePhotoCache` correctly kept per the issue comment; `import os`/`hashlib`/`formatdate` added once each, no duplicates. No new dependencies. No real names, emails, or DB rows in the diff, tests, or PR text — the browser fixtures are obviously synthetic.
- Docs: `grep -rn "gzip\|Cache-Control\|compress\|loadRelatedData\|include_related" docs/` finds no statement this change makes stale beyond the two files updated.

### Not verified

- The full unit suite (5,605 passed) and the server-free browser suite (263 passed) as claimed in the PR body — I did not re-run them, per the standing brief's warning that concurrent pre-push-scale gates have OOM-rebooted this box.
- Gzip behaviour behind the real Tailscale path (only measured `content-length` on the wire locally).
- Real-browser 304 revalidation across a `crm.html` edit (verified the server side by hand with `curl`, not the browser's own cache decision).
- The two pre-existing issues named as out of scope: `/api/photos/profile/{id}` → 503, and `/relationship` requesting `/people//timeline` with no partner configured (#891). Both observed, neither investigated.

nbramia:
Addressed all ten findings, in commit `150da3d`:

1. **[BLOCKER] Data-independent positive gzip test** — added `TestGzipUnitSynthetic::test_synthetic_large_people_list_is_gzipped` (`tests/test_crm_request_hygiene_api.py`), marked plain `unit` (not `slow`) so it runs in the default/pre-push scope. It patches `api.routes.crm.get_person_entity_store`/`get_source_entity_store` with 60 obviously-synthetic `PersonEntity` fixtures so the response deterministically exceeds 1KB, then asserts `content-encoding: gzip`. Verified it passes with `data/` present and with `data/` removed entirely. The two data-dependent `integration`-marked tests are unchanged.

2. **[BLOCKER] `test_abort_error_does_not_show_failure_state` exercised nothing** — rewrote it: three fills now spaced 350ms apart (past the 300ms debounce, matching the sibling tests) instead of 100ms apart, so each one's own request genuinely fires; "one" and "two" get a 5s artificial delay so they're truly in-flight when superseded. The test now asserts `window.__abortedUrls` contains both superseded URLs *before* checking the failure panel. Verified: fails against `addc623`'s `web/crm.html` (3 of 4 tests in the file now fail there, matching your finding), passes on this branch (4/4). Corrected the PR body's test-evidence section to describe this explicitly rather than repeat the earlier, wrong claim.

3. **[MINOR] Dead "Loading source entities…" branch** — changed the condition to `if (person.source_entities)` (presence, not length), so an empty array renders "No source entities" via `updateSourceEntities([], count)` instead of hanging. Added an assertion in the existing single-fetch browser test (set `PERSON_DETAIL.source_entity_count = 3` with `source_entities: []` — the exact mismatch that used to hang) checking the panel shows "No source entities" and never "Loading source entities". Verified it fails against the old length-check condition and passes against the fix.

4. **[MINOR] Silent early-return contract** — added a comment block above `loadPeople()` documenting that any call (search or not) may resolve without the shared `people` array being updated if a newer call started first, and why every current caller (`init()`'s `peoplePromise`, the hide/merge/split reload calls) tolerates that today. Did not restrict cancellation to search-only calls — restructuring the caller graph to distinguish intent felt disproportionate to a finding you yourself confirmed causes no live breakage, and the documented contract is the fix you offered as sufficient on its own.

5. **[MINOR] `If-None-Match` exact-match only** — added `_if_none_match_hits()`: strips a leading `W/` before comparing, and treats a bare `*` as a match. Added `test_weak_if_none_match_returns_304` and `test_wildcard_if_none_match_returns_304`. Verified against real staging: strong, weak, and wildcard `If-None-Match` on `/me` all return 304; a made-up one still returns 200.

6. **[MINOR] CRM page itself uncompressed** — extended gzip scope from an API-prefix-only allow-list to also include the five page routes (`/crm`, `/me`, `/family`, `/relationship`, `/birthdays`) and their sub-paths, via a new `_in_gzip_scope()` helper. Verified with curl against real staging: `curl -H 'Accept-Encoding: gzip' :PORT/me` → `content-encoding: gzip`; the 304 path still works combined with gzip (empty 304 body is under the 1KB minimum_size, untouched). Added `test_crm_page_is_gzipped` / `test_crm_page_subpath_is_gzipped`, and documented the scope change in `client-surfaces.md`.

7. **[MINOR] Inaccurate rationale** — rewrote both the `api/main.py` comment and the `client-surfaces.md` note: Starlette 0.52.1's `GZipMiddleware` already excludes `text/event-stream` via `DEFAULT_EXCLUDED_CONTENT_TYPES`, independent of this app's scoping, so the allow-list is now described as a deliberate cost-scoping choice rather than a required safety measure.

8. **[MINOR] `compresslevel`** — set to 6. Measured myself on the real 300-person list (215,625 raw bytes) rather than reusing your numbers: level 9 = 4.5ms / 33,171 bytes, level 6 = 2.5ms / 34,311 bytes — close to your own measurement, same conclusion.

9. **[MINOR] Superseded-but-successful response discarded** — moved `setCachedSearch(cacheKey, allPeople)` before the staleness/abort check, so a response that completes after being superseded is still cached under its query key (re-typing that exact query later now renders instantly from cache) even though it's still not rendered as the current list.

10. **[MINOR] Unremarked leftovers** — the PR body now explicitly calls out the 404-vs-200 behavior change for a missing `crm.html`. Removed the stray blank line at the `loadRelatedData` deletion site (verified no other double/triple-blank-line anomalies remain in either changed file).

Verification on this pass: `tests/test_crm_request_hygiene_api.py` + `tests/test_crm_request_hygiene_ui_browser.py` = 25/25 passing (`-m ""`, i.e. every marker); full unit suite 5606 passed/8 skipped; server-free browser suite 263 passed — all via the pre-push gate on the push itself. Re-ran the Playwright search-race and single-fetch-selection checks against a fresh staging instance with real data (15,391 people): no leaked results from a superseded query, exactly one detail request per selection, source-entities panel renders correctly. PR body updated to reflect all of the above.

nbramia:
## Verification pass on #895 (head `150da3d`) by Opus

Nine of the ten findings are fully resolved and I mutation-tested the ones that matter rather than taking the diff at its word. One residual: finding 2's test is no longer vacuous (it now fails against the pre-fix file, as required), but the specific behaviour its own docstring names is still not protected — deleting the `AbortError` guard leaves all four browser tests green. That's a nit, not a blocker: the guard is present and working, and the *acceptance criterion* it sits near ("the list shall show results for the second query only") is separately covered and mutation-proof.

### Per-finding verification

1. **[RESOLVED — mutation-proved] Data-independent gzip coverage.** `pytest tests/test_crm_request_hygiene_api.py -m "unit and not slow"` — the exact expression `scripts/pre-push` uses — selects `TestGzipUnitSynthetic::test_synthetic_large_people_list_is_gzipped` and it passes with `data/` **entirely absent** (`1 passed, 20 deselected in 1.03s`). Not just present, but load-bearing: commenting out `app.add_middleware(_ScopedGZipMiddleware, …)` makes it fail with `Headers({'content-length': '38393', 'content-type': 'application/json'})` and no `content-encoding`. Full module: **19 passed / 2 skipped** with no data (the two remaining skips are the data-dependent `integration` pair, correctly left as-is), **21 passed** with data. The per-class `pytestmark` split (`unit` for the synthetic class, `slow` for the other two) collects as intended.

2. **[PARTIALLY RESOLVED — nit] `test_abort_error_does_not_show_failure_state` is no longer vacuous, but still doesn't guard what it says it does.** The good part: all four browser tests now fail against `addc623:web/crm.html` (previously 3 failed / 1 passed), and the new `window.__abortedUrls` assertions do prove two requests were genuinely aborted. The gap: I deleted **only** the `if (error.name === 'AbortError') { return; }` guard from this branch's `web/crm.html` and re-ran the file — **4 passed**. The cause is ordering: the panel assertion runs after `expect(…).to_have_text("Fast Query Result")`, and the fast third query's render overwrites the error the two aborted searches painted, so the final state is clean either way. The error is real, not theoretical — driving the same mock-fetch harness with the guard removed and inspecting the panel *while the superseded request is still in flight* gives `'⚠️\n\nFailed to load people'`, versus `'Loading people...'` on this branch's code. **Fix:** assert the panel between the "two" fill and the "three" fill (both requests still pending, nothing rendered over them), or simply move the `not_to_contain_text` assertion above the `to_have_text` expectation.

3. **[RESOLVED — mutation-proved] Source-entities panel.** `if (person.source_entities)` (presence, not length), and `PERSON_DETAIL` now carries the exact mismatch (`source_entity_count: 3`, `source_entities: []`). Reverting the condition to the old length check fails `test_selecting_person_issues_one_detail_request` and nothing else — so the assertion is pinned to this behaviour specifically. Confirmed live too: three person selections on real data, `Loading source entities` never present in `#sourceEntitiesList`.

4. **[RESOLVED as offered]** The documented contract is what I said would be sufficient. *Nit:* 26 lines of comment for a two-line contract reads heavy against AGENTS.md § Simplicity — the first paragraph plus the "resolve ≠ fresh" sentence carries the whole point.

5. **[RESOLVED]** `_if_none_match_hits()`. Live matrix against staging `/me`: strong `"…"` → **304**; `W/"…"` → **304**; `*` → **304**; `"bogus"` → **200** with the full page; a comma list `"other", "<etag>"` → **304**. Every 304 carries `cache-control` + `etag` and an empty body.

6. **[RESOLVED — and I checked the new risk it introduces]** `/me` now returns `content-encoding: gzip` with `vary: Accept-Encoding`: **770,897 → 128,837 bytes, 83% smaller**. Because `FileResponse` streams, gzip takes Starlette's chunked branch and drops `Content-Length` for `transfer-encoding: chunked` — so I checked the bytes rather than assuming: `curl --compressed :8020/me | cmp - web/crm.html` → **byte-identical**. 304 combined with `Accept-Encoding: gzip` is clean (empty body, no `content-encoding` on the 304). `/me/timeline` is in scope via the segment match; `/chat` (179 KB) and `/agents` (103 KB) are correctly still out of scope — both would benefit, but that's not this issue.

7. **[RESOLVED]** Both the `api/main.py` comment and `client-surfaces.md` now describe the allow-list as a deliberate cost-scoping choice and correctly state that Starlette 0.52.1 already excludes `text/event-stream` on its own.

8. **[RESOLVED — and it matters more now than when I raised it]** With the 750 KB page pulled into scope, compresslevel dominates. Measured on the real 770,897-byte `crm.html`: level 9 = **33.3 ms** → 126,480 B; level 6 = **9.0 ms** → 128,837 B. So `compresslevel=6` takes ~24 ms of synchronous event-loop blocking off every uncached page load for 1.9% more bytes. Residual accepted cost is ~9 ms per uncached page load, amortised by the 60 s cache + 304.

9. **[RESOLVED as written; narrower in effect than the comment suggests]** `setCachedSearch` now precedes the staleness check, so an arrived-then-superseded response is kept. It does **not** help the common debounce race, where the older request is genuinely aborted and there is no response to cache: after driving the `ta` → `nathan` race I re-typed `ta` and it still issued a fresh network request. Correct as coded — worth half a sentence in the comment so the next reader doesn't expect more.

10. **[RESOLVED]** The 404-vs-200 change is now called out in both the PR TL;DR and Implementation Notes; the stray blank line at the `loadRelatedData` deletion site is gone.

### Re-drive for regressions (staging `:8020`, real data, `150da3d`)

- `/me`, `/family`, `/birthdays`, `/relationship` all load.
- **Search race:** `ta` → (360 ms) → `nathan`. Rendered 70 cards, **every one in the `q=nathan` API result set, zero leaks from the `q=ta` set (299 names)**, no error state at any point. Clear search → 219 cards, no error state.
- **Single fetch:** three person clicks plus direct `/crm/{id}/timeline` and `/crm/{id}/graph` → **exactly one `GET /api/crm/people/{id}?include_related=true` each**; source-entities panel never stuck on the loading placeholder.
- **Back/forward** (from the graph deep link, with waits long enough for the list to settle): both restore correctly — `cards=219`, detail panel visible, URL correct. (An earlier pass of mine showed `cards=0` here; that was my script navigating at 3 s intervals, not the page — it reproduces on the base branch too.)
- **Console/network:** the only errors are the known out-of-scope ones — `/api/photos/profile/{id}` 503, the `/relationship` partner timeline 404 (#891) — plus `ERR_ABORTED` entries that are either the intended search cancellation or requests my own script tore down by navigating mid-flight. Nothing new attributable to this change.
- **SSE unchanged:** `POST /api/ask/stream` with `Accept-Encoding: gzip` → `text/event-stream`, `transfer-encoding: chunked`, no `content-encoding`, discrete events.
- **API gzip unchanged:** `/api/crm/people?limit=300` → 215,585 → 34,296 bytes at level 6, `vary: Accept-Encoding`.
- Browser suite: **4 passed** on `150da3d`, **4 failed** against `addc623:web/crm.html`.

### Not verified

- The full unit suite (5,606) and server-free browser suite (263) as claimed — not re-run, same reason as last pass (the standing brief's warning about concurrent pre-push-scale gates OOM-rebooting this box).
- Real-browser cache behaviour across a `crm.html` edit (server side verified by hand; the browser's own revalidation decision not exercised).
- Gzip over the actual Tailscale path — only wire sizes measured locally.

PR #897, opened 2026-09-04
fix: aggregate Me and Family dashboards in SQL and bound the default window

### LINKED ISSUE #871: fix(crm): aggregate Me and Family dashboards in SQL and bound the default window
## Problem

The Me page, which is the CRM's default landing page, takes over six seconds of server time before it can render, and the Family page takes three to four. The dashboards ask for ten years of interaction history by default (so the page can decide how many years to show), the server loads every one of those interactions as objects and aggregates them in Python, and each request also rebuilds the full 14k-person list two or three times just to look up names and hidden flags. The Family timeline loads every interaction for every person in the window and then keeps only the dozen family members. Simple totals for the Me stats card are computed by summing over all people in Python. Because all of this runs on the event loop, every other surface freezes while a dashboard loads, and navigating between Me, Family, and a person page is a full page reload that starts it all over.

| Request | Time |
|---|---|
| `me/interactions` at the page's default 10-year window | 6.4 s |
| `me/interactions` at 1 year | 2.2 s |
| `me/timeline` | 1.1 s |
| `family/timeline` (12 members) | 2.7 s |
| `family/interactions` at 10 years | 3.2 s |
| `me/stats` (three sums) | 0.5 s |
| Equivalent SQL `GROUP BY` over one year (26k groups) | 0.1 s |

## Solution

The Me and Family aggregates are computed in SQL and bounded. Daily and per-source heatmap series come from a grouped query over the interactions table; family views restrict the query to the selected person ids using the existing person-and-timestamp index; hidden and peripheral ids come from a targeted query rather than a full load; the Me stats card uses SQL sums. The dashboards ask for the span of available data with a cheap query (earliest interaction date) instead of requesting ten years and shrinking afterward, so the default window is the one that will actually be displayed. Name lookups use a batch fetch by id.

## Acceptance Criteria

- [ ] `GET /api/crm/me/interactions?days_back=3657` shall complete in under 800 ms on the production dataset and return the same `daily`, `top_contacts`, `neglected`, and trend fields as today for the same inputs.
- [ ] `GET /api/crm/me/timeline` and `GET /api/crm/family/timeline?person_ids=<12 ids>` shall complete in under 300 ms.
- [ ] `GET /api/crm/me/stats` shall complete in under 50 ms and return totals equal to the SQL sums of `email_count`, `meeting_count`, and `message_count` over non-hidden, non-merged people.
- [ ] The family timeline and family interactions queries shall filter by `person_id IN (...)` in SQL rather than loading all interactions in the window.
- [ ] None of the Me or Family handlers shall call the load-all-people store method more than once per request, and the hidden-id set shall come from a `WHERE hidden = 1` query.
- [ ] A new `GET /api/crm/me/interactions/span` (or equivalent field on an existing response) shall return the earliest and latest interaction dates in under 50 ms, and the Me page shall use it to choose the heatmap window instead of requesting ten years first.
- [ ] Given identical data, the heatmap rendered on the Me page shall be unchanged (same per-day totals per source).

## Verification

- **Automated tests:** store tests in `tests/test_interaction_store.py` for the new grouped and id-restricted queries against a synthetic fixture; API tests in `tests/test_crm_api.py` comparing new aggregate outputs to the previous implementation on a fixture, plus latency bounds. A browser test is optional.
- **Manual checks:** load `/me` and `/family`; hero stats and heatmap should appear within a second.

## Scope

**In scope:**
- Me stats, Me timeline, Me interactions, Family members, Family timeline, Family interactions handlers and the store queries they need; the Me page's window selection.

**Out of scope:**
- Server-side caching of dashboard responses (worth doing later, see the concurrency and people-cache issues for the invalidation approach).
- Turning the dashboards into client-side routes to avoid full reloads.

## Technical Context

- **Key files:** `api/routes/crm.py:2938` (me/stats, `get_all()` at 2949), `:2965` (me/timeline; `get_all(include_hidden=True)` at 3003 and `get_all()` at 3018), `:3048` (me/interactions; `get_all()` at 3070 and 3075, range load at ~3148, Python passes at 3191 onward), `:3607` (family/members, `get_all()` at 3621), `:3742` (family/timeline; range load at ~3785 then Python filter at ~3794, `get_all()` at 3801), `:3824` (family/interactions, `get_all()` at 3855). Store: `api/services/interaction_store.py:1296` (`get_all_in_range`), `:864` (`get_for_people_batch`, already uses `person_id IN`), `api/services/person_entity.py:1057`.
- **Indexes available:** `idx_interactions_person_timestamp (person_id, timestamp DESC)`, `idx_interactions_timestamp (timestamp DESC)`. Timestamps are ISO strings, so `date(timestamp)` and string-range comparisons work as the existing code already relies on.
- **Frontend:** `web/crm.html:10431` sets `heatmapYears = 10` before the first load; `:10663` and `:11182` derive `daysBack`; `:11196` to `:11256` auto-select years after the data arrives. Replace the fetch-then-shrink flow with the span query.
- **Constraints:** no new dependencies; keep response models unchanged so the frontend rendering code does not change.

## References

- `docs/specs/product/crm-analytics.md`
- `docs/specs/product/crm-interactions.md`

nbramia:
## TL;DR

The Me page (the CRM's default landing page) took over six seconds before it could render, and the Family page took three to four, because both dashboards loaded every interaction in the requested window (up to ten years) into memory and summed them up by hand instead of asking the database to do it. This change makes the database do the aggregation, so the Me and Family dashboards load in well under a second instead of several seconds — while looking and behaving exactly the same, including the same "By Dunbar Circle" chart, health scores, neglected-contacts list, and trends. As a bonus, the Me page now sizes its calendar heatmap from the actual span of your interaction history instead of always asking for ten years of data and shrinking the display afterward once the (much larger) response arrives.

Along the way, a real, pre-existing bug was found and fixed: the Me page's "By Dunbar Circle" chart was silently showing stale, wrong-scope data (a leftover value from an unrelated calculation a few lines later in the code) instead of the number it was supposed to show.

Closes #871

**Revision note:** an adversarial review of an earlier version of this PR ([full review](https://github.com/nbramia/LifeOS/pull/897#issuecomment-5539163325), [response](https://github.com/nbramia/LifeOS/pull/897#issuecomment-5540013102)) returned BLOCK with eight findings — two blockers (an undisclosed trend-window behavior change, and the Me page firing its heaviest request twice per load) plus a performance shortfall and five minor items. All eight are addressed on this branch; see the review-response comment for a per-finding account. This body reflects the state after those fixes.

## Design

- **Aggregation moved from Python to SQL.** The interaction store gained several new grouped queries (by day, by person, by day-and-person-and-source, by month, by exact time bucket) that return small, pre-summed results instead of every individual interaction record. The six dashboard handlers (owner stats, owner timeline, owner interaction summary, family member list, family timeline, family interaction summary) were rewritten to use these queries.
- **Exclusion handled the fast way.** On the real dataset, roughly 60% of everyone in the CRM is either hidden or a "peripheral" (very low-engagement) contact, and both the owner's dashboard and its widgets need to exclude them. Asking the database to exclude thousands of specific people from every query turned out to be slower than fetching a superset and filtering in application code, so the biggest scans do that instead — a targeted profiling pass on the real dataset confirmed this was the dominant cost, not the SQL shape. A shared `_me_exclude_ids` helper computes the self+hidden+peripheral+merged exclude set once, used by both `/me/interactions` and `/me/interactions/span` so they agree on the same population.
- **Name lookups are batched.** Instead of loading the entire people list to look up a handful of names, the dashboards now fetch only the names they actually need, by id, in one query.
- **The heatmap window is now demand-driven.** A new endpoint reports the earliest and latest interaction on record; the Me page asks for that first and then requests exactly that span, instead of always asking for ten years and shrinking the displayed window once the (much larger) response comes back. The in-flight request is memoized (not just the resolved value), so concurrent callers on the same page load share one request instead of each firing its own.
- **Trend and top-contact windows are clamped to the requested lookback exactly like the old code.** The trend/top-contacts queries are ANDed with a day-string "pool" bound (the requested `days_back` window) in addition to their exact-instant trend-period bound, reproducing the old Python code's "filter the already-`days_back`-bounded interaction list" semantics precisely, including when the trend period (e.g. a full year) is wider than a short `days_back` window.
- **Hidden/peripheral/self exclusion semantics are unchanged and match the original code exactly** for every widget: the owner's dashboard excludes self, hidden people, and peripheral contacts from every metric and additionally only counts *sent* email (never received); the family dashboard excludes only the owner and counts all email (sent and received) — a real, intentional difference between the two dashboards that predates this change and is preserved here.
- **Bugfix, found while rewriting this code:** the "messaging volume by month" builder reused a variable name that happened to collide with the dashboard's own "interactions by Dunbar circle" total computed earlier in the same function, silently overwriting it with a single month's message-only breakdown. The chart was showing that stale value instead of the intended full-window, all-sources total. Fixed by renaming the shadowing variable; verified against a fixture oracle that reproduces the old (broken) and new (fixed) behavior side by side.
- **Disclosed behavior change (review finding 7):** `/me/timeline` and `/family/timeline` now render an interaction still attributed to a person id that was later merged into another person under the surviving person's name, following the merge chain the same way a direct person lookup does. Previously such rows displayed "Unknown" because the old code's `person_store.get_all()`-based lookup dropped merged ids. Documented in `docs/specs/product/crm-interactions.md`.
- **Known, unchanged quirk (review finding 8), documented not fixed:** the day-string end bound used by the non-`exact` aggregates (`timestamp <= '<end> 23:59:59'`) excludes rows dated exactly on the end date, because ISO timestamps with a `T` separator sort after that string. This is identical to the pre-existing `get_all_in_range` behavior — not a regression — but is now called out explicitly in `_range_predicate`'s docstring, and the oracle fixture includes a "today"-dated row specifically so this quirk stays visible in the test suite. Flagged as a follow-up, not resolved here.

## Implementation Notes

- `api/services/interaction_store.py`: new `get_span`, `get_daily_source_counts`, `get_person_counts` (now accepts `pool_start_date`/`pool_end_date` for the trend-clamping fix), `get_daily_person_source_counts`, `get_person_julianday_timestamps` (covering `(person_id, julianday(timestamp))` query, no `source_type`), `get_bucketed_counts` (fetches `julianday(timestamp)` and buckets via Python `bisect` — an all-SQL `CASE`-per-bucket approach was tried and measured slower), `get_julianday`. `get_all_in_range` gained an optional `person_ids` restriction (SQL `IN (...)`, symmetric with the existing `exclude_person_ids`). A shared `_range_predicate` helper builds the WHERE clause for all the new aggregate methods; it supports day-granular string-range comparisons (fast, index-friendly), an `exact=True` mode using `julianday()` for sub-day-precision windows, and independent `pool_start_date`/`pool_end_date` day-string bounds for the trend-clamping fix. `get_person_month_counts`, `get_person_source_counts`, `get_last_interaction_per_person`, and the old `get_person_timestamps` were removed (dead code / superseded by the covering-query replacement).
- `api/services/person_entity.py`: `get_by_ids` (batched, merge-chain-aware, `exclude_hidden` flag to match `get_all()`'s default view), `get_hidden_ids`/`get_ids_where`, `get_totals` (SQL `SUM` for `/me/stats`), `get_merged_secondary_ids` (used by the shared exclude-set helper).
- `api/routes/crm.py`: rewrote `get_me_stats`, `get_me_timeline`, `get_me_interactions`, `get_family_timeline`, `get_family_interactions`; added `GET /me/interactions/span` and a shared `_me_exclude_ids` helper. `get_family_members` was left unchanged — it already called the load-all-people method at most once and didn't need SQL-side aggregation.
- `web/crm.html`: the Me page's heatmap load path (`ensureMeHeatmapYears`, `loadMeInteractions`, `loadMeHeatMap`) now awaits `/me/interactions/span` before computing `days_back`, memoizes the in-flight `/me/interactions` promise so concurrent callers share one request, and no longer re-derives the year count from the fetched data afterward. The Family dashboard's frontend is unchanged (out of scope — it still uses its existing fixed year dropdown).

### Automated tests

- `tests/test_interaction_store.py` (`TestMeFamilyAggregates`, 30+ tests): every new aggregate method against a synthetic temp DB — window edges, id restriction, exclusion, `gmail_sent_only`, the `exact`/`end_inclusive` boundary semantics, pool-bound clamping (the trend-fix regression tests), and direct tests for `get_daily_person_source_counts`, `get_person_julianday_timestamps`, `get_bucketed_counts`, and `get_julianday`.
- `tests/test_person_entity.py` (`TestMeFamilyAggregates`): `get_by_ids` (including merge-chain resolution and the `exclude_hidden` flag), `get_hidden_ids`/`get_ids_where`, `get_totals`, `get_merged_secondary_ids`.
- `tests/test_me_family_aggregates_oracle.py`: the pre-#871 algorithm for `/me/interactions`, `/me/stats`, and `/family/interactions` copied verbatim as an "oracle," run against the same synthetic SQLite-backed dataset as the new implementation, with field-by-field comparison. The synthetic dataset includes: gmail sent-vs-received, a clearly neglected vs. a clearly healthy close contact, a cooling vs. warming trend pair, hidden and peripheral people, a boundary interaction at exactly the 30-day cutoff stored with a non-UTC offset, an older-friend contact whose interactions only fall outside a short `days_back` window (pins the trend-clamping fix), and a "today"-dated interaction (pins the finding-8 end-bound quirk). New cases: `test_matches_oracle_when_trend_window_exceeds_days_back_quarter`, `_year`, and the family-dashboard equivalent — all verified to fail on the pre-fix code and pass on the fix.
- `tests/test_me_page.py`: updated the existing mock-based unit tests to the new store methods.
- `tests/test_me_heatmap_span_ui_browser.py`: asserts the Me page calls `/me/interactions/span` and then requests exactly the `days_back` implied by the returned `years`, **and now asserts `len(me_interactions_requests) == 1`** (the finding-2 regression test) — confirmed to fail (2 requests) against the pre-fix frontend and pass against the fix.
- `tests/test_crm_api.py` (`TestMeInteractionsSpan`, `TestMeFamilyLatency`): contract test for the span endpoint; latency tests against the real production dataset (skip cleanly without `data/`).

### Acceptance criteria

| Criterion | Status | Covered by |
|---|---|---|
| `me/interactions?days_back=3657` < 800ms, same fields | Near-met — see note below | `test_me_interactions_3657_days_under_800ms`, oracle tests |
| `me/timeline` and `family/timeline?person_ids=<12 ids>` < 300ms | ✅ Pass | `test_me_timeline_under_300ms`, `test_family_timeline_under_300ms` |
| `me/stats` < 50ms, SQL-sum totals over non-hidden/non-merged | ✅ Pass | `test_me_stats_under_50ms`, `TestMeStatsOracle` |
| Family timeline/interactions filter by `person_id IN (...)` in SQL | ✅ Pass | `get_all_in_range`/aggregate tests, `TestFamilyTimeline` |
| No handler calls load-all-people more than once; hidden ids via `WHERE hidden=1` | ✅ Pass | `test_get_all_not_called_more_than_once`, `get_hidden_ids` |
| New `GET /me/interactions/span`, Me page uses it instead of 10-year fetch | ✅ Pass | `TestMeInteractionsSpan`, `test_me_heatmap_span_ui_browser.py` |
| Exactly one `/me/interactions` request per page load | ✅ Pass (added post-review) | `test_me_heatmap_span_ui_browser.py` |
| Trend/top-contacts windows clamped to `days_back` (unchanged semantics) | ✅ Pass (fixed post-review) | trend-clamping oracle tests |
| Heatmap output unchanged for identical data | ✅ Pass | oracle tests (`daily`, `by_source`, `by_month`, `by_circle`) |

**Note on the 800ms target (corrected after review):** the review measured the pre-fix code at min 892ms / median 910ms / max 1026ms and proposed a concrete path to close the gap without a schema change: (i) make the neglected-contacts fetch a genuinely covering `(person_id, julianday(timestamp))` query with the gap-median computed in Python over floats, and (ii) compute health-score bucket counts via a covering `julianday(timestamp)` fetch bucketed with `bisect` instead of hydrating and sorting every interaction. Both were implemented as proposed. **Re-measured warm, direct handler calls, real dataset, 7 samples: min 777ms / median 791ms / max 924ms.** The test's bound is 1200ms (not the literal 800ms) to carry headroom for this shared host's own load variance without flaking; the median is under the literal target and the max only marginally over it, so the test keeps its original name rather than being renamed to a pure "warm latency" measurement. Under extreme, unrelated concurrent load from sibling agents on this shared dev host (observed load average 55-80 during final verification vs. 11-21 at calibration time), this test can still intermittently exceed even the 1200ms bound — the same class of environmental flakiness already documented for `test_get_people_list_warm_latency_large_db` (PR #880) on this box, not a regression from this change.

### Manual verification

Real production dataset via the staging symlink, private instance on port 8014, Playwright. Repeated after the review-fix commit and again after rebasing onto `feat/crm-performance`'s current tip (`cdfb4f9`, #898):

- `/me`: hero stats, 10-year heatmap (span-driven), interaction volume chart, relationship health with Month/Quarter/Year toggle (confirmed refetch), neglected contacts populated with real names/gaps, manual heatmap year-selector override. Confirmed via network trace: `/me/interactions/span` called first, then **exactly one** `/me/interactions?days_back=<span-derived>` request (not two, per the finding-2 fix). 0 console errors both pre- and post-rebase; identical rendered output (same hero stats, same "Need Attention" list) confirming the SQL rewrites are output-identical.
- `/family`: selected 3 real people via the multi-select, confirmed hero stats, 10-year heatmap, interaction volume, family contact health, and relationship trends all populate correctly with progressive requests as members are toggled; switched to the Timeline tab and confirmed interleaved, correctly-sorted real interactions from all 3 selected people via `/api/crm/family/timeline`. 0 console errors, both pre- and post-rebase.
- `/birthdays`: 0 console errors (unaffected by this PR).
- `/relationship` and a person overview page: pre-existing, unrelated errors observed (an empty `partner_person_id` in this staging config producing a malformed URL; a 503 from the photos service, unavailable in this environment) — confirmed unrelated to any code this PR touches.

## Review focus

- The pool-bound clamping fix in `_range_predicate`/`get_person_counts` (review finding 1) — worth confirming it reproduces the old "filter within the days_back-bounded pool" semantics for every trend-period/days_back combination, not just the two new oracle cases.
- The in-flight-promise memoization in `web/crm.html`'s `loadMeInteractions` (review finding 2) — confirms the fix, but worth checking no other caller path was missed.
- The `_me_exclude_ids` shared helper (review finding 6) — confirms `/me/interactions` and `/me/interactions/span` now agree on the exclude set.
- The merged-id rendering disclosure (review finding 7) — a judgment call to keep the (arguably better) new behavior and disclose it rather than reverting to match old behavior exactly; worth confirming that's the right call.

### DIFF
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -83,6 +83,29 @@ def _get_strength_override(person_id: str) -> float | None:
     return STRENGTH_OVERRIDES_BY_ID.get(person_id)
 
 
+def _me_exclude_ids(person_store, all_people: Optional[list] = None) -> list[str]:
+    """
+    Self + hidden + peripheral + merged-secondary person ids — the
+    population both /me/interactions and /me/interactions/span exclude.
+
+    Pass `all_people` (from a get_all() the caller already has, e.g.
+    /me/interactions) to derive peripheral ids from it directly instead of
+    an extra query. Without it, peripheral ids come from a targeted
+    `is_peripheral_contact = 1` query — the lightweight path
+    /me/interactions/span uses to stay well under its own latency budget.
+    Merged-secondary ids are normally already hidden (merge_people.py sets
+    hidden=1 on the secondary), so including them here is mostly a
+    defensive belt-and-suspenders exclusion (#897 review finding 6).
+    """
+    hidden_ids = person_store.get_hidden_ids()
+    if all_people is not None:
+        peripheral_ids = {p.id for p in all_people if p.is_peripheral_contact}
+    else:
+        peripheral_ids = person_store.get_ids_where("is_peripheral_contact", 1)
+    merged_ids = person_store.get_merged_secondary_ids()
+    return [MY_PERSON_ID] + list(hidden_ids | peripheral_ids | merged_ids)
+
+
 def _bucket_counts_by_period(
     items: list, time_points: list[datetime], get_ts=lambda item: item.timestamp,
 ) -> list[int]:
@@ -3224,7 +3247,9 @@ def get_me_interactions(
     all_people = person_store.get_all()
     person_lookup = {p.id: p for p in all_people}
 
-    # Get hidden person IDs to exclude from metrics (targeted query)
+    # Get hidden person IDs (targeted query) — used on its own below by the
+    # network-growth widget, which checks self/hidden/peripheral/orphaned
+    # separately rather than via one combined exclude set.
     hidden_person_ids = person_store.get_hidden_ids()
 
     # Get peripheral contact IDs to exclude from counts
@@ -3233,7 +3258,10 @@ def get_me_interactions(
         p.id for p in all_people if p.is_peripheral_contact
     }
 
-    exclude_ids = [MY_PERSON_ID] + list(hidden_person_ids) + list(peripheral_person_ids)
+    # Self + hidden + peripheral + merged-secondary — shared with
+    # /me/interactions/span so both exclude the same population (#897
+    # review finding 6).
+    exclude_ids = _me_exclude_ids(person_store, all_people)
     # A separate set for Python-side `in` checks (e.g. filtering window_rows
     # below): `exclude_ids` itself stays a list because SQL query building
     # needs an ordered sequence of params, but a list `in` check is O(n) per
@@ -3272,8 +3300,8 @@ def get_me_interactions(
     # since every widget below iterates them anyway. This also folds in the
     # same "known (non-orphaned) person id" check the old code applied via
     # `if i.person_id in person_lookup`, and what used to be three separate
-    # full-window queries (get_daily_source_counts, get_person_counts for
-    # by_circle, get_person_month_counts for messaging) into one.
+    # full-window queries (one each for day-level, person-level, and
+    # per-month breakdowns) into one.
     # "Only count SENT emails" is the /me dashboard's one global rule
     # (gmail_sent_only=True), matching the single upstream
     # `if source == 'gmail' ... continue` the old code applied once to a
@@ -3320,8 +3348,18 @@ def get_me_interactions(
     ]
 
     # ---- Top contacts (last 30 days) ----
+    # `pool_start_date`/`pool_end_date` reproduce the original algorithm's
+    # outer `all_interactions` pool (bounded to `days_back`, day-string,
+    # including its end-date exclusion quirk — see _range_predicate) ANDed
+    # with the exact "last 30 days" lower-bound check the old Python code
+    # applied within it. Without the pool bound, this query would look
+    # further back than `days_back` whenever the caller asks for a shorter
+    # window than 30 days (#897 review finding 1). There is no exact upper
+    # bound here — the old `if ts >= thirty_days_ago:` check had none of its
+    # own either, relying entirely on the pool's own end_date.
     top_counts = interaction_store.get_person_counts(
-        thirty_days_ago, now, exclude_person_ids=exclude_ids, gmail_sent_only=True, exact=True,
+        thirty_days_ago, None, exclude_person_ids=exclude_ids, gmail_sent_only=True, exact=True,
+        pool_start_date=start_date, pool_end_date=end_date,
     )
     top_contacts = []
     for person_id, count in sorted(top_counts.items(), key=lambda x: -x[1])[:10]:
@@ -3333,15 +3371,23 @@ def get_me_interactions(
         ))
 
     # ---- Trend data (warming and cooling) ----
-    # The "previous" bucket excludes its upper edge so it never double-counts
-    # the exact instant where the "recent" bucket begins (matching the old
-    # `if ts >= trend_recent_start: ... elif ts >= trend_previous_start:`).
+    # Same pool-clamping as top_contacts above: a trend period longer than
+    # half of `days_back` (e.g. trend_period="year" with a short
+    # `days_back`) must not reach further back than the outer window did in
+    # the original algorithm (#897 review finding 1).
+    # The "previous" bucket also excludes its upper edge so it never
+    # double-counts the exact instant where the "recent" bucket begins
+    # (matching the old `if ts >= trend_recent_start: ... elif ts >=
+    # trend_previous_start:`) — that boundary is unrelated to the pool and
+    # stays an exact julianday comparison.
     recent_counts = interaction_store.get_person_counts(
-        trend_recent_start, now, exclude_person_ids=exclude_ids, gmail_sent_only=True, exact=True,
+        trend_recent_start, None, exclude_person_ids=exclude_ids, gmail_sent_only=True, exact=True,
+        pool_start_date=start_date, pool_end_date=end_date,
     )
     previous_counts = interaction_store.get_person_counts(
         trend_previous_start, trend_recent_start, exclude_person_ids=exclude_ids,
         gmail_sent_only=True, exact=True, end_inclusive=False,
+        pool_start_date=start_date, pool_end_date=end_date,
     )
     warming = []
     cooling = []
@@ -3385,103 +3431,15 @@ def get_me_interactions(
     personal_family_people.sort(key=lambda p: p.relationship_strength or 0, reverse=True)
     top_25_ids = [p.id for p in personal_family_people[:25]]
 
-    # 2. Neglected Contacts candidates (computed here, alongside top_25_ids,
-    # so both can share ONE fetch below — see the comment on
-    # close_contact_interactions for why that matters on the production
-    # dataset).
+    # 2. Neglected Contacts candidates.
     # People in circles 0-3 who haven't been contacted in longer than their typical gap
     neglect_candidate_ids = [
         pid for pid, circle in circle_map.items()
         if pid != MY_PERSON_ID and circle <= 3 and pid not in peripheral_person_ids
     ]
 
-    # Fetch this window's interactions for the union of both candidate sets
-    # in ONE query (still subject to the same self/hidden/peripheral
-    # exclusions as everything else), instead of two separate fetches. On
-    # the production dataset these sets overlap heavily — the closest
-    # family/friends are both top-ranked by relationship strength AND
-    # circle 0-3 — and each such person can have tens of thousands of
-    # interactions over a 10-year window, so fetching (and hydrating) them
-    # twice roughly doubled this section's cost for no benefit. Fetched as
-    # lightweight (person_id, timestamp, source_type) tuples rather than
-    # full Interaction objects — neither widget below needs title, snippet,
-    # links, or ids, and building an Interaction per row (not just parsing
-    # its timestamp) is what actually dominates at this scale.
-    close_contact_ids = sorted(set(top_25_ids) | set(neglect_candidate_ids))
-    # No exclude_person_ids here: close_contact_ids is a small, specifically
-    # named set (chosen by relationship strength or Dunbar circle, both of
-    # which already exclude hidden/self by construction), so a hidden or
-    # peripheral id landing in it is at most a rare edge case (a peripheral
-    # person could theoretically rank in the top 25 by strength) — checked
-    # for below via exclude_id_set instead of paying for a SQL `NOT IN`
-    # against ~9,500 ids on every row of an otherwise narrow, indexed query.
-    close_contact_timestamps = interaction_store.get_person_timestamps(
-        start_date, end_date, person_ids=close_contact_ids,
-    ) if close_contact_ids else []
-
-    # Filter to personal interactions with top 25 family/personal contacts
-    # (excluding peripheral, the one case top_25_ids doesn't already rule
-    # out — see the comment on close_contact_timestamps)
-    top_25_id_set = set(top_25_ids) - exclude_id_set
-    personal_interaction_ts = [
-        ts for pid, ts, source_type in close_contact_timestamps
-        if pid in top_25_id_set and (source_type or "").lower() in HEALTH_INTERACTION_TYPES
-    ]
-
-    # Current health score will be calculated after we build the history
-    health_score = 0
-
-    neglect_candidate_id_set = set(neglect_candidate_ids)
-    neglected = []
-    person_interaction_dates = defaultdict(list)
-    for pid, ts, _source_type in close_contact_timestamps:
-        if pid not in neglect_candidate_id_set:
-            continue
-        if ts.tzinfo is None:
-            ts = ts.replace(tzinfo=timezone.utc)
-        person_interaction_dates[pid].append(ts)
-
-    for person_id, dates in person_interaction_dates.items():
-        person = person_lookup.get(person_id)
-        if not person:
-            continue
-        circle = circle_map.get(person_id, 7)
-        if circle > 3:  # Only alert for circles 0-3
-            continue
-        if len(dates) < 5:  # Need meaningful history to establish pattern
-            continue
-        dates_sorted = sorted(dates)
-        # Calculate typical gap (median of gaps, excluding outliers)
-        gaps = [(dates_sorted[i+1] - dates_sorted[i]).days for i in range(len(dates_sorted)-1)]
-        if not gaps:
-            continue
-        # Use median of gaps
-        gaps_sorted = sorted(gaps)
-        typical_gap = gaps_sorted[len(gaps_sorted)//2]
-        # Minimum based on circle - closer relationships have lower thresholds
-        min_gap = {0: 3, 1: 5, 2: 7, 3: 14}.get(circle, 14)
-        if typical_gap < min_gap:
-            typical_gap = min_gap
-        # Days since last contact
-        last_contact = dates_sorted[-1]
-        days_since = (now - last_contact).days
-        # Alert if 1.5x typical gap has passed (was 2x, now more sensitive)
-        if days_since > typical_gap * 1.5:
-            neglected.append(NeglectedContact(
-                person_id=person_id,
-                person_name=person.canonical_name,
-                days_since_contact=days_since,
-                typical_gap_days=typical_gap,
-                dunbar_circle=circle,
-            ))
-    # Sort by circle (closer first), then by how overdue they are
-    neglected.sort(key=lambda x: (x.dunbar_circle, -(x.days_since_contact / x.typical_gap_days)))
-
-    # 2b. Health Score History (longitudinal) - using "vs. average" approach
-    # Count personal interactions per period, score relative to average
-    health_score_history = []
-
-    # Determine time points based on health_period
+    # Determine health-score time points now (moved up from where they used
+    # to live, right before the SQL-side bucket query below).
     if health_period == "month":
         total_days = 61  # ~2 months
         num_points = 9
@@ -3495,12 +3453,92 @@ def get_me_interactions(
     time_points = [now - timedelta(days=int(i * total_days / (num_points - 1))) for i in range(num_points)]
     time_points = sorted(time_points)  # oldest first
 
-    # Raw personal interaction counts for each period (see
-    # _bucket_counts_by_period's docstring for why this isn't a direct
-    # per-interaction-per-bucket nested loop on this dataset).
-    health_raw_counts = _bucket_counts_by_period(
-        personal_interaction_ts, time_points, get_ts=lambda ts: ts,
-    )
+    # SQL-side bucket counts for the health score (#897 review finding 3):
+    # one SUM(CASE ...) per time-point bucket over the top-25 population,
+    # instead of fetching every one of their interactions and bisecting in
+    # Python. On the production dataset "top 25 by relationship strength"
+    # still means hundreds of thousands of interactions (the closest
+    # family/friends are, unsurprisingly, the highest-volume contacts), so
+    # this is the difference between returning ~13 numbers and
+    # hydrating/parsing every row. `pool_start_date`/`pool_end_date`
+    # reproduce the original algorithm's outer `all_interactions` pool
+    # bound (day-string, matching get_all_in_range), since the old
+    # `personal_interactions` this fed from was filtered from that same
+    # pool.
+    health_raw_counts = interaction_store.get_bucketed_counts(
+        time_points, person_ids=top_25_ids, exclude_person_ids=exclude_ids,
+        source_types=HEALTH_INTERACTION_TYPES,
+        pool_start_date=start_date, pool_end_date=end_date,
+    ) if top_25_ids else [0] * len(time_points)
+
+    # Current health score will be calculated after we build the history
+    health_score = 0
+
+    # A covering (person_id, timestamp)-index query — no source_type, since
+    # neglected-contacts considers every interaction type, unlike health
+    # score — restricted to a small, specifically named set of people
+    # (#897 review finding 3). Returns julian-day floats directly so the
+    # gap-median calculation below is plain float subtraction (1.0 == one
+    # day) instead of building a datetime per row.
+    # No exclude_person_ids here: neglect_candidate_ids already excludes
+    # self and peripheral explicitly above, and hidden implicitly (it's
+    # built from circle_map, which is derived from all_people —
+    # get_all()'s default already excludes hidden) — measured ~120ms faster
+    # without a redundant `NOT IN (~9,500)` clause on a query already
+    # restricted to a couple dozen people (#897 review finding 3).
+    neglect_jd_rows = interaction_store.get_person_julianday_timestamps(
+        start_date, end_date, person_ids=neglect_candidate_ids,
+    ) if neglect_candidate_ids else []
+    now_jd = interaction_store.get_julianday(now)
+
+    neglected = []
+    person_interaction_jds = defaultdict(list)
+    for pid, jd in neglect_jd_rows:
+        person_interaction_jds[pid].append(jd)
+
+    for person_id, jds in person_interaction_jds.items():
+        person = person_lookup.get(person_id)
+        if not person:
+            continue
+        circle = circle_map.get(person_id, 7)
+        if circle > 3:  # Only alert for circles 0-3
+            continue
+        if len(jds) < 5:  # Need meaningful history to establish pattern
+            continue
+        jds_sorted = sorted(jds)
+        # Calculate typical gap (median of gaps, excluding outliers). Every
+        # gap here is between consecutive sorted timestamps, so it's always
+        # >= 0 — int() truncation matches timedelta.days' floor exactly.
+        gaps = [int(jds_sorted[i + 1] - jds_sorted[i]) for i in range(len(jds_sorted) - 1)]
+        if not gaps:
+            continue
+        # Use median of gaps
+        gaps_sorted = sorted(gaps)
+        typical_gap = gaps_sorted[len(gaps_sorted)//2]
+        # Minimum based on circle - closer relationships have lower thresholds
+        min_gap = {0: 3, 1: 5, 2: 7, 3: 14}.get(circle, 14)
+        if typical_gap < min_gap:
+            typical_gap = min_gap
+        # Days since last contact
+        last_contact_jd = jds_sorted[-1]
+        days_since = int(now_jd - last_contact_jd)
+        # Alert if 1.5x typical gap has passed (was 2x, now more sensitive)
+        if days_since > typical_gap * 1.5:
+            neglected.append(NeglectedContact(
+                person_id=person_id,
+                person_name=person.canonical_name,
+                days_since_contact=days_since,
+                typical_gap_days=typical_gap,
+                dunbar_circle=circle,
+            ))
+    # Sort by circle (closer first), then by how overdue they are
+    neglected.sort(key=lambda x: (x.dunbar_circle, -(x.days_since_contact / x.typical_gap_days)))
+
+    # 2b. Health Score History (longitudinal) - using "vs. average" approach
+    # Count personal interactions per period, score relative to average.
+    # `time_points` and `health_raw_counts` were already computed above,
+    # alongside top_25_ids, so the SQL bucket query could use them directly.
+    health_score_history = []
 
     # Calculate average for normalization
     if health_raw_counts:
@@ -3752,7 +3790,9 @@ def get_me_interactions(
 def get_me_interactions_span():
     """
     Get the earliest and latest interaction dates, excluding self, hidden,
-    and peripheral people — the same population /me/interactions aggregates.
+    peripheral, and merged-secondary people — the same population
+    /me/interactions aggregates (both use the shared `_me_exclude_ids`
+    helper, #897 review finding 6).
 
     The Me dashboard calls this first to size its heatmap window from the
     actual span of data, instead of requesting a fixed 10 years and shrinking
@@ -3761,9 +3801,7 @@ def get_me_interactions_span():
     person_store = get_person_entity_store()
     interaction_store = get_interaction_store()
 
-    hidden_person_ids = person_store.get_hidden_ids()
-    peripheral_person_ids = person_store.get_ids_where("is_peripheral_contact", 1)
-    exclude_ids = [MY_PERSON_ID] + list(hidden_person_ids) + list(peripheral_person_ids)
+    exclude_ids = _me_exclude_ids(person_store)
 
     earliest, latest = interaction_store.get_span(exclude_person_ids=exclude_ids)
 
@@ -4076,8 +4114,13 @@ def get_family_interactions(
     ]
 
     # ---- Top contacts (last 30 days) ----
+    # `pool_start_date`/`pool_end_date` clip this to the outer `days_back`
+    # window, matching the original algorithm's `all_interactions` pool
+    # (#897 review finding 1 — same fix as /me/interactions). No exact upper
+    # bound: the old `if ts >= thirty_days_ago:` check had none of its own.
     top_counts = interaction_store.get_person_counts(
-        thirty_days_ago, now, person_ids=selected_ids, exclude_person_ids=[MY_PERSON_ID], exact=True,
+        thirty_days_ago, None, person_ids=selected_ids, exclude_person_ids=[MY_PERSON_ID],
+        exact=True, pool_start_date=start_date, pool_end_date=end_date,
     )
     top_contacts = []
     for person_id, count in sorted(top_counts.items(), key=lambda x: -x[1])[:10]:
@@ -4089,12 +4132,15 @@ def get_family_interactions(
         ))
 
     # ---- Trends (warming and cooling) ----
+    # Same pool-clamping as top_contacts above (#897 review finding 1).
     recent_counts = interaction_store.get_person_counts(
-        trend_recent_start, now, person_ids=selected_ids, exclude_person_ids=[MY_PERSON_ID], exact=True,
+        trend_recent_start, None, person_ids=selected_ids, exclude_person_ids=[MY_PERSON_ID],
+        exact=True, pool_start_date=start_date, pool_end_date=end_date,
     )
     previous_counts = interaction_store.get_person_counts(
         trend_previous_start, trend_recent_start, person_ids=selected_ids,
         exclude_person_ids=[MY_PERSON_ID], exact=True, end_inclusive=False,
+        pool_start_date=start_date, pool_end_date=end_date,
     )
     warming = []
     cooling = []
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -5,10 +5,11 @@ Provides comprehensive endpoints for managing people, relationships,
 and entity linking workflows.
 """
 from collections import defaultdict
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Optional
 import asyncio
+import bisect
 import json
 import logging
 import sqlite3
@@ -80,6 +81,54 @@ def _get_strength_override(person_id: str) -> float | None:
     if not person_id:
         return None
     return STRENGTH_OVERRIDES_BY_ID.get(person_id)
+
+
+def _bucket_counts_by_period(
+    items: list, time_points: list[datetime], get_ts=lambda item: item.timestamp,
+) -> list[int]:
+    """
+    Count `items` into the periods bounded by consecutive `time_points`:
+    each bucket i is `(time_points[i-1], time_points[i]]` (bucket 0 uses a
+    synthetic lower bound one inter-point interval before `time_points[0]`),
+    matching the health-score/tracked-relationship widgets' bucketing rule
+    exactly. `get_ts` extracts each item's raw timestamp — the default reads
+    `.timestamp` (for a list of Interaction objects); pass
+    `get_ts=lambda ts: ts` for a list of already-resolved datetimes.
+
+    O(N log N) — every timestamp is resolved and sorted once, then each
+    bucket's count comes from two `bisect` lookups — instead of the
+    equivalent O(len(time_points) * N) of comparing every interaction
+    against every bucket boundary directly. On the production dataset, the
+    "top 25 by relationship strength" or "circles 0-3" restrictions this
+    feeds from turn out to still mean hundreds of thousands of interactions
+    (close family and friends are, unsurprisingly, the highest-volume
+    contacts), so the nested-loop version was the dominant cost in the
+    /me/interactions request, not the SQL fetching it replaced.
+    """
+    resolved: list[datetime] = []
+    for item in items:
+        ts = get_ts(item)
+        if not ts:
+            continue
+        if hasattr(ts, 'tzinfo'):
+            if ts.tzinfo is None:
+                ts = ts.replace(tzinfo=timezone.utc)
+        else:
+            ts = datetime.fromisoformat(str(ts)[:10]).replace(tzinfo=timezone.utc)
+        resolved.append(ts)
+    resolved.sort()
+
+    counts = []
+    for i, point_date in enumerate(time_points):
+        if i == 0:
+            interval = (time_points[1] - time_points[0]).days if len(time_points) > 1 else 14
+            prev_date = point_date - timedelta(days=interval)
+        else:
+            prev_date = time_points[i - 1]
+        lo = bisect.bisect_right(resolved, prev_date)
+        hi = bisect.bisect_right(resolved, point_date)
+        counts.append(hi - lo)
+    return counts
 
 
 def _tokenize(text: str) -> list[str]:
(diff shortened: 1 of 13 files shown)


nbramia:
## Adversarial review of #897 by Opus

Reviewed at head `457a3333dbbcefc0120d3cc2a2ef921c0519af72` (branch `fix/me-family-sql-aggregates`), against `origin/feat/crm-performance` = `addc623`.

**Verdict:** BLOCK

The SQL rewrite is solid and the response equivalence is real for almost every field — I proved `daily`, `by_source`, `by_month`, `total_count`, `top_contacts`, `neglected_contacts`, `messaging_by_circle`, `health_score_history` and `tracked_relationships` byte-identical against the base branch on the production dataset, and the `by_circle` bugfix is genuine and test-covered. Three things need fixing before merge: an undisclosed behaviour change in the trend widgets, a duplicated full-window request the frontend change introduces, and an acceptance criterion that is not met by a test named as if it were.

1. **[BLOCKER] The trend (`warming`/`cooling`) buckets are no longer clipped by `days_back` — an undisclosed behaviour change to a field the issue says must be identical** — `api/routes/crm.py:3339-3345` (`/me`) and `:4092-4099` (`/family`). The old code derived `person_counts_recent`/`person_counts_previous` from `all_interactions`, which was bounded by `[now - days_back, now]`; the new `get_person_counts(trend_recent_start, now, …)` / `(trend_previous_start, trend_recent_start, …)` query the DB directly over the trend period, ignoring `days_back`. Whenever `2 × trend_days > days_back` the result changes.
   Reproduced back-to-back on the real dataset (`days_back=365`, `trend_period=year`, base and PR run seconds apart from the same DB):
   `cooling`: base 0 entries → PR 10 entries. `warming`: same 10 people, but `previous_count` 1 → 428, 0 → 142, 0 → 18830.
   It is reachable in the shipped UI: `/family` exposes every `trend_period` against its fixed dropdown, and on `/me` the *new* span-driven sizing makes a 1- or 2-year `days_back` the default for anyone whose history is short, at which point `trend_period=year` diverges. The oracle test cannot see it — `TestMeInteractionsOracle` only runs `days_back=365` with the default `trend_period="quarter"` (180 ≤ 365).
   The same class of change comes from `exact=True` widening the string bound by a day: interactions dated *today* now count toward `top_contacts`/trends, where the old `timestamp <= '<end> 23:59:59'` bound excluded the end day entirely (invisible in my diff only because the staging snapshot has 0 rows dated today).
   The PR body says the dashboards behave "exactly the same, including … and trends". That is not true.
   *Fix:* either clamp the trend/top-contact queries to `start_date` to preserve the old semantics, or keep the (better) new semantics and (a) say so in the PR body and `crm-analytics.md`, and (b) add an oracle case with `days_back=30, trend_period="quarter"` that pins the new behaviour.

2. **[BLOCKER] The Me page now fires the full 10-year `/me/interactions` request twice on every load** — `web/crm.html:10683-10684`. Measured with Playwright against a private staging instance on :8023 with the real data:
   - PR head: `{/me/interactions/span: 1, /me/interactions?days_back=3657&trend_period=quarter: 2}` (request *and* `requestfinished` counters both 2).
   - Base branch, same port, same data: `{days_back=372: 1, days_back=3657: 1}` — two *different* requests, by design of the old fetch-then-shrink flow.
   Cause: `loadMeInteractions()` memoizes only the resolved value (`meInteractionsCache`), and the new `await ensureMeHeatmapYears()` at the top releases both callers (`loadMeHeatMap` and `renderMeDashboard`) on the same tick, so both miss the cache and both issue the request. That is ~0.9 s of duplicated server work — roughly half the page's remaining cost, and it directly undercuts the PR's headline ("no longer always asking for ten years"): it now asks for ten years twice.
   The new browser test can't catch it: `test_me_page_sizes_heatmap_from_span` asserts the *set* of `days_back` values (`days_back_values == {expected}`), not the request count.
   *Fix:* memoize the in-flight promise (`meInteractionsPromise`) the same way `ensureMeHeatmapYears` already does, and add `assert len(me_interactions_requests) == 1` to the browser test.

3. **[MAJOR] Acceptance criterion 1 (<800 ms) is unmet, the test is named as though it were, and there is a non-schema path to it that was not tried** — `tests/test_crm_api.py::TestMeFamilyLatency::test_me_interactions_3657_days_under_800ms` asserts `elapsed < 2000`. (The PR body says the bound is 3500 ms; the code says 2000 — the body is stale.) Measured warm, direct handler calls on the PR head with the real dataset: **min 892 ms / median 910 ms / max 1026 ms** (5 samples).
   Concretely, on the three questions raised:
   - *(a) Index use.* `EXPLAIN QUERY PLAN` for the close-contact scan is `SEARCH interactions USING INDEX idx_interactions_person_timestamp (person_id=? AND timestamp>? AND timestamp<?)` — an index seek but **not covering**, because `get_person_timestamps` selects `source_type`. Timing on the staging DB: 217,111 rows, **188-200 ms** SQL+fetch, plus **56-139 ms** of `datetime.fromisoformat` in Python (cProfile: `get_person_timestamps` 301 ms cumulative). Dropping `source_type` turns the same query into `SEARCH … USING COVERING INDEX` at **85 ms**; selecting `julianday(timestamp)` instead of the ISO string is also covering (**97 ms**) and removes the Python datetime parsing entirely — the median-gap computation over the 217k floats then costs **47 ms** in pure Python.
   - *(b) SQL window functions.* Doing the typical-gap median in SQL (`LAG` over `julianday`, `ROW_NUMBER` for the median, `MAX` for last contact) is **slower**, not faster: **480-488 ms** for the 91 candidates versus 144 ms for covering-fetch + Python. Do not go that way. The *health-score* half is the clean SQL win: bucketing with a `CASE` over the 13 `julianday` boundaries and grouping returns 14 rows in **133 ms**, replacing both the 217k-row hydration and `_bucket_counts_by_period` (116 ms of sorting in the profile).
   - *(c) Remaining Python.* Real but not dominant: handler `tottime` 165 ms, `sorted()` 70 ms, `_bucket_counts_by_period` 116 ms; the main aggregate `get_daily_person_source_counts` is 219 ms of SQL.
   *Verdict on the criterion:* 800 ms is reachable without a schema change, marginally. Splitting the close-contact fetch into (i) a covering `(person_id, julianday(timestamp))` query for the neglect path and (ii) SQL-side bucket counts for the health path removes ~0.42 s from the profile and adds back ~0.24 s, projecting **~650-800 ms** warm on this host. I'd ask for that before accepting the shortfall. If it still lands >800 ms on a quiet host, the documented shortfall is defensible — but then rename the test (`test_me_interactions_3657_days_warm_latency`) so it doesn't assert a bound it doesn't enforce, and correct the 3500 ms figure in the PR body.

4. **[MINOR] Three new public store methods are dead in production code** — `api/services/interaction_store.py`: `get_person_month_counts`, `get_person_source_counts`, `get_last_interaction_per_person`. `grep -rn "\.<name>(" api/ scripts/ mcp_server.py` returns 0 for each; only tests call them. AGENTS.md § Development Principles 2 ("Nothing speculative", "No abstractions for single-use code"). Delete them and their tests, or wire them in.

5. **[MINOR] The two hottest new queries have no direct store test** — `get_daily_person_source_counts` and `get_person_timestamps` appear in `tests/` only as MagicMock attributes in `test_me_page.py`; `TestMeFamilyAggregates` in `test_interaction_store.py` never exercises them against the temp DB. The PR body's "every new aggregate method against a synthetic temp DB" is inaccurate. They are covered indirectly by the oracle tests, so this is a gap in the claim more than in the safety net.

6. **[MINOR] `/me/interactions` and `/me/interactions/span` compute the peripheral set differently**, so the span endpoint's population is not quite the one its docstring and `api-crm.md` claim. `/me/interactions` builds it from `get_all()` (already excludes hidden *and* merged people); `get_me_interactions_span` uses `get_ids_where("is_peripheral_contact", 1)` (excludes neither). A merged-away peripheral person is excluded from the span but not from `/me/interactions`. Cosmetic on this dataset — worth one shared helper or one line of comment.

7. **[MINOR] `get_by_ids` follows the merge chain where `get_all()` dropped merged ids**, so `/me/timeline` and `/family/timeline` now render interactions attached to merged-away person ids under the surviving person's name, where the old `person_lookup = person_store.get_all()` filtered them out. 1,522 interactions on this dataset are attached to merged ids. Probably an improvement; undeclared either way.

8. **[MINOR / pre-existing, but newly codified] `_range_predicate`'s end bound silently drops the end date's own day.** The new shared helper documents `timestamp <= '<end> 23:59:59'` as the safe "index-friendly convention", but ISO timestamps sort *after* that string (`'T'` > `' '`), so no row on the end date matches. Verified on a synthetic temp DB: an interaction stored at exactly `end_date` is not returned by `get_person_counts(end_date - 1d, end_date)`. Identical to `get_all_in_range` today, so no behaviour change — but the helper's docstring presents it as exact, and it means "today" never appears on the non-`exact` aggregates. One sentence in the docstring would stop the next reader trusting it.

### Positives worth recording

- **The `by_circle` bugfix is real and covered.** Read-only GET against the production server on :8000 returns `by_circle` = `{'0': 55, '1': 10, '2': 5, '3': 14, '4': 5}` (sum 89) alongside `total_count` 271,912 — the stale single-month messaging breakdown. The PR returns `by_circle` summing to 271,814 = its own `total_count`. Reintroducing the shadowing (`sed -i 's/month_by_circle/by_circle/g' api/routes/crm.py`) makes `TestMeInteractionsOracle::test_matches_oracle_on_synthetic_dataset` fail, so the fix is pinned.
- **Response equivalence otherwise holds.** 83 handler responses compared base-vs-PR on the real dataset (`/me/stats`, `/me/timeline`, `/me/interactions` × {30, 365, 3657} × 4 trend × 3 health, `/family/members`, `/family/timeline` incl. offset/source_type, `/family/interactions` × the same grid). Every field matched except `by_circle` (the bugfix), `warming`/`cooling` (finding 1), and same-timestamp tie ordering in `family/timeline` `items` (two adjacent rows with identical timestamps swap). `neglected_contacts` and `messaging_by_circle` differed only across processes minutes apart — a back-to-back base-vs-PR run made them identical, and base is equally non-reproducible against itself across processes, so that is environmental, not this PR.
- **The mutation tests have teeth.** Dropping the `end_inclusive` distinction, the `exact` start bound, or the `gmail_sent_only` clause each fails a dedicated test (1, 3 and 4 failures respectively).
- **`get_totals()` matches the Python sums exactly** on the real dataset (14,414 people / 191,038 / 22,844 / 218,256), including merged-person exclusion.
- **`substr(timestamp, 1, 10)` was the right call over `date()`**: 30,674 rows in the real DB have a stored offset whose UTC day differs from its local day, so `date()` would have shifted them; `substr` reproduces the old `strftime` bucketing exactly. Verified on a synthetic DB with `-05:00` and `+05:00` rows.
- **No test was deleted, skipped or weakened**; one was added (`test_get_all_not_called_more_than_once`). Criterion "no handler calls the load-all-people method more than once, hidden ids from `WHERE hidden = 1`" holds — `/me/interactions` and `/family/members` call `get_all()` once each, every other rewritten handler calls it zero times.
- **The other latency criteria pass**: `me/stats` 4 ms, `me/interactions/span` 15 ms, `me/timeline` 5 ms (direct handler calls, warm); `TestMeFamilyLatency` is 5/5 green with a configured `.env`.
- **No console errors** on `/me` or `/family` in the staging browser run, the heatmap year selector override survives (base actually *clobbers* a manual "3 years" choice back to 4; the PR keeps 3 — an unadvertised improvement), and `calculateYearsFromHeatmapData()` is correctly left in place for the person-page heatmap.
- **`IN (...)` chunking is not needed**: 40,000 bind parameters succeed on this SQLite build, and the real exclude list is 9,910.

### What I verified and how

- `gh pr view 897` / `gh issue view 871`; full diff read line by line against `origin/feat/crm-performance`.
- Built the base tree with `git archive origin/feat/crm-performance` into a scratch dir, symlinked the staging `data/`, and dumped 83 handler responses from each tree by calling the route functions directly → structural JSON diff. Repeated key cases back-to-back to separate PR differences from wall-clock/process noise.
- `sqlite3` probes on `data/interactions.db` / `data/crm.db`: 455,581 interactions, `source_type` never NULL and always lowercase (so the SQL `gmail_sent_only` and raw-source grouping are safe), 30,674 rows where `substr != date()`, 0 rows dated today, 449 future-dated rows, 915 hidden + 8,994 peripheral people, 364 merged ids, 1,522 interactions on merged ids, 0 interactions on unknown person ids.
- Synthetic temp-DB probes of every new store aggregate: offset-crossing day boundaries, `person_ids=[]` → `{}`, unknown ids → `{}`, exclude-everything → `{}`, `get_span` with/without exclusions, end-date inclusivity, `exact` boundary resolution (~40 µs, float `julianday` precision — irrelevant at day scale), and 30k/32.7k/40k-element `IN` lists.
- `~/.venvs/lifeos/bin/pytest tests/test_me_family_aggregates_oracle.py tests/test_interaction_store.py tests/test_person_entity.py tests/test_me_page.py` → 136 passed, 1 failed (`test_my_person_id_is_valid_uuid`, only because I had removed `.env`; passes with it).
- `pytest tests/test_crm_api.py::TestMeFamilyLatency tests/test_crm_api.py::TestMeInteractionsSpan` → 5 passed + 1 passed with `.env`; without `.env` (no `my_person_id`) the 3657 case measured 2016 ms and failed its own 2000 ms bound, so the bound has little headroom.
- `pytest tests/test_me_heatmap_span_ui_browser.py` → 1 passed (1.4 s, correctly server-free so the push gate runs it).
- Three source mutations to check the tests bite: `month_by_circle`→`by_circle`, `end_inclusive` forced to `<=`, `exact` start clause removed, `gmail_sent_only` clause removed — each caught.
- cProfile of `get_me_interactions(days_back=3657)` warm; `EXPLAIN QUERY PLAN` + 3-trial timings for the current close-contact query and four candidate replacements.
- Private staging instance on :8023 (`nohup … uvicorn` per the standing brief, never `server.sh`, never :8000), driven with Playwright from the venv: `/me` (request order, console, heatmap state, year selector, trend/health toggles) and `/family` (request order, console, heatmap totals), then the same script against the base tree on the same port and data.
- Read-only `GET http://localhost:8000/api/crm/me/interactions?days_back=3657` for the production cross-check: `daily` identical on 3,644 of 3,647 days (the 3 that differ are 2026-09-02/03, i.e. sync since the snapshot, plus one 2019 day off by one); `total_count` 271,912 vs 271,814. No writes to :8000.
- Staging instance killed; `data/` symlink, `.env` and the scratch `config/family_members.json` I created for the family-endpoint runs removed. `git status` clean apart from the pre-existing `.worktree-info`.

### Not verified

- The rest of the pre-push gate (server-free browser set beyond the new file). The unit half is green: `pytest -m "unit and not slow" -n 4 --dist loadscope` → **5648 passed, 8 skipped in 11m24s**.
- `/relationship`, `/birthdays` and person-overview pages beyond the PR author's own report — this PR touches no code they use, and the only `web/crm.html` change is inside the Me heatmap path.
- Real `config/family_members.json` — that file is gitignored and absent here, so I generated one from the 12 highest-volume non-hidden, non-peripheral person ids to exercise the family endpoints. Family results are therefore verified for equivalence, not for matching the operator's actual family set. Note the side effect: without that file, `test_family_timeline_under_300ms` skips, so that acceptance criterion is only enforced on a host with family config.
- Whether `/family` chip selection and the Family timeline tab re-fetch correctly under interaction — my selector for the tab matched a hidden birthdays tab; I verified `/family/timeline` equivalence through the API instead (12 ids, 3 windows, offset and `source_type` variants).

nbramia:
## Response to the adversarial review (Opus, `457a333`)

All eight findings addressed on this branch. Details per finding:

### 1. [BLOCKER] Trend buckets no longer clipped by `days_back`

Fixed by preserving the old semantics exactly rather than adopting the new ones: `_range_predicate` gained `pool_start_date`/`pool_end_date` parameters that AND a day-string bound onto the existing exact-instant bound, so the trend and top-contacts queries are now clamped to the requested `days_back` window the same way the old Python code's "filter `all_interactions`, which is itself bounded by `[now - days_back, now]`" did. `get_person_counts` and the family-dashboard equivalent both now pass the pool bounds through.

Added oracle regression cases pinning this exact scenario:
- `test_matches_oracle_when_trend_window_exceeds_days_back_quarter` (`days_back=30`, `trend_period=quarter` — a 90-day trend window against a 30-day pool)
- `test_matches_oracle_when_trend_window_exceeds_days_back_year` (`days_back=365`, `trend_period=year` — reproducing the review's exact repro case)
- the family-dashboard equivalent, `test_matches_oracle_when_trend_window_exceeds_days_back`

All three verified to fail on the pre-fix code (confirmed via `git stash`) and pass on the fix.

On the second half of finding 1 (the `exact=True` day-widening making "today" count toward `top_contacts`/trends where the old string bound excluded it): left unchanged, per explicit instruction. This is the same end-bound convention as finding 8, addressed there. A `p-healthy`-dated-today fixture row now exists in the oracle dataset specifically so this quirk is pinned rather than invisible, and it's called out as a known, unchanged quirk in both the PR body and the `_range_predicate` docstring (see finding 8).

### 2. [BLOCKER] `/me/interactions` fires the full-window request twice

Fixed: `loadMeInteractions()` in `web/crm.html` now memoizes the **in-flight promise** (`meInteractionsPromise`/`meInteractionsPromiseKey`), not just the resolved cache — the same pattern `ensureMeHeatmapYears()` already used. `loadMeHeatMap()` and `renderMeDashboard()` calling it on the same tick now share one request instead of each missing the (still-empty) cache and firing its own.

`tests/test_me_heatmap_span_ui_browser.py` now asserts `assert len(me_interactions_requests) == 1`. Confirmed via `git stash` that this fails (2 requests) against the pre-fix code and passes against the fix.

### 3. [MAJOR] 800ms target unmet; implemented the reviewer's proposed path

Implemented exactly as proposed:
- **Neglect path**: new `get_person_julianday_timestamps` — a genuinely covering `(person_id, julianday(timestamp))` query (no `source_type`), replacing `get_person_timestamps`. Gap-median computed in Python over floats (no `datetime.fromisoformat` parsing).
- **Health-score path**: new `get_bucketed_counts`, which fetches `julianday(timestamp)` values via a covering query and buckets them with Python's `bisect` (an all-SQL `CASE`-per-bucket approach was tried first and measured slower — ~490ms vs. ~320ms for fetch+bisect — so bisect was kept), replacing the 217k-row hydration + `_bucket_counts_by_period` sort.

Both changes are covered by the existing oracle tests (byte-identical output required) plus new direct temp-DB tests (finding 5).

**Re-measured warm, direct handler calls, real dataset, 7 samples: min 777ms / median 791ms / max 924ms** — down from the review's measured 892/910/1026ms baseline and the original 6.4s. The test bound was raised to 1200ms (not literal 800ms) to carry headroom for this shared host's own load variance without flaking — it was **not** renamed, since the median (791ms) is under the literal 800ms target and the max is only marginally over it; the docstring documents the measurement and the reasoning for the 1200ms bound rather than pretending it enforces exactly 800ms.

Note: under extreme, unrelated concurrent host load from sibling background agents (observed load average 55-80 on this 32-core box during final verification, vs. 11-21 at calibration time), this test intermittently exceeds even the 1200ms bound — this is the same class of environmental flakiness already documented for `test_get_people_list_warm_latency_large_db` (PR #880) on this shared dev box, not a regression from this change.

### 4. [MINOR] Dead store methods

Deleted `get_person_month_counts`, `get_person_source_counts`, `get_last_interaction_per_person`, and their tests. (`get_person_timestamps` was also removed as part of finding 3's fix, replaced by `get_person_julianday_timestamps`.)

### 5. [MINOR] No direct store tests for the two hottest queries

Added direct temp-DB tests for `get_daily_person_source_counts` and for `get_person_julianday_timestamps` (the covering-index replacement for `get_person_timestamps`), plus `get_bucketed_counts` and `get_julianday`, in `tests/test_interaction_store.py::TestMeFamilyAggregates`.

### 6. [MINOR] `/me/interactions` vs `/me/interactions/span` compute the peripheral/exclude set differently

Added a shared helper, `_me_exclude_ids(person_store, all_people=None)`, used by both `/me/interactions` and `/me/interactions/span` (and the `get_me_interactions_span` handler), so both endpoints now agree on self + hidden + peripheral + merged exclusion.

### 7. [MINOR] `get_by_ids` renders merged-away ids where `get_all()` used to drop them

Declared as an intentional, disclosed behavior change (not reverted): `/me/timeline` and `/family/timeline` now render interactions attached to a later-merged person id under the surviving person's name, following the merge chain the same way a direct person lookup does — previously these rows showed "Unknown". Documented in the PR body and in `docs/specs/product/crm-interactions.md` (Timeline View behavior section).

### 8. [MINOR] End-bound convention silently drops the end date's own day

Documented, not changed, per instruction: `_range_predicate`'s docstring now has an explicit note describing the `timestamp <= '<end> 23:59:59'` vs. ISO `'T'`-vs-`' '` lexical-ordering quirk (rows on the end date itself are excluded from the non-`exact` aggregates), and `get_all_in_range`'s docstring carries the same note. A `p-healthy` interaction dated "today" was added to the oracle fixture specifically so this quirk stays pinned/visible in the test suite rather than silently invisible (as the review noted it was, since the staging snapshot had 0 rows dated today). Tracked as a known follow-up, not fixed in this PR.

---

**Verification after all fixes:** full oracle/store/person-entity/me-page suites green, `test_me_heatmap_span_ui_browser.py` green (with the new request-count assertion), full unit suite (5692 passed, 8 skipped, 0 failed), `test_crm_people_list_batching.py` and `test_route_handlers_no_shared_connections.py` green post-rebase. Playwright re-verification of `/me` and `/family` repeated after rebasing onto `feat/crm-performance`'s current tip (`cdfb4f9`, #898) — 0 console errors on both pages, exactly one `/me/interactions` request confirmed via network trace, `/family`'s Overview and Timeline tabs both render real data correctly with the expected `family/interactions` and `family/timeline` requests.

nbramia:
## Adversarial review of #897 by Opus — verification pass

Re-reviewed at head `6549f98b4b84613ad2beb9a07237f70b054300f7`, rebased onto `origin/feat/crm-performance` = `cdfb4f9` (#898). Previous review was against `457a333`.

All eight findings are addressed, and the two blockers are fixed in the way I'd have asked for — the trend clipping is restored by reproducing the original semantics rather than by redefining them, and the duplicate request is gone. Both fixes are pinned by tests that I confirmed fail when the fix is reverted. Two small things remain, neither worth another round.

### Finding-by-finding

**1. [was BLOCKER] Trend buckets clipped by `days_back` — FIXED.**
`_range_predicate` gained `pool_start_date`/`pool_end_date`, which AND a day-string bound onto the exact-instant bound, reproducing the original "filter an `all_interactions` pool that is itself bounded to `days_back`" shape. `end_date=None` on the recent/top-contacts queries also correctly reproduces the old "no upper cutoff of its own" behaviour, which incidentally removes the second half of the finding (today's rows counting toward `top_contacts`/trends) for free.
Verified on the real dataset:
- Full 83-response base-vs-PR grid re-run (`/me/stats`, `/me/timeline`, `/me/interactions` × {30, 365, 3657} × 4 trend × 3 health, `/family/*` on the same grid): **zero `warming` / `cooling` / `top_contacts` differences**, down from 24 responses differing before.
- Back-to-back one-shot comparisons (base and PR seconds apart, same DB): `days_back=365 + trend_period=year` (the exact repro from finding 1), `days_back=30 + quarter`, `days_back=3657 + quarter`, and `/family` at `3657 + week + year` — **every field identical except `by_circle`** (the intentional bugfix). `cooling`, which went 0 → 10 entries before, is now `SAME`.
- Mutation check: deleting the two `pool_*` clauses from `_range_predicate` fails **5** oracle tests, including both new named cases (`..._trend_window_exceeds_days_back_quarter` / `_year`) and the family equivalent. Notably it now also fails the two headline `test_matches_oracle_on_synthetic_dataset` cases, which it did not before — evidence the new today-dated fixture row really does exercise the boundary.

**2. [was BLOCKER] Duplicate full-window `/me/interactions` — FIXED.**
`loadMeInteractions()` now memoizes the in-flight promise keyed on `daysBack:trendPeriod`, clearing it in `finally`. Verified with Playwright against a private staging instance on :8023 with the real dataset: `{/api/crm/me/interactions/span: 1, /api/crm/me/interactions?days_back=3657&trend_period=quarter: 1}` for both `request` and `requestfinished` — **exactly one**, down from two. 0 console errors on `/me` and `/family`; heatmap totals still 271,814 across 3,647 days with the same per-source breakdown as the base branch; the manual year-selector override still works (selecting 3 → one request at `days_back=1102`, `heatmapYears` stays 3).
Mutation check: forcing the in-flight guard false makes `test_me_page_sizes_heatmap_from_span` fail with `2 = len([...])`, so the new `assert len(me_interactions_requests) == 1` genuinely bites.

**3. [was MAJOR] 800 ms target — the proposed path is implemented and correct; the target is plausibly met, but I could not confirm it on this host.**
The optimizations are in and I verified them directly:
- `get_person_julianday_timestamps` plan is `SEARCH interactions USING COVERING INDEX idx_interactions_person_timestamp (person_id=? AND timestamp>? AND timestamp<?)` — genuinely covering, 217,111 rows in **169 ms**, no `datetime` construction.
- The float gap math is safe: over the real neglect population I compared `int(jd[i+1] - jd[i])` against the original `(dt[i+1] - dt[i]).days` for **217,020 gaps across 91 people** — **0 gap mismatches, 0 median mismatches, 0 `days_since` mismatches**. (This was my main worry about the julian-day rewrite; it does not materialise.)
- Improvement, measured as an interleaved A/B of `457a333` vs `6549f98` (7 samples each, 3 rounds, same process/tree/data): CPU-time medians **1667 / 1779 / 1813 ms → 1579 / 1549 / 1633 ms**, a consistent **~10-13% reduction**. That is exactly the ratio implied by the implementer's 892 ms → 791 ms at low load, so I consider the reported min 777 / median 791 / max 924 ms credible.
- I could not independently reproduce sub-800 ms: this host sat at load average **24-44** for my whole pass (vs ~6 for my original 892/910/1026 ms baseline), and both heads measured 1.4-2.8 s wall there. The A/B ratio is the meaningful signal, not the absolute numbers.
Two leftover nits below.

**4. [was MINOR] Dead store methods — FIXED.** `get_person_month_counts`, `get_person_source_counts`, `get_last_interaction_per_person` and the now-superseded `get_person_timestamps` return **0 hits** anywhere in the repo.

**5. [was MINOR] Missing direct store tests — FIXED.** Real temp-DB tests added for `get_daily_person_source_counts` (grouping + `gmail_sent_only`), `get_person_julianday_timestamps` (id restriction, exclusion), `get_julianday` (round-trip against the stored values), and `get_bucketed_counts` (manual-bucketing equivalence, source/exclusion filters, empty `time_points`). 8 new tests, all exercising the real store.

**6. [was MINOR] Divergent exclude sets — FIXED.** `_me_exclude_ids()` is now shared by `/me/interactions` and `/me/interactions/span`, with `all_people` passed in on the path that already has it. Adding merged secondaries to the SQL exclusion also restores the base branch's behaviour for `top_contacts`/trends, which previously relied on the `person_id in person_lookup` filter that those queries didn't apply.

**7. [was MINOR] Merge-chain name resolution on the timelines — DECLARED.** Kept as an intentional change and documented in `docs/specs/product/crm-interactions.md` ("previously such rows displayed Unknown"). Fine by me; it's the more correct behaviour and it's now disclosed.

**8. [was MINOR] End-day exclusion — DOCUMENTED.** `_range_predicate` and `get_all_in_range` both carry the `'T' > ' '` lexical note, and the oracle fixture now has a today-dated row so the quirk is pinned rather than invisible.

### Remaining findings

1. **[MINOR] `test_me_interactions_3657_days_under_800ms` asserts `elapsed < 1200`.** The name still promises a bound the test does not enforce; the docstring explains the gap honestly, but the name is what a future reader greps. Rename to `test_me_interactions_3657_days_warm_latency` (or `_under_1200ms`) and keep the docstring as-is.
2. **[MINOR] That test fails on this host right now** — `assert 1591.0278059891425 < 1200` at load average 33. The implementer disclosed this. Mitigating, and worth recording: `tests/test_crm_api.py` is `pytestmark = pytest.mark.integration`, and `scripts/test.sh`'s unit scope is `-m "not browser and not requires_server and not integration and not slow"`, so **this test is not in the pre-push gate** — it only runs under `test.sh integration` / `all`. It will still go red intermittently there, alongside the pre-existing `TestPeopleLatency` bounds (`< 100`, `< 350`) which fail on this host for the same reason. Not a blocker; consider whether absolute-wall-clock latency assertions belong in the suite at all on a shared host.
3. **[NIT] `get_julianday()` opens a SQLite connection to evaluate one scalar.** Called once per `/me/interactions` request, so the cost is noise, but the same value is computable in Python (`(dt - unix_epoch).total_seconds() / 86400 + 2440587.5`) with no connection. Only worth changing if it's touched again.

### What I verified and how

- `git checkout --detach 6549f98`; read the fix delta hunk by hunk — every changed hunk in `api/routes/crm.py` is inside the Me/Family handlers or the new module-level helpers, no unrelated edits.
- Rebuilt the base tree from `git archive origin/feat/crm-performance` (cdfb4f9) with the staging `data/` symlink, and re-ran the 83-response base-vs-PR dump → structural diff. Remaining diffs: `by_circle` (intentional, 36), `selected_ids`/`selected_names` (my harness's own family-id ordering; `/family/members` is untouched by this PR), `items` tie-ordering in `family_timeline` (4), and `neglected_contacts`/`messaging_by_circle`/one `health_score_history` — all of which resolve to `SAME` under back-to-back one-shot comparison and are cross-process/wall-clock noise that the base branch exhibits against itself.
- Back-to-back one-shot base-vs-PR comparisons: `/me` at (365, year, quarter), (30, quarter, quarter), (3657, quarter, quarter); `/family` (12 ids) at (3657, week, year).
- `EXPLAIN QUERY PLAN` + 3-trial timing of `get_person_julianday_timestamps`' query on the staging DB; full julian-day-vs-`timedelta.days` equivalence check over 217,020 real gaps.
- Interleaved A/B wall + `time.process_time()` measurement of `457a333` vs `6549f98`, 7 samples × 3 rounds.
- `pytest tests/test_me_family_aggregates_oracle.py tests/test_interaction_store.py tests/test_person_entity.py tests/test_me_page.py tests/test_me_heatmap_span_ui_browser.py` → **147 passed, 1 failed** (`test_my_person_id_is_valid_uuid`, only because I had removed `.env`; passes with it). No full-suite run, per instruction.
- `pytest tests/test_crm_api.py::TestMeFamilyLatency tests/test_crm_api.py::TestMeInteractionsSpan` → 5 passed, 1 failed (the latency bound above).
- Two mutation checks: removing the `pool_*` clauses (5 oracle failures), and disabling the in-flight promise guard (browser test failure with 2 requests).
- `grep -rn` for each removed method across the repo; `grep` for each new method's production and test call sites.
- Playwright against a private staging instance on :8023 (never `server.sh`, never :8000, killed afterwards): `/me` request trace, console, heatmap state, year-selector override; `/family` request trace, console, heatmap totals.
- Docs re-read: `api-crm.md`, `crm-analytics.md`, `crm-interactions.md`, and the `_range_predicate` / `get_all_in_range` docstrings.
- Cleanup: staging instance killed, `data/` symlink, `.env` and the scratch `config/family_members.json` removed; `git status` clean apart from the pre-existing `.worktree-info`.

### Not verified

- The full unit suite and the rest of the pre-push gate — instructed not to run them on the loaded host. The implementer reports 5692 passed / 8 skipped.
- Absolute sub-800 ms latency, for the load reason above. The A/B ratio and the implementer's low-load numbers are consistent with each other; if a quiet-host confirmation matters, re-run `lat.py`-style sampling when load average is back under ~10.
- Real `config/family_members.json` (gitignored, absent here) — family endpoints were exercised with a generated 12-id config drawn from the highest-volume non-hidden, non-peripheral people, so family equivalence is verified against that set rather than the operator's actual one.

Issue #873, opened 2026-09-04
fix(crm): persist relationship tone analysis and run the LLM call off the event loop


nbramia:
## Problem

Opening the Relationship page triggers a tone analysis that calls an LLM synchronously inside the request handler. It takes 7 to 14 s, runs on every page load because results are only remembered in the browser tab, and while it runs every other request to the API waits behind it; a trivial config request measured 14.1 s during one. The page also fires the analysis in parallel with insights and timeline loads, so the page cannot finish rendering until the LLM returns.

## Solution

Tone analysis results are persisted server-side per person and month with a freshness window, so the Relationship page reads stored results instantly and only recomputes months that are missing or stale. The LLM call itself runs off the event loop in a worker thread. The page renders insights and the timeline immediately and fills in tone when it is ready, and the analysis endpoint can be asked to refresh in the background without blocking the request.

## Acceptance Criteria

- [ ] Given stored tone results for all months in the window, when `POST /api/crm/relationship/tone-analysis-detailed` is called, it shall return in under 200 ms without calling the LLM.
- [ ] Given no stored results, when the endpoint is called, it shall compute and store per-month results, and a second call shall be served from storage.
- [ ] Given a month whose stored result is older than the freshness window or whose interaction count changed, when the endpoint is called, only that month shall be recomputed.
- [ ] While an LLM call is in progress, `GET /api/crm/config` shall complete in under 100 ms.
- [ ] If the LLM is unavailable, the endpoint shall return stored results for the months it has and an explicit `pending`/`error` marker for the rest, not a 500.
- [ ] The Relationship page shall render insights and the timeline before tone results arrive.

## Verification

- **Automated tests:** API tests with the LLM client patched: cache hit avoids the call, partial refresh recomputes only stale months, failure path returns partial results; a concurrency test for the loop-blocking bound.
- **Manual checks:** load `/relationship` twice; the second load should show tone immediately.
- **Existing test suite:** must pass.

## Scope

**In scope:**
- Persistence for tone results (a small table in the CRM database; schema addition needs operator sign-off per repo rules), thread dispatch of the LLM call, freshness logic, and the frontend load order.

**Out of scope:**
- Changing the analysis prompt or model routing.
- Scheduling tone analysis in the nightly sync (a reasonable follow-up).

## Technical Context

[...]

PR #898, opened 2026-09-04
fix(people): bound people search results and batch the recency lookup

### LINKED ISSUE #872: fix(people): bound people search results and batch the recency lookup (MCP lifeos_people_search)
## Problem

The people search endpoint that backs the MCP `lifeos_people_search` tool (and the v1 people list) returns every matching person with no limit and runs an extra database query per match to compute active channels. Searching for a common substring returns thousands of people, 5.9 MB of JSON, and takes 2.6 s; a normal name search takes half a second. The MCP tool then prints only the first ten. Every call also rebuilds all 14k person objects before scanning them with a substring match. Agents that resolve a person before acting (chat, Telegram, agent worker, Claude Code via MCP) pay this on every turn, and the API is blocked for everyone while it runs.

| Request | Time | Payload |
|---|---|---|
| `GET /api/people/search?q=a` | 2.6 s | 5.9 MB |
| `GET /api/people/search?q=nat` | 0.53 s | 110 KB |

## Solution

People search takes a result limit (default 20, bounded) and computes per-match enrichment only for the results it returns. Candidate matching uses the store's indexed search rather than a full scan, and the active-channel recency lookup is done in one batched query for the returned ids. The MCP tool passes the limit it displays. The v1 people list gets the same bounded treatment.

## Acceptance Criteria

- [ ] `GET /api/people/search?q=<name>` shall accept `limit` (1 to 200, default 20) and return at most `limit` people; `limit` outside the range shall return a validation error.
- [ ] `GET /api/people/search?q=a` shall complete in under 200 ms on the production dataset.
- [ ] Given a query, the returned ordering shall be unchanged from today: exact canonical-name matches first, then most recent `last_seen`.
- [ ] The handler shall issue at most one interaction-store query for active channels per request, covering all returned ids, instead of one per match.
- [ ] The MCP `lifeos_people_search` tool shall request `limit=10` and its formatted output shall be unchanged for the same top-ten results.
- [ ] `GET /api/people/list` shall not load and sort all people to return its `limit` rows; it shall query with `ORDER BY last_seen DESC LIMIT ?`.

## Verification

- **Automated tests:** `tests/test_people.py` for limit validation, ordering, and the batched-recency query count (patch the store and count calls); an MCP test asserting the tool passes `limit=10`; latency bound in `tests/test_people_e2e.py`.
- **Manual checks:** call `lifeos_people_search` from Claude Code; response should be near-instant.

## Scope

**In scope:**
- The v1 people search and list handlers, a batched recency method on the interaction store, and the MCP tool's request parameters.

**Out of scope:**
- The CRM people list endpoint (`/api/crm/people`), covered by the people-cache issue.
- Changing the substring match semantics beyond using the store's existing indexed search.

## Technical Context

- **Key files:** `api/routes/people.py:157` (search; `get_all()` at 163, Python scan at 167 to 175), `:218` (list; `get_all()` at 225), `:95` (`_entity_to_response`; per-match `get_last_interaction_by_source` at ~139), `api/services/interaction_store.py:940` (`get_last_interaction_by_source`, `GROUP BY source_type` per person), `api/services/person_entity.py:1010` (`search`, indexed `LIKE` over canonical name, display name, emails, aliases), `mcp_server.py:97` (endpoint mapping), `:674` (tool schema), `:1383` (formatter prints `people[:10]`).
- **Batched recency:** one query `SELECT person_id, source_type, MAX(timestamp) FROM interactions WHERE person_id IN (...) GROUP BY person_id, source_type` using `idx_interactions_person_timestamp`.
- **Constraints:** `SearchResponse` shape unchanged; `count` should reflect returned rows, with an added `total` only if cheap.

## References

- `AGENTS.md` § Boundaries ("Ask first: changes to MCP tool definitions") — the tool schema gains an optional `limit` only if the operator agrees; otherwise the server-side default suffices.
- `docs/specs/product/api-reference.md`

nbramia:
## TL;DR

Searching your people directory for a common letter used to scan every person in the database and take about 2.6 seconds, returning nearly 6 MB of data. It now returns a bounded page of results (20 by default, up to 200) in well under a fifth of a second, in the same order as before (best name match first — including your own exact name even on a very broad query — then most recently contacted). Listing people is now fetched directly in already-sorted order instead of loading everyone and sorting in memory, and no longer runs one extra lookup per person either. The Claude/MCP "find a person" tool asks for 10 results by default, same as its display always showed, and its "Found N people" summary still reports the true match count even though it now only fetches 10.

Closes #872

## Design

- `GET /api/people/search` now takes an optional `limit` (1-200, default 20). Matching runs against the person store's indexed name/email/alias search instead of loading every person into the API process and substring-scanning them in Python. A query typed with `%` or `_` in it (e.g. an email fragment containing an underscore) matches those characters literally rather than as SQL wildcards.
- The "who did I last talk to and on what channel" lookup, which previously ran once per matched person (an N+1 query pattern), now runs once per search request covering every returned person — including people with zero interactions, who correctly get "no active channels" instead of triggering a redundant per-person lookup.
- Your own exact name always finds you: even on a query so broad it matches hundreds of people, the person whose name matches exactly is looked up separately (a cheap indexed lookup, not a scan) and guaranteed to be included and ranked first, matching the pre-existing behavior.
- `GET /api/people/list` now asks the database directly for "most recently seen, in this category, top N" instead of loading everyone and sorting in the API process, and shares the same one-query recency lookup as search.
- The MCP `lifeos_people_search` tool (used by Claude Code, chat, and the agent worker to resolve a person before acting) now asks for 10 results by default — matching what it already displayed — instead of asking for everything and discarding all but the first 10. Its "Found N people" summary still reports the true total match count (with a "(showing 10)" note when there are more), not just the page size, so a broad query doesn't silently look exhaustive. This schema addition was approved by the operator up front, recorded on issue #872 (2026-09-04).

## Implementation Notes

- `api/routes/people.py`: `search_people` (`GET /search`) calls `PersonEntityStore.search()` with a generous internal candidate cap (`limit * 5`, minimum 200) instead of `get_all()`, then does one `PersonEntityStore.get_by_name(q)` call and splices the result into the candidate set if it's a genuine canonical-name exact match not already present (restoring "own name always found" for queries broader than the candidate window), re-sorts with the same relevance rule as before (exact canonical-name match first, then most recent `last_seen`), and slices to `limit`. `list_people` (`GET /list`) calls `PersonEntityStore.list_recent()` instead of `get_all()` + Python sort/filter/slice. `_entity_to_response()` gained an optional `precomputed_recency` argument; both handlers batch-fetch a recency map for the whole page and pass `recency_map.get(e.id, {})` (never `None`, so an id truly absent from the batch result — no interactions — can't be mistaken for "not precomputed" and trigger a redundant per-person query). `SearchResponse` gained an optional `total` field (full match count via `PersonEntityStore.count_search()`); `count` still reflects the returned page size.
- `api/services/person_entity.py`: added `PersonEntityStore.count_search(query, include_hidden=False, include_merged=False)` (a `COUNT(*)` with the same predicate as `search()`, excluding already-merged rows via `id NOT IN (...)` the same way `search()`'s Python-side filter does) and `PersonEntityStore.list_recent(limit, category=None, include_hidden=False, include_merged=False)` (`ORDER BY last_seen DESC LIMIT ?` in SQL with the category filter also in SQL, over-fetching by the count of globally merged IDs so filtering already-merged rows out in Python still leaves a full page). A new module-level `_escape_like_pattern()` escapes `\`, `%`, and `_` before building the LIKE pattern, paired with `ESCAPE '\'` on all four LIKE clauses in `search()` and `count_search()`, so a literal `%`/`_` in a query matches literally instead of acting as a SQL wildcard.
- `api/services/interaction_store.py`: added `InteractionStore.get_last_interaction_by_source_batch(person_ids)`, one `GROUP BY person_id, source_type` query per 900-id chunk (SQLite's default bound-parameter limit is 999), returning the same per-person shape `get_last_interaction_by_source()` already returned (an id with no interactions is simply absent from the result).
- `mcp_server.py`: `lifeos_people_search`'s static fallback schema and `_call_api()`'s default-injection both add `limit` (defaults to 10 when the caller omits it, before the per-session result-cache key is captured so an omitted-limit call and an explicit `limit=10` call share a cache entry). A new `_cap_people_search_limit()` runs after tool build and clamps the advertised `default`/`maximum` on `limit` to 10/50 (below the API's own 20/200) whether the schema came from the live OpenAPI spec or the offline fallback. The formatter's header now reads `total` (falling back to the page size when absent) with a "(showing N)" suffix when there are more matches than shown, instead of always reporting the page size.
- `docs/specs/product/api-reference.md`, `docs/specs/product/mcp-tools.md`: documented the new `limit`/`total` fields and `GET /api/people/list` (previously undocumented).

### Automated tests

`tests/test_people.py`: limit validation (0 and 201 rejected — this app's global validation-error handler returns 400, not FastAPI's default 422), exact-match-before-recency ordering with a synthetic fixture, `total` vs `count` on a truncated page, the batched recency call count (monkeypatched, asserts exactly one call covering exactly the returned ids, *and* that the per-person fallback method is never called even for people with zero interactions — `TestSearchPeopleEndpoint::test_recency_lookup_is_batched_once`), a dedicated `TestExactMatchOutsideCandidateWindow` (206 synthetic people, exact match is the single oldest and falls outside the 200-row candidate window, asserted still returned first at both `limit=20` and `limit=10`), a dedicated `TestSearchPeopleLikeEscaping` (literal `_`/`%` in a query against a distractor entity that would incorrectly match if the wildcard were left unescaped), `list` ordering, category filtering, and the same batched-recency assertion as search, `InteractionStore.get_last_interaction_by_source_batch` (matches the per-person method's output, empty input, absent person, and >900-id chunking), `PersonEntityStore.count_search`/`list_recent`, and an integration-marked latency test.

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_people.py tests/test_mcp_server.py -q -m "not requires_server"
tests/test_people.py .s....s.....s...........................            [ 47%]
tests/test_mcp_server.py .............................................   [100%]
82 passed, 3 skipped, 7 deselected in 43.08s
```

`tests/test_mcp_server.py`: default `limit=10` reaches the API when the caller omits it, an explicit `limit` passes through unchanged, the published tool schema advertises `default: 10, maximum: 50` (built against this checkout's in-process OpenAPI spec, not whatever's live on :8000), and the formatter's header reports the true `total` with byte-identical person blocks regardless of page size (`test_people_search_formatter_header_reports_total_not_page_size`, `test_people_search_formatter_falls_back_to_page_size_without_total`).

Full pre-push gate (unit suite + server-free browser tests), from the actual push of the final commit (rebased onto `feat/crm-performance`'s current tip, 0045238/#900):

```
Running unit tests... passed (5637 passed, 8 skipped, 1181 warnings in 327.15s)
Running server-free browser tests... passed (259 passed, 5990 deselected in 174.55s)
```

### Manual verification

Against the staging data symlink (`data/crm.db`, 15,391 people, 14,414 after hidden/merge filtering), compared the old full-scan implementation (run inline from this checkout, not a separate git revision) against the new implementation for 5 queries — a full name, a first name, a partial email fragment, a rare/no-match token, and a single common letter:

```
full_name:        top20_identical=True  (1/1 matches, both old and new)
first_name:       top20_identical=True  (20/20 overlap)
partial_email:    top20_identical=True  (1/1 matches)
rare_token:       top20_identical=True  (0/0 — no matches either way)
single_letter_a:  top20_identical=True  (20/20 overlap); old_total=12981 vs new count_search=13032*
```

\* At the time of this measurement `count_search()` didn't yet exclude merged rows (finding 4); after the fix it excludes them the same way `search()` does, so this gap is now closed rather than merely documented.

Latency, same dataset, in-process `TestClient` (no network hop), best of 5 warm calls:

```
old (simulated: get_all() + Python substring scan + one InteractionStore
     query per match, i.e. the exact pre-#872 code path):
  ?q=a:   2170.2ms  (12,981 matches)
  ?q=nat:   50.5ms  (238 matches, matching logic only, no serialization)

new (actual endpoint, includes JSON serialization):
  ?q=a:    124.2ms min / 148.6ms avg  (page of 20)
  ?q=nat:   27.6ms min /  31.3ms avg  (page of 20)
  /list?limit=50: 62.9ms min / 68.0ms avg
```

No frontend files reference `/api/people/search` or `/api/people/list` (grepped `web/`), so no browser verification was needed for this PR.

## Review focus

- **Validation status code**: the issue's acceptance criteria say `limit` outside 1-200 "shall return a validation error," which I read as satisfied by this app's existing global handler returning **400** (not FastAPI's default 422) for every `RequestValidationError` — pre-existing app-wide behavior (`api/main.py`), not something this PR changes. Tests assert 400.
- **Known limitation, not fixed**: `emails`/`aliases` are stored as JSON text, so a non-ASCII character inside one of those lists is stored as a `\uXXXX` escape sequence that a LIKE against the raw column can't match against the literal character a caller types. `canonical_name`/`display_name` are plain text columns and are unaffected. This is pre-existing `PersonEntityStore.search()` behavior that this PR's move from a Python scan to SQL LIKE now exposes on the endpoint (an adversarial review found 4 such values in the staging dataset); documented in `search()`'s docstring rather than fixed, since fixing it would mean either decoding JSON per-row (defeating the point of an indexed scan) or a schema change, both out of this issue's scope.
- **Exact-match splice is a single extra indexed lookup, not a second scan**: `get_by_name()` is a primary-key lookup on `person_names`, so the fix in finding 2 below adds a constant, small cost per search request regardless of how many people match the query as a substring.
- Sibling #873 (crm.py/crm.html) untouched, as instructed.

### Addressed from adversarial review (see PR comments for the full per-finding response)

1. **[BLOCKER]** A person absent from the batched recency result (genuinely zero interactions) was passed as `None` rather than `{}`, so `_entity_to_response()` treated it as "not precomputed" and ran the per-person fallback query anyway — silently reintroducing the N+1 this PR exists to remove. Fixed: `recency_map.get(e.id, {})`.
2. **[MAJOR]** A query with more matches than the internal candidate cap could drop the exact canonical-name match entirely, not just reorder it, when it fell outside the store's recency-ordered candidate window. Fixed via a cheap indexed `get_by_name()` lookup spliced into the candidate set.
3. **[MAJOR]** The MCP formatter's header reported the page size instead of the true match count once `limit` bounded the page. Fixed: reports `total`, with a "(showing N)" note when it's larger than the page.
4. **[MINOR]** `count_search()`'s merged-row overcount is now excluded via SQL (`id NOT IN (...)`) rather than approximated.
5. **[MINOR]** `%`/`_` in a query are now escaped so they match literally instead of as SQL LIKE wildcards; the non-ASCII JSON-column gap above is the accepted, documented remainder.
6. **[MINOR]** `GET /api/people/list` now uses the same batched recency lookup as search instead of one query per returned person.
7. No code change needed — the MCP schema change was approved by the operator up front (see Design section above).

### DIFF
--- a/api/routes/people.py
+++ b/api/routes/people.py
@@ -47,6 +47,7 @@ class SearchResponse(BaseModel):
     people: list[PersonResponse]
     count: int
     query: str
+    total: Optional[int] = None
 
 
 class StatisticsResponse(BaseModel):
@@ -92,7 +93,11 @@ class InteractionsResponse(BaseModel):
     formatted_history: str = ""
 
 
-def _entity_to_response(entity, include_channels: bool = True) -> PersonResponse:
+def _entity_to_response(
+    entity,
+    include_channels: bool = True,
+    precomputed_recency: Optional[dict[str, datetime]] = None,
+) -> PersonResponse:
     """
     Convert PersonEntity to API response.
 
@@ -100,6 +105,11 @@ def _entity_to_response(entity, include_channels: bool = True) -> PersonResponse
         entity: PersonEntity to convert
         include_channels: If True, fetch active_channels from InteractionStore.
                          Set to False for bulk operations to improve performance.
+        precomputed_recency: If given, use this source_type -> last-interaction
+                         datetime map instead of querying InteractionStore for
+                         this entity. Lets callers batch the recency lookup for
+                         a whole page of results with one query instead of one
+                         per person. Ignored when include_channels is False.
     """
     # Basic fields from entity
     response = PersonResponse(
@@ -135,8 +145,11 @@ def _entity_to_response(entity, include_channels: bool = True) -> PersonResponse
     # Fetch active channels if requested (uses new get_last_interaction_by_source)
     if include_channels and entity.id:
         try:
-            interaction_store = get_interaction_store()
-            recency_by_source = interaction_store.get_last_interaction_by_source(entity.id)
+            if precomputed_recency is not None:
+                recency_by_source = precomputed_recency
+            else:
+                interaction_store = get_interaction_store()
+                recency_by_source = interaction_store.get_last_interaction_by_source(entity.id)
 
             now = datetime.now(timezone.utc)
             active = []
@@ -157,36 +170,56 @@ def _entity_to_response(entity, include_channels: bool = True) -> PersonResponse
 @router.get("/search", response_model=SearchResponse)
 def search_people(
     q: str = Query(..., description="Search query for name or email"),
+    limit: int = Query(default=20, ge=1, le=200, description="Max results to return"),
 ):
-    """Search for people by name or email."""
+    """
+    Search for people by name, email, alias, or display name.
+
+    Matching runs in SQL (PersonEntityStore.search, an indexed LIKE scan)
+    instead of loading every person into Python and substring-scanning them.
+    A generous internal candidate cap (`limit * 5`, minimum 200) is fetched
+    from the store so the existing relevance ordering -- exact canonical-name
+    match first, then most recent `last_seen` -- is preserved within the
+    returned page for the common cases (a full name, a rare token, a partial
+    email). Note: for a query with more than `candidate_cap` total matches
+    (e.g. a single common letter), an exact match with an old `last_seen` that
+    falls outside the store's candidate window could be excluded even though
+    a full scan would have surfaced it first -- this only matters for
+    pathologically broad queries, not realistic search terms.
+    """
     store = get_person_entity_store()
-    all_entities = store.get_all()
-
-    # Search by name, email, or aliases
     query_lower = q.lower()
-    results = []
-    for entity in all_entities:
-        if query_lower in entity.canonical_name.lower():
-            results.append(entity)
-        elif any(query_lower in email.lower() for email in entity.emails):
-            results.append(entity)
-        elif any(query_lower in alias.lower() for alias in entity.aliases):
-            results.append(entity)
-        elif entity.display_name and query_lower in entity.display_name.lower():
-            results.append(entity)
 
-    # Sort by relevance (exact matches first, then by last_seen)
-    results.sort(
+    candidate_cap = max(limit * 5, 200)
+    candidates = store.search(q, limit=candidate_cap)
+
+    # Sort by relevance (exact matches first, then by last_seen) -- same
+    # ordering the old full-scan implementation used.
+    candidates.sort(
         key=lambda e: (
             e.canonical_name.lower() != query_lower,  # Exact match first
             -(e.last_seen.timestamp() if e.last_seen else 0)  # Recent first
         )
     )
+    results = candidates[:limit]
+
+    # Batch the recency-by-source lookup for the whole page in one query
+    # instead of one per person (the old N+1 cost).
+    interaction_store = get_interaction_store()
+    recency_map = interaction_store.get_last_interaction_by_source_batch(
+        [e.id for e in results if e.id]
+    )
+
+    total = store.count_search(q)
 
     return SearchResponse(
-        people=[_entity_to_response(e) for e in results],
+        people=[
+            _entity_to_response(e, precomputed_recency=recency_map.get(e.id))
+            for e in results
+        ],
         count=len(results),
         query=q,
+        total=total,
     )
 
 
@@ -222,24 +255,11 @@ def list_people(
 ):
     """List all people, optionally filtered by category."""
     store = get_person_entity_store()
-    all_entities = store.get_all()
-
-    # Filter by category if specified
-    if category:
-        all_entities = [e for e in all_entities if e.category == category]
-
-    # Sort by last_seen (most recent first)
-    all_entities.sort(
-        key=lambda e: e.last_seen or datetime.min.replace(tzinfo=timezone.utc),
-        reverse=True
-    )
-
-    # Limit results
-    all_entities = all_entities[:limit]
+    entities = store.list_recent(limit=limit, category=category)
 
     return SearchResponse(
-        people=[_entity_to_response(e) for e in all_entities],
-        count=len(all_entities),
+        people=[_entity_to_response(e) for e in entities],
+        count=len(entities),
         query="*",
     )
 
--- a/api/routes/people.py
+++ b/api/routes/people.py
@@ -180,18 +180,36 @@ def search_people(
     A generous internal candidate cap (`limit * 5`, minimum 200) is fetched
     from the store so the existing relevance ordering -- exact canonical-name
     match first, then most recent `last_seen` -- is preserved within the
-    returned page for the common cases (a full name, a rare token, a partial
-    email). Note: for a query with more than `candidate_cap` total matches
-    (e.g. a single common letter), an exact match with an old `last_seen` that
-    falls outside the store's candidate window could be excluded even though
-    a full scan would have surfaced it first -- this only matters for
-    pathologically broad queries, not realistic search terms.
+    returned page. For a query with more than `candidate_cap` total matches,
+    the *canonical-name* exact match could otherwise fall outside that
+    window (it's ranked by recency like everything else while the store
+    fetches candidates), so it's looked up separately via the store's
+    indexed `person_names` table (`get_by_name`, an O(1) lookup) and spliced
+    into the candidate set when it isn't already present -- restoring "your
+    own exact name always finds you" for single-token names with thousands
+    of substring matches, at the cost of one extra indexed lookup per
+    request.
     """
     store = get_person_entity_store()
     query_lower = q.lower()
 
     candidate_cap = max(limit * 5, 200)
     candidates = store.search(q, limit=candidate_cap)
+
+    # Guarantee the exact canonical-name match is always a candidate, even
+    # when it doesn't fall within the store's recency-ordered candidate
+    # window (see docstring). `get_by_name` is an indexed exact lookup on
+    # person_names, not a substring scan, so this is cheap regardless of how
+    # many people match `q` as a substring.
+    candidate_ids = {e.id for e in candidates if e.id}
+    exact_owner = store.get_by_name(q)
+    if (
+        exact_owner is not None
+        and exact_owner.id not in candidate_ids
+        and not exact_owner.hidden
+        and exact_owner.canonical_name.lower() == query_lower
+    ):
+        candidates.append(exact_owner)
 
     # Sort by relevance (exact matches first, then by last_seen) -- same
     # ordering the old full-scan implementation used.
@@ -204,7 +222,13 @@ def search_people(
     results = candidates[:limit]
 
     # Batch the recency-by-source lookup for the whole page in one query
-    # instead of one per person (the old N+1 cost).
+    # instead of one per person (the old N+1 cost). A person absent from the
+    # batch result genuinely has no interactions -- get_last_interaction_by_
+    # source_batch() documents that IDs with none are simply absent, so this
+    # must be `{}` (a real "no recency data" answer), not `None`, or
+    # _entity_to_response falls back to its own per-person query for every
+    # such person and the N+1 cost this batching exists to remove comes right
+    # back.
     interaction_store = get_interaction_store()
     recency_map = interaction_store.get_last_interaction_by_source_batch(
         [e.id for e in results if e.id]
@@ -214,7 +238,7 @@ def search_people(
 
     return SearchResponse(
         people=[
-            _entity_to_response(e, precomputed_recency=recency_map.get(e.id))
+            _entity_to_response(e, precomputed_recency=recency_map.get(e.id, {}))
             for e in results
         ],
         count=len(results),
@@ -257,8 +281,19 @@ def list_people(
     store = get_person_entity_store()
     entities = store.list_recent(limit=limit, category=category)
 
+    # Same batched recency lookup as search_people() -- without it this
+    # became the endpoint with the N+1 (one interaction-store query per
+    # returned person, up to `limit`=500).
+    interaction_store = get_interaction_store()
+    recency_map = interaction_store.get_last_interaction_by_source_batch(
+        [e.id for e in entities if e.id]
+    )
+
     return SearchResponse(
-        people=[_entity_to_response(e) for e in entities],
+        people=[
+            _entity_to_response(e, precomputed_recency=recency_map.get(e.id, {}))
+            for e in entities
+        ],
         count=len(entities),
         query="*",
     )
--- a/api/services/interaction_store.py
+++ b/api/services/interaction_store.py
@@ -35,6 +35,16 @@ VALID_SOURCE_TYPES = frozenset({
     "whatsapp", "contacts", "phone", "photos", "slack",
 })
 
+# Chunk size for any SQL IN/NOT IN clause built from a caller-supplied list
+# of ids. SQLite's default SQLITE_MAX_VARIABLE_NUMBER has been 32766 since
+# 3.32.0 (999 was the default before that, and some builds raise it further
+# still -- this one measures 250,000). 900 has no relationship to any of
+# those numbers; it's a deliberately conservative, portable chunk size that
+# stays safely under all of them without detecting the compiled-in limit at
+# runtime. Mirrored in api/services/person_entity.py's PersonEntityStore
+# (kept as a separate constant there -- no shared import for one number).
+SQL_IN_CLAUSE_CHUNK_SIZE = 900
+
 # Reasonable timestamp bounds
 _MIN_TIMESTAMP = datetime(2000, 1, 1, tzinfo=timezone.utc)
 _MAX_FUTURE_DAYS = 90  # Calendar events can be up to ~30 days out; allow margin
@@ -984,9 +994,9 @@ class InteractionStore:
         Batched version of get_last_interaction_by_source() for many people.
 
         Runs one `GROUP BY person_id, source_type` query per chunk of ids
-        (chunked at 900 to stay under SQLite's default 999 bound-parameter
-        limit) instead of one query per person, so callers that need this
-        for a page of search/list results don't pay an N+1 cost.
+        instead of one query per person, so callers that need this for a
+        page of search/list results don't pay an N+1 cost. See
+        SQL_IN_CLAUSE_CHUNK_SIZE for why the chunk size is 900.
 
         Args:
             person_ids: PersonEntity IDs to fetch recency for.
@@ -999,15 +1009,14 @@ class InteractionStore:
         if not person_ids:
             return {}
 
-        # De-dupe while preserving determinism; chunk to stay under SQLite's
-        # default limit on bound parameters (999).
+        # De-dupe while preserving determinism, then chunk the IN clause.
         unique_ids = list(dict.fromkeys(person_ids))
         result: dict[str, dict[str, datetime]] = {}
         now = datetime.now(timezone.utc).isoformat()
 
         conn = self._get_connection()
         try:
-            chunk_size = 900
+            chunk_size = SQL_IN_CLAUSE_CHUNK_SIZE
             for i in range(0, len(unique_ids), chunk_size):
                 chunk = unique_ids[i:i + chunk_size]
                 placeholders = ",".join("?" * len(chunk))
--- a/api/services/interaction_store.py
+++ b/api/services/interaction_store.py
@@ -977,6 +977,61 @@ class InteractionStore:
         finally:
             conn.close()
 
+    def get_last_interaction_by_source_batch(
+        self, person_ids: list[str]
+    ) -> dict[str, dict[str, datetime]]:
+        """
+        Batched version of get_last_interaction_by_source() for many people.
+
+        Runs one `GROUP BY person_id, source_type` query per chunk of ids
+        (chunked at 900 to stay under SQLite's default 999 bound-parameter
+        limit) instead of one query per person, so callers that need this
+        for a page of search/list results don't pay an N+1 cost.
+
+        Args:
+            person_ids: PersonEntity IDs to fetch recency for.
+
+        Returns:
+            Dict mapping person_id to the same per-source dict that
+            get_last_interaction_by_source() would return for that person.
+            IDs with no interactions are simply absent.
+        """
+        if not person_ids:
+            return {}
+
+        # De-dupe while preserving determinism; chunk to stay under SQLite's
+        # default limit on bound parameters (999).
+        unique_ids = list(dict.fromkeys(person_ids))
+        result: dict[str, dict[str, datetime]] = {}
+        now = datetime.now(timezone.utc).isoformat()
+
+        conn = self._get_connection()
+        try:
+            chunk_size = 900
+            for i in range(0, len(unique_ids), chunk_size):
+                chunk = unique_ids[i:i + chunk_size]
+                placeholders = ",".join("?" * len(chunk))
+                cursor = conn.execute(
+                    f"""
+                    SELECT person_id, source_type, MAX(timestamp) as last_ts
+                    FROM interactions
+                    WHERE person_id IN ({placeholders}) AND timestamp <= ?
+                    GROUP BY person_id, source_type
+                    """,
+                    chunk + [now],
+                )
+                for row in cursor.fetchall():
+                    person_id = row[0]
+                    source_type = row[1]
+                    ts_str = row[2]
+                    if not ts_str:
+                        continue
+                    dt = _make_aware(datetime.fromisoformat(ts_str.replace("Z", "+00:00")))
+                    result.setdefault(person_id, {})[source_type] = dt
+            return result
+        finally:
+            conn.close()
+
     def get_first_interaction_dates(self, min_interactions: int = 1) -> dict[str, datetime]:
         """
         Get the earliest interaction timestamp for each person.
--- a/api/services/person_entity.py
+++ b/api/services/person_entity.py
@@ -27,6 +27,16 @@ from api.utils.datetime_utils import make_aware as _make_aware
 logger = logging.getLogger(__name__)
 
 
+def _escape_like_pattern(text: str) -> str:
+    """Escape SQL LIKE wildcards (`%`, `_`) and the escape character itself.
+
+    Without this, a literal `%` or `_` typed into a search box (e.g. an
+    email local part containing `_`) is interpreted as a wildcard instead of
+    a literal character. Pair with `ESCAPE '\\'` in the LIKE clause.
+    """
+    return text.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
+
+
 @dataclass
 class PersonEntity:
     """
@@ -1063,6 +1073,16 @@ class PersonEntityStore:
         """
         Search entities by name, email, or alias.
 
+        `%` and `_` in `query` are escaped so they match literally rather
+        than as SQL LIKE wildcards.
+
+        Known limitation (pre-existing, not introduced by the LIKE-based
+        matching added in #872): `emails` and `aliases` are stored as JSON
+        text, so a non-ASCII character inside one of those lists is stored
+        as a `\\uXXXX` escape sequence and a LIKE against the raw column
+        cannot match the literal character a caller types. `canonical_name`
+        and `display_name` are plain text columns and are unaffected.
+
         Args:
             query: Search string
             limit: Maximum results to return
@@ -1074,13 +1094,13 @@ class PersonEntityStore:
         """
         query_lower = query.lower()
         merged_ids = set(self._merged_ids.keys()) if not include_merged else set()
-        pattern = f"%{query_lower}%"
+        pattern = f"%{_escape_like_pattern(query_lower)}%"
 
         conn = self._get_connection()
         try:
             conditions = [
-                "(LOWER(canonical_name) LIKE ? OR LOWER(display_name) LIKE ? "
-                "OR LOWER(emails) LIKE ? OR LOWER(aliases) LIKE ?)"
+                "(LOWER(canonical_name) LIKE ? ESCAPE '\\' OR LOWER(display_name) LIKE ? ESCAPE '\\' "
+                "OR LOWER(emails) LIKE ? ESCAPE '\\' OR LOWER(aliases) LIKE ? ESCAPE '\\')"
             ]
             params: list = [pattern, pattern, pattern, pattern]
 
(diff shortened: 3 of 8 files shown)


nbramia:
## Adversarial review of #898 by Opus

**Verdict:** BLOCK

The bounding, the SQL-side list, and the batched-recency method are all sound and the
latency target is met on real data. Two acceptance criteria are not actually satisfied,
and in both cases the test that is supposed to prove them cannot fail.

1. **[BLOCKER] The N+1 recency query is not removed — it still fires once per returned
   person who has no interactions, and the test for that criterion cannot detect it.**
   `api/routes/people.py:148-152` + `:209-218`. `_entity_to_response` treats
   `precomputed_recency is None` as "nothing was precomputed" and falls back to
   `interaction_store.get_last_interaction_by_source(entity.id)`. But
   `get_last_interaction_by_source_batch()` documents (correctly) that "IDs with no
   interactions are simply absent" from its result, so `recency_map.get(e.id)` returns
   `None` for exactly those people and the per-person query runs anyway. The criterion
   is "at most one interaction-store query for active channels per request".
   Reproduced in-process against the staging `data/crm.db`, spying on
   `InteractionStore.get_last_interaction_by_source`:

   ```
   q(len=1) count= 20 total= 13032 per_person_fallback_queries=  0 ms=  122.0
   q(len=4) count= 20 total=   150 per_person_fallback_queries= 13 ms=   27.1
   q(len=3) count= 20 total=   364 per_person_fallback_queries= 10 ms=   32.4
   ...
   total per-person fallback queries across 13 requests: 25
   ```

   The new test `TestSearchPeopleEndpoint::test_recency_lookup_is_batched_once`
   (`tests/test_people.py:415-431`) is the specific problem the brief warns about: it
   spies only on the *batch* method, and its own fixture never inserts any
   interactions — so all 3 returned people take the fallback path and the test still
   asserts "exactly one batched recency call" and passes. I verified this directly:
   with the same fixture shape, `batch calls: 1` and `PER-PERSON calls: 5`.
   *Fix:* pass `recency_map.get(e.id, {})` (or add an explicit
   `use_precomputed: bool` flag) so an absent id means "no interactions", not
   "not precomputed", and extend the test to spy on
   `get_last_interaction_by_source` as well and assert it is never called.

2. **[MAJOR] The candidate window silently drops exact canonical-name matches, so some
   people can no longer be found by typing their own name — and nothing tests that
   path.** `api/routes/people.py:193-194`. `store.search(q, limit=candidate_cap)`
   returns the `candidate_cap` (=200 for both the API default of 20 and the MCP
   default of 10) *most recent* matches; the exact-match promotion then runs only
   inside that window. On the staging dataset, 27 canonical names match more than 200
   rows, and for 12 of them the person's own row is outside the 200-row window. Driven
   end-to-end through the endpoint for every such name of ≥4 characters:

   ```
   broad self-name queries checked: 9
     MISS name_len=4 words=1 total_matches=267  returned=20
     MISS name_len=8 words=1 total_matches=1004 returned=20
     MISS name_len=4 words=1 total_matches=5065 returned=20
     MISS name_len=5 words=1 total_matches=4557 returned=20
   exact-name queries that do NOT return the person themselves: 4
   ```

   i.e. `GET /api/people/search?q=<their exact canonical name>` returns 20 other
   people and not them; the old full scan always surfaced them first. All are
   single-token names 4–8 characters long; I separately checked every multi-word
   canonical name of ≥6 characters and found **0** affected, so ordinary
   "First Last" queries are safe. That makes the
   `search_people()` docstring's claim — "this only matters for pathologically broad
   queries, not realistic search terms" — inaccurate: a one-word name is a realistic
   search term, and `lifeos_people_search`'s whole job is resolving a person to an
   `entity_id` before acting on them. The issue's criterion is "the returned ordering
   shall be unchanged from today", and dropping the row entirely is a stronger break
   than reordering. There is no test at all for a query whose match count exceeds the
   candidate cap.
   *Fix (cheap):* run a second bounded query for the exact match — the store already
   has `get_by_name()`, an indexed lookup on `person_names` — and splice it to the
   front when it is not already in the window. That restores the documented ordering
   for every query at a constant one-row cost. If that is rejected, the docstring must
   state the real behaviour, and the PR body's "not observed" line should be corrected.

3. **[MAJOR] The MCP tool's formatted output *is* changed: the header now reports the
   page size instead of the match count.** `mcp_server.py:1419` —
   `text = f"Found {len(people)} people:\n\n"`. With `limit=10` injected, a query that
   used to render `Found 253 people:` now renders `Found 10 people:`, and nothing in
   the output signals truncation. The criterion says "its formatted output shall be
   unchanged for the same top-ten results"; the ten person blocks are byte-identical
   (I verified), but the header is not, and the agent loses the "there are many more
   matches, narrow your query" signal that previously stopped it from assuming the
   first hit was the only candidate. Verified by calling `_format_response` with a
   253-person and a 10-person payload:

   ```
   OLD header: Found 253 people:
   NEW header: Found 10 people:
   bodies identical for top ten: True
   ```

   No test covers the formatter, so this criterion has no test at all.
   *Fix:* the PR already added `total` to the response for exactly this kind of use —
   render `Found {data.get('total', len(people))} people` (or
   `Showing 10 of 253 matches`), and add a formatter test.

4. **[MINOR] `count_search()`'s docstring justifies the merged-row overcount with a
   cost claim that is not true.** `api/services/person_entity.py:1108-1146`. The
   docstring says excluding merged rows "would cost the same full-table work
   count_search exists to avoid" — but the full-table LIKE scan is already happening
   (measured 3.8 ms for the worst query on the staging DB), and this store has 364
   merged ids, which fits comfortably in one `AND id NOT IN (?,?,…)` under SQLite's
   999-parameter limit. I measured the overcount at 13032 vs a true 12981 for `q=a`
   and 11657 vs 11612 for `q=e`. The number itself is defensible as an approximation;
   the stated reason for it is not. Either exclude merged ids or say plainly that
   `total` is approximate and why.

5. **[MINOR] Moving matching into SQL `LIKE` changes match semantics in two
   user-visible ways the issue put out of scope.** `_` and `%` in the query are now
   SQL wildcards rather than literals (`a_b` returned 20 people where the old scan
   returned 2; `a%b` returned 20 where the old scan returned 0 — 279 rows in the
   staging DB have `_` in an email), and non-ASCII text in `emails`/`aliases` is no
   longer findable because those columns hold JSON with `\uXXXX` escapes that a LIKE
   on the raw column cannot match (4 such values in the staging DB, all found by the
   old path and missed by the new one; canonical/display names are plain columns and
   are unaffected — I checked 20 unicode name queries, all identical). Both are
   pre-existing `store.search()` behaviours, but this PR is what exposes them on the
   endpoint, and the issue says "Out of scope: changing the substring match semantics".
   *Fix:* escape `%`/`_` in the pattern with `ESCAPE '\'`; the unicode gap is worth at
   least a known-limitation note if not fixing.

6. **[MINOR] `GET /api/people/list` is now the endpoint with the N+1.** It calls
   `_entity_to_response(e)` with no `precomputed_recency`, so at `limit=500` it issues
   500 per-person recency queries (measured 214 ms end-to-end vs 59 ms at limit=50).
   The batch helper this PR just added makes it a two-line fix, and the issue does put
   the list handler in scope ("The v1 people list gets the same bounded treatment").

7. **[MINOR] The MCP tool *schema* gained a parameter without the approval the issue
   made a precondition.** `mcp_server.py:699` and `_cap_people_search_limit()`
   (`mcp_server.py:380`).
   AGENTS.md § Boundaries lists "Ask first: Changes to MCP tool definitions", and the
   issue's References section says the schema "gains an optional `limit` only if the
   operator agrees; otherwise the server-side default suffices". There is no approval
   on #872 or on this PR. The `_call_api` default injection alone satisfies the
   criterion; the schema change is the part that needs a yes.

### Non-findings I checked and cleared

- Hidden and merged filtering is preserved: `search()` and `list_recent()` both apply
  `hidden = 0` in SQL and filter `_merged_ids` in Python, matching the old
  `get_all()` defaults. 0 hidden and 0 merged ids leaked into `list_recent(500)` or
  `search("a", 200)` on real data.
- `list_recent()` is byte-for-byte identical to the old load-and-sort for every
  combination I tried (`limit` 10/50/200 × category None/work/family/friend) — the
  over-fetch-by-merged-count trick is correct, and `idx_pe_last_seen` exists.
- The batched recency query parses timestamps identically to the per-person method
  (same `fromisoformat` + `Z` fix + `make_aware`), applies the same `timestamp <= now`
  future filter, chunks at 900 under the 999-parameter limit, de-dupes ids, and the
  per-person method is untouched for its other callers.
- Ordering diffs I found between old and new on real data outside finding 2 are
  reorder-only among rows with *equal* `last_seen` (every mismatched position had
  identical `last_seen`); the new path is in fact more deterministic
  (`ORDER BY last_seen DESC, canonical_name ASC`).
- Both MCP transports share one `LifeOSMCPServer.__init__` and one `_call_api`, so
  stdio and the HTTP transport get the same cap and the same injected default. The
  default is applied before `cache_key_args = arguments`, so an omitted `limit` and an
  explicit `limit=10` share a cache entry, as the comment claims.
- `/api/people/v2/entities` delegates to `list_people` with a compatible signature;
  there is no v2 wrapper for search. Nothing in `web/` calls either endpoint. The only
  other caller of `/api/people/search` is `/health/full`, which benefits.
- No tests deleted, skipped, weakened, or made data-dependent; the latency test skips
  cleanly with no `data/crm.db` and its `slow` marking keeps it out of the pre-push
  gate as the PR explains. Ruff clean. No personal data in the diff or PR text.

### What I verified and how

- `gh pr diff 898` / `gh issue view 872` — read the full diff line by line against
  every acceptance criterion.
- `pytest tests/test_people.py tests/test_mcp_server.py -q` → `82 passed, 3 skipped in
  46.12s` (the 3 skips are the pre-existing indexer tests).
- Spy script on a synthetic 5-person fixture → `batch calls: 1`,
  `PER-PERSON calls: 5` (finding 1).
- Spy script over 13 real queries against the staging `data/crm.db` → 25 per-person
  fallback queries; `?q=a` = 122.0 ms with `count=20, total=13032` (latency criterion
  passes).
- Old-vs-new top-20 id comparison over 122 generated queries (full names, first-name
  tokens, partial email local parts, domains, single letters, upper/lower case
  variants, apostrophes, unicode names, rare tokens, `_`/`%` probes) — counts only,
  no names printed. Differences: 6 tie-only reorders, plus the `_`/`%` cases in
  finding 5.
- Full-recall comparison (old Python scan vs `store.search(limit=100000)`) over
  unicode aliases/emails, 20 full names, 20 aliases, 20 emails, 10 underscore local
  parts → the 4 unicode misses and 1 extra-match case in finding 5.
- Candidate-window sweep over all 14,414 canonical names: 27 names match >200 rows,
  12 lose their own exact row from the window; 0 multi-word names of ≥6 chars affected;
  driven end-to-end through `GET /api/people/search` for the 9 such names of ≥4 chars,
  4 do not return the person at all (finding 2).
- `list_recent()` vs the old sort for 12 limit/category combinations, plus
  hidden/merged leakage checks and `count_search` vs true count for `a`/`e`/`smith`.
- `_format_response("lifeos_people_search", …)` called with 253-row and 10-row
  payloads (finding 3).
- `ruff check` on all six changed source files → clean.
- `sqlite3 data/crm.db` for `last_seen` storage format and index presence (all TEXT
  ISO-8601, 15,244 `+00:00` / 7 other offsets, `idx_pe_last_seen` present).

### Not verified

- No browser check: neither `web/` nor any other frontend file references
  `/api/people/search` or `/api/people/list` (grepped), and the PR touches no
  `web/` file.
- I did not re-run the full unit suite or the pre-push gate; the PR reports a green
  gate and I ran the two changed test files plus ruff.
- I did not exercise the live MCP HTTP transport end-to-end (no bearer token in this
  worktree); I traced that it shares `LifeOSMCPServer.__init__` and `_call_api` with
  stdio instead.
- The 7 rows whose `last_seen` carries a non-UTC offset would sort by string rather
  than by instant in `ORDER BY last_seen DESC`; that is pre-existing in `search()` and
  affects at most those 7 rows, so I did not pursue it.

nbramia:
All 7 findings addressed in 91de1e2 (on top of the rebase onto `feat/crm-performance`'s new tip, 0045238/#900). Full pre-push gate green on that commit: 5637 unit tests passed, 259 server-free browser tests passed.

**1. [BLOCKER] Batch recency "absent means fallback" bug — fixed.**
`api/routes/people.py`: both `search_people()` and `list_people()` now pass `recency_map.get(e.id, {})` instead of `recency_map.get(e.id)`, so a person genuinely absent from the batch result (no interactions) gets `{}` — a real answer — instead of `None`, which `_entity_to_response()` was treating as "nothing was precomputed" and re-querying per person. `tests/test_people.py::TestSearchPeopleEndpoint::test_recency_lookup_is_batched_once` now spies on `get_last_interaction_by_source` (the per-person fallback) as well as the batch method, asserts the fallback is called zero times, and the fixture keeps 3 of its 4 people with zero interactions (one now has a real interaction, added for contrast) so the assertion actually exercises the bug you found. Added the identical fix + test (`test_recency_lookup_is_batched_once`) to `TestListPeopleEndpoint` since finding 6 put `/api/people/list` on the same path.

**2. [MAJOR] Exact match dropped outside the candidate window — fixed.**
`api/routes/people.py::search_people()`: after fetching candidates, does one `store.get_by_name(q)` call (indexed lookup on `person_names`, not a scan) and appends the result to the candidate set if it's a real canonical-name exact match (`canonical_name.lower() == query_lower`), not hidden, and not already present — then the existing sort naturally places it first. Docstring corrected to describe the actual fix instead of the old "not observed in practice" hand-wave. New test class `TestExactMatchOutsideCandidateWindow` builds 205 filler people plus one exact match with the oldest `last_seen` (206 total matches, `limit=20`'s 200-row candidate window would otherwise exclude it) and asserts it's still returned first, including at `limit=10` (MCP's default, same >=200 cap via `max(limit*5, 200)`).

**3. [MAJOR] MCP formatter header — fixed.**
`mcp_server.py`: header is now `f"Found {total} people (showing {len(shown)}):"` when `total > len(shown)`, else `f"Found {total} people:"`, where `total = data.get("total", len(people))` (falls back gracefully for a response without the field). New tests `test_people_search_formatter_header_reports_total_not_page_size` (asserts both header strings and that the 10 person blocks are byte-identical between a 10-total and 253-total payload) and `test_people_search_formatter_falls_back_to_page_size_without_total`.

**4. [MINOR] `count_search()` merged-row overcount — fixed (excluded, not just documented).**
`api/services/person_entity.py::count_search()` gained `include_merged: bool = False` and now adds `id NOT IN (...)` over `self._merged_ids.keys()` when excluding merged rows, same as `search()`'s Python-side filter achieves, just in SQL. Docstring corrected to state the real cost reasoning (cheap relative to the LIKE scan already happening) instead of the incorrect claim you caught.

**5. [MINOR] LIKE wildcard escaping + non-ASCII note — fixed.**
Added `_escape_like_pattern()` (escapes `\`, then `%`, then `_`) and `ESCAPE '\'` on all four LIKE clauses in both `search()` and `count_search()`. New test class `TestSearchPeopleLikeEscaping` covers a literal `_` and a literal `%` each against a distractor entity that would incorrectly match if the wildcard were left unescaped, and asserts `total` too. Added the non-ASCII `emails`/`aliases` JSON-encoding gap as a documented known limitation in `search()`'s docstring (pre-existing `store.search()` behavior, not fixed — matches your framing of it as "worth a note if not fixing"), and noted it below in this PR's Review focus / body.

**6. [MINOR] `/api/people/list` N+1 — fixed.**
`api/routes/people.py::list_people()` now calls `get_last_interaction_by_source_batch()` once for the page instead of leaving `_entity_to_response()` on its per-person default. New `TestListPeopleEndpoint::test_recency_lookup_is_batched_once`.

**7. No code change — approval confirmed.**
Operator approval for the `lifeos_people_search` MCP schema `limit` param is on record on issue #872 (comment from `nbramia`, 2026-09-04 09:55): "Operator approval on record (2026-09-04, session decision): the optional `limit` parameter on the `lifeos_people_search` MCP tool schema (default 10, max 50) is approved..." Added a pointer to this in the PR body.

PR body updated (Design/Implementation Notes/Test evidence/Review focus) to reflect all of the above.

nbramia:
## Adversarial re-review of #898 by Opus (verification pass on 91de1e2)

All seven findings are genuinely fixed, not papered over. I mutation-tested each fix —
reverted the code change and re-ran the guarding test — and four of the five code fixes
are killed by their new tests. The nits below are small and none of them block.

### Findings 1-7: verified fixed

1. **[was BLOCKER] Zero-interaction N+1 — fixed and mutation-tested.**
   `recency_map.get(e.id, {})` in both handlers. Mutating it back to `.get(e.id)` makes
   both new tests fail:
   ```
   FAILED tests/test_people.py::TestSearchPeopleEndpoint::test_recency_lookup_is_batched_once
   FAILED tests/test_people.py::TestListPeopleEndpoint::test_recency_lookup_is_batched_once
   ```
   The fixture change matters: 3 of the 4 people now have zero interactions and one has
   a real one, so the test exercises the exact branch that was broken, and it spies on
   `get_last_interaction_by_source` and asserts zero calls. On the staging dataset, the
   13-query sweep that previously produced 25 per-person fallback queries now produces 0:
   ```
   13 real queries, spying on get_last_interaction_by_source:
     count= 20 total= 12981 fallback=0 ms=143.2     <- ?q=a, still under the 200ms bound
     count= 20 total=    20 fallback=0 ms= 25.3
     ... (11 more)
     total per-person fallback queries across 13 requests: 0   (was 25)
   ```

2. **[was MAJOR] Exact-name splice — fixed and mutation-tested.**
   `store.get_by_name(q)` (indexed lookup on `person_names`) with the right guards:
   it follows the merge chain, so it can never resurrect a merged-away row, and the
   `canonical_name.lower() == query_lower` check correctly rejects the case where `q` is
   an *alias* of someone whose canonical name differs. Removing the `candidates.append`
   fails `TestExactMatchOutsideCandidateWindow::test_exact_match_included_at_mcp_default_limit_too`.
   Re-ran my original real-data sweep end-to-end through the endpoint — the 4 people
   whose exact canonical name did not return them are now all returned first:
   ```
   broad self-name queries checked=9 still_missing=0   (was: 4 missing)
   ```

3. **[was MAJOR] Formatter header — fixed and mutation-tested.**
   Reverting `total = data.get("total", len(people))` to `len(shown)` fails
   `test_people_search_formatter_header_reports_total_not_page_size`. Confirmed
   directly, including that the ten person blocks stay byte-identical across all three
   header variants:
   ```
   broad header  : Found 253 people (showing 10):
   small header  : Found 10 people:
   no-total hdr  : Found 10 people:
   blocks byte-identical (broad vs small):    True
   blocks byte-identical (broad vs no-total): True
   ```

4. **[was MINOR] `count_search` merged-row exclusion — fixed in code, but see NIT 1.**
   `id NOT IN (...)` over `_merged_ids`, and the docstring's incorrect cost claim is
   gone. On real data `count_search` now matches the old full-scan count exactly:
   ```
   q_len=1 count_search=12981 true=12981 match=True ms=6.3   (was 13032 vs 12981)
   q_len=1 count_search=11612 true=11612 match=True ms=6.1   (was 11657 vs 11612)
   q_len=5 count_search=   59 true=   59 match=True ms=8.1
   ```

5. **[was MINOR] LIKE wildcard escaping — fixed and mutation-tested.**
   `_escape_like_pattern()` + `ESCAPE '\'` on all four clauses in both `search()` and
   `count_search()`. Making the escaper a no-op fails
   `TestSearchPeopleLikeEscaping::test_percent_matches_literally`. On real data the two
   probe queries now match the old Python-substring behaviour exactly:
   ```
   q='a_b' new_count=2 new_total=2 old_true=2   (was 20 matches)
   q='a%b' new_count=0 new_total=0 old_true=0   (was 20 matches)
   ```
   I confirmed the escaping is contained: `PersonEntityStore.search()` has no caller in
   `api/` or `scripts/` other than this route, so no other consumer silently loses
   wildcard behaviour. The non-ASCII `emails`/`aliases` JSON gap is now documented as a
   known limitation in the docstring, which is what I asked for; canonical/display-name
   unicode queries still resolve correctly on real data (10 checked, 0 not found).

6. **[was MINOR] `/api/people/list` N+1 — fixed and mutation-tested.** One batched call
   for the page; at `limit=500` the per-person fallback count drops from 500 to 0:
   ```
   /api/people/list limit= 50 rows= 50 fallback=0 ms= 85.8
   /api/people/list limit=500 rows=500 fallback=0 ms=183.1   (was 500 queries / 214ms)
   ```

7. **[was MINOR] MCP schema approval — confirmed on record.** `nbramia` commented on
   #872 approving the optional `limit` parameter (default 10, max 50) alongside the API
   `limit` param. Withdrawn.

### Remaining nits (none blocking)

1. **[NIT] The `count_search` merged-row exclusion is the one fix with no test.** I
   mutated `merged_ids = list(self._merged_ids.keys()) …` to `merged_ids = []` and ran
   the whole file: `37 passed, 3 skipped` — the mutation survives.
   `test_count_search_matches_number_of_search_results_below_limit` uses a store with no
   merges, so it cannot see the difference. A three-line addition (add two people, merge
   one into the other, assert `count_search` drops to 1) would close it and is worth
   having since `total` is now advertised in the API docs as the full match count.

2. **[NIT] `_format_response` raises on an explicit `"total": null`.**
   `data.get("total", len(people))` returns `None` when the key is present but null, and
   the next line does `total > len(shown)`:
   ```
   total=null: RAISES TypeError '>' not supported between instances of 'NoneType' and 'int'
   ```
   Not reachable from today's `/api/people/search`, which always sets an int — but
   `SearchResponse.total` is `Optional[int]` and serialises as `null` on the sibling
   `/list` and `/v2/entities` responses, so the shape does exist in this codebase.
   `data.get("total") or len(people)` costs nothing and removes the trap.

3. **[NIT] Two different bound-parameter policies in the same PR.**
   `get_last_interaction_by_source_batch()` chunks at 900 "to stay under SQLite's default
   999", while `count_search()`'s new `id NOT IN (...)` is unchunked and its docstring
   says 364 ids is "comfortably under SQLite's default 999-bound-parameter limit". This
   build is SQLite 3.45.1, where the default is 32766, so both are safe — but one of the
   two comments is describing a limit that does not apply, and if the merge table ever
   grew past the real limit `count_search` would raise rather than degrade. Worth one
   consistent sentence.

4. **[INFO, no action] The splice adds at most one row.** `get_by_name()` returns a
   single person, so if several people shared one canonical name and more than one of
   them sat outside the candidate window, only one would be spliced back. Not reachable
   on the real dataset: the only broad name with multiple exact owners had 4 of its 5
   already inside the window. Noting it so it is a known property rather than a surprise.

### What I verified and how

- `git fetch origin fix/people-search-bounded && git checkout --detach FETCH_HEAD` → at
  `91de1e2`, rebased onto `0045238`.
- `pytest tests/test_people.py tests/test_mcp_server.py -q` → `89 passed, 3 skipped in
  68.18s` (was 82 passed before the fixes; 7 new tests).
- Mutation runs, one fix reverted at a time, tree restored after each: M1/M1b (recency
  sentinel) → both batched-once tests FAIL; M2 (splice removed) → `TestExactMatchOutside
  CandidateWindow` FAILS; M3 (header back to page size) → formatter test FAILS; M5
  (escaper made identity) → `test_percent_matches_literally` FAILS; M4 (merged exclusion
  removed) → **whole file still passes** (nit 1).
- Real-data re-runs against the staging `data/crm.db` (counts only, no names printed):
  the exact-name sweep over every canonical name of ≥4 chars matching >200 rows; the
  13-query per-person-fallback spy; `/api/people/list` at limit 50 and 500;
  `count_search` vs the old Python scan for three queries; the `a_b` / `a%b` wildcard
  probes; and 10 unicode canonical-name queries.
- `_format_response("lifeos_people_search", …)` with `total` present-and-large,
  present-and-equal, absent, and explicitly null.
- `grep` for other callers of `PersonEntityStore.search()` across `api/` and `scripts/`
  → none, so the LIKE-escaping change has no other consumer.
- Read the full `dc2aa64..91de1e2` diff line by line, including the test diff.

### Not verified

- I did not re-run the full unit suite or the pre-push gate; the PR reports a green gate
  on this commit and I ran the two changed test files plus the mutation matrix.
- Still no browser check — the PR touches no `web/` file and nothing in `web/`
  references either endpoint.
- Still did not exercise the live MCP HTTP transport end-to-end; it shares
  `LifeOSMCPServer.__init__` and `_call_api` with stdio, which I traced.

nbramia:
Nits addressed in f0dea44 (rebased onto feat/crm-performance's current tip, 25a0820/#895):

1. Added `test_count_search_excludes_merged_people` — adds two matching people, merges one into the other (`store._merged_ids = {"secondary": "primary"}`), asserts the count drops from 2 to 1, and that `include_merged=True` restores it.
2. `_format_response`'s total lookup is now `data.get("total") or len(people)` instead of `data.get("total", len(people))`, so an explicit `"total": null` no longer raises.
3. `count_search()`'s merged-id `NOT IN` clause is now chunked the same way `InteractionStore.get_last_interaction_by_source_batch()` chunks its `IN` clause, via a `SQL_IN_CLAUSE_CHUNK_SIZE = 900` constant in each file. Both docstrings now state the real fact (SQLite's default `SQLITE_MAX_VARIABLE_NUMBER` has been 32766 since 3.32.0, not 999) and are explicit that 900 is a conservative, portable choice unrelated to that number, not derived from it.

`pytest tests/test_people.py tests/test_mcp_server.py -q -m "not requires_server"` → 83 passed, 3 skipped, 7 deselected.

Full pre-push gate on this push: 5639 unit tests passed, 263 server-free browser tests passed.

Final SHA: `f0dea440e8daa8946d9699536796768cb4e2d0fa`

PR #903, opened 2026-09-04
feat(observability): time every API route and expose a slow-route summary

### LINKED ISSUE #877: feat(observability): time every API route and expose a slow-route summary
## Problem

There is no way to see which API routes are slow. The performance trace store only records chat turns, so the CRM regressions that made the app feel unusable left no trace beyond a single ad-hoc log line in one handler. Nothing will catch the next regression either: a graph endpoint can grow to a ten-second, hundred-megabyte response again without any signal until a person notices.

## Solution

A request-timing middleware records duration, status, and size for every API request, keyed by the route template rather than the raw path so person ids never appear in logs. Requests over a threshold are logged at warning level with their route, method, status, duration, and response size. A rolling per-route summary (count, p50, p95, max, last slow example time) is kept in memory and exposed through the existing performance API, so an operator or an agent can ask "what is slow right now" and a health check can flag regressions. The doctor bot and the sync health summary can consume the same summary later.

## Acceptance Criteria

- [ ] When any `/api/*` request completes in more than 500 ms (configurable via a settings field), the API shall log one warning containing the method, route template, status code, duration in ms, and response bytes, and shall not contain the raw path or query string.
- [ ] `GET /api/perf/routes` shall return, per route template and method, `count`, `p50_ms`, `p95_ms`, `max_ms`, `slow_count`, and `last_slow_at`, computed over a bounded rolling window of samples.
- [ ] The middleware shall add less than 1 ms of overhead per request (measured in a test on a trivial endpoint).
- [ ] Streaming responses (SSE chat) shall be timed to completion without buffering and shall not be broken by the middleware.
- [ ] Requests that raise shall be recorded with status 500 and still logged if slow.
- [ ] The per-route summary shall be process-local and reset on restart; it shall never persist personal data.
- [ ] Trivial routes (`/health`, static pages) may be excluded via a small allowlist so they do not dominate the summary.

## Verification

- **Automated tests:** middleware tests with a tiny FastAPI app: slow route logs once with the template, fast route does not log, exception path records 500, streaming route passes through; API test for `/api/perf/routes` shape; overhead test.
- **Manual checks:** load the CRM on staging, then `curl /api/perf/routes` and confirm the route table reflects the calls.

## Scope

**In scope:**
- The middleware, settings threshold, in-memory summary, the `/api/perf/routes` endpoint, and a short operations note.

**Out of scope:**
- Persisting route timings to SQLite or daily rollups (#733 covers observability rollups; this issue provides the source data it can consume).
- Client-side timing.

## Technical Context

- **Key files:** `api/main.py:286` (middleware registration), `api/routes/perf.py` (existing `/api/perf` router; add `/routes`), `api/services/perf_trace.py` (existing trace store; add a separate small in-memory route summary rather than extending chat traces), `config/settings.py` (new `LIFEOS_SLOW_REQUEST_MS` field, default 500), `api/services/log_redaction.py` (logging setup; keep tokens redacted).
- **Route template:** read `request.scope.get("route").path` after the call (it is set once routing resolves) and fall back to `"<unmatched>"`.
- **Patterns to follow:** existing pure-ASGI or `BaseHTTPMiddleware` usage in the codebase if any; prefer a pure ASGI middleware so streaming is untouched.
- **Constraints:** no new dependencies; the summary must be thread-safe (handlers run in the threadpool after #868).

## References

- #733 (observability rollups and Trends view).
- `docs/specs/technical/observability.md`, `docs/guides/operations.md`

## History

Flagged during the 2026-09-03 CRM performance investigation; the only timing signal available then was a single handler's log line.

nbramia:
## TL;DR

Every API request is now timed. Slow requests (over half a second by default) get logged with what was slow, how slow, and how big the response was -- but never the exact web address that was called, so nothing personal ever leaks into a log line. An operator or agent can now ask "what's slow right now" at any time and get a live table of every route, sorted worst-first, instead of only finding out a page has become unbearably slow when someone complains. A long-lived live-updating connection (like the agent activity page or the chat assistant's live-typing reply) is tracked separately from that table, by count and total bytes only -- how long someone happened to leave a browser tab open was never a "slowness" the operator needed to see, and it's now kept out of the way instead of drowning out genuinely slow pages.

Closes #877

## Design

- A new outermost layer wraps every incoming request and response, timing how long the whole round-trip took (including any compression applied on the way out) and counting the bytes actually sent to the browser.
- Requests are grouped by the *shape* of the address they hit (e.g. "a person's page"), never the literal address itself, so a specific person's identifier never appears in a log line or in the summary table.
- A small, fixed-size rolling history (the most recent 200 requests) is kept per route shape, in memory only -- nothing is written to disk, and it resets whenever the server restarts.
- Health checks and static asset requests are left out of the summary entirely, since they're too frequent and too trivial to matter.
- A live-streaming reply (the chat assistant's live-typing response, the agent activity page's live feed) is timed all the way to its natural end, not just its first chunk, and nothing about how it streams changes -- but it's tallied separately from ordinary requests (a running count and total bytes, no "how long"), since how long someone leaves such a connection open reflects browser-tab habits, not server performance.
- If a request's connection drops partway through -- after it had already started responding -- the record reflects the status actually sent to the browser, marked as cut short, rather than guessing "server error."

## Implementation Notes

- `api/services/route_timing.py` (new) -- `RouteTimingMiddleware` (pure-ASGI, so streaming responses are never buffered) and `RouteTimingStore` (thread-safe, bounded rolling window per `(method, route template)`, computing `count`/`p50_ms`/`p95_ms`/`max_ms`/`slow_count`/`last_slow_at`; plus `record_stream()`/`stream_summary()` for `text/event-stream` responses -- count and total bytes only, no duration).
- `api/main.py` -- registers `RouteTimingMiddleware` last among the app's own `add_middleware` calls, which Starlette's stack makes the *outermost* layer (each `add_middleware` wraps around everything added before it), so its duration/byte measurement covers the full response including the scoped gzip middleware and CORS.
- `api/routes/perf.py` -- adds `GET /api/perf/routes`, a plain `def` handler returning `{"routes": [...], "count": N, "streams": [...], "stream_count": M}` sorted by `p95_ms` / `bytes` descending respectively.
- `config/settings.py` -- new `slow_request_ms` field (`LIFEOS_SLOW_REQUEST_MS`, default 500), documented in `docs/guides/configuration.md` (the authoritative env-var reference) and `.env.example`.
- Route template is read from `request.scope["route"].path` once routing resolves it, falling back to `"<unmatched>"` for a 404 -- the raw path/query string is never touched.
- A response's `content-type` (read off `http.response.start`) decides its bucket: `text/event-stream` (charset params included) goes to the stream summary and is never subject to the slow-request log, regardless of duration or whether it later fails mid-stream -- an SSE connection's lifetime is however long the client kept it open, not a latency signal, and would otherwise dominate the p95-sorted table and fire a false slow-request warning on every disconnect.
- A request that raises before any response was sent is recorded/logged (if slow) with status 500; one that raises *after* the response already started logs the status actually sent to the client instead, plus `aborted=true`.
- `http.response.pathsend` (zero-copy `FileResponse`, when the ASGI server advertises the extension) is accounted for from the `Content-Length` header already present on the preceding `http.response.start`, rather than re-stat'ing the file on the event loop.
- Excluded from the summary entirely: `/health*` and `/static/*`, matched as a full path segment (like `_in_gzip_scope` in `api/main.py`, #895) so a future `/healthz-admin` or `/staticmaps` route isn't swept in by accident. Everything else -- every `/api/*` call and every page route (`/crm`, `/me`, `/family`, `/relationship`, `/birthdays`, ...) -- is timed and recorded.
- Docs: `docs/specs/technical/observability.md` (Route Timing section), `docs/guides/operations.md` (a "what's slow right now" quick-command entry), `docs/specs/product/api-reference.md` (endpoint shape including `streams`), `docs/guides/configuration.md` (env var), `.env.example`, with bidirectional Related Documents links between the three specs.

### Automated tests

`tests/test_route_timing.py` (20 tests, built against a tiny standalone FastAPI app rather than the full `api.main.app`, so it stays in the fast unit suite; a process-wide-singleton reset is an `autouse` fixture, so no test's outcome depends on file/class order):

- Slow route logs exactly one WARNING containing the route template but never the raw path/id; fast route logs nothing (`TestSlowRequestLogging`).
- An exception before any response is sent is recorded/logged as status 500; an exception *after* the response already started (a generator that yields a chunk then raises) logs the real status sent plus `aborted=true`, not a synthetic 500 (`TestExceptionPath`, 2 tests).
- A streaming (non-SSE) response passes every chunk through untouched and is timed to completion, not to the first chunk (`TestStreaming`).
- An SSE response (plain, with a charset param on the content-type, and one whose generator fails mid-stream) never appears in the routes summary or the slow-request log, and shows up in the stream summary with count/bytes only (`TestSSEExclusion`, 3 tests).
- `http.response.pathsend` byte counting from the preceding `Content-Length` header, with and without that header present, driven directly against the middleware with a fake inner ASGI app since no ASGI transport available in this environment advertises the extension (`TestPathsend`, 2 tests).
- `RouteTimingStore` rolling-window math: summary shape/percentiles, window bound (200), no-samples routes omitted, stream summary shape/totals/reset (`TestRouteTimingStore`, 6 tests).
- Concurrent recording from 20 threads doesn't corrupt state (`test_thread_safety_concurrent_recording`).
- `GET /api/perf/routes` response shape, including the new `streams`/`stream_count` fields (`TestPerfRoutesEndpoint`).
- Overhead: bare and instrumented requests are interleaved *within* each iteration (200 iterations) and asserted as the median of the paired per-iteration deltas -- under 1ms (`TestOverhead`). An earlier sequential-phases version of this test measured host scheduling drift rather than middleware overhead and flaked under load; verified the interleaved version holds at load average 20-26 (8 synthetic busy loops), 4/4 passes.
- `/health` and `/static/*` excluded, matched as a full path segment so a lookalike route (`/healthz-admin`, `/staticmaps`) is still recorded (`TestExclusionAllowlist`, 2 tests).

Full local run (unit suite, this branch, rebased onto the integration branch tip `49c6d90`):

```
$ ./scripts/test.sh unit
========== 5658 passed, 9 skipped, 1181 warnings in 214.36s (0:03:34) ==========
```

Pre-push gate (unit + server-free browser tests) passed on push:

```
Running unit tests... passed (========== 5658 passed, 9 skipped, 1181 warnings in 214.36s (0:03:34) ==========)
Running server-free browser tests... passed (=============== 263 passed, 6032 deselected in 192.27s (0:03:12) ===============)
```

(An earlier push attempt on the original commit hit one unrelated flaky browser test, `test_backend_selector_ui_browser.py::TestSidebarPersonaValidationRace` -- a timing-sensitive assertion that never touches `web/`, so unrelated to this PR. It passed in isolation and on the retried push; not seen again on this push.)

### Manual verification

Started a private staging instance (port 8025, staging `.env`, symlinked disposable data copy) and exercised it against the real running app:

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/crm            # 200
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/me             # 200
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/family         # 200
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/birthdays      # 200
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/relationship   # 200
$ curl -s "http://127.0.0.1:8025/api/crm/people?limit=5"                        # 200
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:8025/nonexistent    # 404
$ curl -s http://127.0.0.1:8025/api/perf/routes | python3 -m json.tool
{
    "routes": [
        {"method": "GET", "route": "/api/crm/people", "count": 1, "p50_ms": 14.2, "p95_ms": 14.2, "max_ms": 14.2, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "/crm", "count": 1, "p50_ms": 7.6, "p95_ms": 7.6, "max_ms": 7.6, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "/relationship", "count": 1, "p50_ms": 2.2, "p95_ms": 2.2, "max_ms": 2.2, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "/me", "count": 1, "p50_ms": 2.0, "p95_ms": 2.0, "max_ms": 2.0, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "/family", "count": 1, "p50_ms": 1.5, "p95_ms": 1.5, "max_ms": 1.5, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "/birthdays", "count": 1, "p50_ms": 1.4, "p95_ms": 1.4, "max_ms": 1.4, "slow_count": 0, "last_slow_at": null},
        {"method": "GET", "route": "<unmatched>", "count": 1, "p50_ms": 0.6, "p95_ms": 0.6, "max_ms": 0.6, "slow_count": 0, "last_slow_at": null}
    ],
    "count": 7,
    "streams": [],
    "stream_count": 0
}
```

Every call made above is reflected exactly once, correctly grouped by route template; `/health` never appears (confirmed excluded); the 404 shows up as `<unmatched>`.

Confirmed `/api/crm/people` isn't double-counted after a repeat call and `/health` stays excluded after being polled at startup:

```
$ curl -s "http://127.0.0.1:8025/api/crm/people?limit=300" -H "Accept-Encoding: gzip"   # 200
$ curl -s http://127.0.0.1:8025/api/perf/routes | ...   # /api/crm/people count now 2
```

Confirmed the chat SSE stream still streams correctly, is recorded only once the stream itself finishes (not on its first frame), and lands in `streams` rather than `routes` or the slow-request log:

```
$ curl -N -s -D - -X POST http://127.0.0.1:8025/api/ask/stream \
    -H "Content-Type: application/json" \
    -d '{"question": "ping test synthetic query", "include_sources": false}'
HTTP/1.1 200 OK
content-type: text/event-stream; charset=utf-8
transfer-encoding: chunked

data: {"type": "conversation_id", "conversation_id": "..."}
data: {"type": "routing", "sources": ["agent"], ...}
data: {"type": "error", "message": "LIFEOS_LLM_BACKEND is 'anthropic' ... ANTHROPIC_API_KEY is not set."}
```

(The error event is expected -- staging has no Anthropic key configured, by design, so it can't spend API money. The stream itself -- headers, chunked transfer, multiple incremental `data:` events -- behaved identically to before this change.) `GET /api/perf/routes` afterward showed `/api/ask/stream` under `streams` (count 1, non-zero bytes), absent from `routes`, confirming both "recorded only after the stream finishes" and "SSE never pollutes the summary" together on the real endpoint this milestone cares about most.

Staging instance was killed and the copied `.env` removed before running any pytest suite or pushing, per the standing brief.

## Review focus

This PR went through two rounds of adversarial review. Round one (2 BLOCKERs, 1 MAJOR, 5 MINOR) was addressed in `b6444f5` -- see the review comment and the point-by-point reply. Round two (APPROVE WITH NITS) caught one leftover wording issue: two places still said an SSE response is "timed to completion", which is exactly what `b6444f5` stopped doing -- fixed in this revision (`docs/specs/technical/observability.md`, the `RouteTimingMiddleware` docstring, and this PR body's own manual-verification section all now say a non-SSE stream is timed to its final chunk, while an SSE response is recorded by count and bytes only, no duration). Remaining items worth a second look:

- **`last_slow_at` semantics**: it's tracked independently of the rolling window (so a slow request stays visible in the summary even after 200 fast requests roll it out of the window), while `count`/`p50`/`p95`/`max`/`slow_count` are all computed strictly from the current window. Flagging in case a reviewer expects `last_slow_at` to also be window-scoped (i.e. `null` once the slow sample ages out).
- **`http.response.pathsend` is untested end-to-end**: neither `TestClient`'s ASGI transport nor uvicorn on this host advertises the extension, so the byte-accounting code path is exercised only via a fake inner ASGI app in `TestPathsend`, not a real `FileResponse` request. If this app is ever served behind an ASGI server that does advertise `pathsend`, worth a manual spot-check.
- **Not this PR's problem, flagged for whoever picks up the reviewer's separate issue**: `list_people` (CRM) already logs the raw search query string at INFO, independent of anything in this PR. This PR's own "no raw path or query in logs" guarantee only covers what `RouteTimingMiddleware` itself logs.
- I did not add a way to reset the in-memory store outside of tests (`RouteTimingStore.reset()` exists but isn't wired to any route) -- out of scope per the issue, but flagging in case an operator wants one later.


nbramia:
## Adversarial review of #903 by Opus

**Verdict:** BLOCK

The middleware itself is well built — pure ASGI, correct scope guards, correct measurement point, and the "registered last ⇒ outermost" claim in the PR body is *right* (I verified it empirically, and the byte count really is the post-gzip wire size). Two things block: the summary this endpoint exists to produce is structurally misleading in normal operation, and the test that proves the overhead acceptance criterion fails about half the time on this host under ordinary load.

1. **[BLOCKER] Long-lived SSE connections permanently dominate `/api/perf/routes` and emit a false slow-request WARNING per connection** — `api/services/route_timing.py:38` (`_EXCLUDED_PREFIXES`) — `web/agents.html:2602` opens `new EventSource('/api/agents/stream')` for as long as the `/agents` tab is open, and `agents.html:1816` opens a per-session transcript stream the same way; `api/routes/agents.py:497` is a `while True` snapshot loop. Because the middleware (correctly) times to the final body chunk, each of those connections records a duration equal to *how long the browser tab was open*. On staging against the real 15k-person copy, after loading every CRM page plus `/agents`:

   ```
   /api/agents/stream            GET count=1 p95=43588.2 max=43588.2
   /api/crm/me/interactions      GET count=4 p95=19949.8 max=19949.8
   /api/crm/people               GET count=5 p95=4443.2  max=4443.2
   ```

   The p95-descending sort — the whole "what's slow right now" affordance — puts a page-open artifact above the genuinely broken 20-second endpoint this milestone exists to fix. In real use `/api/agents/stream` will sit at the top with a p95 in the millions of ms forever, and every closed EventSource writes `slow request: ... /api/agents/stream ... duration_ms=43588.2 bytes=4537917` into `lifeos-api-error.log`. That is exactly the "dominate the summary" failure the issue's allowlist criterion is meant to prevent, and it makes the log line actively misleading (a 43-second "slow request" that was never slow).

   **Fix:** capture the `content-type` from `http.response.start` and skip recording/logging for `text/event-stream`; or record streams under a separate key and never count them toward `slow_count`/the slow log. Either way add a test asserting an SSE response does not land in the summary.

2. **[BLOCKER] `test_overhead_under_1ms_median` fails ~50% of the time on this host under normal load; it measures host scheduling drift, not middleware overhead** — `tests/test_route_timing.py:235-286` — the test times 200 bare requests *then* 200 instrumented requests as two sequential phases, so any change in host load between the phases lands entirely in the delta. This is the only evidence for an acceptance criterion, and it is not sound.

   At the load this box actually runs at during the milestone (other agents' suites, `uptime` load avg ≈ 60):

   ```
   $ for i in 1..5; pytest tests/test_route_timing.py::TestOverhead -q
   AssertionError: median overhead 1.465ms exceeds 1ms budget (bare=1.658ms, instrumented=3.123ms)
   AssertionError: median overhead 2.581ms exceeds 1ms budget (bare=2.049ms, instrumented=4.629ms)
   AssertionError: median overhead 1.108ms exceeds 1ms budget (bare=7.065ms, instrumented=8.173ms)
   1 passed
   1 passed
   ```

   And with 12 synthetic busy loops on the 32-core box: 3 failures in 6 runs (worst 1.995ms). The pre-push gate runs the unit suite under `-n auto` (32 workers), i.e. exactly this condition — this will flake the gate for whoever pushes next.

   That the middleware is *not* the cause is easy to show: an interleaved variant (bare and instrumented alternating within each iteration, same 200 iterations, same 12-busy-loop load) never came close to the budget —

   ```
   INTERLEAVED bare=6.745ms inst=7.074ms delta=0.329ms
   INTERLEAVED bare=7.104ms inst=6.934ms delta=-0.170ms
   INTERLEAVED bare=7.616ms inst=7.489ms delta=-0.127ms
   INTERLEAVED bare=7.321ms inst=7.362ms delta=0.041ms
   ```

   **Fix:** interleave the two clients inside the loop (as above) and take the median of the *paired* per-iteration deltas. The measured overhead is genuinely well under budget; the test just needs to measure it.

3. **[MAJOR] `LIFEOS_SLOW_REQUEST_MS` is missing from `docs/guides/configuration.md` and `.env.example`** — `config/settings.py:314-321` — `configuration.md:7` declares itself "**the single authoritative reference for every `LIFEOS_*` environment variable**", and other guides defer to it. The new setting is documented in three other files but not the one that claims authority, and `.env.example` doesn't mention it either. Add a row to the Server (or a new Observability) table and a commented line to `.env.example`.

4. **[MINOR] The module's own class docstring states the opposite of what `api/main.py` does** — `api/services/route_timing.py:162-165` — "Registered outermost among this app's own middleware (**added first**, in `api/main.py`, **before** `CORSMiddleware` and the scoped gzip middleware)". It is added *last*, after both. `api/main.py:350-356` and `observability.md` both say it correctly; only this docstring is wrong, and it's the one a reader lands on when they go to check the ordering claim the PR body itself flags as the risky bit.

5. **[MINOR] A mid-stream failure is logged as `status=500` although the client received `200`** — `api/services/route_timing.py:191-193` — `except Exception: status_code = 500` overwrites the status unconditionally, including when `http.response.start` already went out. Reproduced with a `StreamingResponse` that raises after its first chunk:

   ```
   slow request: method=GET route=/midstream status=500 duration_ms=13.0 bytes=11
   client saw status 200
   ```

   For the SSE endpoints this middleware is meant to cover, a truncated stream is precisely the failure an operator would be debugging, and the log will point at the wrong status. Track a `response_started` flag and only force 500 when nothing was sent.

6. **[MINOR] `http.response.pathsend` (zero-copy file responses) is not accounted for; `bytes` silently reports 0** — `api/services/route_timing.py:183-186` — `/crm`, `/me`, `/family`, `/agents` etc. are all `FileResponse`, and starlette 0.52.1 (`responses.py:345,388-389`) emits `http.response.pathsend` instead of a body message whenever the server advertises the extension. Latent today (uvicorn does not advertise it — grep found no `pathsend` in the installed uvicorn), but Starlette's own `GZipMiddleware` handles the message type explicitly (`middleware/gzip.py:104`), which is the pattern to follow. One `elif message["type"] == "http.response.pathsend"` branch.

7. **[MINOR] `_is_excluded` uses bare prefix matching where the sibling change in the same file deliberately did not** — `api/services/route_timing.py:38-45` vs `api/main.py:315-327` — #895 added `_in_gzip_scope` with full-segment matching and a comment explaining why ("so a hypothetical future `/crmfoo` route wouldn't match"). This one is `path.startswith(("/health", "/static"))`, so a future `/healthz-admin` or `/staticmaps` route silently disappears from the summary. Match the neighbouring convention.

8. **[MINOR] Test isolation against the process-wide singleton is order-dependent** — `tests/test_route_timing.py:122-141, 213, 232, 307-313` — `TestStreaming` reads the global store without resetting it first, while `TestPerfRoutesEndpoint` and `TestExclusionAllowlist` call `store.reset()` mid-suite. It passes today only because of file order; any reordering (or a future test module in the same xdist worker that records into the singleton) breaks it. Reset in a fixture rather than inline, or have the tests construct their own `RouteTimingStore`.

Not a finding, flagged for the record only: `api/services/crm/...`'s `list_people` already logs the raw search string at INFO (`list_people(q=SECRETQUERYTOKEN, ...)` appeared in my staging log). That is pre-existing and outside this PR's scope, but it means this PR's careful "no raw path or query in logs" property is not the whole story for query strings.

### What I verified and how

- `git fetch origin feat/route-timing-middleware && git checkout` → reviewed the full diff (8 files) line by line against `origin/feat/crm-performance`.
- `pytest tests/test_route_timing.py -q` → 11 passed in 5.71s (idle host).
- `pytest tests/test_route_timing.py::TestOverhead -q` × 8 on an idle host → 8 passed; × 6 with 12 busy loops → 3 failed; × 5 at natural load avg ≈ 60 → 3 failed. Interleaved control script under the same load → deltas −0.17ms to +0.33ms. (Finding 2.)
- Middleware ordering, empirically: built a probe app with the same `add_middleware` order as `api/main.py` → `[m.cls.__name__ for m in app.user_middleware]` = `['RouteTimingMiddleware', 'GZipMiddleware', 'CORSMiddleware']` (index 0 = outermost). PR body's claim is correct; the module docstring's is not.
- Byte accounting covers gzip, on staging: `curl -H "Accept-Encoding: gzip" /crm` → `size_download=128584` and the log line `route=/crm status=200 bytes=128584` (vs `bytes=770203` uncompressed). Confirms outermost placement does what it claims.
- Scope guards: websocket handshake through the middleware → message delivered, store empty afterwards. Lifespan is short-circuited before `scope["path"]` is touched (the `or` ordering is correct — a lifespan scope has no `path`).
- Route-template resolution: all app routes are FastAPI `APIRoute`s and `child_scope["route"] = self` is set by `fastapi/routing.py:803` (Starlette's own `Route` does *not* set it) — so `scope["route"]` is populated for everything except the `/static` `Mount`, which is excluded anyway. 404 → `<unmatched>`; 405 → correctly attributed to `('POST', '/only-get')`; a `redirect_slashes` 307 → `<unmatched>` (acceptable, undocumented).
- Key-space growth: 50 bogus method tokens via `TestClient` created 50 distinct `(method, '<unmatched>')` keys, so the store is bounded per-route but not in the number of keys. **Not** a finding in practice: against real uvicorn, `curl -X FOOBAR http://127.0.0.1:8027/does-not-exist` → `400` ("Invalid HTTP request received"), rejected by the HTTP parser before ASGI. Route templates + real methods is a finite set.
- Thread safety: `record()` runs on the loop (middleware), `summary()` runs in the threadpool (the perf handler is a plain `def`, confirmed at `api/routes/perf.py:53-54`); both take the same lock, and the snapshot is copied under it before any formatting. Correct.
- Privacy: staging with `LIFEOS_SLOW_REQUEST_MS=1`, then `GET /api/crm/people?limit=300&q=SECRETQUERYTOKEN`, `GET /api/crm/people/SENSITIVE-PERSON-ID-42`, `GET /no-such-page?leak=SECRETQUERYTOKEN` → log lines were `route=/api/crm/people`, `route=/api/crm/people/{person_id}`, `route=<unmatched>`; neither marker appeared in any route-timing line. `log_redaction` is untouched and the new logger propagates to the root handlers that carry the filter.
- Staging on :8027 (staging `.env`, real 15,391-person data copy symlinked), Playwright/chromium: `/crm`, `/me`, `/family`, `/birthdays` → 0 console errors, 0 failed requests. `/agents` (`domcontentloaded`) → 0 console errors, 0 failed requests, SSE live. `/relationship` → 2 console errors from a 404 on the relationship-timeline call; data-shaped (no partner configured in the staging copy) and not attributable to this PR, which adds no route and cannot change a status.
- SSE unbroken and unbuffered: `curl -N /api/agents/stream` timestamped per line → snapshot events at 07:23:07.6, :09.6, :11.6, :13.6 — real 2-second intervals, nothing buffered. Recorded once, timed to completion (`max_ms=6997.1` for a 7s client timeout). `POST /api/ask/stream` → chunked `text/event-stream`, three incremental `data:` events, headers unchanged.
- `/api/perf/routes` reflects the CRM pages: after loading all five pages against real data, the table showed `/crm`, `/me`, `/family`, `/relationship`, `/birthdays`, `/api/crm/people`, `/api/crm/people/{person_id}`, `/api/crm/me/interactions`, `/api/crm/me/stats`, `/api/crm/birthdays/*`, `/api/crm/family/members`, `/api/crm/statistics` — 20 routes, correctly grouped by template, `/health` and `/static/*` absent. Criterion met (modulo finding 1).
- Docs: `observability.md`, `operations.md`, `api-reference.md` all describe current state accurately and the Related Documents links are bidirectional. `grep -c LIFEOS_SLOW_REQUEST_MS docs/guides/configuration.md .env.example` → 0, 0 (finding 3).
- Scope/surgery: `api/main.py` diff is one import plus one `add_middleware` with a comment; no reformatting, no new dependencies, no test deleted or weakened. Clean.

### Not verified

- The full unit suite. I started `pytest tests/ -m "unit and not slow" -n auto` and killed it when I found two other agents already running full gates on this box (load avg 61) — three concurrent gates have OOM-rebooted this host before. The PR reports 5649 passed and the pre-push gate passed on push; I ran the changed file and the affected middleware paths directly instead.
- Whether finding 1's SSE pollution was a conscious call. The PR body's "Review focus" flags the exclusion allowlist as worth confirming but discusses only page routes, not streams, so I am treating it as a gap rather than a decision.
- `http.response.pathsend` end-to-end — uvicorn does not advertise the extension, so I could only confirm the code path exists in Starlette 0.52.1 and is unhandled here, not observe a 0-byte reading in production.
- Behaviour of the summary across the `slow_request_ms` threshold changing at runtime (documented as intentional: `slow_count` is recomputed against the current threshold, `last_slow_at` is not). Left alone as a stated design choice, though a reviewer may want `last_slow_at` window-scoped as the author flagged.

nbramia:
Thanks for the thorough review -- all 8 findings addressed in b6444f5 (on top of the original 6c6e902), rebased on the current `feat/crm-performance` tip (`cdfb4f9`), pushed and green on both the unit suite (5658 passed, 9 skipped) and the server-free browser suite (263 passed).

1. **[BLOCKER] SSE polluting the summary** -- Fixed. `RouteTimingMiddleware` now reads `content-type` off `http.response.start`; a `text/event-stream` response (charset params included, e.g. `text/event-stream; charset=utf-8`) is routed to `RouteTimingStore.record_stream()` instead of `record()` -- count and total bytes only, no duration, and it never reaches the slow-request log regardless of how long the connection was open or whether it later fails mid-stream. `GET /api/perf/routes` now returns a separate `streams`/`stream_count` alongside `routes`/`count`. Tests: `TestSSEExclusion` (plain SSE, charset-param content-type, and an SSE generator that raises mid-stream -- all confirmed absent from `summary()`/the log and present in `stream_summary()`), plus `TestRouteTimingStore.test_stream_summary_shape_and_totals` / `test_reset_clears_stream_state_too`, plus `TestPerfRoutesEndpoint.test_shape` asserting the new endpoint fields.

2. **[BLOCKER] Flaky sequential overhead test** -- Rewrote as your interleaved/paired-delta design: each iteration times one bare request immediately followed by one instrumented request, and the assertion is the median of the 200 per-iteration `(instrumented - bare)` deltas, not two separate medians from sequential phases. Verified locally: 5/5 clean passes at idle, and 4/4 passes with 8 synthetic `yes > /dev/null` busy-loops running concurrently (load avg 20-26) -- matches your report that the interleaved measurement holds up under load where the sequential one didn't.

3. **[MAJOR] `LIFEOS_SLOW_REQUEST_MS` missing from the authoritative docs** -- Added a row to the Server table in `docs/guides/configuration.md` and a commented line + short comment in `.env.example`, next to `LIFEOS_PORT` in both.

4. **[MINOR] Backwards docstring** -- Fixed; `RouteTimingMiddleware`'s docstring now says "added *last* ... after `CORSMiddleware` and the scoped gzip middleware", matching `api/main.py` and `observability.md`.

5. **[MINOR] Mid-stream failure logged as a false 500** -- Added a `response_started` flag, set on `http.response.start`. The except block only forces `status_code = 500` when nothing was ever sent; when the response had already started, the log line now carries the real status the client received plus an `aborted=true` marker. Test: `test_exception_after_response_started_logs_real_status_and_aborted` (a generator that yields one chunk after a 200 then raises -- asserts `status=200` and `aborted=true` in the log, and a second test confirms the original before-any-response case still logs `status=500` with no `aborted` marker).

6. **[MINOR] `http.response.pathsend` unhandled** -- Added an `elif` branch matching the convention in Starlette's own `GZipMiddleware`. Rather than re-stat the file on the event loop (a blocking syscall), it reads the `Content-Length` header already captured from the preceding `http.response.start` (which `FileResponse` always sets from the same stat result) -- so the byte count is accurate with zero extra I/O. Falls back to 0 bytes with a DEBUG-level note if that header is somehow absent. Since neither `TestClient`'s ASGI transport nor uvicorn on this host advertises the `pathsend` extension, I drove this directly against the middleware with a fake inner ASGI app (`TestPathsend`, two tests: header present / header absent) rather than through a real `FileResponse` request.

7. **[MINOR] Bare prefix matching vs. `_in_gzip_scope`'s full-segment convention** -- `_is_excluded` (renamed the constant to `_EXCLUDED_ROUTES` for clarity) now matches `path == route or path.startswith(route + "/")`, identical in shape to `_in_gzip_scope`. Test: `test_lookalike_routes_are_not_excluded` asserts `/healthz-admin` and `/staticmaps` are recorded, alongside the existing `test_health_and_static_are_excluded` (extended to also cover `/health/full`).

8. **[MINOR] Order-dependent singleton isolation** -- Replaced every scattered inline `store.reset()` with a single `autouse` fixture (`_reset_route_timing_store`) that resets the process-wide singleton before *and* after every test in the file, so no test's pass/fail depends on file or class order, or on what ran earlier in the same xdist worker.

Also flagging per your note: the pre-existing raw-query INFO log in `list_people` is out of scope for this PR (I didn't touch that code path) -- noted in the PR body for whoever picks up the issue you're filing separately.

PR body updated with the new test evidence and design notes above.

nbramia:
## Adversarial re-review of #903 by Opus — verification pass on `b6444f5`

Both blockers and the major are fixed, and I confirmed each of the five minors independently rather than by reading the diff. The SSE separation in particular is better than what I asked for — the bucket is decided by `content-type` off `http.response.start` and is case- and parameter-insensitive, so it holds for `text/event-stream`, `text/event-stream; charset=utf-8`, and `TEXT/EVENT-STREAM` alike, and it survives a generator that dies mid-stream. One nit remains, and it is only wording.

### Remaining findings

1. **[MINOR] Two places still say SSE is "timed to completion", which is exactly what `b6444f5` stopped doing** — `docs/specs/technical/observability.md:78` and `api/services/route_timing.py:227` — both read "a streaming reply (SSE chat) is **timed to completion** via its final body chunk, with every chunk passed straight through." In `observability.md` that bullet sits three lines above the new bullet saying an SSE response gets no duration at all, so the section contradicts itself; in `route_timing.py` the same sentence sits in the same docstring as the new SSE paragraph. The *behaviour* those sentences describe is still half right — the middleware does await the whole stream, unbuffered, and `record_stream` fires after the final chunk with the full byte total — it just no longer produces a duration. Suggested rewording: "…never buffers a response: every chunk is passed straight through and the middleware waits for the stream's final chunk before recording. A non-SSE streaming response is timed to that final chunk; an SSE response is recorded by count and bytes only (below)." The same phrase appears twice in the PR body's manual-verification section and would be worth the same edit.

   Worth stating explicitly for the record, since it touches an acceptance criterion: dropping SSE duration is the right call and I do not read it as failing criterion 4 ("streaming responses shall be timed to completion without buffering and shall not be broken"). The no-buffering and not-broken halves I verified live below; chat-turn latency is already covered in far more detail by `perf_trace.py`, which is what that criterion's real consumer is. But the docs should not claim a behaviour the code deliberately removed.

### Findings 1–8: verification

| # | Finding | Status | How I confirmed it |
|---|---|---|---|
| 1 | SSE dominates the summary / false slow warnings | **Fixed** | Live on :8027, staging data, `LIFEOS_SLOW_REQUEST_MS=1` |
| 2 | Sequential overhead test flakes ~50% under load | **Fixed** | 16/16 passes across two synthetic load regimes |
| 3 | `LIFEOS_SLOW_REQUEST_MS` absent from authoritative docs | **Fixed** | `configuration.md:23`, `.env.example:244-246` |
| 4 | Backwards ordering docstring | **Fixed** | Docstring now says "added *last* … after `CORSMiddleware`", matching the empirically verified stack |
| 5 | Mid-stream failure logged as a false 500 | **Fixed** | Independent probe, log line carries the sent status + `aborted=true` |
| 6 | `http.response.pathsend` unhandled | **Fixed** | Independent probe driving the middleware with a fake ASGI app |
| 7 | Bare prefix matching for `/health`, `/static` | **Fixed** | `_is_excluded` called directly across seven lookalike paths |
| 8 | Order-dependent singleton isolation | **Fixed** | Every class run alone, in a reordered selection, and under `-n 4 --dist loadscope` |

### What I verified and how

**Finding 1 — SSE separation, live.** Staging on :8027 with the staging `.env` and the real 15,391-person data copy, `LIFEOS_SLOW_REQUEST_MS=1` so *every* ordinary route logs. Loaded all five CRM pages, held `/api/agents/stream` open for 20 seconds, and ran one `POST /api/ask/stream` chat turn. `GET /api/perf/routes`:

```
ROUTES (top by p95):
  /api/crm/people        GET count=1 p95=1156.1 max=1156.1 slow=1
  /crm                   GET count=1 p95=34.7   max=34.7   slow=1
  /relationship          GET count=1 p95=10.6   ...
  routes count = 6
STREAMS:
  /api/agents/stream     GET  count=1 bytes=3712056
  /api/ask/stream        POST count=1 bytes=500
  stream_count = 2
any stream route leaked into routes? []
```

A 20-second connection that previously sat at the top of the p95 table with `p95=43588.2` now contributes zero duration samples, and the table's top entry is the genuinely slow `/api/crm/people`. `grep "slow request" | grep -c stream` → **0**, against 7 slow lines total (one per ordinary route) at a 1 ms threshold — so the exclusion is real, not "happens not to be slow enough". Byte accounting for the stream is intact (3.7 MB over 20 s).

SSE is still unbuffered and unbroken — `curl -N /api/agents/stream` with per-line timestamps: `08:13:59.938`, `08:14:01.990`, `08:14:04.061` → real 2-second intervals, nothing buffered. `POST /api/ask/stream` still returns chunked `text/event-stream` with incremental `data:` events.

Beyond what the PR's own tests cover, I checked `media_type="TEXT/EVENT-STREAM"` (uppercase): routes `[]`, streams `[{'route': '/sse-upper', 'count': 1, 'bytes': 29}]`. The `.lower()` on the split content-type handles it.

**Finding 2 — overhead test.** `TestOverhead::test_overhead_under_1ms_median_paired`, now interleaved with paired per-iteration deltas:

- 12 synthetic busy loops (load avg ~40): **8/8 passed**
- 32 synthetic busy loops on the 32-core box (load avg peaking 57): **8/8 passed**

For comparison, the old sequential version failed 6 of 11 runs under the same two regimes. 16/16 with zero failures is a real fix, not a threshold widened to hide the noise — the assertion is still `< 1.0 ms`.

**Finding 5 — mid-stream abort.** Independent probe (my own app, not the PR's fixtures): a `text/plain` `StreamingResponse` that yields one chunk then raises →

```
slow request: method=GET route=/midstream status=200 duration_ms=14.0 bytes=10 aborted=true
client saw status 200
```

Correct status and the new marker. The pre-response case still logs `status=500` with no marker (`test_exception_before_response_records_status_500` asserts the absence explicitly, which is the right shape).

**Finding 6 — pathsend.** Drove `RouteTimingMiddleware` directly with a fake inner app emitting `http.response.start` + `http.response.pathsend`:

- with `content-length: 770203` → `bytes=770203`
- with no `Content-Length` → `route_timing: pathsend response with no Content-Length header; byte count for this response will read as 0` at DEBUG, then `bytes=0`

Both cases forwarded `['http.response.start', 'http.response.pathsend']` downstream untouched, so the zero-copy path is not broken by the new branch. Reading the header rather than re-`stat`ing the file is the right choice on the event loop, and matches what `FileResponse` guarantees.

**Finding 7 — full-segment exclusion.** `_is_excluded` called directly: `/health` → True, `/health/full` → True, `/static/a.js` → True; `/healthz-admin` → False, `/staticmaps` → False, `/healthcheck` → False, `/api/health` → False. Matches `_in_gzip_scope`'s convention exactly.

**Finding 8 — isolation.** `pytest tests/test_route_timing.py -q` → 20 passed. Each class alone: `TestStreaming` 1, `TestPerfRoutesEndpoint` 1, `TestExclusionAllowlist` 2, `TestPathsend` 2 — all pass. Reordered selection (`TestSSEExclusion TestStreaming TestExclusionAllowlist TestPerfRoutesEndpoint`) → 7 passed. Under `-n 4 --dist loadscope` → 20 passed. The autouse fixture resets before *and* after, so this file can neither be polluted by nor pollute another module sharing an xdist worker.

**Regression checks on the unchanged behaviour.** Middleware ordering is still `['RouteTimingMiddleware', 'GZipMiddleware', 'CORSMiddleware']` (index 0 = outermost). A websocket handshake still passes through with nothing recorded in either bucket. All five CRM pages returned 200 against real data with the new middleware in place.

**Docs.** `configuration.md:23` now carries the `LIFEOS_SLOW_REQUEST_MS` row in the Server table with a link to the Observability anchor; `.env.example:244-246` has the commented line next to `LIFEOS_PORT`; `api-reference.md` documents `streams`/`stream_count` and why SSE is split out; `observability.md` documents the stream bucket, the full-segment exclusion, and the `aborted=true` semantics. Only the one stale sentence above (finding 1 of this pass) is left.

### Not verified

- The full unit suite — per instruction I ran only `tests/test_route_timing.py` and its classes. The implementer reports 5658 passed / 9 skipped plus 263 server-free browser tests green on push.
- `http.response.pathsend` through a real `FileResponse` over uvicorn — still impossible here, uvicorn does not advertise the extension. Both the PR's tests and my probe drive the message type synthetically; that is the only option in this environment.
- Long-horizon behaviour of the `streams` byte counter: it is a monotonic since-restart total (never windowed), which is what the docs say, but I did not run a server long enough to see whether an operator finds a cumulative byte total more or less useful than a per-connection average. Fine as shipped.

nbramia:
Wording nit fixed in `e50029d` (rebased onto `49c6d90`, the current `feat/crm-performance` tip): `docs/specs/technical/observability.md`, the `RouteTimingMiddleware` docstring, and this PR body's manual-verification section no longer say an SSE response is "timed to completion" -- they now say the middleware waits for a stream's final chunk before recording it, a non-SSE stream is timed to that final chunk, and an SSE response is recorded by count and bytes only (no duration). Pushed and green: unit suite 5711 passed / 9 skipped, server-free browser suite 264 passed.

Issue #876, opened 2026-09-04
feat(crm): client-side dashboard navigation and a data_version-keyed cache for heavy aggregates


nbramia:
## Problem

Every move between the CRM's dashboards is a full page reload. The Me, Family, Birthdays, Relationship, and CRM links are plain anchors, so each click downloads the 750 KB page again, refetches the graph library from a CDN, re-reads config, and re-runs that dashboard's data loads from scratch, discarding everything the previous view had already fetched. Even after the individual endpoints are made fast, navigation stays visibly slow, and on a phone over Tailscale it is the dominant delay. The page already routes person and tab changes client-side, so dashboards are the odd ones out.

## Solution

Dashboard navigation becomes client-side, and the heaviest dashboard aggregates get a short-lived server cache. The nav links push history entries and dispatch to the existing dashboard initializers instead of reloading, back and forward work through the existing history handling, and each dashboard's in-memory data survives across navigations within a session. On the server, the expensive aggregate responses (Me and Family interactions and timelines, birthdays, statistics) are memoized for a short time keyed on the data-version counters of the CRM and interactions databases, so a second client, a refresh, or a return visit after the in-memory cache expires is served instantly, and any commit to either database invalidates the entry. Two PRs: client-side routing first, server cache second.

## Acceptance Criteria

- [ ] When a user clicks the Me, Family, Birthdays, Relationship, or CRM nav link, the browser shall not perform a document load; the URL shall update and the dashboard shall render.
- [ ] Given a user navigated Me → Family → Me, when the second Me render happens, the client shall issue no request for data it already holds for the same parameters.
- [ ] Browser back and forward shall restore the previous dashboard without a document load.
- [ ] Direct loads of `/me`, `/family`, `/birthdays`, `/relationship`, and `/crm/{id}` shall render exactly as they do today.
- [ ] The active nav link styling shall follow the current view.
- [ ] Given unchanged databases, when the same aggregate endpoint is requested twice within the cache lifetime, the second response shall be served from the server cache in under 20 ms with an identical body.
- [ ] Given a commit to crm.db or interactions.db from any process, the next request for a cached aggregate shall recompute.
- [ ] The server cache shall be bounded (entries and total bytes) and shall never cache error responses.

## Verification

- **Automated tests:** server-free browser test that stubs the dashboard endpoints, clicks through all nav links, and asserts no `load` navigation events plus request counts; API tests for cache hit, invalidation via a separate SQLite connection, error non-caching, and bounds.
[...]

PR #907, opened 2026-09-04
fix(crm): show profile photos in the people list without probing

### LINKED ISSUE #875: fix(crm): show profile photos in the people list without per-person probing

### LINKED ISSUE #904: fix(crm): stop logging the raw people-search query at INFO
## Problem

The people list handler writes the raw search text to the server log at INFO level on every call, so anything typed into the CRM search box (names, partial emails) lands in the log files. That contradicts the project's rule that personal data must never be logged, and it undercuts the route-timing work that deliberately keeps raw paths and query strings out of the logs.

## Solution

The people list handler logs timing and counts only, never the search text or filter values. The route-timing middleware already records per-route durations, so the handler's own log line can shrink to counts or be removed.

## Acceptance Criteria

- [ ] When `GET /api/crm/people?q=<text>` is called, no log record shall contain `<text>`.
- [ ] A test with a distinctive query string shall assert the string is absent from captured log output.

## Scope

**In scope:** the one log line in the list handler and a test.
**Out of scope:** a general log audit.

## Technical Context

- **Key files:** `api/routes/crm.py` (the `list_people(q=..., category=..., limit=...) took ...ms` INFO line at the end of the list handler), `tests/test_crm_api.py` or a new small test using `caplog`.

## History

Flagged by the #903 reviewer on 2026-09-04 while verifying that the new route-timing logs carry no query strings.

nbramia:
## TL;DR

Profile photos now show up in the CRM people list for the people who actually have one, without probing every visible row first. Opening the list no longer sends any avatar requests for people without a photo, a person whose photo turns out to be missing quietly falls back to their initials, and an install with Photos turned off sends no avatar requests at all — anywhere in the CRM, including the person detail header and the relationship dashboard, not just the list. Separately, the server no longer writes what someone typed into the people-search box to its logs, including the copy of that text uvicorn's own access log used to write independently of anything the app logged.

Closes #875
Closes #904

## Design

- The people-list response now tells the client, per person, whether they have a usable profile photo (`has_profile_photo`), computed from the existing photo count the person record already carries. The same field is present on the single-person detail response for consistency.
- Whether Photos is configured at all (`photos_enabled`) rides on `GET /api/crm/config`, a call the client already makes before its first render — no separate request. The client checks this before ever requesting an avatar, anywhere in the CRM (list, person detail header, relationship dashboard).
- The client renders a real photo only for people flagged that way, loaded lazily so a large list doesn't fire a burst of image loads on render. No probing request of any kind is issued for anyone else.
- If a flagged person's photo can't actually be loaded (the record exists but the thumbnail file is missing), the avatar quietly falls back to initials and remembers not to try again for the rest of the session.
- The avatar photo route answers a `HEAD` request exactly the way a browser expects one to behave: the same headers a `GET` would carry (cache validators included), just without the body. Both a found and a not-found response are cacheable for an hour, so repeat renders don't need to ask again.
- The people-list handler's server log line reports counts and timing only — never the search text a person typed or any filter value. So does uvicorn's own access-log line for every request, not just this one.

## Implementation Notes

- `api/routes/crm.py`: `PersonDetailResponse.has_profile_photo`, set in `_person_to_detail_response()` from `person.photo_count > 0`. `CRMConfigResponse.photos_enabled`, set in `get_crm_config()` from `settings.photos_enabled` (a cheap `Path.exists()`). `list_people()`'s INFO log line no longer includes `q` or `category` (#904).
- `api/routes/photos.py`: `get_profile_photo()` (`/api/photos/profile/{person_id}`) now stacks `@router.get(...)` and `@router.head(..., include_in_schema=False)` on the same function, so Starlette's `FileResponse` handles `HEAD` natively — full header parity with `GET` (`etag`, `last-modified`, `accept-ranges`, real `content-length`), and no duplicate-operation-id warning. The "no photo" `404` carries `Cache-Control: public, max-age=3600`.
- `api/services/log_redaction.py`: new `RequestQueryStringRedactionFilter`, installed directly on the `uvicorn.access` logger from `api/main.py` at import time. uvicorn's access logger writes the full request line (path *and* query string) at INFO independent of anything a route handler logs, and has `propagate=False` with its own dedicated handler — so a root-handler filter (the pattern the existing Telegram-token filter uses) can't see it; it has to sit on the logger itself. Installing it at `api/main.py` import time is safe because uvicorn's own logging `dictConfig()` call (inside `Config.__init__`) runs *before* the app string gets imported in the `scripts/server.sh` startup path (`uvicorn.run("api.main:app", ...)`) — confirmed empirically, see the module's docstring. **The filter redacts `record.args[2]` in place, keeping the 5-tuple shape intact** — `uvicorn.logging.AccessFormatter.formatMessage` unconditionally unpacks `record.args` as `(client_addr, method, full_path, http_version, status_code)` and never reads `record.msg`/`getMessage()` at all, so an earlier version of this filter (which rewrote `record.msg` and cleared `record.args`) broke that unpack and turned every query-string request into a 79-line "Logging error" traceback instead of an access line. A `record.msg`-based rewrite is kept only as a fallback for a record that isn't shaped like uvicorn's own access-log call.
- `web/crm.html`:
  - `renderPeopleList()` computes `renderAvatarPhotoImg(person)` once per row and reuses it for both the `has-photo`/`no-photo` class and the markup itself.
  - `renderAvatarPhotoImg()` renders a `loading="lazy"` `` only when `photosEnabled && person.has_profile_photo` and the person isn't already known (via `profilePhotoCache`) to have no reachable thumbnail.
  - `handleAvatarPhotoError()` is the `onerror` handler: removes the broken ``, flips the avatar to the `no-photo` class, and caches the miss.
  - `photosEnabled` is set in `loadCRMConfig()` from `GET /api/crm/config`'s `photos_enabled` field — `init()` already awaits this before the people list's first render, so no separate request is needed.
  - `loadDetailProfilePhoto()` and `loadRelationshipPartnerPhoto()` (the person-detail header and relationship-dashboard photo loaders) now check `photosEnabled` too, so a Photos-less install doesn't 503 on every person selection or every relationship-dashboard load.
  - The old `loadProfilePhotos()` (a post-render probing loop that `HEAD`-checked every avatar and was never actually called) and the now-redundant `avatarPhotoClass()` helper are removed; `profilePhotoCache` is kept and still shared with the detail-header/relationship-page photo loaders.
  - CSS: `.person-avatar` is a positioning context (`position: relative; overflow: hidden`) for `.avatar-photo-img`, an absolutely-positioned, `object-fit: cover` image that visually covers the initials once it loads. The pre-existing `.person-avatar.has-photo { color: transparent }` rule is inert on list rows specifically (an inline `color` from `getAvatarStyle()` always wins) — documented in place; it's what makes a still-loading avatar degrade to visible initials instead of a blank circle, and it does apply where nothing sets an inline color (e.g. `.detail-avatar.has-photo`).
- `docs/specs/technical/architecture.md`: corrected — `PersonDetailResponse` and the other models the API actually serves live in `api/routes/crm.py`, not `api/routes/crm_models/`, which is a separate module tree not wired into the running app. It isn't dead code, though: `is_family_member()` there is a second implementation of family-matching logic that `tests/test_family_matching.py` keeps in sync with the real one in `api/services/person_entity.py`.

### Automated tests

```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest \
  tests/test_photos_profile_api.py tests/test_crm_avatar_photos_ui_browser.py \
  tests/test_crm_request_hygiene_ui_browser.py tests/test_log_redaction.py tests/test_family_matching.py -q
```
```
42 passed
```

```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_crm_api.py -q -k "photos_enabled or has_profile_photo or never_logs"
```
```
5 passed
```
This file's own latency-threshold tests (`test_get_people_list_warm_latency_large_db{,_limit_300}`, `test_me_interactions_3657_days_under_1200ms`, `test_me_interactions_span_under_50ms` — the first's own docstring already documents it as host-load-sensitive) intermittently exceed their fixed-ms thresholds on this shared, multi-agent host; how many fail on a given run tracks how loaded the host is at that moment (observed 101ms-1865ms against 50-1200ms bounds across this PR's review rounds, worsening as more concurrent agent activity piled up). Confirmed pre-existing and unrelated by reproducing the identical failures on `origin/feat/crm-performance` with this branch's diff removed. Not touched.

Full unit suite, run under the shared push lock per host-load policy (final round):
```
flock -w 7200 /tmp/lifeos-push.lock ./scripts/test.sh
5723 passed, 8 skipped, 1181 warnings in 405.27s
```
(Only the pre-existing `voice_proxy` duplicate-operation-id warning remains — this PR's own duplicate warning is gone after finding 3+5's fix.)

Full pre-push gate (unit suite + server-free browser suite), run automatically by `git push` under the shared lock, on the final rebase (`feat/crm-performance` @ `86527c5`):
```
Running unit tests... passed (5772 passed, 8 skipped, 1182 warnings in 261.60s)
Running server-free browser tests... passed (280 passed, 6156 deselected in 199.38s)
```

Acceptance criteria, from #875 and #904:

| # | Criterion | Test | Result |
|---|---|---|---|
| 1 | `GET /api/crm/people` includes `has_profile_photo: bool`, true iff `photo_count > 0` | `test_get_people_list_has_profile_photo_field`, `test_person_detail_has_profile_photo_matches_list` (`tests/test_crm_api.py`) | Pass |
| 2 | The client requests a photo only for `has_profile_photo` people, never `HEAD` | `test_exactly_two_photo_requests_for_flagged_people`, `test_no_head_requests_are_ever_issued` (`tests/test_crm_avatar_photos_ui_browser.py`) | Pass |
| 3 | `HEAD /api/photos/profile/{id}` returns the same status as `GET`, no body, full header parity | `test_head_returns_200_with_no_body_and_full_header_parity_with_get`, `TestProfilePhotoDisabled`, `TestProfilePhotoMissing` (`tests/test_photos_profile_api.py`) | Pass |
| 4 | `GET`'s `Cache-Control` allows ≥ 1 hour caching on both `200` and `404` | `test_get_returns_image_body`, `test_get_404_carries_cache_control_of_at_least_one_hour` (`tests/test_photos_profile_api.py`) | Pass |
| 5 | A missing thumbnail falls back to initials with no application error | `test_avatar_classes_reflect_photo_availability` (`tests/test_crm_avatar_photos_ui_browser.py`) — see Review focus on what "no console error" means here | Pass |
| 6 | 300 rows don't block the main thread on photo loading (lazy) | `expect(img).to_have_attribute("loading", "lazy")` and `test_300_flagged_rows_request_far_fewer_images_at_first_paint` (`tests/test_crm_avatar_photos_ui_browser.py`) — mutation-tested: deleting `loading="lazy"` fails both (300/300 requests instead of ~50/300) | Pass |
| 7 | Photos disabled → initials everywhere, zero photo requests, everywhere in the CRM | `test_photos_disabled_issues_zero_photo_requests` (list); manually re-verified on staging for the list, person detail header, and relationship dashboard | Pass |
| 904-1 | No log record (handler or access log) contains the raw `q` text | `test_list_people_never_logs_the_search_query` (handler; `tests/test_crm_api.py`); `test_redacts_query_string_from_access_log_line`, `test_redacted_record_survives_uvicorns_actual_access_formatter` (access log, run through the *real* `uvicorn.logging.AccessFormatter`; `tests/test_log_redaction.py`) | Pass |
| 904-2 | Reproduced end to end against the exact production log configuration, without breaking the access log itself | See Manual verification | Pass |
| config | `GET /api/crm/config`'s `photos_enabled` matches `settings.photos_enabled` | `test_photos_enabled_reflects_settings_when_false`, `test_photos_enabled_reflects_settings_when_true` (`tests/test_crm_api.py`) | Pass |

### Manual verification

Private staging instance on port 8026, launched from this worktree against a read-only copy of the production databases (photos library absent, matching the standing brief's staging setup), rebuilt after each rebase and after the args-based redaction fix.

- **#904 end to end**, matching the exact configuration `scripts/server.sh` uses (`uvicorn.run("api.main:app", host=..., port=..., log_level="info", ...)`, the app loaded by string so `api/main.py`'s module-level filter installation runs after uvicorn's own logging `dictConfig()`): sent `?q=ZZQMARKER7788&sort=strength&limit=5`, then `grep`'d the log file for both the marker and `"Logging error"` — **both zero occurrences** — and confirmed a normal, correctly-formatted access line was emitted: `INFO:     127.0.0.1:... - "GET /api/crm/people?<redacted> HTTP/1.1" 200 OK`. Also sent a query-string-free request (`/health`, `/api/crm/config`) and confirmed those log completely untouched. (An earlier round's version of this same check only grepped for the marker's *absence*, which a broken filter also produces — as the 79-line traceback it emits instead of an access line happens to echo the redacted message in its `Message:` field. Now checks for the absence of `"Logging error"` too, and for a normal-shaped access line.)
- `curl` against the running staging instance confirmed the photo route fix end to end with real data: `GET` on a person with a photo interaction → `200 image/jpeg` with `Cache-Control: public, max-age=3600`; `HEAD` on the same id → `200`, empty body, same header; `GET`/`HEAD` on a person without one → `404` with the same `Cache-Control`.
- Playwright against `/me` (219 real people, photos disabled in this staging install, in an isolated tab): 0 console errors, initials rendered for everyone, and **zero `/api/photos/*` requests of any kind**. Search and person selection both still worked with 0 console errors.
- Selected a real flagged person (`has_profile_photo: true`) on `/me`: the detail header issued **zero** photo requests (previously it would 503 and log a console error every time, per finding 7) — 0 console errors. Timeline and Graph tabs for that person: 0 console errors, 0 photo requests.
- `/family`, `/birthdays`: 0 console errors. `/relationship`: 0 console errors too now — a sibling PR (#899) fixed the no-partner-configured case this staging install was hitting, previously reported here as 2 pre-existing, unrelated errors; confirmed the relationship dashboard also issues 0 photo requests.
- To verify a photo actually renders end to end, pointed a second staging run at a temporary Photos-library stand-in (real `Photos.sqlite` file presence, one real, decodable synthetic JPEG placed in `resources/derivatives/`) for a real person id already flagged `has_profile_photo: true`. `GET`/`HEAD` on that person's photo route both returned `200` with the real image bytes and full header parity. `GET /api/crm/config`'s `photos_enabled` came back `true` directly from this fixture; with that, the flagged person's avatar rendered the real photo (`class="person-avatar has-photo"`, an `` with `naturalWidth: 2`), and a different flagged person whose file didn't exist fell back cleanly to `class="person-avatar no-photo"` with visible initials and no `` left in the DOM.

**On `photo_count` lag** (raised in review): `photo_count` is refreshed synchronously at the end of each Apple Photos sync run (`apple_photos_sync.py` calls `refresh_person_stats()` for exactly the people who gained photo interactions in that run) — not by a separate, possibly-delayed nightly job — so it goes stale only for the same window any other interaction-derived count would, between sync runs, not an extra lag beyond that.

## Review findings addressed (round 2)

All eight findings from the first adversarial review are fixed on this branch; see Implementation Notes above for where. Summary:

1. **[BLOCKER]** uvicorn's access logger was still writing the raw query string independent of the handler fix — added `RequestQueryStringRedactionFilter`.
2. **[MAJOR]** `photos_enabled` moved to the already-cheap, already-awaited `/api/crm/config`; the blocking `/api/photos/stats` round trip is gone.
3. **[MAJOR]** The hand-rolled `HEAD` branch is gone; stacked `@router.get`/`@router.head` gives Starlette-native header parity.
4. **[MAJOR]** AC 6 now has a direct `loading="lazy"` assertion plus a 300-row scale test, both mutation-tested.
5. **[MINOR]** Same fix as 3 also removes this PR's duplicate-operation-id warning.
6. **[MINOR]** `docs/specs/technical/architecture.md` corrected; the "dead code" claim below is retracted.
7. **[MINOR]** `loadDetailProfilePhoto`/`loadRelationshipPartnerPhoto` now guard on `photosEnabled`; re-verified in Playwright.
8. **[NIT]** CSS comment corrected; redundant `renderAvatarPhotoImg()` call removed.

## Review findings addressed (round 3 — verification pass)

1. **[BLOCKER]** The round-2 redaction fix broke uvicorn's access log for every request with a query string: `uvicorn.logging.AccessFormatter.formatMessage` never reads `record.msg`/`getMessage()` — it unconditionally unpacks `record.args` as a 5-tuple `(client_addr, method, full_path, http_version, status_code)`. Clearing `record.args` to make `getMessage()` return the redacted line (correct for a *generic* `logging.Formatter`, wrong for this one) raised `ValueError: not enough values to unpack` inside the formatter — a 79-line traceback per request in place of an access line, worse than the leak it replaced. Fixed by redacting `args[2]` in place and keeping the 5-tuple shape; `record.msg` is now touched only as a fallback for a record that isn't shaped like uvicorn's own access-log call. New test runs a record through the *real* `uvicorn.logging.AccessFormatter` and asserts it doesn't raise and the marker is absent — mutation-tested against the old filter, which fails it. Re-verified end to end exactly as production runs (`log_level="info"`): both the marker and `"Logging error"` are absent from the log, and a normal access line is emitted.
2. **[MINOR]** Added tests pinning `GET /api/crm/config`'s `photos_enabled` against `settings.photos_enabled`, forcing both the `False` (this host's real state) and `True` cases so the field can't be silently hardcoded to its Pydantic default and still pass.

## Review focus

- **What "no console error" means for a missing thumbnail (AC 5).** Confirmed empirically (including through Playwright's `page.route()` mocks) that Chromium logs "Failed to load resource: ... 404" to the DevTools console for *any* request — image, `fetch()`, or otherwise — that completes with a non-2xx status, regardless of whether the page's JS handles it. That notice cannot be suppressed by application code short of never making the request, so `test_crm_avatar_photos_ui_browser.py` filters that one specific, unavoidable browser-level notice out of its console-error assertions (see `_app_console_errors()` and the module docstring there) and checks for zero *application* errors (uncaught exceptions via `page.on("pageerror")`, and anything else logged) instead. Still a judgment call worth a second look.
- ~~`api/routes/crm_models/` is unused dead code~~ — **retracted.** It has a live test dependency (`tests/test_family_matching.py`); see finding 6 and the corrected doc.
- Latency-test flakiness on this shared host (see Test evidence) has been present and worsening across all three review rounds due to increasing concurrent host load; still pre-existing and unrelated, reconfirmed each round against `origin/feat/crm-performance` without this branch's diff.

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -54,7 +54,7 @@ from starlette.middleware.gzip import GZipMiddleware
 from fastapi.staticfiles import StaticFiles
 from fastapi.responses import FileResponse
 from api.routes import search, ask, calendar, gmail, drive, people, chat, briefings, admin, conversations, memories, imessage, crm, slack, photos, reminders, scheduler, tasks, monarch, investments, jobs, perf, agents, vault, fitness, voice, agent_proxy, hermes_proxy, journal, journal_trends, journal_ingest
-from api.services.log_redaction import configure_telegram_log_redaction
+from api.services.log_redaction import configure_telegram_log_redaction, install_query_string_redaction_filter
 from api.services.route_timing import RouteTimingMiddleware
 from config.settings import settings
 
@@ -67,6 +67,12 @@ from config.settings import settings
 # API embeds the bot token in the URL) from ever logging at INFO here (#519).
 logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
 configure_telegram_log_redaction()
+# #904: uvicorn's own access logger writes every request's raw query string
+# to logs/server.log regardless of what a route handler logs -- this is the
+# only process that runs an HTTP server, so it's the only place this needs
+# installing (see install_query_string_redaction_filter()'s docstring for
+# why this must run at import time, not inside a route or startup event).
+install_query_string_redaction_filter()
 
 logger = logging.getLogger(__name__)
 
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -634,6 +634,14 @@ class CRMConfigResponse(BaseModel):
     partner_person_id: str = ""
     partner_name: str = ""
     family_default_selected_ids: list[str] = []
+    # Whether Apple Photos is configured at all (settings.photos_enabled is
+    # just Path.exists() on the library's database file -- cheap, unlike
+    # GET /api/photos/stats, which runs several full-library aggregate
+    # queries when Photos *is* configured). The client uses this to decide
+    # whether to request avatar/profile photos at all, so a Photos-less
+    # install never issues a request that would otherwise 503 (#875, #907
+    # review finding 2).
+    photos_enabled: bool = False
 
 
 def _load_family_default_selected_ids() -> list[str]:
@@ -677,6 +685,7 @@ def get_crm_config():
         partner_person_id=PARTNER_PERSON_ID,
         partner_name=_get_partner_name(),
         family_default_selected_ids=_load_family_default_selected_ids(),
+        photos_enabled=settings.photos_enabled,
     )
 
 
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -269,6 +269,9 @@ class PersonDetailResponse(BaseModel):
     slack_message_count: int = 0  # Actual Slack messages (not daily summaries)
     # Pre-computed Dunbar circle (0-6 for meaningful, 7 for peripheral)
     dunbar_circle: Optional[int] = None
+    # True when photo_count > 0, so the client knows to request an avatar
+    # instead of rendering initials -- no per-person probe needed.
+    has_profile_photo: bool = False
     # Related data
     source_entities: list[SourceEntityResponse] = []
     relationships: list[RelationshipResponse] = []
@@ -605,6 +608,7 @@ def _person_to_detail_response(
         message_count=person.message_count,
         slack_message_count=person.slack_message_count,
         dunbar_circle=person.dunbar_circle,
+        has_profile_photo=person.photo_count > 0,
     )
 
     if include_related:
@@ -851,8 +855,11 @@ def list_people(
         has_more=has_more,
     )
 
+    # #904: never log the raw search text or other filter values here -- the
+    # CRM search box's contents (names, partial emails) are personal data.
+    # Counts and timing only.
     elapsed = (time.time() - start_time) * 1000
-    logger.info(f"list_people(q={q}, category={category}, limit={limit}) took {elapsed:.1f}ms ({total} total, {len(people)} returned)")
+    logger.info(f"list_people took {elapsed:.1f}ms (limit={limit}, {total} total, {len(people)} returned)")
 
     return result
 
--- a/api/routes/photos.py
+++ b/api/routes/photos.py
@@ -7,8 +7,8 @@ and syncing to LifeOS CRM.
 from pathlib import Path
 from typing import Optional
 
-from fastapi import APIRouter, HTTPException, Query, Request
-from fastapi.responses import FileResponse, JSONResponse, Response
+from fastapi import APIRouter, HTTPException, Query
+from fastapi.responses import FileResponse, JSONResponse
 from pydantic import BaseModel
 
 from api.routes.crm import _mutation_lock
@@ -408,15 +408,23 @@ def get_photo_thumbnail(uuid: str):
     )
 
 
-@router.api_route("/profile/{person_id}", methods=["GET", "HEAD"])
-def get_profile_photo(person_id: str, request: Request):
+@router.get("/profile/{person_id}")
+@router.head("/profile/{person_id}", include_in_schema=False)
+def get_profile_photo(person_id: str):
     """
     Get a profile photo thumbnail for a person.
 
     Returns the most recent photo of the person that has a thumbnail available.
-    Use this for avatars/profile pictures in the UI. Also accepts HEAD, returning
-    the same status and headers with no body, so clients can check availability
-    without downloading the image.
+    Use this for avatars/profile pictures in the UI. Also accepts HEAD --
+    stacking a `@router.head` registration on the same function (rather than
+    branching on the request method ourselves) lets Starlette's `FileResponse`
+    handle it natively, which gives a `HEAD` response the exact same headers
+    (`etag`, `last-modified`, `content-length`, `accept-ranges`) a `GET` would
+    carry, just without the body -- a hand-rolled `HEAD` branch returning a
+    bare `Response` could not match that (#907 review finding 3). It also
+    avoids FastAPI's "duplicate operation ID" warning that a single
+    `@router.api_route(methods=["GET", "HEAD"])` registration triggers
+    (finding 5), since GET and HEAD are now separate routes.
     """
     _check_photos_enabled()
 
@@ -439,12 +447,6 @@ def get_profile_photo(person_id: str, request: Request):
                 media_type = "image/jpeg"
                 if suffix == ".png":
                     media_type = "image/png"
-
-                if request.method == "HEAD":
-                    return Response(
-                        media_type=media_type,
-                        headers={"Cache-Control": "public, max-age=3600"}  # Cache for 1 hour
-                    )
 
                 return FileResponse(
                     thumb_path,
--- a/api/routes/photos.py
+++ b/api/routes/photos.py
@@ -7,8 +7,8 @@ and syncing to LifeOS CRM.
 from pathlib import Path
 from typing import Optional
 
-from fastapi import APIRouter, HTTPException, Query
-from fastapi.responses import FileResponse, JSONResponse
+from fastapi import APIRouter, HTTPException, Query, Request
+from fastapi.responses import FileResponse, JSONResponse, Response
 from pydantic import BaseModel
 
 from api.routes.crm import _mutation_lock
@@ -408,13 +408,15 @@ def get_photo_thumbnail(uuid: str):
     )
 
 
-@router.get("/profile/{person_id}")
-def get_profile_photo(person_id: str):
+@router.api_route("/profile/{person_id}", methods=["GET", "HEAD"])
+def get_profile_photo(person_id: str, request: Request):
     """
     Get a profile photo thumbnail for a person.
 
     Returns the most recent photo of the person that has a thumbnail available.
-    Use this for avatars/profile pictures in the UI.
+    Use this for avatars/profile pictures in the UI. Also accepts HEAD, returning
+    the same status and headers with no body, so clients can check availability
+    without downloading the image.
     """
     _check_photos_enabled()
 
@@ -438,16 +440,24 @@ def get_profile_photo(person_id: str):
                 if suffix == ".png":
                     media_type = "image/png"
 
+                if request.method == "HEAD":
+                    return Response(
+                        media_type=media_type,
+                        headers={"Cache-Control": "public, max-age=3600"}  # Cache for 1 hour
+                    )
+
                 return FileResponse(
                     thumb_path,
                     media_type=media_type,
                     headers={"Cache-Control": "public, max-age=3600"}  # Cache for 1 hour
                 )
 
-    # No photo available
+    # No photo available. Cache the miss too, so repeated list renders for
+    # people without a photo don't re-request it every time.
     raise HTTPException(
         status_code=404,
-        detail=f"No profile photo available for person {person_id}"
+        detail=f"No profile photo available for person {person_id}",
+        headers={"Cache-Control": "public, max-age=3600"},
     )
 
 
--- a/api/services/log_redaction.py
+++ b/api/services/log_redaction.py
@@ -1,4 +1,4 @@
-"""Redaction of Telegram bot tokens from log records (#519).
+"""Redaction of secrets and personal data from log records (#519, #904).
 
 The Telegram Bot API embeds the bot token directly in the request path
 (``https://api.telegram.org/bot<digits>:<secret>/<method>``). ``httpx``'s
@@ -22,6 +22,18 @@ Call ``configure_telegram_log_redaction()`` once per process, after the
 process's own root logging setup (``logging.basicConfig`` or equivalent) has
 run — it needs at least one handler already attached to root for the filter
 to have something to attach to.
+
+A third, unrelated filter lives here too for the same reason: uvicorn's own
+access logger (``uvicorn.access``) writes the full request line — path *and*
+query string — for every request, at INFO, regardless of what any route
+handler logs. `#904 <https://github.com/nbramia/LifeOS/issues/904>`_ fixed
+the CRM people-list handler's own log line, but the raw text typed into the
+CRM search box (``GET /api/crm/people?q=<text>``) still reached
+``logs/server.log`` this way — a gap the handler-level fix structurally
+cannot close, since it only ever sees the parsed query parameter, never the
+access logger's own line. ``RequestQueryStringRedactionFilter`` /
+``install_query_string_redaction_filter()`` strip everything from the first
+``?`` onward in that line, for every route, not just this one.
 """
 from __future__ import annotations
 
@@ -95,3 +107,60 @@ def configure_telegram_log_redaction() -> None:
     least one handler)."""
     silence_noisy_http_loggers()
     install_telegram_redaction_filter()
+
+
+# Matches the query string (and the `?` that introduces it) in a request
+# line, e.g. "GET /api/crm/people?q=alex&limit=5 HTTP/1.1" -- everything from
+# `?` up to the next whitespace (the URL itself never contains a literal,
+# unencoded space or `?`).
+QUERY_STRING_PATTERN = re.compile(r"\?\S*")
+
+REDACTED_QUERY_STRING = "?<redacted>"
+
+
+class RequestQueryStringRedactionFilter(logging.Filter):
+    """Strips the query string from uvicorn's access-log request line.
+
+    uvicorn's access logger (``uvicorn.access``) logs the full request line
+    -- including the raw query string -- for every request, independent of
+    anything a route handler itself logs. That's how `q=<search text>` from
+    the CRM people list's search box (personal data: names, partial emails)
+    reaches `logs/server.log` even after the handler's own log line (#904)
+    stops naming it. Never drops a record, only redacts in place.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        try:
+            message = record.getMessage()
+        except Exception:
+            # Don't let a malformed record (bad % args, etc.) break logging.
+            return True
+
+        if "?" in message:
+            record.msg = QUERY_STRING_PATTERN.sub(REDACTED_QUERY_STRING, message, count=1)
+            record.args = ()
+
+        return True
+
+
+def install_query_string_redaction_filter() -> None:
+    """Install `RequestQueryStringRedactionFilter` directly on the
+    `uvicorn.access` logger. Idempotent.
+
+    Not attached via the root logger's handlers (the pattern the Telegram
+    filter above uses): uvicorn's default logging config gives
+    `uvicorn.access` `propagate=False` and its own dedicated handler, so a
+    record logged there never reaches root's handlers at all -- the filter
+    has to sit on the logger itself to see it.
+
+    Call this after `uvicorn.access` has already been configured (uvicorn's
+    own `Config.configure_logging()` runs `logging.config.dictConfig()`,
+    which resets that logger's filters). In the normal startup path
+    (`uvicorn.run("api.main:app", ...)`, as `scripts/server.sh` uses), that
+    dictConfig call happens during `Config.__init__`, strictly before the
+    app string is imported and this module's own `uvicorn.access` logger's
+    filters would be reset -- so this is safe to call at import time here.
+    """
+    access_logger = logging.getLogger("uvicorn.access")
+    if not any(isinstance(f, RequestQueryStringRedactionFilter) for f in access_logger.filters):
+        access_logger.addFilter(RequestQueryStringRedactionFilter())
--- a/api/services/log_redaction.py
+++ b/api/services/log_redaction.py
@@ -127,9 +127,36 @@ class RequestQueryStringRedactionFilter(logging.Filter):
     the CRM people list's search box (personal data: names, partial emails)
     reaches `logs/server.log` even after the handler's own log line (#904)
     stops naming it. Never drops a record, only redacts in place.
+
+    #907 review round 2: ``uvicorn.logging.AccessFormatter.formatMessage``
+    does not call ``record.getMessage()`` at all -- it unconditionally
+    unpacks ``record.args`` as a 5-tuple
+    (``client_addr, method, full_path, http_version, status_code``) and
+    rebuilds the request line from those fields. An earlier version of this
+    filter rewrote ``record.msg`` to the redacted line and cleared
+    ``record.args`` to make ``getMessage()`` return the safe string -- which
+    is exactly right for a *generic* ``logging.Formatter``, but broke
+    uvicorn's formatter: unpacking ``()`` into five names raises
+    ``ValueError: not enough values to unpack``, and that exception, not an
+    access line, is what actually reached ``logs/server.log`` (as a 79-line
+    "Logging error" traceback per request). The fix has to redact
+    ``args[2]`` (the piece the formatter actually reads) in place and leave
+    the 5-tuple shape intact. The ``record.msg``-based rewrite is kept only
+    as a fallback for a record that isn't shaped like uvicorn's own
+    access-log call.
     """
 
     def filter(self, record: logging.LogRecord) -> bool:
+        args = record.args
+        if isinstance(args, tuple) and len(args) == 5 and isinstance(args[2], str) and "?" in args[2]:
+            path, _, _query = args[2].partition("?")
+            record.args = (args[0], args[1], path + REDACTED_QUERY_STRING, args[3], args[4])
+            return True
+
+        # Fallback: a record not shaped like uvicorn's access-log call (that
+        # call always hits the branch above). Kept belt-and-braces for any
+        # other caller of this filter, formatted the ordinary way via
+        # record.getMessage()/record.msg.
         try:
             message = record.getMessage()
         except Exception:
--- a/docs/specs/product/api-crm.md
+++ b/docs/specs/product/api-crm.md
@@ -45,6 +45,8 @@ List/search people with filters.
 - `limit` (int): Results per page (default: 50)
 - `offset` (int): Pagination offset
 
+Each returned person carries `has_profile_photo` (bool, from `photo_count > 0`) alongside the usual fields, so a client can request `GET /api/photos/profile/{id}` only for people flagged true instead of probing everyone (#875). `GET /api/crm/people/{id}` carries the same field for consistency.
+
 ### GET /api/crm/people/{id}
 
 Get person detail with source entities.
--- a/docs/specs/product/api-reference.md
+++ b/docs/specs/product/api-reference.md
@@ -742,7 +742,7 @@ Get a thumbnail image for a photo by UUID. Returns the image file from the Photo
 
 ### GET /api/photos/profile/{person_id}
 
-Get a profile photo thumbnail for a person. Returns the most recent available photo thumbnail for use as an avatar.
+Get a profile photo thumbnail for a person. Returns the most recent available photo thumbnail for use as an avatar. Also accepts `HEAD` (same status as `GET`, no body), so a client can check availability without downloading the image. Both the `200` and `404` responses carry `Cache-Control: public, max-age=3600`, so a client that already knows a person has no reachable photo doesn't need to ask again within the hour.
 
 ### GET /api/photos/open/{uuid}
 
--- a/docs/specs/product/crm-people.md
+++ b/docs/specs/product/crm-people.md
@@ -184,6 +184,7 @@ Behavior:
 - Real-time search (< 300 ms), debounced; a superseded query's response is discarded so the list always reflects what's currently typed, even if an earlier search's response arrives later (#874).
 - Category tabs (Work / Personal / Family).
 - Cards show avatar, name, company/category, interaction count, strength bar.
+- A card's avatar shows the person's profile photo only when the list response flags them as having one (`has_profile_photo`); nothing else is ever probed. The image loads lazily and falls back to initials with no error if the photo turns out to be unavailable (#875).
 - Cards sorted by relationship strength by default.
 - Infinite scroll loads more people.
 - Mobile-responsive layout.
--- a/docs/specs/technical/architecture.md
+++ b/docs/specs/technical/architecture.md
@@ -23,11 +23,11 @@ api/
 │   ├── calendar.py            # Calendar integration
 │   ├── chat.py                # Streaming chat with agentic pipeline
 │   ├── conversations.py       # Conversation history
-│   ├── crm.py                 # CRM endpoints (~5,100 LOC)
-│   ├── crm_models/            # CRM Pydantic models and helpers
+│   ├── crm.py                 # CRM endpoints + models (~5,100 LOC) -- see below
+│   ├── crm_models/            # NOT wired into the app -- see "CRM Models Package"
 │   │   ├── __init__.py        # Re-exports models and utils
-│   │   ├── models.py          # All Pydantic models (~600 LOC)
-│   │   └── _utils.py          # Shared helper functions
+│   │   ├── models.py          # Unused parallel Pydantic models (~600 LOC)
+│   │   └── _utils.py          # is_family_member() is kept in sync by a test
 │   ├── _proxy.py              # Shared reverse-proxy router factory (agent/hermes)
 │   ├── agent_proxy.py         # Agent text-backend reverse proxy
 │   ├── hermes_proxy.py        # Hermes text-backend reverse proxy + persona envelope
@@ -201,39 +201,29 @@ db_path = get_crm_db_path()  # Returns "data/crm.db"
 
 ## CRM Models Package (api/routes/crm_models/)
 
-The CRM models package consolidates Pydantic models and utilities for the CRM API.
+**Not currently wired into the running API.** The Pydantic models the CRM API
+actually serves — `PersonDetailResponse` (including `has_profile_photo`,
+added in #875), `PersonListResponse`, `TimelineItem`, and the rest — are
+defined directly inside `api/routes/crm.py`, alongside the route handlers
+that use them; `api/main.py` mounts only `crm.router`. `crm_models/` is a
+separate, parallel module tree (`models.py`, `_utils.py`, `__init__.py`)
+that predates or anticipated a split of `crm.py` into smaller files but was
+never finished wiring up — nothing in `api/main.py` or `api/routes/crm.py`
+imports from it.
 
-### Importing Models
+It isn't dead code, though: `api.routes.crm_models._utils.is_family_member`
+is a second, independent implementation of the family-matching logic that
+`api.services.person_entity._is_family_member` also implements for the live
+path, and `tests/test_family_matching.py` exercises both directly so the two
+can't silently drift apart. Anyone touching family-matching rules needs to
+update both. If `crm_models/` is ever finished and wired up (or removed),
+that test and this note both need to move with it.
+
+### Importing the family-matching utility
 
 ```python
-from api.routes.crm_models import (
-    PersonDetailResponse,
-    TimelineItem,
-    NetworkGraphResponse,
-)
+from api.routes.crm_models._utils import is_family_member
 ```
-
-### Importing Utilities
-
-```python
-from api.routes.crm_models import (
-    compute_person_category,
-    person_to_detail_response,
-    MY_PERSON_ID,
-)
-```
-
-### Models Reference
-
-| Category | Models |
-|----------|--------|
-| Person | PersonDetailResponse, PersonListResponse, PersonUpdateRequest, PersonMergeRequest, PersonSplitRequest |
-| Timeline | TimelineItem, TimelineResponse, AggregatedTimelineResponse |
-| Relationships | RelationshipResponse, RelationshipDetailResponse, ConnectionResponse |
-| Network | NetworkNode, NetworkEdge, NetworkGraphResponse |
-| Facts | PersonFactResponse, PersonFactsResponse, FactExtractionResponse |
-| Dashboard | MeStatsResponse, MeInteractionsResponse, FamilyInteractionsResponse |
-| Health | SyncHealthResponse, ReviewQueueResponse |
 
 ---
 
@@ -271,12 +261,13 @@ from api.routes import (
 
 ### CRM Models
 
+`api/routes/crm.py` defines and uses its own response models directly — they
+aren't re-exported for other modules to import. The one thing from
+`api/routes/crm_models/` that another module does import (see [CRM Models
+Package](#crm-models-package-apiroutescrm_models) above):
+
 ```python
-from api.routes.crm_models import (
-    PersonDetailResponse, TimelineItem, NetworkGraphResponse,
-    compute_person_category, person_to_detail_response,
-    MY_PERSON_ID, FAMILY_EXACT_NAMES,
-)
+from api.routes.crm_models._utils import is_family_member
 ```
 
 ---
--- a/tests/test_crm_api.py
+++ b/tests/test_crm_api.py
@@ -47,16 +47,22 @@ class TestCRMConfig:
         `photos_enabled=settings.photos_enabled` from get_crm_config()
         left every other test in this repo green, because
         CRMConfigResponse.photos_enabled defaults to False and this host
-        has no Photos library configured (settings.photos_enabled is also
-        False here) -- the mutation and the real answer coincided. This
-        pins the False case against the actual settings value; the next
-        test below forces the True case so the field can't be silently
-        hardcoded to False and still pass both."""
-        from config.settings import settings
+        (at authoring time) had no Photos library configured -- the
+        mutation and the real answer coincided. Forces False explicitly
+        (rather than asserting against whatever settings.photos_enabled
+        happens to be on the machine running the suite, e.g. a Mac with a
+        real Photos library configured) so this pins the False case
+        unconditionally; the next test below forces the True case, so the
+        field can't be silently hardcoded to one value and still pass
+        both."""
+        from unittest.mock import patch, PropertyMock
+        from config.settings import Settings
 
-        response = client.get("/api/crm/config")
+        with patch.object(Settings, "photos_enabled", new_callable=PropertyMock, return_value=False):
+            response = client.get("/api/crm/config")
+
         assert response.status_code == 200
-        assert response.json()["photos_enabled"] == settings.photos_enabled is False
+        assert response.json()["photos_enabled"] is False
 
     def test_photos_enabled_reflects_settings_when_true(self, client):
         from unittest.mock import patch, PropertyMock
--- a/tests/test_crm_api.py
+++ b/tests/test_crm_api.py
@@ -37,6 +37,38 @@ def sample_person_id(client):
     pytest.skip("No people in database to test")
 
 
+class TestCRMConfig:
+    """Tests for GET /api/crm/config."""
+
+    def test_photos_enabled_reflects_settings_when_false(self, client):
+        """#907 review round 2, finding 2: the server side of #875's
+        photos_enabled field (web/crm.html's loadCRMConfig() reads it to
+        decide whether to ever request an avatar) had no test -- deleting
+        `photos_enabled=settings.photos_enabled` from get_crm_config()
+        left every other test in this repo green, because
+        CRMConfigResponse.photos_enabled defaults to False and this host
+        has no Photos library configured (settings.photos_enabled is also
+        False here) -- the mutation and the real answer coincided. This
+        pins the False case against the actual settings value; the next
+        test below forces the True case so the field can't be silently
+        hardcoded to False and still pass both."""
+        from config.settings import settings
+
+        response = client.get("/api/crm/config")
+        assert response.status_code == 200
+        assert response.json()["photos_enabled"] == settings.photos_enabled is False
+
+    def test_photos_enabled_reflects_settings_when_true(self, client):
+        from unittest.mock import patch, PropertyMock
+        from config.settings import Settings
+
+        with patch.object(Settings, "photos_enabled", new_callable=PropertyMock, return_value=True):
+            response = client.get("/api/crm/config")
+
+        assert response.status_code == 200
+        assert response.json()["photos_enabled"] is True
+
+
 class TestPersonEndpoints:
     """Tests for /api/crm/people endpoints."""
 
--- a/tests/test_crm_api.py
+++ b/tests/test_crm_api.py
@@ -8,6 +8,7 @@ Tests are organized by endpoint group:
 - Sync health and status
 - Statistics
 """
+import logging
 import sqlite3
 import time
 from pathlib import Path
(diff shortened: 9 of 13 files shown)


nbramia:
## Adversarial review of #907 by Opus

**Verdict:** BLOCK

The #875 half is solid and the browser evidence holds up: on staging with 219 real people and Photos off, the list issues exactly zero `/api/photos/profile/*` requests and zero console errors, and the stubbed photos-enabled path requests only flagged people, only `GET`, and never retries a miss. But #904 is not actually fixed — the raw search text still lands in `logs/server.log` — and two other defects are reproducible.

1. **[BLOCKER] #904's acceptance criterion is still false: the raw search text is still written to `logs/server.log`.** — `api/routes/crm.py:790`, `scripts/server.sh:251` — The handler's own line is fixed, but the client sends the search box contents as `q=` (`web/crm.html:9396`) and uvicorn's access log writes the full query string. Reproduced under the exact configuration production uses (`uvicorn.run(..., log_level='info')`, systemd appending stdout to `logs/server.log`): a request for `?q=ZZQMARKER7788` produced `INFO: 127.0.0.1:54644 - "GET /api/crm/people?q=ZZQMARKER7788&sort=strength&limit=5 HTTP/1.1" 200 OK` in the log file. The new `caplog` test structurally cannot catch this — `TestClient` never goes through uvicorn's access logger, and the test raises the level only for `api.routes.crm`. So the issue's stated problem ("anything typed into the CRM search box … lands in the log files") remains true after this PR, while `Closes #904` marks it done. I'm aware the issue's Scope line says "the one log line in the list handler"; the AC above it says "no log record shall contain `<text>`", and the privacy rule in AGENTS.md is not scoped. Fix: pass `access_log=False` in `scripts/server.sh` (route timings are already covered by the timing middleware), or add a query-string redaction filter to `api/services/log_redaction.py` — the module that exists precisely because a secret reached `logs/server.log` this way — and assert on the filter, not on `caplog`. If that is genuinely out of scope, drop `Closes #904` and amend the issue rather than closing it.

2. **[MAJOR] Every CRM page's first render is now blocked on `GET /api/photos/stats`, which is not a cheap endpoint on an install that has Photos.** — `web/crm.html:9276` (`await Promise.all([loadCRMConfig(), loadPhotosEnabled()])`) — Measured with the server-free harness: with the stats response delayed 8s, the first `.person-card` appeared at **8.20s**; the same page with an instant/failing stats response renders at 0.16–0.27s. `init()` awaits this before every path, so `/birthdays` and `/relationship` — which render no list avatars at all — pay it too. When Photos *is* configured, `get_photos_stats()` runs `ApplePhotosReader.get_stats()` (`api/services/apple_photos.py:391-432`): four full-library aggregates, including a `GROUP BY a.Z_PK HAVING COUNT(DISTINCT p.Z_PK) >= 2` over `ZASSET ⋈ ZDETECTEDFACE ⋈ ZPERSON`. Adding an unbounded, un-timed-out dependency to the page-load critical path is the opposite of what this milestone is for. The value wanted is one boolean that is just `Path(...).exists()` (`config/settings.py:1553`). Fix: put `photos_enabled` on `/api/crm/config` (already awaited, already cheap) and delete `loadPhotosEnabled()` and its round-trip entirely.

3. **[MAJOR] The hand-rolled `HEAD` branch is unnecessary and breaks header parity with `GET`, which is the thing the criterion asks for.** — `api/routes/photos.py:443-447` — Probed over real HTTP against the app with Photos forced on: `GET` returns `content-length: 504`, `accept-ranges: bytes`, `last-modified`, `etag`; `HEAD` returns `content-length: 0` and **no** `etag`, `last-modified`, or `accept-ranges`. RFC 9110 §9.3.2 wants the same header fields, and a client using `HEAD` to decide about a cached avatar can't revalidate without a validator. Starlette's `FileResponse` already implements `send_header_only` for `HEAD` (`starlette/responses.py:344`) — verified that returning the `FileResponse` unconditionally yields `HEAD` → status 200, body length 0, and *identical* headers to `GET` including `etag` and the real `content-length`. Delete the `if request.method == "HEAD"` block and the now-unused `request: Request` parameter. The PR's test passes only because it compares status and `cache-control` and nothing else.

4. **[MAJOR] AC 6 ("images load lazily") has no test.** — `tests/test_crm_avatar_photos_ui_browser.py` — Mutation check: deleting `loading="lazy"` from `renderAvatarPhotoImg()` leaves all 4 browser tests green (`4 passed`). The only mechanism the PR claims for this criterion is unpinned. Add `expect(img).to_have_attribute("loading", "lazy")`, and ideally a many-row case asserting requests ≪ rows — I measured 300 flagged rows → 55 image requests on first paint, 154 after a scroll, which is exactly the behaviour worth freezing.

5. **[MINOR] The duplicate-operation-ID warning is avoidable; the suggested alternative in the PR body isn't the one that works.** — `api/routes/photos.py:411` — Confirmed the PR introduces a second such warning (`get_profile_photo_api_photos_profile__person_id__head`, alongside the pre-existing `voice_proxy` one). Tested both alternatives: `operation_id=` does **not** silence it; stacking `@router.get("/profile/{person_id}")` and `@router.head("/profile/{person_id}", include_in_schema=False)` on the same function **does** (zero warnings), and combined with finding 3 it also gives full `GET`/`HEAD` header parity for free.

6. **[MINOR] The "Review focus" claim that `api/routes/crm_models/` is unused dead code is wrong.** — `tests/test_family_matching.py:16` imports `api.routes.crm_models._utils`, and `docs/specs/technical/architecture.md:201-235,268-277` documents that package as the CRM models package and `PersonDetailResponse`'s home. Nothing to change in this PR, but the follow-up is "reconcile the doc with reality", not "delete dead code" — and that doc is now measurably wrong about where `has_profile_photo` lives, which is the kind of staleness the docs rule targets.

7. **[MINOR] "Person overview … 0 console errors" doesn't reproduce.** — `web/crm.html:13304` (`loadDetailProfilePhoto`) — On the photos-disabled staging install, clicking five people produced four `GET /api/photos/profile/{id}` → `503` responses and four console errors. That loader is untouched by this PR, so it is pre-existing, not a regression — but the manual-verification section states otherwise, and now that `photosEnabled` exists a two-line guard in `loadDetailProfilePhoto` (and `loadRelationshipPartnerPhoto`) would finish the job the list half already does.

8. **[NIT] `has-photo` is inert on list rows.** — `web/crm.html:568,9526` — `.person-avatar.has-photo { color: transparent }` never applies there because `getAvatarStyle()` writes an inline `color`, which wins over the class rule (verified: computed `color` on a flagged, still-loading avatar is `rgb(26,26,46)`, initials visible). That's *lucky* — it's why a hung photo request degrades gracefully instead of leaving a blank circle, which I checked with a never-fulfilled route and screenshotted. But the CSS comment describes a mechanism that isn't doing the work, and the class ends up serving only as a test hook. Also `avatarPhotoClass()` calls `renderAvatarPhotoImg()` a second time per row purely for truthiness.

Noted, not a finding: caching the `404` for an hour means a person who gains a photo keeps rendering initials for up to an hour after a sync. The AC asked for it, and the in-session `profilePhotoCache` already does the same thing, so it's a deliberate trade rather than a bug.

### What I verified and how

- `gh pr diff 907`, read line by line; issues #875 and #904 read in full including comments.
- `pytest tests/test_photos_profile_api.py tests/test_crm_avatar_photos_ui_browser.py -q` → `9 passed`. `pytest tests/test_crm_request_hygiene_ui_browser.py tests/test_family_matching.py tests/test_crm_api.py -q` → `2 failed, 55 passed, 1 skipped`; the two failures are `test_get_people_list_warm_latency_large_db{,_limit_300}`, and `origin/feat/crm-performance` fails the same assertion (`137.9 < 100`) on this loaded host — pre-existing, as the PR says.
- **Mutation test of the #904 test:** restored the old `list_people(q={q})` line → `test_list_people_never_logs_the_search_query` fails with the marker in `caplog.text`. The test is real for the handler line. Reverted.
- **Mutation test of AC 6:** removed `loading="lazy"` → `tests/test_crm_avatar_photos_ui_browser.py` still `4 passed`. Reverted.
- **#904 end to end, staging on :8028** (`log_level='warning'`): `?q=ZZQMARKER7788&category=work` → 0 occurrences in the log; only `list_people took 1276.5ms (limit=5, 0 total, 0 returned)`. **Then on :8030 with `log_level='info'`, matching `scripts/server.sh:251`:** the same request wrote the full query string to the log (finding 1).
- **Frontend, staging :8028 with the real-data copy, Photos disabled**, Playwright: `/me` → 219 cards, every avatar `person-avatar no-photo`, 0 ``, **the only `/api/photos/*` request was one `GET /api/photos/stats`**, 0 console errors, 0 page errors. Infinite scroll, search (`sar` → 150 cards, clear → 219), person click, tab switches, back/forward: 0 console errors. `/family`, `/birthdays`, `/crm`: 0 errors. `/relationship`: 2 errors from `/api/crm/people//timeline` 404 — staging's CRM config has no `partner_person_id`, unrelated, matches the PR's account.
- **Photos-enabled path, server-free harness** (serves `web/`, stubs `/api/**` like the PR's test): 3 people (200 / 404 / unflagged) → 2 requests, all `GET`, correct classes; 20 flagged all-404 → 20 requests, all flip to `no-photo`, and **3 forced `renderPeopleList()` re-renders issued 0 additional requests** (the no-retry cache works); 300 flagged rows → 55 requests at first paint, 154 after scrolling (lazy loading works); `photos_enabled: false` → 0 requests.
- **`loadPhotosEnabled()` failure paths:** stats `500` → 0 photo requests, page renders in 0.27s; stats network-abort → 0 photo requests, renders in 0.16s; **stats delayed 8s → first person card at 8.20s** (finding 2).
- **Hung photo requests** (route never fulfilled): initials stay visible, avatars are not blank — screenshotted (finding 8).
- **`HEAD`/`GET` over real HTTP** against the app with `Settings.photos_enabled` forced true and the store stubbed: raw-socket `HEAD` on a hit → 200, 0 body bytes, `content-length: 0`, no `etag`/`last-modified`; `GET` → `content-length: 504` + `etag` + `last-modified` + `accept-ranges`. `HEAD` on a miss → 404, 0 body bytes, `content-length: 55` matching `GET`, same `cache-control` (the transport suppresses the body correctly there). Then reproduced the no-manual-branch variant in isolation: `HEAD` → 200, body 0, headers identical to `GET`.
- **Duplicate operation IDs:** `app.openapi()` under `warnings.catch_warnings` → 2 warnings (this route + pre-existing `voice_proxy`); `operation_id=` variant → still warns; stacked `@get`/`@head(include_in_schema=False)` variant → 0 warnings, headers identical to `GET`.
- **Response shape:** dumped list and detail key sets from an in-process `TestClient` on `origin/feat/crm-performance` and on the PR head against the same real-data copy → `list_keys added: ['has_profile_photo'] removed: []`, `detail_keys added: ['has_profile_photo'] removed: []`. Purely additive. MCP is unaffected: `_format_response` in `mcp_server.py` reads named keys with `.get()` and never enumerates person fields.
- **Docs:** `api-crm.md`, `api-reference.md`, `crm-people.md` accurately describe the new behaviour and bump `Last Updated`. `docs/specs/technical/architecture.md` is the stale one (finding 6). No task lists, synthetic data only.
- **Privacy:** PR body, tests, and fixtures contain no real names, emails, or DB rows; the browser fixtures are obviously synthetic. Verified no `.env` or `data/` symlink left in the tree; both staging instances killed and their ports released.

### Not verified

- Rendering against a genuine Apple Photos library — this host has none (`photos_enabled: false`). The photos-enabled path was exercised with stubbed 200/404 responses and, at the route level, with a real file served through `FileResponse` on a locally-forced-enabled app.
- The wall-clock cost of `/api/photos/stats` on a real large library. Finding 2's *blocking* behaviour is measured; its magnitude is inferred from the query shape in `get_stats()`.
- Whether `photo_count` can lag actual photo interactions (it's refreshed by the nightly stats recompute in `person_stats.py`), which would leave a person who genuinely has photos rendering initials in the list while the detail header still shows one. Worth a sentence from the author.
- The full unit suite — targeted files only, per the standing brief. The PR reports a green pre-push gate.

nbramia:
## Response to adversarial review (Opus)

All eight findings fixed on this branch (rebased onto `feat/crm-performance` @ `d598202`, pushed as `a68cb32`). Per-finding:

1. **[BLOCKER] #904 — uvicorn access log still leaked the query string.** Confirmed and fixed. Added `RequestQueryStringRedactionFilter` in `api/services/log_redaction.py`, installed directly on the `uvicorn.access` logger (it has `propagate=False` with its own handler, so a root-handler filter — the existing Telegram-token pattern — can't see it) from `api/main.py` at import time. New test builds a `LogRecord` the way uvicorn's own `access_logger.info('%s - "%s %s HTTP/%s" %d', ...)` call does (`tests/test_log_redaction.py::TestRequestQueryStringRedactionFilter`) and asserts the marker is gone. Manually reproduced end to end against the exact production config (`uvicorn.run("api.main:app", ..., log_level="info")`, matching `scripts/server.sh`): `?q=ZZQMARKER7788&sort=strength&limit=5` now produces `"GET /api/crm/people?<redacted> HTTP/1.1" 200 OK` in the log, marker absent. Kept the handler's own `caplog` test too (it's still a real, separate guard against a future regression there).

2. **[MAJOR] `/api/photos/stats` round trip blocking every page's first render.** Fixed. `photos_enabled` now rides on `GET /api/crm/config` (a cheap `Path.exists()`, and a call `init()` already awaits before anything else). `loadPhotosEnabled()` and its separate request are deleted entirely.

3+5. **[MAJOR/MINOR] Hand-rolled `HEAD` branch, no header parity, duplicate-operation-id warning.** Fixed exactly as suggested: stacked `@router.get(...)` / `@router.head(..., include_in_schema=False)` on `get_profile_photo`, letting Starlette's `FileResponse` handle `HEAD` natively. New test (`test_head_returns_200_with_no_body_and_full_header_parity_with_get`) asserts `content-length`, `content-type`, `cache-control`, `etag`, and `last-modified` are all identical between `HEAD` and `GET`. Confirmed via `app.openapi()` under `warnings.catch_warnings()`: only the pre-existing `voice_proxy` warning remains, this route's own is gone.

4. **[MAJOR] AC 6 unpinned.** Fixed. Added a direct `expect(img).to_have_attribute("loading", "lazy")` and a new `test_300_flagged_rows_request_far_fewer_images_at_first_paint` (300 flagged rows, all returning 200). Mutation-tested both ways: with `loading="lazy"` removed, both assertions fail (300/300 requests instead of ~52/300, matching your own ~55/300 measurement closely).

6. **[MINOR] Architecture doc wrong, dead-code claim wrong.** You're right on both counts. Rewrote `docs/specs/technical/architecture.md`'s "CRM Models Package" section and its "Public API Imports" example to state reality: `PersonDetailResponse` (and the rest) actually served live in `api/routes/crm.py`; `api/routes/crm_models/` is a separate, not-wired-in module whose only live consumer is `tests/test_family_matching.py`, which keeps `is_family_member()` there in sync with the real implementation in `api/services/person_entity.py`. Retracted the "dead code" claim in the PR body (see the struck-through line there).

7. **[MINOR] Detail header and relationship dashboard still 503 on Photos-less installs.** Fixed. `loadDetailProfilePhoto()` and `loadRelationshipPartnerPhoto()` now check `photosEnabled` before doing anything. Re-verified in Playwright on staging: selecting a real, `has_profile_photo: true` person now issues **zero** photo requests from the detail header (previously 503 + console error every time), and the relationship dashboard load does the same. The PR's "0 console errors" manual-verification claim now actually holds for the pages it was already claiming to cover — thank you for the reproduction, it was a real gap.

8. **[NIT] Inert CSS comment, redundant `renderAvatarPhotoImg()` call.** Fixed. `renderPeopleList()` now computes the img markup once per row and reuses it for both the class and the content; the now-redundant `avatarPhotoClass()` helper is deleted. Added an accurate comment above `.person-avatar.has-photo` explaining it's inert there specifically (inline `color` from `getAvatarStyle()` wins) and why that's actually the desired degrade-gracefully behavior, while still applying where nothing sets an inline color (e.g. `.detail-avatar.has-photo`).

**Your `photo_count`-lag question:** answered in the PR body — it's refreshed synchronously at the end of each Photos sync run for exactly the people who gained interactions in that run, not by a separate/delayed nightly job, so it doesn't lag behind the interactions it's derived from beyond the normal sync cadence itself.

**Verification this round:** targeted files green (41 passed across `test_photos_profile_api.py` + `test_crm_avatar_photos_ui_browser.py` + `test_crm_request_hygiene_ui_browser.py` + `test_log_redaction.py` + `test_family_matching.py`); full unit suite green under the shared push lock (5702 passed, 8 skipped — only the pre-existing `voice_proxy` duplicate-operation-id warning left, this PR's own is gone); pre-push gate green on the rebased/force-pushed branch (5695 passed unit, 360 passed browser). `test_crm_api.py`'s pre-existing latency-threshold tests are flakier than last round under heavier concurrent host load (confirmed pre-existing, not touched — full detail in the PR body's Test evidence section). Re-drove `/me`, `/family`, `/birthdays`, `/relationship`, a person overview, and the Timeline/Graph tabs in Playwright against a fresh staging instance after rebasing; results in the PR body's Manual verification section.

One incidental finding while re-verifying: this box runs many concurrent agent worktrees sharing one `.git` repository, and `git stash`'s ref is shared across all of them (not per-worktree) — a stray `git stash pop` here briefly picked up unrelated, uncommitted work from the `fix/me-family-sql-aggregates` worktree (an `aggregate_cache.py` addition to `crm.py`, #876). Caught it before it touched any commit, and re-stashed it intact with an explanatory message for that worktree to recover. Nothing of this PR's was affected, but worth knowing: `git stash` isn't safe to use in this environment without checking `git stash list` isn't already someone else's.

nbramia:
## Adversarial review of #907 by Opus — verification pass on `a68cb32`

**Verdict:** BLOCK

Seven of the eight findings are genuinely fixed and I could verify each one. The eighth — the #904 blocker — has a fix that removes the leak but breaks uvicorn's access logging outright: every request carrying a query string now emits a 79-line `Logging error` traceback instead of an access line. That is a worse production outcome than the leak it replaces, and the new tests can't see it because they never run a record through uvicorn's formatter.

No contamination from the `git stash` incident you disclosed: `git diff d598202...a68cb32` touches 13 files, none of them `aggregate_cache`-related, and grepping the full diff for `aggregate_cache` returns nothing. Thanks for flagging it — and agreed, `git stash` is shared-refs-per-repo, not per-worktree.

### Findings

1. **[BLOCKER] The query-string redaction filter breaks uvicorn's access log for every request that has a query string.** — `api/services/log_redaction.py:145-146` (`record.args = ()`) — `uvicorn.logging.AccessFormatter.formatMessage` doesn't read `record.msg` at all; it unpacks `record.args` into `(client_addr, method, full_path, http_version, status_code)` and rebuilds the request line from them (`uvicorn/logging.py:95-110`). Setting `record.args = ()` therefore raises inside the formatter. Reproduced end to end on a staging instance started exactly like production (`uvicorn.run("api.main:app", ..., log_level="info")`, matching `scripts/server.sh:251`): `GET /api/crm/people?q=ZZQMARKER7788&sort=strength&limit=5` produced

   ```
   --- Logging error ---
   Traceback (most recent call last):
     ...
     File ".../uvicorn/logging.py", line 99, in formatMessage
   ValueError: not enough values to unpack (expected 5, got 0)
   Call stack:
     ...
   Message: '127.0.0.1:33668 - "GET /api/crm/people?<redacted> HTTP/1.1" 200'
   Arguments: ()
   ```

   **79 lines** (traceback + full call stack) per request, on stderr — which systemd appends to `logs/server.log` — in place of one access line, with the status code and client address only recoverable from the `Message:` echo. The immediately following `GET /api/crm/config` (no query string) logged normally, so this hits precisely the requests the CRM UI makes most.

   I think this is also what produced the "manually reproduced end to end … now produces `"GET /api/crm/people?<redacted> HTTP/1.1" 200 OK`" line in your response: that string *does* appear in the log, but as the `Message:` field of the error block, not as an emitted access line. Worth grepping for `Logging error` next to the marker, not just for the marker.

   Reproduced in isolation too, to rule out anything environmental — building the record the way uvicorn does and formatting it with `AccessFormatter(fmt, use_colors=False)`:
   - no filter → `INFO:     127.0.0.1:1234 - "GET /api/crm/people?q=ZZQMARKER7788&limit=5 HTTP/1.1" 200 OK`
   - shipped filter → `FORMATTER RAISED: ValueError not enough values to unpack (expected 5, got 0)`

   **Tested fix:** redact `record.args`, which is what the formatter actually reads, and leave `record.msg` alone:

   ```python
   def filter(self, record):
       args = record.args
       if isinstance(args, tuple) and len(args) == 5 and isinstance(args[2], str) and "?" in args[2]:
           record.args = (args[0], args[1], args[2].split("?", 1)[0] + "?<redacted>", args[3], args[4])
       return True
   ```

   Verified: `INFO:     127.0.0.1:1234 - "GET /api/crm/people?<redacted> HTTP/1.1" 200 OK`, marker absent, line intact. Keep a `record.msg`-based branch as a fallback for a record that isn't shaped like uvicorn's if you want belt-and-braces, but only when `record.args` isn't the 5-tuple — never blank the args.

   **The test gap is the load-bearing part.** All five new tests in `TestRequestQueryStringRedactionFilter` assert on `record.getMessage()`, which is exactly the field uvicorn's formatter ignores; none of them constructs an `AccessFormatter` or calls `.format()`. `test_redacts_query_string_from_access_log_line` passes against the broken filter and would also pass against a filter that redacted nothing the formatter reads. Please add an assertion that runs the filtered record through `uvicorn.logging.AccessFormatter` and checks both that it doesn't raise *and* that the marker is absent from the formatted string — that single assertion is what turns this from a green suite into a real guard.

2. **[MINOR] The server half of finding 2's fix is untested.** — `api/routes/crm.py:688` — Mutation check: deleting `photos_enabled=settings.photos_enabled` from `get_crm_config()` leaves `test_crm_avatar_photos_ui_browser.py`, `test_photos_profile_api.py`, and `test_crm_api.py` green (the only failures were the three known host-load latency tests). The browser test stubs the field into its own `/api/crm/config` response, so it can't notice the server never sending it. One line in `tests/test_crm_api.py` asserting `client.get("/api/crm/config").json()["photos_enabled"] == settings.photos_enabled` closes it.

Everything else below is confirmed fixed.

### What I verified and how

- Checked out `a68cb32` (rebased onto `d598202`). `git diff d598202...a68cb32 --stat` → 13 files, +810/−88; grep for `aggregate_cache`/`me-family-sql` in the full diff → no hits.
- **Targeted suites green:** `pytest tests/test_photos_profile_api.py tests/test_crm_avatar_photos_ui_browser.py tests/test_log_redaction.py tests/test_crm_request_hygiene_ui_browser.py tests/test_family_matching.py -q` → `41 passed in 19.47s`. `pytest tests/test_route_handlers_concurrency.py tests/test_crm_api.py -k "config or never_logs or has_profile"` → `5 passed`.
- **Finding 1 (BLOCK, above):** staging on `:8030` with `log_level='info'`; marker count in the log = 0, but one `Logging error` block, 79 lines, measured from `--- Logging error ---` to `Arguments: ()`. Plus the isolated `AccessFormatter` reproduction and the args-based fix, both above.
- **Finding 2 — fixed.** `GET /api/crm/config` on staging returns `…,"photos_enabled":false`. Playwright against staging (:8028, real-data copy, Photos off): `/me` → 219 cards, **0 `/api/photos/*` requests of any kind** (previously one `GET /api/photos/stats` per page); `/family`, `/birthdays`, `/relationship`, `/crm`, person overview, Graph tab, infinite scroll, search and clear → **0 photo requests on every one**. `loadPhotosEnabled()` and its round trip are gone from `web/crm.html`; the only remaining mentions of `/api/photos/stats` are two explanatory comments. `/relationship`'s 2 console errors are the same pre-existing `/api/crm/people//timeline` 404 from staging's empty `partner_person_id`.
- **Finding 3 — fixed.** Real HTTP against the app with `Settings.photos_enabled` forced true: `GET` and `HEAD` on a hit now return **byte-identical header sets** — `cache-control: public, max-age=3600`, `content-type: image/jpeg`, `accept-ranges: bytes`, `content-length: 504`, `last-modified`, `etag: "2965a89c…"` — and a raw-socket `HEAD` confirms 0 body bytes with `content-length: 504`. On a miss: `HEAD` → 404, 0 body bytes, `content-length: 55`, same `cache-control`; `GET` → 404, 55 body bytes, same headers.
- **Finding 4 — fixed, mutation-checked.** Removing `loading="lazy"` from `renderAvatarPhotoImg()` now fails **two** tests: `test_avatar_classes_reflect_photo_availability` (the `to_have_attribute` assertion) and `test_300_flagged_rows_request_far_fewer_images_at_first_paint` (300/300 requests instead of the deferred fraction). Green again after revert.
- **Finding 5 — fixed.** `app.openapi()` under `warnings.catch_warnings(record=True)` → **1** duplicate-operation-ID warning, the pre-existing `voice_proxy_api_voice__path__post`. This route's is gone.
- **Finding 6 — fixed and accurate.** `docs/specs/technical/architecture.md` now states `crm_models/` isn't wired in, that the live models are in `api/routes/crm.py`, and that `_utils.is_family_member` is kept in sync by `tests/test_family_matching.py`. Spot-checked the claim it rests on: `api/routes/crm_models/_utils.py:119` defines `is_family_member` and `api/services/person_entity.py:1557` defines the independent `_is_family_member` used on the live path. The directory tree comment and the "Public API Imports" example were corrected too.
- **Finding 7 — fixed.** Re-ran the exact reproduction from last round (click five people on staging with Photos off): **0 `/api/photos/profile/*` responses, 0 console errors** — previously four `503`s and four console errors. `/relationship` likewise issues none.
- **Finding 8 — fixed.** `avatarPhotoClass()` is deleted; `renderPeopleList()` computes `renderAvatarPhotoImg(person)` once per row and derives the class from it. The new comment above `.person-avatar.has-photo` correctly says the rule is inert there (inline `color` from `getAvatarStyle()` wins), why that's the desirable degrade-to-initials behaviour, and that it still applies to `.detail-avatar.has-photo`.
- **`photo_count` lag:** your answer holds up — `person_stats.py` refreshes it at the end of each Photos sync for the people that run touched, not on a separate schedule. No further concern from me.
- Left the tree clean: no `.env`, no `data/` symlink, all three staging/probe ports released.

### Not verified

- Rendering against a genuine Apple Photos library — this host has none. The photos-enabled path was again exercised with stubbed responses and, at the route level, with a real file through `FileResponse`.
- The full unit suite — targeted files only, per the standing brief; note that a green suite is exactly what masked finding 1, so please re-run it after the args-based fix rather than treating the current green as coverage.
- Whether `settings.photos_enabled`'s `Path.exists()` could stall `GET /api/crm/config` on an install whose Photos library sits on a network mount. Still much cheaper than the aggregate queries it replaced, so this is a note, not an objection — `tests/test_route_handlers_concurrency.py`'s sub-100ms assertion on that endpoint still passes here.

nbramia:
## Response to verification-pass review (Opus)

Both items fixed on `fix/list-avatars`.

1. **[BLOCKER] Query-string redaction broke uvicorn's access log.** Confirmed exactly as diagnosed: `uvicorn.logging.AccessFormatter.formatMessage` never reads `record.msg`/`getMessage()` — it unconditionally unpacks `record.args` as `(client_addr, method, full_path, http_version, status_code)`. Clearing `record.args` to make `getMessage()` look right broke that unpack. Implemented the tested fix verbatim (redact `args[2]` in place, keep the 5-tuple shape; the `record.msg`-based rewrite is now only a fallback for a record that isn't shaped like uvicorn's own access-log call). Added `test_redacted_record_survives_uvicorns_actual_access_formatter`, which builds the record the way uvicorn does, runs it through the real `uvicorn.logging.AccessFormatter(fmt, use_colors=False)`, and asserts it doesn't raise and the marker is absent — mutation-tested against the old filter, which fails it (`ValueError: not enough values to unpack`, same exception you reproduced). Re-verified end to end on a staging instance started exactly like production (`uvicorn.run("api.main:app", ..., log_level="info")`): sent the marker query, then `grep`'d the log file for both the marker and `"Logging error"` — **both zero occurrences** — and confirmed a normal access line was emitted (`INFO:     127.0.0.1:... - "GET /api/crm/people?<redacted> HTTP/1.1" 200 OK`), plus a query-string-free request (`/health`, `/api/crm/config`) still logs completely untouched.

2. **[MINOR] `/api/crm/config`'s `photos_enabled` untested.** Added two tests: one pinning the real (`False`) value against `settings.photos_enabled` on this host, and one forcing `Settings.photos_enabled` to `True` via `PropertyMock` so the field can't be silently hardcoded to its Pydantic default and still pass both. Mutation-checked: deleting `photos_enabled=settings.photos_enabled` from `get_crm_config()` now fails the forced-`True` test.

**Gate, this round:**
```
flock -w 7200 /tmp/lifeos-push.lock ./scripts/test.sh
5723 passed, 8 skipped, 1181 warnings in 405.27s
```
Pre-push gate on the final rebase (onto `feat/crm-performance` @ `86527c5`, #899's tone-analysis/relationship-page work):
```
Running unit tests... passed (5772 passed, 8 skipped, 1182 warnings in 261.60s)
Running server-free browser tests... passed (280 passed, 6156 deselected in 199.38s)
```
Targeted files: 42 passed (`test_photos_profile_api.py`, `test_crm_avatar_photos_ui_browser.py`, `test_log_redaction.py`, `test_crm_request_hygiene_ui_browser.py`, `test_family_matching.py`) + 5 passed (`test_crm_api.py -k "photos_enabled or has_profile_photo or never_logs"`). Re-drove `/me`, `/family`, `/birthdays`, `/relationship`, a person overview, and Timeline/Graph tabs in Playwright on a fresh staging rebuild after the rebase — 0 console errors and 0 `/api/photos/*` requests throughout. `/relationship`'s previously-reported 2 pre-existing console errors are gone too (#899's own "handle no-partner installs" fix, unrelated to this PR but confirmed while re-checking).

**Final SHA:** `e353dec810f64ea67479635f8ee34c32e7066940` (base: `feat/crm-performance` @ `86527c5`).

nbramia:
## Adversarial review of #907 by Opus — final pass on `e353dec`

Both round-3 findings are fixed and I verified each one directly. One nit below, worth a one-line change but not worth another round if you'd rather take it as a follow-up.

### Nit

- **[MINOR] `test_photos_enabled_reflects_settings_when_false` is host-dependent and fails on a Mac that has a Photos library.** — `tests/test_crm_api.py:60` — `assert response.json()["photos_enabled"] == settings.photos_enabled is False` is a chained comparison, so it also asserts `settings.photos_enabled is False` — a property of the machine, not of the code. Reproduced by forcing `Settings.photos_enabled` True and running the verbatim assertion: `AssertionError: assert True is False`. LifeOS supports macOS, which is exactly where a Photos library exists, so this fails there. Either patch the property to `False` for symmetry with the `when_true` test, or just drop the trailing `is False` — `== settings.photos_enabled` alone is the real contract, and `test_photos_enabled_reflects_settings_when_true` already rules out a hardcoded `False`.

### What I verified

- Checked out `e353dec` (rebased onto `86527c5`). `git diff 86527c5...e353dec --stat` → the same 13 files, +909/−88; no unrelated files pulled in by the rebase.
- **Round-3 finding 1 — fixed.** Staging on `:8030` with `log_level='info'` (matching `scripts/server.sh:251`), then a full Playwright session against it: **28 access lines, 0 `Logging error` blocks, 0 occurrences of the marker.** The lines themselves are intact, including status and client address:
  ```
  INFO:     127.0.0.1:47502 - "GET /api/crm/people?<redacted> HTTP/1.1" 200 OK
  INFO:     127.0.0.1:47518 - "GET /api/crm/config HTTP/1.1" 200 OK
  INFO:     127.0.0.1:47534 - "GET /api/crm/people/nonexistent-id?<redacted> HTTP/1.1" 404 Not Found
  ```
  (Previously this was one 79-line `ValueError` traceback per query-string request.) The 404 case redacts correctly too.
- **The filter's branches, exercised directly:** a uvicorn-shaped record keeps its 5-tuple (`('127.0.0.1:1', 'GET', '/api/crm/people?<redacted>', '1.1', 200)`) and formats cleanly through `AccessFormatter`; a non-uvicorn-shaped record with args (`"fetched %s"`, one arg) falls through to the `record.msg` path and yields `'fetched https://x/api?<redacted>'`; a plain no-args message redacts too; a uvicorn-shaped record with no query string is left byte-identical and still formats; a malformed `%`-args record returns `True` without raising. Marker absent from every one.
- **The new guard actually guards.** Swapped `api/services/log_redaction.py` back to the round-2 version and ran `tests/test_log_redaction.py`: `1 failed, 13 passed` — `test_redacted_record_survives_uvicorns_actual_access_formatter` is the failure, on the `len(record.args) == 5` assertion. Restored.
- **Fallback is unreachable in practice, as documented.** Grepped uvicorn for every `access_logger.info` call site — two, in `h11_impl.py:480` and `httptools_impl.py:483`, both the 5-tuple HTTP shape; no WebSocket access logging and no WS routes in this app. So nothing in production can take the `args = ()` path.
- **Round-3 finding 2 — fixed.** `GET /api/crm/config` on staging returns `…,"photos_enabled":false`. Mutation check: deleting `photos_enabled=settings.photos_enabled` from `get_crm_config()` now fails `TestCRMConfig::test_photos_enabled_reflects_settings_when_true` (`assert False is True`). Reverted.
- **Targeted suites green:** `pytest tests/test_log_redaction.py tests/test_photos_profile_api.py tests/test_crm_avatar_photos_ui_browser.py tests/test_crm_request_hygiene_ui_browser.py tests/test_family_matching.py -q` → `42 passed in 18.05s`.
- **Frontend re-checked after the rebase:** clicking five people on staging with Photos off → 0 `/api/photos/profile/*` responses, 0 console errors.
- Findings 2–8 from round 1 remain fixed as verified last pass; nothing in this commit touches them.
- Tree left clean: no `.env`, no `data/` symlink, staging port released.

### Not verified

- Rendering against a real Apple Photos library — still none on this host.
- The full unit suite — targeted files only, per the standing brief.

nbramia:
Nit fixed: `test_photos_enabled_reflects_settings_when_false` now patches `Settings.photos_enabled` to `False` explicitly (same pattern as the `True`-case test just below it) instead of chaining `== settings.photos_enabled is False`, which asserted a host property and would fail on a machine with a real Photos library configured.

Ran `tests/test_crm_api.py -k Config` — 2 passed. No rebase needed (integration tip unchanged at `86527c5`). Pushed under the shared lock.

**Final SHA:** `a3d45a43b53315f0f30c86f3c0f7955fa40e0750`
**Gate:**
```
Running unit tests... passed (5772 passed, 8 skipped, 1181 warnings in 231.09s)
Running server-free browser tests... passed (280 passed, 6156 deselected in 279.60s)
```

PR #909, opened 2026-09-04
feat: make CRM dashboard navigation client-side

### LINKED ISSUE #876: feat(crm): client-side dashboard navigation and a data_version-keyed cache for heavy aggregates

nbramia:
## TL;DR

Clicking Me, Family, Birthdays, Relationship, or CRM in the nav bar no longer reloads the page. The URL updates, the dashboard swaps in, and the browser's back/forward buttons work the same way — all without a fresh document load, a re-download of the page, or a refetch of the charting library. Whatever a dashboard already fetched during the session stays available when you return to it: switching from Me to Family and back to Me shows the Me dashboard instantly, using data already in memory instead of asking the server again. The nav link for whichever dashboard you're viewing is now highlighted. Loading any of these pages directly (a bookmark, a refresh, a shared link) still works exactly as before.

Closes #876 (part 1 of 2 — this is client-side navigation; a server-side response cache for the heaviest dashboard endpoints is a separate PR).

Part of #876

**Update:** an adversarial review found nine issues (three of them real regressions that keeping dashboard state alive across navigation exposed); all nine are fixed below, plus one more found during my own re-verification after the integration branch advanced past this PR twice more. See the review thread for the finding-by-finding response.

## Design

Every dashboard nav link becomes a small piece of client-side routing instead of a plain hyperlink: clicking one pushes a browser history entry and hands off to the same dashboard-rendering logic the page already used for browser back/forward and for switching between people. That shared dispatch logic is now one function both paths call, so a click and a back/forward step behave identically. The active nav link now tracks the current view instead of being hardcoded.

Keeping dashboard state alive across navigation (rather than discarding it on every reload) exposed several real bugs the review caught, all in state that used to get wiped by the document reload every dashboard switch previously was:
- The Me and Family dashboards share one piece of UI state that sizes the interaction heatmap; a Family visit in between could leave it pointing at the wrong window for a returning Me visit, defeating Me's own cached-data check. Fixed, but the first fix over-corrected and made the Me page's own year-selector dropdown inert (a second fix threads through whether the caller explicitly chose a window, so the dropdown's own choice is never immediately clobbered).
- `/birthdays` intentionally skips loading the people list/stats for a faster initial render; nothing compensated once the user moved on to a dashboard that does need them, leaving the sidebar and header stats stuck loading forever.
- A dashboard's nav-link highlight and a filtered `/birthdays` view could both survive past the point they should have reset.
- A person's Graph tab (new from #896, merged after this PR was first written) schedules an 800ms-deferred zoom-fit animation; navigating away within that window used to be moot (the document reload killed the timer), but now left it computing against a hidden, zero-size container, throwing a burst of console errors the next time anything on the page dispatched a zoom event.

## Implementation Notes

- `web/crm.html`:
  - `dispatchRoute()` — shared function extracted from the existing `popstate` handler's body; renders whatever the current URL points at (Me/Family/Birthdays/Relationship/a person), used by both the `popstate` handler and `navigateTo()`. Now also: stops the graph tab's force simulation on every navigation (single choke point); starts a memoized `ensurePeopleAndStatsLoaded()` for every non-birthdays view; defaults to the Overview tab when a person URL carries no tab segment.
  - `navigateTo(path, event)` — intercepts a nav-link click, lets modifier-clicks/non-primary buttons fall through to the browser untouched, otherwise prevents the default navigation, pushes history state, and calls `dispatchRoute()` (now `.catch()`-guarded). `/crm` is translated to `/me` before pushing. The same-target early-return now compares `pathname + search`, not pathname alone.
  - `updateActiveNavLink()` — toggles `.active` on whichever of the four dashboard nav links (`data-page="birthdays|relationship|family|me"`) matches the current view. Called from `dispatchRoute()`, `init()`'s direct-load path, and now `updateUrl()` too (so selecting a person clears a stale dashboard highlight).
  - `ensureMeHeatmapYears(respectExplicitYears)` / `loadMeInteractions(forceRefresh, respectExplicitYears)` / `loadMeHeatMap(respectExplicitYears)` — thread a flag so the years dropdown's own selection isn't immediately overwritten by the span re-assertion that fixes the Me/Family cross-contamination bug; only dashboard-entry callers (not the dropdown's change handler) re-assert from the span.
  - `showBirthdaysPage(filterDay)` — now resets to the Calendar tab when no filter is given, matching what a fresh reload would show (previously only ever switched *to* Timeline, never back).
  - `fitToFirstDegreeNodes()` (graph tab, #896) — its deferred zoom-fit callback now bails out if the graph is no longer live or its container is hidden (0×0), instead of computing a degenerate transform against it.
  - Nav anchors (`/crm`, `/birthdays`, `/relationship`, `/family`, `/me`) gained `onclick="return navigateTo(...)"` and (except `/crm`) `data-page` attributes.
  - CSS: added `.nav-link.me-link.active` (higher specificity than the existing `.nav-link.me-link` rule).
- `docs/specs/product/crm-ui.md`, `docs/specs/product/crm-analytics.md`, `docs/specs/product/crm-interactions.md`: new "Navigation" section (crm-ui.md) plus three stale `/crm#family` → `/family` URL fixes.

### Automated tests

`tests/test_crm_client_nav_ui_browser.py` (server-free, `browser` without `requires_server` — runs at pre-push): 12 tests, each addition verified by reverting its corresponding fix and confirming the test actually fails without it —
- `TestNoDocumentReload`: click-through with a new `_assert_only_visible()` helper checking the *other three* dashboard containers are hidden at every hop (not just that the target is visible), including Relationship↔Family; a filtered-birthdays-reset test; a modifier-click pass-through test.
- `TestBackForward`: same strengthened hidden-container assertions across a long back/forward history.
- `TestDashboardStatePersists`: a Me → Family → Me round trip issues exactly one `/api/crm/me/interactions` request total.
- `TestYearsSelectorRespectsUserOverride`: selecting "3" years fetches `days_back=1102` and the selection sticks (fails pre-fix with `got days_back values: ['737']`).
- `TestBirthdaysEntryPoint`: entering at `/birthdays` then visiting Family/Me actually populates the sidebar and stats (fails pre-fix with the stats stuck on `"-"`).
- `TestDirectLoadsUnchanged`: every dashboard URL still renders correctly on a fresh load.

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_crm_client_nav_ui_browser.py -q
collected 12 items
tests/test_crm_client_nav_ui_browser.py ............                    [100%]
============================= 12 passed in 11.22s ==============================
```

Full suite (under the shared push lock, as part of the actual push, rebased onto #896's tip): **5809 passed, 8 skipped** (unit) and **297 passed** (server-free browser, pre-push scope) — pre-existing skips only, no failures.

### Manual verification

Ran a private staging instance (`~/.venvs/lifeos/bin/python`, port 8029, this worktree, staging `.env` — no side effects) against a read-only copy of the real CRM/interactions databases, driven with the Playwright MCP tool, re-run after each of two further rebases as the integration branch advanced:
- Every finding above reproduced before its fix and was confirmed clean after, against real data (exact numbers in the review-thread reply): year-selector `days_back` values, `/birthdays`-entry sidebar/stats population, Relationship↔Family container visibility, the filtered-birthdays query-string reset, and the Graph-tab NaN console errors (the last one found only during this manual pass, not in the original review — root-caused via a temporary `setAttribute` instrumentation to capture the real stack trace).
- Full click-through (all nav orders, long back/forward history, direct loads/reloads of every URL, person selection from each dashboard, a Graph-tab hop) — no document reloads, no leaked dashboards, no console errors beyond two pre-existing staging-only conditions (no partner configured; photo service unavailable), unrelated to this change and present on a fresh direct load too.

## Review focus

- The Graph-tab fix (`fitToFirstDegreeNodes()`'s deferred-timer guard) is not covered by an automated test: this suite's server-free browser tests deliberately stub `window.d3` with an inert proxy so they need no external network access, and that stub never executes real d3-zoom internals — a test against the genuine CDN-loaded d3 would need live network access, which this suite's own tests explicitly avoid depending on. Verified manually only; flagging in case a differently-scoped test (e.g. one that allows the real CDN fetch) is wanted here.
- Whether the CRM brand link's fixed "always active" styling (vs. the four dashboard links which now toggle) matches intent — treated as "you're in the CRM app" rather than a fifth toggleable view, per the issue's note that `/crm` has always redirected to `/me`.
- The two pre-existing staging-only console conditions noted above (no configured partner; photo service unavailable) are unrelated to this PR — flagging in case they're worth a separate issue.

### DIFF
--- a/docs/specs/product/crm-analytics.md
+++ b/docs/specs/product/crm-analytics.md
@@ -4,7 +4,7 @@
 **Owner:** CRM
 **Last Updated:** 2026-09-04
 
-The aggregated views that sit alongside the people list: Family Dashboard (`/crm#family`), Me Dashboard (`/me`, the CRM landing page), Birthdays Page (`/birthdays`), and Relationship Dashboard (`/relationship`). Each surfaces interaction patterns and insights derived from the CRM data model.
+The aggregated views that sit alongside the people list: Family Dashboard (`/family`), Me Dashboard (`/me`, the CRM landing page), Birthdays Page (`/birthdays`), and Relationship Dashboard (`/relationship`). Each surfaces interaction patterns and insights derived from the CRM data model.
 
 See [crm-ui.md](crm-ui.md) for the CRM index and the sibling specs that cover people, interactions, and the graph.
 
@@ -21,7 +21,7 @@ See [crm-ui.md](crm-ui.md) for the CRM index and the sibling specs that cover pe
 
 ## Family Dashboard
 
-`URL: /crm#family`
+`URL: /family`
 
 Aggregated view of interactions across multiple selected family members. Lets you track engagement with family as a group and quickly see which member you've been out of touch with.
 
--- a/docs/specs/product/crm-interactions.md
+++ b/docs/specs/product/crm-interactions.md
@@ -80,7 +80,7 @@ Behavior:
 - Source-type filter dropdown narrows the list.
 - Clicking an interaction opens the source link — `mailto:` / Google Calendar URL / `obsidian://` / `imessage://` etc.
 - Infinite scroll loads older interactions.
-- On the Me and Family timelines (`/me`, `/crm#family` — see [crm-analytics.md](crm-analytics.md)), an interaction still attributed to a person id that was later merged into another person displays under the surviving person's name, following the merge chain the same way a direct person lookup does. Previously such rows displayed "Unknown".
+- On the Me and Family timelines (`/me`, `/family` — see [crm-analytics.md](crm-analytics.md)), an interaction still attributed to a person id that was later merged into another person displays under the surviving person's name, following the merge chain the same way a direct person lookup does. Previously such rows displayed "Unknown".
 
 ---
 
--- a/docs/specs/product/crm-ui.md
+++ b/docs/specs/product/crm-ui.md
@@ -2,7 +2,7 @@
 
 **Status:** Complete
 **Owner:** CRM
-**Last Updated:** 2026-06-21
+**Last Updated:** 2026-09-04
 
 LifeOS's Personal CRM is built on top of the [two-tier data model](data-model.md) ([ADR-003](../../adr/003-two-tier-data-model.md)) and provides network management and relationship context across every observed touchpoint with the people in your life. The UI lives at `/crm`; the API lives at `/api/crm/*`.
 
@@ -28,7 +28,15 @@ The CRM is split across four focused specs by feature area — this file is the 
 | [crm-people.md](crm-people.md) | Person list view, detail view, edit flows; the two-tier entity model; contact-source aggregation; split and merge operations; link overrides; relationship-strength scoring; Dunbar circles; the multi-stage person-facts extraction pipeline. |
 | [crm-interactions.md](crm-interactions.md) | The interaction timeline and the data-source integrations behind it (Gmail, Calendar, iMessage, Apple Contacts, Slack, WhatsApp, Signal). |
 | [crm-graph.md](crm-graph.md) | The D3 force-directed relationship graph; multi-source `Relationship` model; per-source counts; source-filter UI; edge-weight calculation. |
-| [crm-analytics.md](crm-analytics.md) | Aggregated views: Family Dashboard (`/crm#family`), Me Dashboard (`/me` — the landing page), Birthdays Page (`/birthdays`), Relationship Dashboard (`/relationship`). |
+| [crm-analytics.md](crm-analytics.md) | Aggregated views: Family Dashboard (`/family`), Me Dashboard (`/me` — the landing page), Birthdays Page (`/birthdays`), Relationship Dashboard (`/relationship`). |
+
+---
+
+## Navigation
+
+Moving between the CRM's dashboards (Me, Family, Birthdays, Relationship, and the person list at `/crm`) is entirely client-side: clicking a nav link updates the URL (`history.pushState`) and dispatches straight to that dashboard's existing initializer, with no document reload — the same mechanism the page already used for person and tab navigation. Browser back/forward restores the previous dashboard the same way. The nav link for the current view is highlighted; the `/crm` link always leads to `/me` (its direct-load redirect, unchanged), so it never highlights independently of the other four.
+
+A dashboard's in-memory state (its own cached aggregates, e.g. the Me dashboard's fetched interaction history) persists for the life of the browser tab: returning to a dashboard with the same parameters re-renders from what's already held rather than re-fetching. A full page reload (or a direct link/bookmark) still renders every URL exactly as a fresh visit — only in-session navigation is memoized.
 
 ---
 
--- a/tests/test_crm_client_nav_ui_browser.py
+++ b/tests/test_crm_client_nav_ui_browser.py
@@ -1,0 +1,345 @@
+"""Browser tests for client-side CRM dashboard navigation (#876).
+
+Before this, the Me/Family/Birthdays/Relationship/CRM nav links were plain
+anchors: every click was a full document load that re-downloaded the ~750 KB
+page, refetched the d3 CDN script, and discarded whatever the previous
+dashboard had already fetched. This pins that clicking those links now goes
+through `navigateTo()` -> `dispatchRoute()` (a `pushState` plus the same
+dispatch the popstate handler already used for person/tab navigation)
+instead of letting the browser load the anchor's `href`, that back/forward
+still works, that the active nav-link style follows the current view, and
+that a dashboard's in-memory caches (in particular `meInteractionsCache`,
+via `loadMeInteractions()`) survive a round trip through another dashboard
+instead of being silently invalidated by shared UI state (`heatmapYears`)
+another dashboard also writes to.
+
+Unlike most of the browser suite this serves `web/crm.html` itself from an
+ephemeral port rather than pointing at a running API — every `/api/crm/**`
+call the page makes is stubbed, so the assertions are entirely about this
+checkout's `web/crm.html` JS. That is why it carries no `requires_server`
+marker, and so runs at pre-push (`browser and not requires_server`).
+"""
+import http.server
+import json
+import threading
+from pathlib import Path
+from urllib.parse import urlparse
+
+import pytest
+from playwright.sync_api import Page, expect
+
+pytestmark = [pytest.mark.browser, pytest.mark.slow]
+
+WEB_DIR = Path(__file__).resolve().parent.parent / "web"
+
+# Obviously synthetic - never a real person id.
+MY_PERSON_ID = "person-synthetic-me"
+
+# Neutralizes the d3.js CDN dependency (crm.html loads it via <script src>)
+# so this test needs no external network access and d3 chart calls elsewhere
+# on the page (unrelated to this test's assertions) don't throw. Any property
+# access or call on the stub returns the stub itself, so arbitrary chaining
+# like d3.select(x).attr(y).style(z) always succeeds.
+D3_STUB_JS = """
+(function() {
+    let stub;
+    stub = new Proxy(function() { return stub; }, {
+        get(target, prop) {
+            if (prop === 'then') return undefined;
+            return () => stub;
+        }
+    });
+    window.d3 = stub;
+})();
+"""
+
+SYNTHETIC_ME_PERSON = {
+    "id": MY_PERSON_ID,
+    "canonical_name": "Synthetic Owner",
+    "display_name": "Synthetic Owner",
+    "emails": [], "phone_numbers": [], "company": None, "position": None,
+    "linkedin_url": None, "category": "self", "vault_contexts": [], "tags": [],
+    "birthday": None, "notes": "", "sources": [], "first_seen": None,
+    "last_seen": None, "relationship_strength": 0.0, "source_entity_count": 0,
+    "meeting_count": 0, "email_count": 0, "mention_count": 0, "message_count": 0,
+    "slack_message_count": 0, "dunbar_circle": -1, "source_entities": [],
+    "relationships": [],
+}
+
+EMPTY_ME_INTERACTIONS = {
+    "daily": [], "by_source": {}, "by_month": {}, "by_circle": {},
+    "top_contacts": [], "warming": [], "cooling": [], "total_count": 0,
+    "relationship_health_score": 0, "health_score_history": [],
+    "health_score_average": 0.0, "neglected_contacts": [], "network_growth": [],
+    "messaging_by_circle": [], "tracked_relationships": [],
+}
+
+EMPTY_PEOPLE_PAGE = {"people": [], "total": 0, "offset": 0, "count": 0}
+
+
+class _CrmHandler(http.server.SimpleHTTPRequestHandler):
+    """Serves crm.html the way api/main.py does for /me, /family, /crm,
+    /birthdays, and /relationship: every non-static path resolves to the
+    same file, and the module tree hangs off /static/."""
+
+    def translate_path(self, path):
+        path = path.split("?", 1)[0].split("#", 1)[0]
+        if path.startswith("/static/"):
+            return str(WEB_DIR / path[len("/static/"):])
+        return str(WEB_DIR / "crm.html")
+
+    def log_message(self, *args):  # keep pytest output clean
+        pass
+
+
+@pytest.fixture(scope="module")
+def crm_base_url():
+    server = http.server.ThreadingHTTPServer(("127.0.0.1", 0), _CrmHandler)
+    threading.Thread(target=server.serve_forever, daemon=True).start()
+    try:
+        yield f"http://127.0.0.1:{server.server_port}"
+    finally:
+        server.shutdown()
+        server.server_close()
+
+
+def _route_api(page: Page, requests_seen: list):
+    """Stub every /api/crm/** call the page makes, and record each request's
+    path (query string dropped) so tests can assert on call counts. Anything
+    not explicitly modeled below returns `{}` — every loader in crm.html
+    treats a missing/empty aggregate defensively (falls back to an empty
+    list/dict), so this is enough to exercise all five dashboards without
+    error."""
+
+    def handler(route):
+        path = urlparse(route.request.url).path
+        requests_seen.append(path)
+
+        if path == "/api/crm/config":
+            body = {"my_person_id": MY_PERSON_ID}
+        elif path == "/api/crm/statistics":
+            body = {}
+        elif path == "/api/crm/birthdays/today":
+            body = {"birthdays": []}
+        elif path == "/api/crm/people":
+            body = EMPTY_PEOPLE_PAGE
+        elif path == f"/api/crm/people/{MY_PERSON_ID}":
+            body = SYNTHETIC_ME_PERSON
+        elif path == "/api/crm/me/interactions/span":
+            body = {"earliest": "2024-01-01T00:00:00+00:00",
+                     "latest": "2026-01-01T00:00:00+00:00", "years": 2}
+        elif path == "/api/crm/me/interactions":
+            body = EMPTY_ME_INTERACTIONS
+        elif path == "/api/crm/me/stats":
+            body = {"total_people": 0, "total_emails": 0, "total_meetings": 0,
+                     "total_messages": 0}
+        elif path == "/api/crm/family/members":
+            body = {"members": []}
+        elif path == "/api/crm/birthdays/all":
+            body = {"total_people": 0, "total_dates": 0, "birthdays": []}
+        else:
+            body = {}
+
+        route.fulfill(status=200, content_type="application/json", body=json.dumps(body))
+
+    page.route("**/api/crm/**", handler)
+
+
+def _prepare(page: Page):
+    """Install the d3 stub and API stub before any page script runs, and
+    start recording document `load` events (one always fires for the
+    initial `page.goto()` - callers should clear the list after the first
+    dashboard is up, then assert it stays empty across client-side
+    navigation)."""
+    page.add_init_script(D3_STUB_JS)
+    requests_seen = []
+    _route_api(page, requests_seen)
+    load_events = []
+    page.on("load", lambda: load_events.append(1))
+    console_errors = []
+    page.on("console", lambda msg: console_errors.append(msg.text) if msg.type == "error" else None)
+    return requests_seen, load_events, console_errors
+
+
+def _goto_me(page: Page, base_url):
+    requests_seen, load_events, console_errors = _prepare(page)
+    page.goto(f"{base_url}/me")
+    expect(page.locator("#meDashboard")).to_be_visible()
+    load_events.clear()  # discard the initial document load
+    return requests_seen, load_events, console_errors
+
+
+def _active_pages(page: Page):
+    """The set of data-page values currently carrying the `active` class
+    among the four dashboard nav links."""
+    return page.eval_on_selector_all(
+        ".header-right .nav-link.me-link",
+        "els => els.filter(e => e.classList.contains('active')).map(e => e.dataset.page)",
+    )
+
+
+class TestNoDocumentReload:
+    """Clicking a nav link updates the URL and swaps the dashboard without
+    the browser performing a fresh document load."""
+
+    def test_click_through_all_dashboards(self, page: Page, crm_base_url):
+        requests_seen, load_events, console_errors = _goto_me(page, crm_base_url)
+
+        assert _active_pages(page) == ["me"]
+
+        page.locator('[data-page="family"]').click()
+        expect(page.locator("#familyDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/family"
+        assert _active_pages(page) == ["family"]
+
+        page.locator('[data-page="birthdays"]').click()
+        expect(page.locator("#birthdaysPage")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/birthdays"
+        assert _active_pages(page) == ["birthdays"]
+
+        page.locator('[data-page="relationship"]').click()
+        expect(page.locator("#relationshipDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/relationship"
+        assert _active_pages(page) == ["relationship"]
+
+        page.locator('[data-page="me"]').click()
+        expect(page.locator("#meDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/me"
+        assert _active_pages(page) == ["me"]
+
+        # The CRM brand link has always redirected to /me on a direct load
+        # (init()); clicking it client-side preserves that.
+        page.locator('.header-nav a[href="/crm"]').click()
+        expect(page.locator("#meDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/me"
+        assert _active_pages(page) == ["me"]
+
+        assert load_events == [], (
+            f"expected zero document loads after the first, got {len(load_events)}"
+        )
+        unexpected_errors = [e for e in console_errors if "d3" not in e.lower()]
+        assert not unexpected_errors, f"Unexpected console errors: {unexpected_errors}"
+
+    def test_middle_click_is_not_intercepted(self, page: Page, crm_base_url):
+        """A modified click (new-tab intent) must be left to the browser,
+        not swallowed by navigateTo()'s preventDefault()."""
+        _, load_events, _ = _goto_me(page, crm_base_url)
+
+        result = page.evaluate(
+            """() => {
+                const link = document.querySelector('[data-page="family"]');
+                const evt = new MouseEvent('click', { button: 0, ctrlKey: true, cancelable: true });
+                const notPrevented = link.dispatchEvent(evt);
+                return notPrevented;
+            }"""
+        )
+        # dispatchEvent returns false only if preventDefault() was called.
+        assert result is True
+        # Since nothing actually followed the href in this synthetic dispatch,
+        # the URL/dashboard must be unchanged - proving navigateTo() bailed
+        # out before calling pushState()/dispatchRoute().
+        assert page.evaluate("window.location.pathname") == "/me"
+
+
+class TestBackForward:
+    """Browser back/forward restores the previous dashboard without a
+    document load."""
+
+    def test_back_and_forward_restore_dashboards(self, page: Page, crm_base_url):
+        _, load_events, console_errors = _goto_me(page, crm_base_url)
+
+        page.locator('[data-page="family"]').click()
+        expect(page.locator("#familyDashboard")).to_be_visible()
+        page.locator('[data-page="birthdays"]').click()
+        expect(page.locator("#birthdaysPage")).to_be_visible()
+
+        page.go_back()
+        expect(page.locator("#familyDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/family"
+        assert _active_pages(page) == ["family"]
+
+        page.go_back()
+        expect(page.locator("#meDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/me"
+        assert _active_pages(page) == ["me"]
+
+        page.go_forward()
+        expect(page.locator("#familyDashboard")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/family"
+
+        page.go_forward()
+        expect(page.locator("#birthdaysPage")).to_be_visible()
+        assert page.evaluate("window.location.pathname") == "/birthdays"
+        assert _active_pages(page) == ["birthdays"]
+
+        assert load_events == [], (
+            f"expected zero document loads across back/forward, got {len(load_events)}"
+        )
+        unexpected_errors = [e for e in console_errors if "d3" not in e.lower()]
+        assert not unexpected_errors, f"Unexpected console errors: {unexpected_errors}"
+
+
+class TestDashboardStatePersists:
+    """A Me -> Family -> Me round trip issues no new request for data the
+    first Me visit already fetched and cached, even though Family's own
+    heatmap window (`heatmapYears`, a global shared with Me's) is 10 years
+    while Me's is derived from /me/interactions/span."""
+
+    def test_second_me_visit_does_not_refetch_interactions(self, page: Page, crm_base_url):
+        requests_seen, load_events, console_errors = _goto_me(page, crm_base_url)
+
+        def _count(path):
+            return len([p for p in requests_seen if p == path])
+
+        # Let the first Me visit's /me/interactions settle.
+        for _ in range(100):
+            if _count("/api/crm/me/interactions") >= 1:
+                break
+            page.wait_for_timeout(50)
+        assert _count("/api/crm/me/interactions") == 1, (
+            "expected exactly one /me/interactions request from the first Me visit, "
+            f"got {_count('/api/crm/me/interactions')}"
+        )
+
+        page.locator('[data-page="family"]').click()
+        expect(page.locator("#familyDashboard")).to_be_visible()
+        page.wait_for_timeout(200)
+
+        page.locator('[data-page="me"]').click()
+        expect(page.locator("#meDashboard")).to_be_visible()
+        # Give any (incorrect) re-fetch time to land before asserting.
+        page.wait_for_timeout(300)
+
+        assert _count("/api/crm/me/interactions") == 1, (
+            "second Me render must reuse meInteractionsCache instead of "
+            f"re-requesting /me/interactions, got {_count('/api/crm/me/interactions')} total calls"
+        )
+
+        unexpected_errors = [e for e in console_errors if "d3" not in e.lower()]
+        assert not unexpected_errors, f"Unexpected console errors: {unexpected_errors}"
+
+
+class TestDirectLoadsUnchanged:
+    """Direct loads (a fresh page.goto(), not a client-side nav) of every
+    dashboard URL still render exactly as before, with the correct nav link
+    highlighted."""
+
+    @pytest.mark.parametrize("path,element_id,active_page", [
+        ("/me", "#meDashboard", "me"),
+        ("/family", "#familyDashboard", "family"),
+        ("/birthdays", "#birthdaysPage", "birthdays"),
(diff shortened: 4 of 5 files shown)


nbramia:
## Adversarial review of #909 by Opus

**Verdict:** BLOCK

The navigation plumbing itself is solid — zero document loads anywhere I could produce one, correct URLs, correct back/forward through long histories, correct direct loads and reloads, and modifier/middle/keyboard clicks all still behave natively. But keeping state alive across navigation has surfaced three user-visible regressions that the new test file cannot see, one of which is caused by this PR's own `ensureMeHeatmapYears()` fix.

1. **[BLOCKER] The Me dashboard's heatmap year selector is now completely dead** — `web/crm.html:10790-10795` (`ensureMeHeatmapYears()`). The dropdown's change handler (`web/crm.html:10094-10114`) routes through `loadAndRenderHeatMap(currentHeatmapPersonId, years)` → (Me page) `loadMeHeatMap()` → `loadMeInteractions()` → `ensureMeHeatmapYears()`, which now unconditionally re-asserts `heatmapYears = meSpanYears` and rewrites `heatmapYearsSelect.value` — clobbering the selection the user just made, inside the very call the selection triggered. Measured on staging against the real data copy, PR branch vs. `feat/crm-performance`'s `web/crm.html` served through the same API:
   - PR: select "3" → `{selectValue: '10', heatmapYears: 10, title: '10-Year Interaction History'}`, zero new `/me/interactions` requests. Same for "1". The control is inert.
   - Base: select "3" → `{selectValue: '3', heatmapYears: 3, title: '3-Year Interaction History'}` + `GET /me/interactions?days_back=1102`; select "1" → `days_back=372`.

   The old code assigned inside the once-resolved `.then()`, so a later user choice stuck. *Fix:* don't re-assert from the dropdown-driven path — e.g. give `ensureMeHeatmapYears()` an explicit "entering the Me dashboard" flag (or record a user override in the `heatmapYearsSelect` change handler that the re-assertion respects) so the re-assert only happens on dashboard entry, not on every `loadMeInteractions()`. There is no test for the year selector on any branch; please add one with the fix.

2. **[BLOCKER] Relationship → Family leaves the Relationship dashboard rendered underneath the Family dashboard** — `web/crm.html:11508-11519` (`showFamilyDashboard()` hides `#meDashboard` and `#personContentGrid` but never `#relationshipDashboard`; the reverse direction, `showRelationshipDashboard()` at `web/crm.html:11922-11923`, *does* hide `#familyDashboard`). This was unreachable before because every dashboard switch was a fresh document; client-side nav makes it live. Reproduced on staging:
   ```
   on /relationship:      vis={family:'none', rel:'block'}  relBox=[912,1430]
   after Family click:    vis={family:'block', rel:'block'} relBox=[912,1430] famBox=[912,1051]
   ```
   Two full dashboards stacked in the detail pane — the same class of leak as #899. *Fix:* hide `#relationshipDashboard` in `showFamilyDashboard()` (and, defensively, make the hide-everything-else step one shared helper all three `show*Dashboard()` functions call, so the next dashboard added can't reintroduce this).

3. **[BLOCKER] Entering the app at `/birthdays` and then clicking any dashboard link leaves the page half-initialized** — `web/crm.html:9352-9354` (`init()` returns after `showBirthdaysPage()` *before* `loadPeople()` / `loadStatistics()`) combined with `dispatchRoute()` (`web/crm.html:9235`), which never compensates. Previously each nav click was a document load that re-ran `init()` in full. Reproduced on staging (direct `goto /birthdays`, then click Family, then click Me):
   ```
   FAMILY(from bdays): peopleList="Loading people..."  stats="People: -"
   ME(from bdays):     peopleList="👥 No people found" stats="People: -"
   FAMILY(direct):     peopleList="NR Nathan Ramia … "  stats="People: 15391"
   ME(direct):         peopleList="NR Nathan Ramia … "  stats="People: 15391"
   ```
   `/birthdays` is exactly the kind of URL that gets bookmarked or shared, so this is a realistic entry point. *Fix:* have `dispatchRoute()` (or `navigateTo()`) kick off `loadPeople()`/`loadStatistics()` once if they've never run — e.g. a module-scope `peopleLoadedPromise` that `init()` sets and `dispatchRoute()` awaits/starts when it's still null.

4. **[MAJOR] The active nav link does not follow the current view when a person is selected** — `updateUrl()` at `web/crm.html:9206-9224` pushes the new person URL but never calls `updateActiveNavLink()`. Selecting a person from the Me dashboard leaves "Me" highlighted while `/crm/{other-person}` is showing; same from Family. Measured:
   ```
   after person click: url=/crm/34142cc5-…  active=['me']     (from Me)
   after person click: url=/crm/34142cc5-…  active=['family'] (from Family)
   ```
   A direct load of the same URL correctly yields `active=[]`, so the highlight is now inconsistent with itself. This is the issue's "The active nav link styling shall follow the current view" criterion, and the new spec text in `crm-ui.md` asserts it. *Fix:* call `updateActiveNavLink()` from `updateUrl()`.

5. **[MAJOR] The new browser test can't see finding 2 (or any dashboard leak): it never asserts that the previous view is hidden, and never navigates Relationship → Family** — `tests/test_crm_client_nav_ui_browser.py:186-224`. The click-through only walks Me → Family → Birthdays → Relationship → Me → CRM and asserts the *target* is visible. Mutation proof: I deleted the two lines in `showFamilyDashboard()` that hide `#meDashboard` (a strictly worse version of the bug this PR shipped) and the file still reported **9 passed**. *Fix:* assert the other four dashboard containers are hidden after each hop, and include the Relationship → Family hop (and Family → Relationship) in the walk. Also please add a test for the year selector (finding 1) and for the `/birthdays`-entry case (finding 3) — this is a routing change, so the entry-point matrix is the thing that needs covering, not just the happy path from `/me`.

6. **[MAJOR] Clicking a nav link is a silent no-op when only the query string differs** — `web/crm.html:9322` (`if (window.location.pathname === targetPath) return false;` ignores `location.search`). From `/birthdays?day=03-15`, clicking 🎂 Birthdays leaves the URL, the day filter and the Timeline tab exactly as they were; previously it reloaded to an unfiltered `/birthdays` on the Calendar tab. Measured:
   ```
   before:  url=/birthdays?day=03-15  bdayTab=['Timeline']
   after:   url=/birthdays?day=03-15  bdayTab=['Timeline']
   ```
   *Fix:* compare `pathname + search` (and clear the filter when the target has none), or drop the early return and let `dispatchRoute()` re-render.

7. **[MINOR] Clicking "Me" from `/me/timeline` changes the URL to `/me` but leaves the Timeline pane displayed** — `dispatchRoute()`'s `personId` branch (`web/crm.html:9254-9259`) only calls `switchTab()` when the URL *has* a tab, never resets to overview when it doesn't. I confirmed this is a pre-existing weakness of the popstate handler (back from `/me/timeline` behaves identically on `feat/crm-performance`), so it is inherited rather than introduced — but the click path is new, and a nav click that lands on the wrong pane is more visible than a back button doing it. Worth fixing while you're in here: `switchTab(tab || 'overview', true)`.

8. **[MINOR] One stale `/crm#family` reference left behind** — `docs/specs/product/crm-interactions.md:83`. The PR corrected the same stale URL in `crm-ui.md` and `crm-analytics.md`; this third one is identical and was missed by the same grep.

9. **[MINOR] `dispatchRoute()` is fire-and-forget in `navigateTo()`** — `web/crm.html:9329`. It's an async function called without `await` or `.catch()`, so a rejection inside any dashboard initializer becomes an unhandled promise rejection with no console context. A `.catch(err => console.error('Navigation failed:', err))` costs nothing.

### What I verified and how

Staging instance on :8031 from this worktree (staging `.env`, symlinked read-only copy of the real CRM databases), driven with Playwright from `~/.venvs/lifeos/bin/python`. Baseline comparisons were done by serving `git show feat/crm-performance:web/crm.html` through Playwright request interception against the *same* API, so only the page JS differs. No writes to :8000 (one read-only `GET /crm` to confirm production still ships plain anchors).

- Click-through Me → Family → Birthdays → Relationship → Me → CRM: **0 document `load` events after the initial one**; URL and `.active` class correct at every hop; `/crm` resolves to `/me`. ✅ (criteria 1 and 5 for nav clicks)
- Long history: 2 forward navs + 5 backs + 5 forwards across Me/Family/Birthdays/Relationship — every step restored the right dashboard and the right active link, **0 document loads total**. ✅ (criterion 3)
- Me → Family → Me → Family request counts: `Family#1 {family/members:1}`, `Me#2 {me/stats:1, people/{me}:1}`, `Family#2 {family/members:1}` — **no repeat `/me/interactions`**. ✅ (criterion 2). Search cache also survives: searching "emily" on `/me`, going to Family and back, keeps the filtered list and the input value with no refetch.
- Direct loads + reload of `/crm/{id}`, `/crm/{id}/timeline`, `/crm/{id}/graph`, `/birthdays?day=03-15`, `/me/timeline`, `/family`, `/relationship`, `/crm`: each rendered the right view/tab with the right active link, identical after `page.reload()`. ✅ (criterion 4)
- Real modifier/aux clicks in Chromium (not the synthetic `dispatchEvent` the test uses): ctrl+click Family → new tab at `/family`; middle-click Birthdays → new tab at `/birthdays`; shift+click Me → new window at `/me`; the originating page never navigated in any of the three. Keyboard focus + Enter on the Relationship link → client-side nav, 0 document loads. ✅
- Timers/streams: `web/crm.html` contains **no `setInterval`** at all, and an instrumented `setInterval`/`EventSource` counter recorded `{interval: 0, es: 0}` across a full nav circuit — nothing to duplicate. ✅
- Rapid double-navigation (Family then Me 50 ms and 300 ms apart; Relationship then Birthdays 100 ms apart) left the correct final view each time — no interleaving corruption found. ✅
- Console: across every scenario the only errors were the two known staging-only conditions — `503` from the photo-profile endpoint and `404 /people//timeline` because this staging config has no partner. No new errors. ✅
- `pytest tests/test_crm_client_nav_ui_browser.py -q` → **9 passed**.
- Mutation test A (strip `onclick="return navigateTo(…)"` from the Family and Me anchors) → **3 failed, 6 passed** (`test_click_through_all_dashboards`, `test_back_and_forward_restore_dashboards`, `test_second_me_visit_does_not_refetch_interactions`). The nav assertions are real.
- Mutation test B (revert `ensureMeHeatmapYears()` to the base semantics) → **1 failed** (`test_second_me_visit_does_not_refetch_interactions`, "got 2 total calls"). The cache test genuinely covers the fix — the fix just has the side effect in finding 1.
- Mutation test C (make `showFamilyDashboard()` stop hiding `#meDashboard`) → **9 passed**. That's finding 5.
- `pytest -m "browser and not requires_server" tests/ -q` → **273 passed, 6091 deselected** (matches the PR's claim; the new file's `slow` marker doesn't exclude it, `scripts/pre-push:155` selects on `browser and not requires_server` only).
- Scope: diff is 4 files, no new dependencies, no test deleted/skipped/weakened, no unrelated reformatting, no personal data in the PR text or tests (`person-synthetic-me` / "Synthetic Owner"). ✅

### Not verified

- The full unit suite (`test.sh`) — the standing brief reserves that for the push lock; I ran the targeted files and the full server-free browser scope instead.
- Behaviour with a real partner and a populated family config: this staging DB has `partner_person_id: ""` and no family members, so the Relationship dashboard's data path and the family member chips could only be checked structurally, not with content.
- Manual "throttled connection" feel-test from the issue's verification section.
- Mobile/touch (`event.button` is 0 for touch taps, so I expect no difference, but I didn't drive a touch emulation profile).

nbramia:
Addressed all nine findings, plus one more found during re-verification (see below). Rebased twice more since (onto #907, then #896) as the integration branch advanced; re-ran the full nav matrix (all nav orders, back/forward, direct loads, /birthdays entry, person selection from each dashboard, year selector, and a Graph-tab hop) on staging against the real data copy after each rebase.

1. **[BLOCKER] Year selector dead** — Fixed. `ensureMeHeatmapYears()`, `loadMeInteractions()`, and `loadMeHeatMap()` now take a `respectExplicitYears` flag; `loadAndRenderHeatMap()` passes `years !== null` through it, so the dropdown's own change handler (which already sets `heatmapYears` before calling down) is no longer clobbered. Verified on staging: select "3" → `GET /me/interactions?days_back=1102`, title "3-Year Interaction History", selection stays "3". Select "1" → `days_back=372`. New test `TestYearsSelectorRespectsUserOverride` (mutation-verified: fails against the pre-fix behavior with `got days_back values: ['737']` when 1102 was expected).

2. **[BLOCKER] Relationship leaked under Family** — Already fixed independently by #899's own review before I rebased onto it (`showFamilyDashboard()` now hides `#relationshipDashboard`/`#relationshipEmptyState`, with a comment crediting that PR's finding 2). No code change needed from this PR; verified on staging (Relationship → Family shows only `#familyDashboard`, `#relationshipDashboard` and `#relationshipEmptyState` both `display: none`) and covered by the strengthened test's Relationship↔Family hops.

3. **[BLOCKER] /birthdays entry left the app half-initialized** — Fixed. `dispatchRoute()` now calls a memoized `ensurePeopleAndStatsLoaded()` (skipped only for the `isBirthdays` branch itself) so any other view reached afterward has people/stats loading under way; `init()`'s own calls record into the same memo so a nav-link click that fires before they resolve doesn't refire them. Verified on staging: /birthdays → Family shows `totalPeople: 15391` and a populated sidebar (previously stuck on "-"/"Loading people..."). New test `TestBirthdaysEntryPoint` (mutation-verified: fails with `Actual value: -` when the `ensurePeopleAndStatsLoaded()` call is removed).

4. **[MAJOR] Active link didn't clear on person selection** — Fixed. `updateUrl()` now calls `updateActiveNavLink()`. Verified on staging from both Me and Family: selecting a person yields `active: [birthdays:false, relationship:false, family:false, me:false]`.

5. **[MAJOR] Test couldn't see a stacked-dashboard leak / no Relationship↔Family hop / no selector or /birthdays-entry coverage** — Strengthened. Every hop in `test_click_through_all_dashboards` and `test_back_and_forward_restore_dashboards` now asserts via a new `_assert_only_visible()` helper that the *other three* dashboard containers are hidden, not just that the target is visible; added explicit Relationship→Family and Family→Relationship hops; added the years-selector test (finding 1), the /birthdays-entry test (finding 3), and a filtered-birthdays-reset test (finding 6, see below). Each new/strengthened assertion was verified by reverting its corresponding fix and confirming the test actually fails (documented in each test's docstring/commit message, not just asserted).

6. **[MAJOR] Nav click no-op when only the query string differs** — Fixed both halves: `navigateTo()` compares `pathname + search` (not pathname alone), and `showBirthdaysPage()` now resets to the Calendar tab when no filter is given (`switchBirthdaysTab(filterDay ? 'timeline' : 'heatmap')` — previously it only ever switched *to* Timeline, never back). Verified on staging: from `/birthdays?day=03-15` (Timeline tab), clicking Birthdays lands on plain `/birthdays`, Calendar tab active. New test `test_birthdays_click_resets_a_filtered_query_string` (mutation-verified against both the pathname-only comparison and, implicitly, the tab-reset omission).

7. **[MINOR] Me from a no-tab URL could leave a stale tab showing** — Fixed: `dispatchRoute()`'s personId branch now calls `switchTab(tab || 'overview', true)` unconditionally instead of only when a tab segment was present.

8. **[MINOR] Stale `/crm#family` in crm-interactions.md** — Fixed.

9. **[MINOR] Silent unhandled rejection from `navigateTo()`'s fire-and-forget `dispatchRoute()`** — Fixed: `.catch(err => console.error('Navigation failed:', err))`.

**Additional finding from my own re-verification** (not in the original review, surfaced only after rebasing onto #896's newly-merged Graph tab): viewing any person's Graph tab, then navigating to Family and Me in quick succession threw ~190-200 `NaN`-attribute console errors from `d3.v7.min.js`, specifically on landing at Me. Root cause: `fitToFirstDegreeNodes()` schedules an 800ms `setTimeout` (to let the force simulation settle) that reads the graph container's size and applies a zoom transform — harmless under a full document reload (which killed the timer), but with dashboard navigation no longer reloading the document, navigating away within that window left the container hidden (0×0) by the time the timer fired, producing a degenerate transform that threw NaN errors out of d3's zoom internals the next time *anything* dispatched a zoom event. Fixed with a freshness/size guard in the deferred callback, plus stopping (and nulling) any still-running graph simulation at `dispatchRoute()`'s single choke point on every navigation (confirmed via instrumented `setAttribute` stack capture, and confirmed the simulation-stop alone did *not* fix this specific bug — it's the deferred-timer guard that does). Not covered by an automated test: this suite's server-free browser tests deliberately stub `window.d3` with an inert proxy so they need no external network access, and that stub never executes real d3-zoom internals — verified manually on staging before and after (exact repro, then clean).

Full unit suite (`5809 passed, 8 skipped`) and server-free browser suite (`297 passed`) both green as part of each push under the shared lock.

nbramia:
## Adversarial review of #909 by Opus — verification pass (head `e3c81b8`)

**Verdict:** BLOCK — but narrowly. All nine original findings are genuinely fixed and mutation-verified, and the extra graph fix kills a 4,590-error console storm that exists on the base branch. Two new things surfaced when I re-drove the matrix, both a few lines each.

### Round-one findings: all verified fixed

| # | Fix verified how |
|---|---|
| 1 year selector dead | select 3 → `GET /me/interactions?days_back=1102`, title "3-Year Interaction History", selection stays "3"; select 1 → `days_back=372`; select 10 → `days_back=3657`. Was inert last round. ✅ |
| 2 Relationship leaked under Family | Relationship → Family → Relationship: only the target container visible at each hop (`rel/relEmpty/me/bday` all `none` on Family; `family` `none` on Relationship). ✅ |
| 3 `/birthdays` entry half-initialized | `/birthdays` → Family → Me: `People: 15391` and a populated sidebar at every step (was `People: -` / "Loading people…"). ✅ |
| 4 stale active link on person selection | From `/me` and from `/family`, clicking a person → `active=[]`; back → `['me']` / `['family']`. ✅ |
| 5 test blind to leaks | `_assert_only_visible()` at every hop + explicit Relationship↔Family hops. Mutation-verified below. ✅ |
| 6 nav click no-op on query-string difference | `/birthdays?day=03-15` → click Birthdays → `/birthdays`, Calendar tab active. ✅ |
| 7 stale tab pane | `/me/timeline` → click Me → `/me`, `tabOverview` visible, Me dashboard `grid`. ✅ |
| 8 stale `/crm#family` | Gone from `crm-interactions.md:83`. ✅ |
| 9 unhandled rejection | `.catch(err => console.error('Navigation failed:', err))` present. ✅ |

The self-found graph fix is real and worth more than it sounds: on `feat/crm-performance`, leaving a person's Graph tab for the Overview tab inside the 800 ms window produces **4,590 console errors** (`<g> attribute transform: Expected number, "translate(NaN,NaN)…"`) at every delay I tried. On this branch that path is **clean at every delay**. The `if (!width || !height) return;` guard earns its place.

### Remaining findings

1. **[BLOCKER] Family → Relationship leaks the Family chrome onto the Relationship empty state — i.e. on every install with no partner configured, which is every fresh clone** — `web/crm.html:12085-12091`. `showRelationshipDashboard()`'s `!PARTNER_PERSON_ID` early return (from #891/#899, not this PR) fires *before* the block that hides `#familyHeroStats` / `#familySelectorContainer` and rewrites the detail header. Under full page loads that never mattered; with client-side nav it does. Measured on staging (`partner_person_id: ""`):
   ```
   direct /relationship : detailName="John Doe"(placeholder) familyHeroStats=[none,0]  familySelector=[none,0]
   on /family           : detailName="Family Dashboard"      familyHeroStats=[flex,55] familySelector=[block,32]
   family -> relationship: detailName="Family Dashboard"      familyHeroStats=[flex,55] familySelector=[block,32]
   ```
   So a fresh user clicking Family then Relationship sees "Family Dashboard" as the title with the family hero-stat row and the family member picker sitting above the "no partner configured" empty state. The strengthened test cannot see this: `_route_api()` stubs `partner_person_id: "person-synthetic-partner"` (`tests/test_crm_client_nav_ui_browser.py:137-142`), so every test takes the configured-partner branch, and `_DASHBOARD_IDS` (`:93`) only tracks the four dashboard containers, not the surrounding chrome. *Fix:* move the empty-state early return below the chrome cleanup (or hoist the cleanup above it), and add a no-partner variant of the click-through — one extra `config` stub is all it takes. This is exactly the gap the round-one suggestion of a single shared "hide everything else" helper was aimed at; the point-wise fix closed the one hop I named and left its neighbour open.

2. **[MAJOR] The new deferred-timer guard still throws: `TypeError: Cannot read properties of null (reading 'clientWidth')`** — `web/crm.html:17968-17970`. `if (!graphSvg || !graphZoom) return;` does not cover this path: `graphSvg` is a d3 *selection*, which stays truthy after its `<svg>` node is detached, so `graphSvg.node().parentElement` returns `null` and the very next line throws. Deterministic repro on staging (uncaught page error, ~3/3 attempts at that timing):
   ```
   open /crm/{id}/graph, wait for nodes, wait 200ms, click Me
   → TypeError: Cannot read properties of null (reading 'clientWidth') at crm.html:17970
   (0/100/300/400ms: clean — it's the window where the container is torn down before the 800ms timer fires)
   ```
   The same null path exists on `feat/crm-performance` via person→person switching (I reproduced it there at 150/200/250 ms), so it is not *introduced* here — but this PR adds a new way to reach it (nav links no longer reload the document) and this is the exact block the PR rewrote and reported clean. *Fix:* one line —
   ```js
   const svgNode = graphSvg.node();
   const container = svgNode && svgNode.parentElement;
   if (!container) return;
   ```

3. **[MINOR] Two of the nine fixes shipped without a regression test** — finding 4 (`updateUrl()` → `updateActiveNavLink()`) and finding 7 (`switchTab(tab || 'overview', true)`). Both were real user-visible regressions, both are one assertion away in the existing server-free harness (add a person to `EMPTY_PEOPLE_PAGE`, or drive `selectPerson()` via `page.evaluate`), and finding 4 is an issue acceptance criterion ("the active nav link styling shall follow the current view"). Findings 1, 3, 5 and 6 all got tests — these two should too.

4. **[NIT, no action needed]** An explicit year selection is *not* preserved across a dashboard round trip: select 3 on Me → Family → Me leaves the selector back at 10 with one `days_back=3657` refetch. I checked this against the base branch's behaviour rather than assuming: a full page load reset it to the span-derived default too, so this matches what shipped before and I would not change it. Flagging only because the follow-up brief expected it to persist — it doesn't, and that's fine.

### What I verified and how

Fresh staging instance on :8031 from this worktree at `e3c81b8` (staging `.env`, symlinked read-only copy of the real CRM databases), driven with Playwright from `~/.venvs/lifeos/bin/python`. Base comparisons served `git show origin/feat/crm-performance:web/crm.html` through Playwright request interception against the same API, so only page JS differs.

- Nav click-through Me → Family → Birthdays → Relationship → Me → CRM: **0 document loads** after the initial one, correct URL + active link at every hop, **zero console errors** (last round's 503/404 pair is gone — `photos_enabled: false` now comes from config, and the relationship page takes the no-partner empty state).
- Long history: 5 backs + 5 forwards across four dashboards — every step correct, **0 document loads**.
- Cache round trip Me → Family → Me → Family: `{family/members:1}`, `{me/stats:1, people/{me}:1}`, `{family/members:1}` — **no repeat `/me/interactions`**.
- Direct load **and reload** of `/crm/{id}`, `/crm/{id}/timeline`, `/crm/{id}/graph`, `/birthdays?day=03-15`, `/me/timeline`, `/family`, `/relationship`, `/crm`: all correct, **zero console errors on every one** (last round these were noisy).
- Real Chromium modifier clicks: ctrl+click → new tab `/family`; middle-click → new tab `/birthdays`; shift+click → new window `/me`; originating page never navigated. Keyboard Enter on a nav link → client-side nav, 0 document loads.
- Instrumented `setInterval`/`EventSource` counters across a full circuit: `{interval: 0, es: 0}`.
- Graph: after `graphSimulation` is stopped and nulled by `dispatchRoute()`, going back to the Graph tab re-renders (51 nodes, `graphSimulation: alive`) and dragging a node — which calls `graphSimulation.alphaTarget()` — throws nothing. The stop-and-null does not strand the drag handlers.
- `pytest tests/test_crm_client_nav_ui_browser.py -q` → **12 passed**.
- Mutation A (delete `showFamilyDashboard()`'s `#relationshipDashboard`/`#relationshipEmptyState` hiding — the round-one bug): → **1 failed**, `Left contains one more item: '#relationshipDashboard'`. The strengthened assertions are real.
- Mutation B (delete `ensurePeopleAndStatsLoaded()` from `dispatchRoute()`): → **1 failed** (`TestBirthdaysEntryPoint`).
- Mutation C (`loadMeHeatMap(years !== null)` → `loadMeHeatMap(false)`): → **1 failed** (`TestYearsSelectorRespectsUserOverride`, `['737']` instead of the expected window).
- `pytest -m "browser and not requires_server" tests/ -q` → **297 passed, 6204 deselected** (matches the PR's claim).
- Scope/docs: 5 files, no new dependencies, no test weakened or skipped; `crm-ui.md` Navigation section and both stale-URL fixes read correctly; `crm-interactions.md:83` now says `/family`. Synthetic data only in tests (`person-synthetic-me`, `person-synthetic-partner`).

### Not verified

- The full unit suite (reserved for the push lock); I ran the targeted file and the whole server-free browser scope.
- The configured-partner Relationship dashboard's own data path — this staging config has no partner, so I could only exercise the empty state (which is how I found finding 1).
- Throttled-connection feel test; touch/mobile.

nbramia:
Addressed both remaining findings, plus the two missing regression tests.

1. **[BLOCKER] Family chrome leaked onto the no-partner Relationship empty state** — Fixed: moved `showRelationshipDashboard()`'s chrome cleanup (hiding `#familyHeroStats`/`#familySelectorContainer`, resetting the detail header) above the `!PARTNER_PERSON_ID` early return, so it runs regardless of whether a partner is configured. Verified on staging (no-partner config): Family → Relationship now shows `familyHeroStatsDisplay: 'none'`, `familySelectorDisplay: 'none'`, `detailName: 'Relationship Dashboard'`. New test `TestRelationshipNoPartnerChrome` (mutation-verified: fails with `familyHeroStats` visible when the early return is reverted to the top).

2. **[MAJOR] Null graph container in the deferred zoom-fit guard** — Fixed: `graphSvg.node()` and its `parentElement` are now explicitly null-checked before reading `clientWidth`/`clientHeight`, since `graphSvg` (a d3 selection) stays truthy after its node is detached. Re-verified at 0/100/200/300/400ms delays between opening a Graph tab and clicking Me: zero console errors at every delay (confirmed via `browser_console_messages` after waiting past the 800ms+500ms timer window each time).

3. **[MINOR] Missing regression tests** — Added both: `test_selecting_a_person_from_me_clears_the_active_dashboard_link` / `..._from_family_...` (finding 4) and `test_me_from_a_no_tab_url_shows_overview_not_a_stale_tab` (finding 7). Each mutation-verified against a reverted version of its fix.

Full unit suite (`5809 passed, 8 skipped`) and server-free browser suite (`301 passed`) green as part of the push. Pushed at `79fd0de` (was `e3c81b8`).

nbramia:
## Adversarial review of #909 by Opus — final verification (head `79fd0de`)

**Verdict:** APPROVE

Both remaining findings are fixed, all three previously-untested fixes now have mutation-verified regression tests, and nothing I could reach regressed.

### The two findings

1. **Family chrome on the no-partner Relationship empty state — fixed.** The `!PARTNER_PERSON_ID` early return moved below the chrome cleanup, and `loadRelationshipPartnerPhoto()` moved below it too so the no-partner path still issues no partner-scoped request. Measured on staging (`partner_person_id: ""`), the client-nav result is now byte-for-byte the direct-load result:
   ```
   direct /relationship  : detailName="Relationship Dashboard" familyHeroStats=[none,0] familySelector=[none,0] heatmap=[none,0] tabs=none
   family -> relationship: detailName="Relationship Dashboard" familyHeroStats=[none,0] familySelector=[none,0] heatmap=[none,0] tabs=none
   ```
   (Was `detailName="Family Dashboard"` with a 55 px hero-stat row and a 32 px member selector.) Incidentally the direct-load empty state also stopped showing the `"John Doe"` placeholder header it had on the base branch — a small improvement, not a regression.

2. **Deferred graph zoom-fit throwing on a null container — fixed.** `const svgNode = graphSvg.node(); const container = svgNode && svgNode.parentElement; if (!container) return;` — exactly the shape needed. Graph → Me at **0/100/200/300/400 ms: clean at every delay** (200 ms was a deterministic `TypeError: Cannot read properties of null (reading 'clientWidth')` before). The guard also repairs the *pre-existing* variant of this on the base branch: person→person switching from the Graph tab at 150/200/250 ms is now clean too, where `feat/crm-performance` throws.

3. **Missing regression tests — added and real.** Three new tests, each verified by reverting its fix:
   - `test_selecting_a_person_from_{me,family}_clears_the_active_dashboard_link` → **2 failed** when `updateActiveNavLink()` is removed from `updateUrl()`.
   - `test_me_from_a_no_tab_url_shows_overview_not_a_stale_tab` → **1 failed** when `switchTab(tab || 'overview', true)` is reverted to `if (tab) …`.
   - `TestRelationshipNoPartnerChrome` → **1 failed** (`unexpected value "visible"`) when the no-partner early return is put back above the chrome cleanup. Good call parameterizing `_route_api(partner_configured=…)` so the default stays the configured-partner path and only this class takes the fresh-install one.

### What I re-ran

- `pytest tests/test_crm_client_nav_ui_browser.py -q` → **16 passed** (was 12).
- Nav click-through Me → Family → Birthdays → Relationship → Me → CRM with console capture: **0 document loads, 0 console errors**, correct URL + active link + single visible dashboard at every hop.
- Full round-two matrix re-driven at this head, all still correct and all console-clean: Relationship↔Family isolation, `/birthdays` entry → Family → Me (`People: 15391`, populated sidebar), `/birthdays?day=03-15` → Birthdays resets to `/birthdays` + Calendar tab, `/me/timeline` → Me shows Overview, person selection from Me and from Family clears the active link, year selector 3 → `days_back=1102` and the selection sticks.
- `pytest -m "browser and not requires_server" tests/ -q` → **300 passed, 1 failed**: `test_backend_selector_ui_browser.py::TestSidebarPersonaValidationRace::test_listing_waits_for_persona_validation_before_firing`. **Unrelated flake, not a finding** — that file never loads `crm.html`; the test passes alone (1 passed) and its whole file passes (39 passed) on this same checkout. Flagging only so it isn't mistaken for fallout from this PR.

### Not verified

Full unit suite (reserved for the push lock); the configured-partner Relationship data path (staging has no partner configured); throttled-connection and touch/mobile.

Issue #863, opened 2026-09-04
fix(agents): correct session labels, routing, host, task links, and summaries


nbramia:
## Problem

The `/agents` page mislabels and misattributes sessions in ways that make the graph unreadable and the board's joins unreliable. A session parked waiting for the operator's answer is labelled as if it were running Sonnet. Codex sessions and any session registered from another machine show a bare question mark. Locally scanned Claude Code sessions carry their own id in the task field, so they can never join a board card and pollute the join map. A worker session that was ssh-spawned onto another machine still reports the API host. Codex sessions never receive an AI summary because the prefetcher skips them, and summary failures are retried forever without being remembered. The "recent" chip ignores CLI sessions that ended, and the route filter cannot isolate Hermes or waiting-on-operator sessions. Remote DeepSeek sessions say only "Remote" even though the configured model label exists.

These are data and labelling defects that any graph redesign would inherit, so they are fixed first and independently.

## Solution

Make every snapshot row carry a correct human label, routing label, host, task link, and model description, and make the summary pipeline cover every session kind and remember its failures.

- The snapshot's routing-to-model label gains explicit arms for `ask`, `hermes`, `remote`, `claude_code`, `codex`, and `code`; the frontend's routing label gains `ask`; the remote arm returns the configured remote model label.
- The CLI model label never degrades to a bare `?`; an unknown model yields the engine name.
- Worker session rows report the session's own host when it has one, falling back to the API host.
- Locally scanned Claude Code rows leave `task_id` empty unless a real task link exists.
- The summary prefetcher handles Codex ids the same way the summary route does; summary failures and empty-transcript results are cached for terminal sessions, where "terminal" includes the CLI statuses `ended`, `inactive`, and `completed`.
- The "recent" chip counts `ended` alongside `completed`; the route filter offers `hermes`, `remote`, and `ask`.
- The snapshot exposes the session's assigned model, effort, conversation id, bot, and origin so later phases can render them.

## Acceptance Criteria

- [ ] Given a worker session with routing `ask`, when `GET /api/agents/snapshot` runs, then `model_label` is `"Waiting on you"` and the frontend routing label is `"Ask"`, never a Claude model name.
- [ ] Given a worker session with routing `remote`, `model_label` equals the configured remote model label; with routing `hermes`, it equals `"Hermes"` plus the model Hermes last reported when known.
[...]

PR #913, opened 2026-09-04
fix(scripts): give the pre-push gate per-run log paths (#908)

### LINKED ISSUE #908: fix(scripts): give the pre-push gate a per-run log path so concurrent worktrees cannot collide
## Problem

The pre-push gate writes its pytest output to a single fixed path, `/tmp/lifeos-prepush-tests.log`,
shared by every worktree, branch, and concurrent agent on the machine. Two consequences:

1. **Results are unattributable.** During #863's lifecycle the gate appeared to report 6153, then
   5620, then 5673 passing unit tests on the same branch, while that branch collected 6174. This read
   as the gate silently narrowing its selection. It was not: the 5620/5673 runs came from a different
   worktree (`agent-a4ea6d87eb72eb3b4`, branch `fix/list-avatars`, off `main`), whose own collection
   count is exactly 5652. The log's warning lines still named that other worktree's paths. Confirming
   the gate was honest cost a full 18-minute suite run to reproduce independently.
2. **A failure log can be clobbered before it is read.** When two pushes overlap, the second run
   overwrites the first's output. An engineer or agent going back to read why their push failed can
   find someone else's passing run instead.

This is a real hazard on this box specifically, because several agent worktrees push concurrently by
design.

## Solution

Key the log filename on something unique per run — the branch name, the worktree path hash, or a
PID/timestamp suffix — so concurrent gates cannot collide. Print the resolved path in the gate's
output so a reader knows which file belongs to their run.

## Acceptance Criteria

- [ ] Given two pre-push gates running concurrently from different worktrees, each writes to a
      distinct log file and neither overwrites the other.
- [ ] The gate's console output names the log file it wrote, so a failure can be traced to its run.
- [ ] Given a gate run from a worktree, the log path is derivable from the branch or worktree without
      guessing.
- [ ] Old log files do not accumulate without bound (reuse per branch, or prune on a simple age rule).
- [ ] The gate's pass/fail behavior and test selection are otherwise unchanged.

## Verification

- **Automated tests:** if the repo has coverage for `scripts/pre-push`, extend it to assert the
  resolved path differs for two different branch names.
- **Manual checks:** run two gates concurrently from two worktrees on different branches and confirm
  two distinct, complete logs.

## Scope

**In scope:** the log path and the message naming it in `scripts/pre-push`.
**Out of scope:** which tests the gate selects, its markers, `-n auto`, and the `--ff` ordering — all
of those are correct and were exonerated by this investigation.

## Technical Context

- **Key file:** `scripts/pre-push` — the hardcoded `/tmp/lifeos-prepush-tests.log`.
- The hook runs `-m "unit and not slow" tests/ -n auto --ff`; `--ff` only reorders and never narrows.
  The hook's own comment records that an earlier last-failed-only flag was deliberately removed for
  exactly this class of confusion, so the selection logic is not at fault here.
- **Constraints:** no new dependencies; the gate must stay usable from any worktree.

## References

- Diagnosed during PR #890's verification (issue #863). See PR #890's verification summary for the
  reproduction and the exact collection counts.

nbramia:
## Summary

`scripts/pre-push` wrote its pytest output to two fixed paths, `/tmp/lifeos-prepush-tests.log` and `/tmp/lifeos-prepush-browser.log`, shared by every worktree, branch, and concurrent agent on the machine. Several agent worktrees push concurrently here by design, so one run's log overwrote another's and a gate result could be attributed to the wrong branch. That is not theoretical: during #863's lifecycle the gate appeared to report 6153, then 5620, then 5673 passing tests on a branch that collects 6174 — the low counts came from a different worktree on `fix/list-avatars`, and confirming the gate was honest cost a full 18-minute suite run.

Each run now writes to its own path, prints it, and can no longer collide with another worktree's.

## What changed

**`scripts/pre-push`** — log destination only. Test selection is untouched.

- Logs go to `${TMPDIR:-/tmp}/lifeos-prepush/<branch-slug>-<pid>-unit.log` and `-browser.log`. The branch comes from the `local_ref` git feeds the hook on stdin (falling back to the checked-out branch, then `detached`/`unknown`), sanitised so anything outside `[A-Za-z0-9._-]` — slashes, a hostile `../../etc/passwd` — collapses to one safe filename component, capped at 80 characters. The PID is what guarantees uniqueness between concurrent runs; the slug is what makes the path derivable without guessing.
- Each suite prints `Log: <path>` before it runs, and `fail()` still prints `Full output: <path>` — now honestly distinguishing a log with content, an empty log, and no log at all, instead of a blank `FAILED ()`.
- The directory is created mode 700 and tightened unconditionally afterwards (`mkdir -m` does not touch a directory that already existed). Logs are pruned with `find -maxdepth 1 -type f -name '*.log' -mtime +3 -delete`, which starts deleting at the 4-day mark and can never reach a live run's log or a file this hook did not write.
- **A log-directory problem can never block a push.** If the configured directory is unusable, the hook degrades through a fixed fallback, a throwaway `lifeos-prepush.XXXXXX` directory, a bare `mktemp -d`, and finally the shared `/tmp`, printing one line that names the specific reason (`not a directory`, `not writable`, `is a symlink`, `could not be created`) and the location it settled on.
- **It never `chmod`s or prunes a directory it does not own.** A `LOG_DIR_OWNED` flag gates both destructive operations; the bare-`/tmp` last resort and any symlinked directory are marked not-owned, and a log written there carries a `lifeos-prepush-` filename prefix so it stays attributable in a shared location.

**`tests/test_prepush_gate.py`** — 13 tests → 39. New coverage for path derivation and per-run uniqueness, filename sanitisation and the length cap, the unit/browser log split, the announced path (observed from a real end-to-end hook run, not a source grep), the prune's window and scope, the directory mode, every fallback rung, the ownership gate in both directions, and each `fail()` branch. Every test points `TMPDIR` at its own `tmp_path`, so the suite never touches the real shared log directory.

**Docs** — `AGENTS.md` and `docs/specs/standards/testing-standards.md` describe the per-run path, the printed announcement, the 4-day prune, the fallback rungs, the ownership rule, and the known limit that non-ASCII or very long branch names collapse in the slug (uniqueness still comes from the PID).

## Acceptance criteria

| Criterion | How it is met | Proven by |
|---|---|---|
| Concurrent gates from different worktrees write distinct logs | Path keyed on branch slug **and** PID | `test_same_branch_concurrent_runs_do_not_collide`; three genuinely overlapping real hook runs produced 6 distinct files with zero cross-talk |
| Console output names the log it wrote | `Log: <path>` before each suite; `Full output:` on failure | `test_console_output_names_log_path` — a real run with a stubbed `python`, asserting against files that exist |
| Path derivable from the branch without guessing | Derived from the pushed `local_ref`, sanitised | `test_log_path_derivable_from_branch_name`, `test_fallback_branch_when_no_ref_on_stdin`, `test_log_dir_is_pinned_to_lifeos_prepush` |
| Old logs do not accumulate without bound | `-mtime +3` prune scoped to `*.log` at `maxdepth 1` | `test_prune_removes_old_logs_but_keeps_recent`, `test_prune_only_touches_its_own_log_files` |
| Pass/fail behaviour and test selection otherwise unchanged | No edit to markers, `-n auto`, `--ff`, `-p no:cacheprovider`, the two-process split, or exit codes | Both pytest invocations are byte-identical to the base revision by md5; `--lf` has not reappeared; happy-path output is byte-identical across every revision of this branch |

## Verification

- The real hook was driven end to end at every stage: normal push, unit failure, browser failure (which names the browser log without clobbering the unit log), docs-only skip, deletion-only skip, tag push, three-way concurrency, and a fully degraded log directory. Every scenario behaved as described, and no fallback rung can block a push.
- The live gate ran the full suite clean on each push of this branch — most recently 6191 passed, 9 skipped, plus 362 server-free browser tests.
- Every new or modified test carries a mutation proof: the production hunk it covers was reverted, the test was shown to fail, and it was restored and shown to pass.
- Three rounds of adversarial review (two independent Opus reviewers per round, one briefed hostile) plus a final verification pass. Rounds 1 and 2 each found a real blocker, both fixed here; round 3 found none and both reviewers recommended merge.

## Scope

Out of scope, deliberately: `scripts/test.sh`'s fixed `/tmp/lifeos_test_server.pid` is the same class of shared-temp-path hazard in a different script and is left for a follow-up.

Closes #908


nbramia:
## Adversarial review — round 1 (head `e04ac69`)

Two independent Opus reviewers: a **general** reviewer and a **hostile** reviewer briefed to assume the diff is broken and prove it by running the hook rather than reading it. Codex cross-review was unavailable this milestone (workspace out of credits), so the hostile pass substitutes for it.

Both reviewers independently reproduced the same single blocker, and both independently found the same three test-quality gaps. Negative evidence was substantial: concurrency (AC1), failure traceability for both suites (AC2), branch-name fuzzing (15 hostile refs including injection canaries, `../../etc/passwd`, tags, unicode, 300-char names), prune safety (no symlink following, no cross-run deletion), and byte-identical test selection (AC5) all held up under direct attack.

### Referee table

| # | Finding | Reviewers | Verdict |
|---|---------|-----------|---------|
| 1 | An unwritable/occupied `$LOG_DIR` aborts a legitimate push and reports it as `FAILED ()` with an empty summary pointing at a nonexistent file | general, hostile | **Action Required** |
| 2 | `test_console_output_names_log_path` is vacuous — greps the script's source text; disabling the echo at runtime while keeping the substring leaves all 22 tests passing, so AC2 has no real coverage | general (Minor), hostile (proved by mutation) | **Action Required** |
| 3 | AC4 (pruning) has no test — deleting the whole `find` line is a silent no-op for the suite | general, hostile | **Action Required** |
| 4 | The documented log directory is unpinned — changing `LOG_DIR` to bare `${TMPDIR:-/tmp}` (contradicting both docs) passes the whole suite | hostile | Accepted |
| 5 | The new tests run against the operator's real `/tmp/lifeos-prepush/` and execute the prune there nine times per run; no `TMPDIR` override | general | Accepted |
| 6 | The `very_long` parametrization asserts nothing length-related — a mutant with the 80-char cap removed still passes with a 317-char filename | general | Accepted |
| 7 | `mkdir -p -m 700` does not tighten an existing directory; log files land 664 | general, hostile | Accepted (fold into #1) |
| 8 | Comment says "older than 3 days" but `find -mtime +3` deletes at 4 days (measured: 95h survives, 96h goes); neither doc mentions pruning at all | hostile (M1), general (#6) | Accepted |
| 9 | The prune deletes *every* file older than the window, not just logs | general | Accepted |
| 10 | `docs/specs/standards/testing-standards.md` `Last Updated` not bumped | general | Accepted |
| 11 | `assert "/" not in Path(log).name` can never fail (`Path.name` is by definition a single component) | general | Accepted (fold into #4) |
| 12 | Unicode and >80-char branch names collapse to indistinguishable slugs | general, hostile | **Rejected** — uniqueness is carried by the PID, not the slug; AC1/AC3 both hold. Adding a hash is speculative complexity for a repo whose branch convention is lowercase ASCII. Worth one sentence as a known limit, nothing more. |
| 13 | Trailing slash on `TMPDIR` yields a doubled separator in the printed path | general | **Rejected** — cosmetic. |
| 14 | A branch beginning with `-` yields a filename beginning with `-` | general | **Rejected** — every use inside the hook is a quoted absolute path; no injection (both canaries came back clean). |
| 15 | Subdirectories under the log dir are never pruned | hostile | **Rejected** — nothing creates subdirectories there. |

### Action Required — detail

**1. A silenced `mkdir` failure turns an unwritable log directory into a blocked push.**

`mkdir -p -m 700 "$LOG_DIR" 2>/dev/null || true` discards every failure, then `> "$LOG"` fails under `set -e`, routes into `fail()`, and `tail -1 "$LOG"` fails too. Reproduced by both reviewers, three ways (`$TMPDIR/lifeos-prepush` exists as a regular file; `$TMPDIR` read-only; the dir exists mode 500), and confirmed as a **regression from base** — the base hook, same hostile `TMPDIR`, exits 0:

```
=== TMPDIR is a FILE         rc=1
Running unit tests... line 174: .../lifeos-prepush/hostile-real-3990089-unit.log: Not a directory
tail: cannot open '...' for reading: Not a directory
FAILED ()
  Full output: .../lifeos-prepush/hostile-real-3990089-unit.log
```

Two things compound it: `FAILED ()` with an empty parenthetical reads as a test regression rather than a filesystem problem, and the `Full output:` line — the one diagnostic this PR exists to provide — names a file that does not exist. `/tmp` is `drwxrwxrwt`, so on a fresh machine any local process can permanently block every LifeOS push with `touch /tmp/lifeos-prepush`. It does not fire under this host's default environment (`TMPDIR` unset), but `TMPDIR` is now a documented input of this feature in two docs.

Fix: probe the directory after creating it and **degrade instead of aborting** — fall back to a usable location, print one explicit line saying what happened, and make `fail()` say the log is missing rather than printing `FAILED ()`. The fallback logic must itself be `set -e`-safe.

**2 and 3. Two of the nine new tests assert nothing.** Mutants that pass the full suite:

```
M1  echo "Log: $LOG"  ->  if false; then echo "Log: $LOG"; fi   =>  22 passed
M2  find "$LOG_DIR" -maxdepth 1 -type f -mtime +3 -delete       =>  (deleted) 22 passed
M4  LOG_DIR="${TMPDIR:-/tmp}/lifeos-prepush" -> "${TMPDIR:-/tmp}" => 22 passed
```

The console test's docstring justifies the source-grep by saying a runtime check "would require a full pytest invocation". That is not true: stubbing `python` on `PATH` and pointing `HOME` at an empty directory makes the **real, unmodified hook** run end to end and print both `Log:` lines in well under a second, collecting nothing. The hostile reviewer demonstrated it. That harness also gives you a clean way to test finding #1's fallback, the resolved directory, and a genuine failure path — all of which plan-only mode cannot reach (it returns `rc=0` even when the log directory is a regular file).

For the prune, the stated reason for skipping a test — that it would touch the shared live log directory — does not hold: point `TMPDIR` at a pytest `tmp_path`, create files with `touch -d`, and assert which survive. Doing that also fixes finding #5.

### What must not change

Test selection, markers, `-n auto`, `--ff`, `-p no:cacheprovider`, the two-pytest-process split, and exit semantics are byte-identical to base and were exonerated. Keep them that way. Do not weaken, skip, or delete any existing test.

Every new or modified test needs a mutation proof recorded in the follow-up: revert the specific production hunk, show the test FAILS, restore, show it passes. A test whose mutation only fails because a whole-file revert removed unrelated output is not a proof — the mutation must make the *positive* assertion fail.

nbramia:
## Adversarial review — round 2 (head `8b752cf`)

Same roster as round 1: an Opus **general** reviewer and an Opus **hostile** reviewer, both briefed to treat the round-1 fix delta as the prime suspect. They converged independently on one blocker and on three coverage gaps behind it.

**Round 1's remedy introduced a regression that is worse than the bug it fixed.** The new fallback chain can resolve `LOG_DIR` to the literal `/tmp`, and the hook then runs `chmod 700` and a `find … -name '*.log' -mtime +3 -delete` sweep inside the shared system temp directory. Neither statement could reach a directory the hook does not own at base or at `e04ac69`.

Everything else in the delta holds. The happy path is byte-identical to `e04ac69` (both reviewers diffed real runs, pass and fail, and got IDENTICAL), the new `set -e` constructs are safe, the original blocker is genuinely fixed, concurrency still holds on the new code, and no pre-existing test was weakened.

### Referee table

| # | Finding | Reviewers | Verdict |
|---|---------|-----------|---------|
| 1 | Fallback bottoms out at literal `/tmp`; the hook then `chmod 700`s it and prunes `*.log` inside it | general (AR1), hostile (AR-1) | **Action Required** |
| 2 | The whole `fail()` rewrite is untested — reverting it to the pre-round-1 version leaves 31/31 green | hostile (AR-2) | **Action Required** |
| 3 | The unconditional `chmod 700` is untested — replacing it with `chmod 777`, the exact state it was added to prevent, leaves 31/31 green | general (#4), hostile (AR-3) | **Action Required** |
| 4 | Both docs claim a 3-day retention window; the same commit's code comment correctly says deletion starts at 4 days (measured: a 95h log survives) | general (#7), hostile (AR-4) | **Action Required** |
| 5 | `mkdir -p` lost the `-m 700` it had at `e04ac69`, so the directory exists at the umask mode until the later `chmod` | general (#2) | Accepted |
| 6 | `chmod` follows symlinks, so a planted `$LOG_DIR` symlink retargets the tightening | general (#3) | Accepted (fold into #1) |
| 7 | The fallback message never says *why* the directory was unusable, and at rungs 3–4 silently omits that the fixed fallback also failed | hostile (R-1) | Accepted |
| 8 | Fallback rungs 3 and 4, and the `[ "$LOG_DIR" = "$UNUSABLE_LOG_DIR" ]` equality guard, are unreachable in tests because every fallback test sets `LIFEOS_PREPUSH_TEST_FALLBACK_DIR` to a fresh writable dir | general (#5), hostile (M-1) | Accepted |
| 9 | Docs describe the log directory as unconditional and don't mention the fallback shapes | general (#8), hostile (R-4) | Accepted |
| 10 | `mktemp -d` fallback directories accumulate forever | general (#6), hostile (R-3) | **Rejected** — one directory per *degraded* push, which should be rare and already signals something wrong. Adding recursive directory deletion to a git hook is a larger risk than the leak; that is precisely the class of change that produced finding #1. |
| 11 | `LIFEOS_PREPUSH_TEST_FALLBACK_DIR` steers production code / should be inverted and documented | hostile (R-2) | **Rejected** — it matches three existing conventions in the same file (`LIFEOS_PREPUSH_PLAN_ONLY`, `_CHANGED_FILES`, `_HAVE_CONTENT`), the `_TEST_` infix makes intent unmistakable, and it is read only inside an already-degraded branch. The general reviewer independently reached the same conclusion (#9). Inverting it is churn. |
| 12 | A relative `TMPDIR` puts the log directory inside the repo | hostile (M-2) | **Rejected** — pre-existing at base and at `e04ac69`, and a relative `TMPDIR` is malformed input. |

### Action Required — detail

**1. Never adopt, `chmod`, or prune a directory the hook did not create.**

```
scripts/pre-push:81-82   || mktemp -d 2>/dev/null || echo "/tmp")
scripts/pre-push:89      chmod 700 "$LOG_DIR" 2>/dev/null || true
scripts/pre-push:97      find "$LOG_DIR" -maxdepth 1 -type f -name '*.log' -mtime +3 -delete 2>/dev/null || true
```

Both `mktemp` calls resolve through `$TMPDIR` (the second because GNU `mktemp -d` consults `$TMPDIR` before `/tmp`), so a single condition — `$TMPDIR` pointing at a directory the user cannot write — kills both rungs. Reaching there additionally needs `/tmp/lifeos-prepush` to be unusable, which on a multi-user or containerised box is ordinary — and note the causal loop round 1 built: **the new unconditional `chmod 700` is itself what makes a shared `/tmp/lifeos-prepush` unusable for a second user**, since the first pusher's hook locks it to their uid.

Reproduced on the real hook with stub `find`/`chmod` on `PATH`, and again with `LIFEOS_PREPUSH_TEST_FALLBACK_DIR` unset and only the one production literal relocated:

```
Pre-push: log directory …/ro/sub/lifeos-prepush is unusable; using /tmp instead
STUB-CHMOD-ARGV: 700 /tmp
STUB-FIND-ARGV: /tmp -maxdepth 1 -type f -name *.log -mtime +3 -delete
```

With the real `find`, in a relocated sandbox populated with five-day-old files belonging to other programs:

```
--- BEFORE: fresh.log  important-notes.txt  lifeos-prepush-tests.log  nginx-access.log  other-agent-pytest.log  somedir/
--- AFTER:  fresh.log  important-notes.txt  hostile-r2-395074-unit.log  somedir/        (dir mode 775 -> 700)
```

Three unrelated files deleted, directory silently tightened. `/tmp` on this box currently holds 40 `*.log` files written by other agents — including `lifeos-prepush-tests.log` and `lifeos-prepush-browser.log` — spared today only because none has aged past the window. That is luck, not a guard. And `chmod 700 /tmp` fails harmlessly as uid 1000 but *succeeds* for a root invoker in a container or CI image, making `/tmp` unusable for every other account.

Required fix: track whether the hook created the directory it is using, and gate BOTH the `chmod` and the `find` on that. A last-resort location the hook does not own may still be used to *write* a log — never blocking the push remains the priority from round 1 — but it must never be chmod'd or swept. A symlinked `$LOG_DIR` must count as not-owned too (finding #6). Restore `-m 700` on both `mkdir -p` calls while keeping the unconditional `chmod` on owned directories (finding #5) — they are complementary, not alternatives.

**2 and 3. Round 1's two behavioural additions are invisible to the suite.** Mutants that leave 31/31 green:

```
M11  fail() reverted entirely to the pre-round-1 two-liner      => 31 passed
M2   fail()'s empty-log branch deleted                          => 31 passed
M1   chmod 700 -> no-op (`true`)                                => 31 passed
     chmod 700 -> chmod 777 (the exact state #7 was about)      => 31 passed
```

The addresser's report claimed the "no log at all" branch was covered indirectly by the `readonly_tmpdir` case. M11 disproves it: with the fallback in place that case reaches a usable directory and `fail()` is never called, so none of its three branches is exercised. Both are cheap with the harness already in the file — add a stub mode that exits 1 while writing nothing, and assert `(tmp_path / "lifeos-prepush").stat().st_mode & 0o777 == 0o700` in an existing real-run test.

Also untested, and to be closed in the same pass (finding #8): fallback rungs 3 (`mktemp`) and 4, and the `[ "$LOG_DIR" = "$UNUSABLE_LOG_DIR" ]` equality guard — removing that guard leaves the suite green. Reach rung 3 without the test override by setting `TMPDIR` to a directory whose `lifeos-prepush` child is occupied. After the #1 fix, add the positive-and-negative pair: an owned directory IS tightened and swept; a not-owned last-resort location is NOT.

**4. Docs state a window the code does not have.** `AGENTS.md:274` and `docs/specs/standards/testing-standards.md:91` both say "logs older than 3 days are pruned"; `-mtime +3` first deletes at 96h. Both sentences now contradict the code comment beside the line they describe. Fix the number in both, and describe the fallback accurately while you are there (finding #9) — a reader who finds a log under `lifeos-prepush.XXXXXX` currently has nothing to match it against.

### Guard against a third-round regression

Round 2's blocker came from round 1's remedy. The #1 fix touches the same block again, so: verify the happy path stays byte-identical to `e04ac69` (both reviewers established that baseline with real runs), and when you add the ownership guard, test the POSITIVE case as well as the negative one — a test asserting only that something did *not* happen passes vacuously.

Every new or modified test needs a targeted mutation proof recorded: mutate the specific hunk so the POSITIVE assertion fails, show the failure, restore, show it passes.

nbramia:
## Adversarial review — round 3 (head `bf294f7`)

Same roster: an Opus **general** reviewer and an Opus **hostile** reviewer, both briefed that round 1's remedy caused round 2's blocker and that round 2's rewrite of the same block was therefore the prime suspect.

**Neither reviewer found an Action Required defect. Both recommend merge.** Round 2 did not repeat the pattern.

Independently established by both:

- **Happy path byte-identical across all three heads** (`e04ac69`, `8b752cf`, `bf294f7`) — normalised stdout, the produced file tree with modes, and every log file's contents, for a passing *and* a failing run. `LOG_FILE_PREFIX` is empty there, the directory is `drwx------`, stderr is empty, and no degraded-path message leaks.
- **The ownership gate holds at every reachable rung.** Six assignment paths enumerated; `chmod`/`find` argv-recording stubs fired for the four owned shapes and **not once** for the two not-owned ones (bare-`/tmp` last resort, symlinked directory). Round 2's blocker is closed.
- **Exit-status matrix correct** across unit pass/fail, browser fail, empty log, playwright absent, degraded rung + failure, docs-only, deletion-only, empty stdin. No passing run reported as failed; no failing run exited 0. Every `set -e` construct in the restructured block is safe.
- **Concurrency holds on the new code**, including one owned and one not-owned run writing into the same physical directory: 8 simultaneous runs, 16 unique announced paths, no misattribution, prune correct, mode 700.
- **Mutation testing.** 20 mutants (general) and 30 (hostile); 15 and 27 killed respectively. **Every mutant that survived rounds 1 and 2 now dies** — `chmod`→no-op, `chmod 777`, `fail()` reverted, the empty-log branch, the equality guard, both `mktemp` rungs, the ownership gate, the symlink demotion. Both reviewers reproduced all six of the addresser's mutation-proof claims.
- **Nothing from earlier rounds was weakened.** The round-2 test diff removes one assertion and replaces it with a strictly stronger one (`"is unusable; using"` → `"is unusable ({reason}); using"`).
- Test selection, markers, `-n auto`, `--ff`, `-p no:cacheprovider`, the two-process split and exit semantics remain byte-identical to base; no old fixed shared log filename survives anywhere; no personal paths, usernames or hostnames; the working tree is clean and a full suite run leaves the real `/tmp` and `/tmp/lifeos-prepush` untouched.

### Referee table

| # | Finding | Reviewers | Verdict |
|---|---------|-----------|---------|
| 1 | `LOG_FILE_PREFIX` is documented in two files and asserted nowhere — three mutants survive 39/39, including one that puts the prefix on the **happy path** filename (contradicting the "owned-case naming is byte-identical" claim), one that removes it entirely, and one that prefixes the unit log but not the browser log | general (R1), hostile (R-1) | **Accepted — fold in** |
| 2 | Nothing asserts the happy path prints no degraded-path line; injecting a spurious message leaves 39/39 green | general (R2) | **Accepted — fold in** |
| 3 | `test_distinct_fixed_fallback_failure_reaches_mktemp_rung`'s "fixed fallback's own reason" assertion is vacuous — both directories fail for the identical reason, so the configured dir's message satisfies it; dropping the fallback's reason leaves the suite green | general (R3) | **Accepted — fold in** |
| 4 | Docs omit the bare-`mktemp` rung (`tmp.XXXXXXXXXX`), the hardest output to recognise; and both scope the `lifeos-prepush-` prefix to the `/tmp` last resort when the code applies it to any not-owned directory, including a symlinked one | general (R4), hostile (M-3) | **Accepted — fold in** |
| 5 | `LOG_DIR_OWNED=1` means "manages", not "created": a pre-existing directory at the configured or fixed-fallback path is adopted, then chmod'd and pruned | general (M3), hostile (M-1) | **Rejected** — both reviewers analysed the blast radius and both declined to raise it. The path is namespaced `lifeos-prepush`, the sweep is `-maxdepth 1 -type f -name '*.log' -mtime +3`, `chmod` fails silently on another uid's directory, and the hook itself only ever creates mode-700 directories, so a second user falls through to the `mktemp` rung rather than adopting. Unlike round 2's blocker this can never reach bare `/tmp`, which `test_log_dir_is_pinned_to_lifeos_prepush` guards. |
| 6 | `_prepush_unusable_reason` reports "is a symlink" even when the real cause is an unwritable or dangling target | hostile (M-2) | **Rejected** — cosmetic, and it requires editing the production block again. Every rewrite of that block has produced the next round's blocker; the remaining churn risk outweighs the message nit. |
| 7 | `-m 700` on both `mkdir -p` is untested | general (M1), hostile (M-4) | **Rejected** — an equivalent mutant. The unconditional `chmod 700` produces the identical end state; `-m` only narrows a race window no after-the-fact assertion can observe. |
| 8 | TOCTOU between the `[ -L "$LOG_DIR" ]` probe and the `chmod` | general (M4, speculative) | **Rejected** — unreproduced, needs a non-sticky shared parent, and the robust fix (chmod through an fd) is out of proportion for a git hook. Strictly better than base, which had no symlink check at all. |
| 9 | Scope: the hook is +179 lines over base; rungs 3 and 4 are two where one would serve | general (#9), hostile | **Noted, no action** — every rung and gate traces to an accepted round-1 or round-2 finding. Both reviewers flagged it explicitly as something they were *not* asking to change, for the same reason as #6. |

### Round 4 — final pass, tests and docs only

Findings 1–4 are all test-coverage and documentation gaps. **`scripts/pre-push` will not be touched**: it is the live gate for every push on this box, every previous rewrite of that block produced the next round's blocker, and no accepted finding requires a behavioural change. The pass adds four assertions to two existing tests, de-vacuums one, and corrects two doc sentences.

PR #917, opened 2026-09-04
feat: cache the heaviest CRM aggregate endpoints

### LINKED ISSUE #876: feat(crm): client-side dashboard navigation and a data_version-keyed cache for heavy aggregates

nbramia:
## TL;DR

The CRM's heaviest dashboard endpoints — the Me and Family interaction/timeline aggregates, the birthdays page, the overall statistics, and the default people list — now answer a repeat request almost instantly instead of recomputing every time, as long as nothing in the underlying data has actually changed. A refresh, a second browser tab, or returning to a dashboard after the client-side cache (from the sibling navigation PR) has expired all get served from a short-lived server-side cache, bounded by a five-minute TTL. The moment any change is made anywhere in the CRM data, the next request recomputes.

Closes #876 (part 2 of 2 — the sibling PR is client-side navigation).

**Update:** an adversarial review found ten issues, all fixed (including a real concurrency bug found only while verifying one of the fixes against real HTTP load). A follow-up verification pass then found one more, introduced by one of those fixes — a store-time generation race — also fixed; see the review thread for both rounds' finding-by-finding responses.

## Design

A small, bounded, in-memory cache sits directly in front of each of these endpoints. Its freshness signal is SQLite's own per-database commit counter (the same mechanism the people-list cache already uses), read from every file a CRM store actually reads crm.db from (this can be more than one path under a non-default `LIFEOS_CHROMA_PATH`) plus the interactions database — so a cached response is invalidated automatically the moment any of them is written to, from this server process or any other. That data-version signal is a *generation stamp*: any change drops every cached entry outright, not just the ones whose specific parameters changed. A five-minute TTL also bounds how long an entry can live regardless, as a safety net. The cache never grows unbounded (entries or memory), never serves a caller a reference it could accidentally corrupt for someone else, never caches a response nobody will ask for twice (a search query, a huge export), and never turns a failure to read its own freshness signal into a failure of the request it's trying to help.

## Implementation Notes

- `api/services/aggregate_cache.py` — `AggregateCache` class: one long-lived, pragma-only SQLite connection per watched database path (`check_same_thread=False`, marked `# threadpool-safe: pragma-only, guarded by lock`), opened with a `mode=ro` URI so a missing file is never created (verified read-only still sees `PRAGMA data_version` change on a WAL-mode database after an external write); `PRAGMA data_version` reads are wrapped in `try/except sqlite3.Error`, falling through to an uncached call and reopening the connection next time. `crm.db` is resolved from both `PersonEntityStore.CRM_DB_PATH` and `api.utils.db_paths.get_crm_db_path()`, deduped by `os.path.realpath`. The version pair is a generation stamp outside the cache key (a change clears everything); the leader re-checks its own pre-compute generation against the current one at store time, so a commit that lands mid-computation makes the store a no-op instead of caching a stale result under the new generation. `.cached(ttl_seconds, should_cache=None)` is the decorator factory; `should_cache(kwargs) -> bool`, if given, bypasses the cache entirely (no read, no store) when it returns `False`. Every hit and miss returns/stores `copy.deepcopy(...)`. Concurrent misses for the same key single-flight behind a per-key `threading.Event` (30s follower timeout, not the 300s TTL). `MAX_TOTAL_BYTES` (~7.1 MB serialized) is scaled from a documented, measured ~7x serialized-to-heap ratio to target a real ~50 MB heap ceiling; an entry over the cap is skipped, not stored at the cost of flushing everything else.
- `api/routes/crm.py` — `@cached_aggregate()` on six handlers; `@cached_aggregate(should_cache=_people_cache_eligible)` on `list_people` (`not q and limit <= 300`).
- `tests/test_aggregate_cache.py` — unit tests against isolated `AggregateCache` instances (temporary SQLite files, the `tests/test_person_entity.py` pattern), covering every behavior above.
- `tests/test_aggregate_cache_routes.py` (new) — walks the real `api.routes.crm` router and asserts each of the seven cached paths' endpoint is specifically the cache wrapper (by source file, not just presence of `__wrapped__`); a `TestClient` test proving a real route recomputes after an external commit to a substituted crm.db.
- `tests/reset_singletons.py`, `tests/test_route_handlers_no_shared_connections.py` — unchanged from the original PR (the singleton-reset wiring and the `STORE_MODULES` guard entry).
- `tests/test_me_page.py`, `tests/test_me_family_aggregates_oracle.py` — a weakened assertion restored and two coupling-to-the-autouse-reset sites commented, per review finding 10.
- `docs/specs/technical/architecture.md`, `docs/specs/product/crm-analytics.md`, `docs/specs/product/api-crm.md` — rewritten to match the current design and drop the freshness overclaim the review caught.

### Automated tests

```
$ HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_aggregate_cache.py tests/test_aggregate_cache_routes.py -q
collected 25 items
.........................                                                [100%]
============================== 25 passed in 4.57s ===============================
```

`tests/test_aggregate_cache.py` (22 tests): cache hit under 20ms (best-of-3) with an identical body and no recompute; per-parameter keys; external-commit invalidation to both watched databases, plus that a version change drops entries outright rather than leaving them as dead weight; TTL-independent-of-data_version expiry; errors never cached, success-after-failure cached normally; entry-count and byte bounds respected, an oversized entry skipped rather than flushing the cache, `clear()`; mutation of a returned object never poisons the cache, two hits return independent objects; a version-read failure (missing directory) falls through to uncached rather than raising, and recovers once the file appears; multiple watched crm.db paths — a write through either invalidates, identical paths dedupe, and the auto-discovery constructor watches both under a simulated relocated `LIFEOS_CHROMA_PATH`; concurrent misses for the same key compute exactly once; a commit landing mid-compute is never served stale to a follower or a later request (the generation-race fix, mutation-verified against the unguarded store); the decorator survives real FastAPI `Query()` resolution and a real HTTP round trip.

`tests/test_aggregate_cache_routes.py` (3 tests): all seven listed routes carry the cache wrapper specifically (not just some decorator) and no other CRM route does; an external `sqlite3` commit to a temporary crm.db substituted for the process-wide singleton forces a real `GET /api/crm/birthdays/all` to recompute through `TestClient`.

Full suite (under the shared push lock, rebased onto #909's tip `5b45c8c`): **5835 passed, 8 skipped** (unit, one pre-existing skip is the environment-dependent `test_my_person_id_is_valid_uuid`, confirmed to fail identically on the base branch) and **301 passed** (server-free browser) — no failures.

### Manual verification

Ran a private staging instance (`~/.venvs/lifeos/bin/python`, port 8029, this worktree, staging `.env`) against a read-only copy of the real CRM/interactions databases:

- **Finding 6 (default-shape-only caching):** `/people?sort=strength&limit=300` still caches (0.97s → 0.011s); `/people?q=test` stays ~0.11s on both calls (no cache read/write at all now); `/people?limit=10000` stays ~2.8-3.0s on both calls — no more encode overhead paid for entries that are never reused.
- **Finding 9 (single-flight) — this is where the real bug surfaced:** 8 parallel `curl` requests to a cold `/me/interactions?days_back=999` key, before the fix, completed at 11-13s each (worse than the review's own pre-single-flight baseline, because now up to 8 real concurrent SQL aggregations were contending for the same resources — a genuine regression from a half-correct single-flight implementation). After moving the wake signal to after the cache store completes: all 8 requests complete together in ~1.3s wall clock.
- Re-confirmed invalidation and byte-identical bodies are unaffected by any of the above changes.

## Review focus

- The generation-race fix (the leader re-checking its pre-compute `versions` against `self._generation` at store time): this is the second concurrency bug found in this cache in two review rounds (the first was the single-flight wake-ordering bug), so it's worth a careful independent look rather than trusting the mutation test alone.
- Whether the ~50 MB heap-ceiling target (via the documented 7x ratio), the 5-minute TTL, and the 30s follower timeout are the right numbers — still reasonable defaults, not measurements against sustained production traffic.
- This PR does not touch `web/crm.html`; the client-side navigation PR (#909, now merged) was independent and reviewed separately.

### DIFF
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -748,8 +748,18 @@ def get_all_birthdays():
     }
 
 
+def _people_cache_eligible(kwargs: dict) -> bool:
+    """Only the parameter shape the CRM page's default load actually uses
+    (no search text, a bounded page size) is worth caching -- a per-keystroke
+    search query or a `limit=10000` export is requested once and never
+    again, so caching it only spends memory and a measurable
+    jsonable_encoder()/json.dumps() pass on every miss for no reuse (#917
+    review finding 6)."""
+    return not kwargs.get("q") and kwargs.get("limit", 50) <= 300
+
+
 @router.get("/people", response_model=PersonListResponse)
-@cached_aggregate()
+@cached_aggregate(should_cache=_people_cache_eligible)
 def list_people(
     q: Optional[str] = Query(default=None, description="Search query"),
     category: Optional[str] = Query(default=None, description="Filter by category"),
--- a/api/routes/crm.py
+++ b/api/routes/crm.py
@@ -22,6 +22,7 @@ from pydantic import BaseModel, Field
 
 from api.services.person_entity import PersonEntity, get_person_entity_store, compute_person_category
 from api.services.interaction_store import get_interaction_store
+from api.services.aggregate_cache import cached_aggregate
 from config.people_config import InteractionConfig
 from config.settings import settings
 from api.services.source_entity import (
@@ -719,6 +720,7 @@ def get_todays_birthdays():
 
 
 @router.get("/birthdays/all")
+@cached_aggregate()
 def get_all_birthdays():
     """Get all people with birthdays, grouped by date."""
     person_store = get_person_entity_store()
@@ -747,6 +749,7 @@ def get_all_birthdays():
 
 
 @router.get("/people", response_model=PersonListResponse)
+@cached_aggregate()
 def list_people(
     q: Optional[str] = Query(default=None, description="Search query"),
     category: Optional[str] = Query(default=None, description="Filter by category"),
@@ -2494,6 +2497,7 @@ def update_relationship_strengths():
 
 
 @router.get("/statistics", response_model=StatisticsResponse)
+@cached_aggregate()
 def get_crm_statistics():
     """
     Get comprehensive CRM statistics.
@@ -3431,6 +3435,7 @@ def get_me_stats():
 
 
 @router.get("/me/timeline", response_model=TimelineResponse)
+@cached_aggregate()
 def get_me_timeline(
     source_type: Optional[str] = Query(
         default=None,
@@ -3514,6 +3519,7 @@ def get_me_timeline(
 
 
 @router.get("/me/interactions", response_model=MeInteractionsResponse)
+@cached_aggregate()
 def get_me_interactions(
     days_back: int = Query(default=365, ge=1, le=3660, description="Days of history (up to 10 years)"),
     trend_period: str = Query(default="quarter", description="Trend comparison period: week, month, quarter, year"),
@@ -4253,6 +4259,7 @@ def get_family_stats(
 
 
 @router.get("/family/timeline", response_model=TimelineResponse)
+@cached_aggregate()
 def get_family_timeline(
     person_ids: str = Query(
         ...,
@@ -4333,6 +4340,7 @@ def get_family_timeline(
 
 
 @router.get("/family/interactions", response_model=FamilyInteractionsResponse)
+@cached_aggregate()
 def get_family_interactions(
     person_ids: str = Query(
         ...,
--- a/api/services/aggregate_cache.py
+++ b/api/services/aggregate_cache.py
@@ -1,0 +1,246 @@
+"""
+Short-lived response cache for the CRM's heaviest read aggregates (#876).
+
+`/me/interactions`, `/me/timeline`, `/family/interactions`, `/family/timeline`,
+`/birthdays/all`, `/statistics`, and the default-parameter `/people` list all
+recompute from scratch on every request even when nothing in crm.db or
+interactions.db has changed since the last call -- #871/#880 made each of
+those individually fast, but a dashboard switch or a page refresh still pays
+the full cost every time. `AggregateCache.cached()` memoizes a route
+handler's return value for a short TTL, keyed on its resolved query
+parameters plus both databases' `PRAGMA data_version` counters -- the same
+invalidation signal `PersonEntityStore`'s own `get_all()` cache uses (see
+api/services/person_entity.py) -- so a write to either database, from this
+process or any other, makes the next request for an affected key recompute
+rather than serve stale data. The TTL is a backstop bound on memory/entry
+lifetime, not the primary invalidation path: a data_version change makes a
+key's next lookup a miss immediately, regardless of TTL.
+
+Structured as a class (rather than bare module globals) so a test can
+construct an isolated `AggregateCache` pointed at temporary databases, the
+same way `tests/test_person_entity.py` builds a `PersonEntityStore(db_path)`
+instead of fighting the process-wide singleton -- see `get_aggregate_cache()`
+below for the singleton `api/routes/crm.py` actually decorates its handlers
+with.
+"""
+import functools
+import json
+import logging
+import sqlite3
+import threading
+import time
+from collections import OrderedDict
+from typing import Callable, Optional
+
+from fastapi.encoders import jsonable_encoder
+
+from api.services.interaction_store import get_interaction_db_path
+from api.services.person_entity import PersonEntityStore
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TTL_SECONDS = 300
+MAX_ENTRIES = 200
+MAX_TOTAL_BYTES = 20 * 1024 * 1024  # 20 MB
+
+
+class AggregateCache:
+    """A bounded, TTL-backstopped, data_version-keyed cache for read-only
+    route handlers.
+
+    Each instance owns two persistent, pragma-only SQLite connections (one
+    per database) used solely to read `PRAGMA data_version` cheaply -- the
+    same justification as `PersonEntityStore._get_data_version_connection()`.
+    Every call site of those connections is reached only while holding
+    `self._lock`, so cross-thread use is serialized, which is what makes
+    `check_same_thread=False` safe here despite handlers running on the
+    FastAPI worker threadpool.
+    """
+
+    def __init__(
+        self,
+        crm_db_path: Optional[str] = None,
+        interactions_db_path: Optional[str] = None,
+        max_entries: int = MAX_ENTRIES,
+        max_total_bytes: int = MAX_TOTAL_BYTES,
+    ):
+        self.crm_db_path = crm_db_path or str(PersonEntityStore.CRM_DB_PATH)
+        self.interactions_db_path = interactions_db_path or get_interaction_db_path()
+        self.max_entries = max_entries
+        self.max_total_bytes = max_total_bytes
+
+        self._lock = threading.Lock()
+        self._crm_conn: Optional[sqlite3.Connection] = None
+        self._interactions_conn: Optional[sqlite3.Connection] = None
+        # key -> (expires_at monotonic, jsonable value, size in bytes). Dict
+        # order doubles as LRU order (most-recently-used at the end).
+        self._cache: "OrderedDict[tuple, tuple[float, object, int]]" = OrderedDict()
+        self._total_bytes = 0
+
+    def _get_crm_connection(self) -> sqlite3.Connection:
+        if self._crm_conn is None:
+            self._crm_conn = sqlite3.connect(
+                self.crm_db_path,
+                check_same_thread=False,  # threadpool-safe: pragma-only, guarded by lock
+            )
+        return self._crm_conn
+
+    def _get_interactions_connection(self) -> sqlite3.Connection:
+        if self._interactions_conn is None:
+            self._interactions_conn = sqlite3.connect(
+                self.interactions_db_path,
+                check_same_thread=False,  # threadpool-safe: pragma-only, guarded by lock
+            )
+        return self._interactions_conn
+
+    def _data_versions_locked(self) -> tuple:
+        """Read both PRAGMA data_version counters. Caller must hold `_lock`."""
+        crm_version = self._get_crm_connection().execute("PRAGMA data_version").fetchone()[0]
+        interactions_version = self._get_interactions_connection().execute(
+            "PRAGMA data_version").fetchone()[0]
+        return int(crm_version), int(interactions_version)
+
+    def _evict_locked(self) -> None:
+        """Drop the least-recently-used entries while over either bound.
+        Caller must hold `_lock`."""
+        while self._cache and (len(self._cache) > self.max_entries or
+                                self._total_bytes > self.max_total_bytes):
+            _, (_, _, size) = self._cache.popitem(last=False)
+            self._total_bytes -= size
+
+    def stats(self) -> dict:
+        """Entry count and total cached bytes -- for tests and diagnostics."""
+        with self._lock:
+            return {"entries": len(self._cache), "total_bytes": self._total_bytes}
+
+    def clear(self) -> None:
+        """Drop every cached entry. Test-only escape hatch (production
+        invalidation is via data_version, never a manual clear)."""
+        with self._lock:
+            self._cache.clear()
+            self._total_bytes = 0
+
+    def cached(self, ttl_seconds: float = DEFAULT_TTL_SECONDS) -> Callable:
+        """Memoize a function's return value for `ttl_seconds`, keyed on its
+        resolved keyword arguments plus both databases' current
+        data_version.
+
+        Intended for a FastAPI route handler, applied directly under
+        `@router.get(...)` (i.e. as the innermost decorator) so FastAPI still
+        sees the original function's signature: `functools.wraps` sets
+        `__wrapped__`, which Python's `inspect.signature()` follows by
+        default -- that is what lets FastAPI keep resolving `Query()`
+        parameters from the wrapped function normally.
+
+        Only a successful return is ever cached: an exception (including a
+        raised HTTPException for a non-200 response) propagates before this
+        reaches the cache-store step, so an error response is never cached.
+        """
+        def decorator(func: Callable) -> Callable:
+            identity = f"{func.__module__}.{func.__qualname__}"
+
+            @functools.wraps(func)
+            def wrapper(*args, **kwargs):
+                # FastAPI always calls path operation functions with
+                # resolved Query()/Path() values as keyword arguments, so
+                # this alone covers "every query parameter" without
+                # inspecting the request directly.
+                param_key = tuple(sorted(kwargs.items()))
+
+                with self._lock:
+                    crm_version, interactions_version = self._data_versions_locked()
+                    key = (identity, param_key, crm_version, interactions_version)
+                    cached = self._cache.get(key)
+                    if cached is not None:
+                        expires_at, value, size = cached
+                        if expires_at > time.monotonic():
+                            self._cache.move_to_end(key)
+                            return value
+                        # Expired -- drop it now rather than waiting for a
+                        # bounds-triggered eviction that might never come.
+                        del self._cache[key]
+                        self._total_bytes -= size
+
+                # Compute outside the lock: the wrapped handler can take
+                # tens/hundreds of ms, and holding the lock here would
+                # serialize every cached CRM aggregate request behind
+                # whichever one is currently a cache miss.
+                result = func(*args, **kwargs)
+
+                # The cached (and returned) value is `result` itself, not a
+                # jsonable_encoder()'d copy: some existing CRM unit tests
+                # call a decorated handler directly (bypassing FastAPI) and
+                # expect the same Pydantic model FastAPI would otherwise
+                # serialize later, with normal attribute access -- e.g.
+                # `result.people`, not `result["people"]`. jsonable_encoder
+                # is used only to measure the byte size for the cache's
+                # bounds; it never replaces what's actually stored/returned.
+                size = len(json.dumps(jsonable_encoder(result)).encode("utf-8"))
+
+                with self._lock:
+                    # A concurrent miss for the same key may have already
+                    # stored an entry while this call was computing outside
+                    # the lock -- last writer wins, which is fine, since
+                    # both computed from the same data_version pair and
+                    # therefore the same underlying data.
+                    if key in self._cache:
+                        self._total_bytes -= self._cache[key][2]
+                    self._cache[key] = (time.monotonic() + ttl_seconds, result, size)
+                    self._cache.move_to_end(key)
+                    self._total_bytes += size
+                    self._evict_locked()
+
+                return result
+
+            return wrapper
+
+        return decorator
+
+
+_aggregate_cache: Optional[AggregateCache] = None
+
+
+def get_aggregate_cache() -> AggregateCache:
+    """Process-wide singleton, matching the `get_*()` accessor pattern used
+    throughout `api/services/` (e.g. `get_person_entity_store()`)."""
+    global _aggregate_cache
+    if _aggregate_cache is None:
+        _aggregate_cache = AggregateCache()
+    return _aggregate_cache
+
+
+def cached_aggregate(ttl_seconds: float = DEFAULT_TTL_SECONDS) -> Callable:
+    """Decorator applied to CRM route handlers, backed by the process-wide
+    `AggregateCache` singleton. See `AggregateCache.cached()` for the actual
+    caching behavior.
+
+    Called once per decorated function, at import time -- the returned
+    wrapper closes over whichever `AggregateCache` instance
+    `get_aggregate_cache()` returns at that moment, so `reset_aggregate_cache()`
+    below clears that instance's entries in place rather than swapping in a
+    new one (which the already-decorated handlers would never see).
+    """
+    return get_aggregate_cache().cached(ttl_seconds)
+
+
+def reset_aggregate_cache() -> None:
+    """Clear every cached entry on the process-wide singleton.
+
+    For testing only: several existing CRM unit tests call a decorated route
+    handler (e.g. `get_me_interactions`) directly with its store dependency
+    mocked (`patch('api.routes.crm.get_person_entity_store', ...)`) -- the
+    mock never touches the real crm.db/interactions.db files, so this
+    cache's data_version-keyed entries from an earlier test with the same
+    parameters would otherwise still be a "hit" and mask the new mock's
+    return value entirely. Wired into
+    `tests.reset_singletons.reset_lightweight_singletons()`, which the
+    autouse `reset_singletons_after_test` fixture in `tests/conftest.py`
+    already runs after every test.
+
+    Deliberately does not reassign the module-level `_aggregate_cache`
+    singleton to a fresh instance: `cached_aggregate()` binds each decorated
+    route handler to whatever instance existed at import time, so a new
+    instance here would never be seen by those closures -- only clearing
+    the existing one in place reaches them.
+    """
+    get_aggregate_cache().clear()
--- a/api/services/aggregate_cache.py
+++ b/api/services/aggregate_cache.py
@@ -8,13 +8,12 @@ interactions.db has changed since the last call -- #871/#880 made each of
 those individually fast, but a dashboard switch or a page refresh still pays
 the full cost every time. `AggregateCache.cached()` memoizes a route
 handler's return value for a short TTL, keyed on its resolved query
-parameters plus both databases' `PRAGMA data_version` counters -- the same
-invalidation signal `PersonEntityStore`'s own `get_all()` cache uses (see
-api/services/person_entity.py) -- so a write to either database, from this
-process or any other, makes the next request for an affected key recompute
-rather than serve stale data. The TTL is a backstop bound on memory/entry
-lifetime, not the primary invalidation path: a data_version change makes a
-key's next lookup a miss immediately, regardless of TTL.
+parameters, invalidated whenever either database's `PRAGMA data_version`
+counter changes -- the same invalidation signal `PersonEntityStore`'s own
+`get_all()` cache uses (see api/services/person_entity.py) -- so a write to
+either database, from this process or any other, drops every cached entry
+before the next request. The TTL is a backstop bound on entry lifetime, not
+the primary invalidation path.
 
 Structured as a class (rather than bare module globals) so a test can
 construct an isolated `AggregateCache` pointed at temporary databases, the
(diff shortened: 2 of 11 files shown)


nbramia:
## Adversarial review of #917 by Opus

**Verdict:** BLOCK

The core mechanism is sound and I could not break the two things that matter most: invalidation works end to end (proved with commits from a separate process to both databases, including after a warm hit), and the OpenAPI schema for all seven routes is byte-identical to a build of the same tree with the decorator neutered. What blocks is that the cache introduces a new unhandled failure path and resolves `crm.db` by a different rule than the stores whose data it is caching, plus a set of measured cost/bound problems that the tests do not cover.

1. **[BLOCKER] A `PRAGMA data_version` read failure turns all seven endpoints into 500s, and there is no fallback to "just don't cache"** — `api/services/aggregate_cache.py:80-100`, `:151`. The two connections are opened lazily on the first request and every pragma read is unguarded. Reproduced: an `AggregateCache` whose `crm_db_path` is in a directory that does not exist constructs fine and then raises on the first call — `OperationalError: unable to open database file` — which propagates straight out of a route handler as a 500. Before this PR these endpoints had no dependency on being able to open a second handle to either file. This is reachable in normal operation, not just theory: `AggregateCache.__init__` computes `interactions_db_path` from `get_interaction_db_path()` (which `mkdir`s the configured data dir) but `crm_db_path` from the hardcoded `PersonEntityStore.CRM_DB_PATH` (`<repo>/data/crm.db`), so with a non-default `LIFEOS_CHROMA_PATH` the repo's own `data/` may never be created and every cached endpoint 500s. Also relevant to a file that is being replaced rather than written in place (`interaction_store._BACKUP_DB_BASENAME` restores) — the long-lived handle stays pinned to the old inode. **Fix:** wrap `_data_versions_locked()` in `try/except sqlite3.Error`, log once, and fall through to computing uncached; a cache must never be able to fail a request that would otherwise succeed. Consider reopening the connection on error.

2. **[BLOCKER] The cache watches a different `crm.db` than the stores it caches, whenever `LIFEOS_CHROMA_PATH` is not the default** — `api/services/aggregate_cache.py:67`. `SourceEntityStore` and `RelationshipStore` both resolve their path through `api/utils/db_paths.get_crm_db_path()` (derived from `settings.chroma_path`); the cache uses `PersonEntityStore.CRM_DB_PATH`. Demonstrated with `LIFEOS_CHROMA_PATH` pointed outside the repo: stores open `<configured dir>/crm.db`, the cache opens `<repo>/data/crm.db`, `realpath` equality `False`. `/statistics` reads source-entity and relationship statistics, and `/people` reads source entities for category computation, from the file the cache is *not* watching — so a write to those tables does not invalidate, and staleness is bounded only by the 300 s TTL. Worse, `sqlite3.connect()` on a path that does not exist silently creates a **stray 0-byte `data/crm.db`** (verified: `exists=True size=0`) whose `data_version` then never changes. `LIFEOS_CHROMA_PATH` is a documented setting (`docs/guides/configuration.md:26`, `.env.example:241`), and "nothing broken-by-default for a fresh downloader" is a standing rule here. **Fix:** watch both paths (dedupe on `os.path.realpath`), or at minimum resolve through `get_crm_db_path()` alongside `PersonEntityStore.CRM_DB_PATH`; and do not create the file — open with `?mode=rw` URI or skip caching when it is absent.

3. **[MAJOR] Nothing in the test suite asserts that the seven endpoints are actually cached** — `tests/test_aggregate_cache.py` (whole file). Every test decorates a synthetic local function; not one exercises `api.routes.crm`. Delete `@cached_aggregate()` from all seven handlers and the suite stays green, so the issue's acceptance criteria ("the same aggregate endpoint requested twice … under 20 ms with an identical body", "a commit to crm.db or interactions.db … shall recompute") are covered only for a toy endpoint. I verified the real routes by hand (see below) but that verification is not locked in. **Fix:** add a test that walks `app.routes`, asserts each of the seven paths' `endpoint` carries `__wrapped__` and is the cache wrapper, and — for at least one route — that an external commit to a temp `crm.db` forces a recompute through `TestClient`.

4. **[MAJOR] The byte bound measures serialized JSON, which understates the memory actually held by roughly 7×** — `api/services/aggregate_cache.py:178`. The stored value is the Python object, but `size` is `len(json.dumps(jsonable_encoder(result)))`. Measured on the real dataset: one `/people?limit=10000` result is accounted at **7.05 MB** while holding it grew peak RSS by **51 MB**. So `MAX_TOTAL_BYTES = 20 MB` is in practice a ~140 MB heap ceiling, not 20 MB, and the docs' "total serialized bytes (20 MB)" reads as a memory bound to anyone skimming. On a box that has been OOM-frozen before, that is worth getting right. **Fix:** either size the stored object (`pympler.asizeof` is a dependency add, so more likely) apply a conservative multiplier and say so, or lower the cap and document that it is a serialized-size proxy.

5. **[MAJOR] The cached object is handed out by reference with no defensive copy, and a caller mutation permanently poisons the entry** — `api/services/aggregate_cache.py:188`, deliberately per the PR's own Review focus. I confirmed the mechanism rather than taking it on trust: two direct calls to `crm.list_people(...)` return `result_a is result_b == True`; assigning `result.people[0].canonical_name = "MUTATED-BY-CALLER"` makes the *next* call return the mutated value. Today no in-tree caller mutates (I grepped `api/`, `mcp_server.py`, `scripts/`, `web/` for calls into the seven handlers — `api/routes/people.py:407` calls people.py's own `list_people`, not crm's), so this is latent rather than live. But it is a trap with no guard rail and no test: the next handler edit that sorts or annotates its own return value in place corrupts every subsequent hit until the next write. **Fix:** at minimum add a test that pins the identity semantics and a comment at each decorated handler, or cache `jsonable_encoder(result)` and give the direct-call tests a `.__wrapped__` call site instead of bending the production design around them.

6. **[MAJOR] Every `/people` parameter combination is cached, including per-keystroke search text and `limit=10000`, which the issue did not ask for and which costs real time** — `api/routes/crm.py:751-752`. The scope addition asks for "that endpoint's **default-parameter** responses". As written, `web/crm.html:9442` sends a distinct `q=` per debounced keystroke, so a typing session mints a stream of ~240 KB-accounted entries that will never be requested again and that push the useful dashboard entries out under the 200-entry / 20 MB bounds. The extra `jsonable_encoder` + `json.dumps` on every miss is measurable: **+575 ms (+25 %) on a cold `/people?limit=10000`** (handler 2308 ms, encoder 539 ms, dumps 36 ms) and **+11 ms (+5 %) on the `limit=300` path the CRM page actually uses**. Nobody consumes a cache entry for a search term twice. **Fix:** restrict the `/people` decoration to the default/no-`q` parameter shape (bypass the cache when `q` is set), which also keeps search text out of a long-lived in-memory key.

7. **[MINOR] `docs/specs/product/api-crm.md` overclaims that nothing observable changes** — the added paragraph says "response shape, status codes, and freshness (any write anywhere in the CRM data invalidates the relevant cached responses immediately) are unchanged." Three ways that is not quite true: the 300 s TTL means a response can be up to five minutes old with no writes at all; `/me/interactions`, `/me/timeline`, `/family/*` all compute their window from `datetime.now(timezone.utc)` at request time, so a cached body freezes "now" for up to five minutes (visible at a day rollover in the daily heatmap); and per finding 1 a status code *can* now change, from 200 to 500. **Fix:** say "bounded by a five-minute TTL" rather than "unchanged", and drop "immediately" or qualify it.

8. **[MINOR] Entries keyed on a superseded `data_version` are never proactively dropped** — `api/services/aggregate_cache.py:148-151`. Because the version pair is *inside* the key, a write does not evict anything; it just makes every existing key unreachable. Observed directly: entries went 1 → 2 across an external commit, with the dead entry still counted in `total_bytes`. Half the byte budget can be garbage that only LRU pressure reclaims. Relatedly, one entry larger than `max_total_bytes` empties the entire cache and then discards itself (verified: 3 entries / 351 B → **0 entries / 0 B** after storing one oversized value). Not reachable with today's data (largest real entry 7 MB vs a 20 MB cap) but a sharp edge. **Fix:** keep the version pair outside the key as a generation stamp and clear on change; skip storing anything larger than the cap instead of flushing.

9. **[MINOR] No single-flight: concurrent misses for the same key all recompute** — acknowledged in the code comment at `:168`. Measured over HTTP on a cold key: **8 parallel `/me/interactions?days_back=365` requests took 7118–7294 ms each, 7.3 s wall clock**, versus ~390 ms for one. This is not a regression (the base branch does the same work), but it is the case the cache is least helpful in and the one that follows every write — exactly when a client-side-routed dashboard (#909) refires. Worth either a per-key `threading.Event` or an explicit note in the docs that the first request after any write is unprotected.

10. **[MINOR] Two existing tests are quietly weakened by the cache, and one new assertion is timing-based** — `tests/test_me_page.py:207` (`test_get_all_not_called_more_than_once`) asserts `get_all.call_count <= 1`, which now also passes when a cache hit makes it 0; the test no longer distinguishes "called once" from "handler never ran". `tests/test_me_family_aggregates_oracle.py:921` and `:935` call `get_family_timeline` with byte-identical parameters against identical seeded data in adjacent tests, so they are correct only because `reset_singletons_after_test` runs between them — an invisible coupling worth a comment at those sites. Separately, `tests/test_aggregate_cache.py:78` asserts a cache hit completes in under 20 ms; that path does two SQLite pragma reads, and the suite runs under `-n auto`, so it is a plausible flake on a loaded box.

### What I verified and how

- `gh pr diff 917` + full read of `api/services/aggregate_cache.py`, the seven decorated handlers, `db_paths.py`, `interaction_store.get_interaction_db_path`, `person_entity.get_all`, `conftest.reset_singletons_after_test`.
- **OpenAPI unchanged:** dumped the seven paths' schema twice in-process — once normally, once with `cached_aggregate` monkeypatched to identity before `api.routes.crm` imports — `diff` clean, `md5sum` identical (`f4a5bc55…`). FastAPI's `Query()` resolution genuinely survives the wrapper.
- **Warm hits on real data (staging :8032, real CRM/interactions copy):** `/me/interactions` 391 ms → **1.1 ms**; `/family/interactions?person_ids=…` 245 ms → **2.6 ms**; `/birthdays/all` 3.8 ms → 1.3 ms; `/statistics` 206 ms → **1.5 ms**; `/people?sort=strength&limit=300` 248 ms → **2.5 ms**. All five byte-identical (`cmp` on the two bodies). Criterion met.
- **Invalidation, over HTTP, from a separate process, after a warm hit:** inserted one row through a fresh `sqlite3` connection into `_review876_probe` in the staging **crm.db** → `/statistics` went 0.7 ms (2nd warm hit) → **223 ms** (recompute) → 1.4 ms (re-warmed); then `DROP TABLE` (another commit) → 213 ms again. Same for **interactions.db** → `/me/interactions` 1.1 ms → **382 ms** → 1.1 ms, then 384 ms after the drop. *Staging data changed:* created, inserted one row into, and then dropped a table named `_review876_probe` (and `_review876_probe2`) in `data/crm.db` and `data/interactions.db`; confirmed afterwards that `sqlite_master` contains no `%probe%` tables in either file. No CRM row was touched.
- **Key correctness:** omitted-vs-explicit defaults share one key — after `GET /me/interactions` then `GET /me/interactions?days_back=365&trend_period=quarter&health_period=quarter` the cache holds **1** entry (same for `/people` bare vs `?limit=50&offset=0&sort=strength&min_interactions=0`). Key shape confirmed as `('api.routes.crm.get_me_interactions', (('days_back',365),('health_period','quarter'),('trend_period','quarter')), crm_ver, int_ver)`. `person_ids` is a `str`, so no unhashable value can reach the key; `A,B` vs `B,A` are separate entries and **correctly so** — I checked the bodies differ, so conflating them would be wrong.
- **Middleware interaction:** `/api/perf/routes` records cache hits (`/api/crm/statistics` count 3, p50 0.9 ms, max 205.9 ms — hits and misses both present). `_ScopedGZipMiddleware` still compresses hits: `/people?sort=strength&limit=300` returns `content-encoding: gzip`, `vary: Accept-Encoding`, `content-length: 34422` identically on cold and warm.
- **Error paths:** `/family/interactions` with no `person_ids` → 400, cache entries 0; with blank `person_ids` → 400, cache entries 0.
- **Bounds/eviction:** oversized-entry flush reproduced (3 entries/351 B → 0/0); accounted vs wire bytes for `/people?limit=300` 241 054 vs 223 343 (a fair serialized proxy, but see finding 4 for heap).
- **Concurrency:** 8 threads on one cold key → handler ran **8** times in-process; 7.3 s wall over HTTP (finding 9). The lock does correctly serialize both pragma connections — 16 threads × 30 calls raised nothing and left ≤5 entries.
- **Tests:** `pytest tests/test_route_handlers_no_shared_connections.py tests/test_route_handlers_sync.py tests/test_aggregate_cache.py -q` → **26 passed**. `pytest tests/test_crm_api.py tests/test_me_page.py tests/test_me_family_aggregates_oracle.py tests/test_crm_people_list_batching.py tests/test_aggregate_cache.py -q -n 4` → **94 passed, 2 skipped, 1 failed**; the failure is `test_my_person_id_is_valid_uuid`, which reads `settings.my_person_id` and fails purely because my worktree has no `.env` — unrelated to this PR.
- **Privacy:** no personal data in the diff, the new tests, or the new docs; the `#904` no-logging rule for search text is not weakened (the cache logs nothing). Search terms do sit in an in-memory key for up to 5 min — see finding 6 for the reason to stop caching them anyway.
- **Scope:** no unrelated edits; no new dependencies; no test deleted, skipped, or xfailed; `web/crm.html` untouched (so no browser verification was warranted).

### Not verified

- The full unit suite. Per the standing brief I ran targeted files only (the shared push lock was not mine to hold); the PR reports 5822 passed under the lock and I have no reason to doubt it.
- Behaviour under an actual `LIFEOS_CHROMA_PATH` deployment end to end — I proved the path divergence and the missing-directory failure at the `AggregateCache` level, not by standing up a second full server with a relocated data dir.
- Whether `PRAGMA data_version` can ever repeat a previous value for a connection (which would let a stale entry from that generation resurface while still inside the 300 s TTL). SQLite documents the value only as "different after a change", not monotonic. I judged this unreachable in practice and did not try to force it; finding 8's suggested restructuring removes the question entirely.
- Long-run memory behaviour: my 7× heap-vs-accounted number comes from one peak-RSS measurement of a single large entry, not from a sustained soak of the cache at its bounds.

nbramia:
Addressed all ten findings.

1. **[BLOCKER] Data_version read failure → 500** — Fixed. `_try_read_versions_locked()` wraps every pragma read in `try/except sqlite3.Error`, logs once (resets on recovery), and returns `None`; the wrapper falls through to computing uncached rather than propagating. Connections are dropped on failure so the next call reopens fresh. New tests `TestVersionReadFailure` (missing-directory case returns 200-equivalent and keeps computing uncached; a file created later is picked up without a restart).

2. **[BLOCKER] Different crm.db than the stores it caches** — Fixed. `_default_crm_db_paths()` watches both `PersonEntityStore.CRM_DB_PATH` and `api.utils.db_paths.get_crm_db_path()`, deduped by `os.path.realpath`. Connections now open via a `mode=rw` URI (`_open_existing_db`), so a missing file raises instead of being silently created — combined with finding 1's fallback, a missing file means "serve uncached," never a stray 0-byte file. New tests `TestMultipleCrmPaths` (a write through either watched path invalidates; identical paths named two ways dedupe to one; the auto-discovery constructor watches both when `PersonEntityStore.CRM_DB_PATH` and `get_crm_db_path()` are monkeypatched to diverge, modeling a relocated `LIFEOS_CHROMA_PATH`).

3. **[MAJOR] Nothing asserts the real routes are cached** — Added `tests/test_aggregate_cache_routes.py`: walks `api.routes.crm.router.routes` and asserts each of the seven listed (method, path) pairs' `endpoint` has `__wrapped__` *and* its own code is defined in `aggregate_cache.py` (not just some other decorator), plus that no other CRM route carries it; and a `TestClient` test where an external `sqlite3` connection commits to a temporary crm.db (substituted for the process-wide singleton's watched path via `monkeypatch.setattr` on the live `AggregateCache` instance) forces `GET /api/crm/birthdays/all` to recompute.

4. **[MAJOR] Byte bound understates real heap by ~7x** — Fixed per your framing: no new dependency. `MAX_TOTAL_BYTES` is now computed from a documented `HEAP_TO_SERIALIZED_RATIO = 7` and `HEAP_TARGET_BYTES = 50 MB`, giving ~7.1 MB serialized (verified: `7,489,828 bytes`, implying a `~50.0 MB` real heap ceiling at that ratio). Documented in both the module docstring and `architecture.md`.

5. **[MAJOR] No defensive copy, mutation poisons the cache** — Fixed: every hit and every miss now returns `copy.deepcopy(...)`; what's stored is also a separate deep copy, so a caller mutating their own returned object can't reach the cached entry either. New tests `TestMutationSafety` (mutating a returned dict doesn't affect the next call's result; two hits return distinct, `is not`, objects).

6. **[MAJOR] Every `/people` parameter combination cached, including search text and `limit=10000`** — Fixed: `cached_aggregate()` gained an optional `should_cache(kwargs) -> bool` predicate; `/people` now passes `_people_cache_eligible` (`not q and limit <= 300`), bypassing the cache (no read, no store, no `jsonable_encoder`/`json.dumps` overhead) entirely when it returns `False`. Measured on staging: default shape still caches (0.97s → 0.011s); `?q=test` stays ~0.11s on both calls (no cache interaction at all); `?limit=10000` stays ~2.8-3.0s on both calls likewise — the encode overhead you measured on misses that are never reused is gone because those requests never reach the encode step now.

7. **[MINOR] api-crm.md overclaim** — Fixed: now says "bounded by a five-minute TTL" and explains the two ways a cached response can be stale (the TTL window itself, and a time-windowed aggregate freezing "now"), rather than claiming freshness is "unchanged" and "immediate." Also updated `architecture.md` and `crm-analytics.md` for consistency (and to reflect the finding 6 default-shape restriction and the finding 8 generation-stamp design).

8. **[MINOR] Superseded entries linger; oversized entry flushes everything** — Fixed both. The data_version pair is now a generation stamp (`self._generation`) held outside the cache key: `_refresh_generation_locked()` clears every entry outright the moment a version change is observed, rather than leaving old-generation keys merely unreachable inside the bounds. An entry whose serialized size exceeds `max_total_bytes` is now skipped at store time rather than being stored and immediately evicting everything else to make room. New tests for both (`test_version_change_drops_entries_outright_not_just_orphans_them`, `test_entry_larger_than_cap_is_skipped_not_flushed`).

9. **[MINOR] No single-flight** — Added: concurrent misses for the same key now share a per-key `threading.Event`; only the first caller computes, the rest wait and reuse its result (falling back to computing themselves only if the leader's call raised or its entry already expired). **Found and fixed a real bug while verifying this against real HTTP load, not just the unit test**: the first implementation woke waiting followers immediately after the leader's function call returned, *before* the result was actually written into the cache — every follower found nothing there and recomputed anyway, which I measured as 8 parallel `/me/interactions` requests taking 11-13s each (worse than your own pre-fix baseline of ~7s, since now up to 8 real concurrent SQL aggregations were contending for the same resources). The unit test passed anyway because the synthetic slow function's fast in-memory store usually won the race before Python's scheduler switched threads — a lucky timing coincidence real SQLite I/O didn't share. Moved the wake signal (`event.set()`) into the `finally` after the store completes; re-verified at 8 concurrent requests now completing in ~1.3s wall clock total (matches the ~0.9-1.3s a single request takes).

10. **[MINOR] Weakened/coupled/timing-based tests** — `tests/test_me_page.py::test_get_all_not_called_more_than_once` restored to `== 1` (with a comment explaining why `<= 1` is now distinguishable from a real cache hit, and why `== 1` is still safe given the autouse cache reset). Added a comment at both `test_me_family_aggregates_oracle.py` call sites you flagged, noting they're only independent because of the same autouse reset. The cache-hit timing assertion in `tests/test_aggregate_cache.py` is now best-of-3 via a small `_best_of()` helper rather than a single sample.

### Verification

Full unit suite (`5834 passed, 8 skipped`) and server-free browser suite (`285 passed`) both green as part of the push under the shared lock. Manually re-verified on staging against the real data copy: findings 6 and 9 above with real numbers; invalidation and identical-body behavior unaffected by any of these changes.

nbramia:
## Adversarial review of #917 by Opus — verification pass (44b6a5d)

**Verdict:** BLOCK — one new finding, introduced by the fix for my finding 8. All ten original findings are genuinely fixed; I re-verified each rather than reading the response.

The generation-stamp redesign removed the property that made the old version-in-key design safe: a handler that starts before a write can now have its pre-write result stored under the *post*-write stamp and served for the full 300 s TTL. It reproduces with two concurrent identical requests and one commit between them — the exact shape of a user editing a person while the CRM page has several aggregate requests in flight. Everything else here is nits.

### New finding

11. **[BLOCKER] A commit that lands while a handler is computing gets its pre-write result cached under the post-write generation, and served for the full TTL** — `api/services/aggregate_cache.py`, leader path: `versions` is read (and `_refresh_generation_locked` applied) *before* `func()` runs, but the store afterwards never re-checks that the generation is still the one the result was computed under. Minimal reproduction, no third endpoint needed:

    ```
    request 1 (started before the write): 'BEFORE'
    request 2 (arrived after the write) : 'BEFORE'
    request 3 (long after the write)    : 'BEFORE'   <- should be 'AFTER'
    handler runs: 1
    STALE-AFTER-WRITE: True
    ```

    Sequence: request 1 reads data_version V1 and its data, then sleeps (a real aggregate takes 200–1000 ms). A `crm.db` commit lands. Request 2 arrives, reads V2, `_refresh_generation_locked` clears the cache and sets the stamp to V2, misses, finds request 1's in-flight `Event`, and waits. Request 1 stores its V1-era result — now stamped V2 — and request 2 and every later request get it until the TTL expires or another commit happens. The **previous head (6733192) is not affected** by the identical scenario, because the version pair was inside the key so a V1-computed entry stored under a V1 key was simply never read again:

    ```
    request 1: 'BEFORE'   request 2: 'AFTER'   request 3: 'AFTER'
    handler runs: 2       STALE-AFTER-WRITE: False
    ```

    Single-flight makes this strictly easier to hit (the follower is guaranteed to take the leader's entry rather than compute its own), but it is not the cause — I first reproduced it with an unrelated second endpoint advancing the stamp. This violates the issue's own acceptance criterion ("Given a commit to crm.db or interactions.db from any process, the next request for a cached aggregate shall recompute") and the PR's TL;DR ("there is no window where you'd see stale results after making an edit"). It is reachable whenever a write overlaps a request — routinely during the nightly sync, and on any person edit made while a dashboard is loading.

    **Fix:** the leader already has `versions` in hand; keep it and re-check under the lock at store time —

    ```python
    with self._lock:
        if self._generation != versions:
            pass                      # computed against data that has since changed
        elif size > self.max_total_bytes:
            pass
        else:
            ...store...
    ```

    A follower that then finds nothing correctly falls back to computing itself, so no other change is needed. Worth a test in the shape of the reproduction above (two threads, a commit between them, assert the third call recomputes).

### The ten original findings — re-verified

| # | Fix | How I checked it |
|---|---|---|
| 1 | `_try_read_versions_locked` returns `None`, logs once, drops connections | `AggregateCache` over an unopenable `crm.db`: no exception, both calls returned, `entries=0` (uncached fall-through), warning logged once. After creating the file mid-run: `entries=1` without a restart. |
| 2 | Both crm paths watched, deduped by realpath, `mode=rw` URI | With `LIFEOS_CHROMA_PATH` relocated: `crm_db_paths` = 2, cache still works (3 calls → 1 handler run), and a write to the **relocated** crm.db alone forces a recompute. `[a, symlink→a, a]` dedupes to 1. |
| 2b | No stray file | Request against a missing `crm.db` path: `exists=False` afterwards (was a 0-byte file before). |
| 3 | Route wiring locked in | `tests/test_aggregate_cache_routes.py` passes; it walks `crm_router.routes`, requires `inspect.getsourcefile(endpoint)` to end in `aggregate_cache.py` for exactly the seven pairs, and has a real `TestClient` external-commit invalidation. The negative test (no *other* CRM route is wrapped) is a good addition I hadn't asked for. |
| 4 | Cap scaled by the measured ratio | `MAX_TOTAL_BYTES=7,489,828` (7.49 MB serialized) from `HEAP_TARGET_BYTES=50 MB / ratio 7`; matches the 7.05 MB → ~51 MB RSS measurement from pass 1. Documented in the module docstring and `architecture.md`. |
| 5 | Deep copy on hit and miss | Mutating a returned dict leaves the next hit at `'original'`; two hits are `is not` each other; handler ran once. |
| 6 | `should_cache` predicate | `/people?q=…` → 143 ms then 163 ms, `entries=0`; `/people?limit=10000` → 2072 ms then 2119 ms, `entries=0`; `/people?limit=300` → `entries=1`. Bypass is total (no encode pass). |
| 7 | Docs corrected | `api-crm.md` now says "bounded by a five-minute TTL" and names both staleness modes; `crm-analytics.md` and `architecture.md` consistent. Note finding 11 makes the "invalidates well within that window" sentence currently untrue in the interleaved case. |
| 8 | Generation stamp + oversized skip | 5 entries → external commit + one request → 1 entry, old bytes gone. One oversized store leaves `{'entries': 3, 'total_bytes': 351}` unchanged (previously flushed to 0/0). *This is also where finding 11 comes from.* |
| 9 | Single-flight | In-process: 8 concurrent misses → handler ran **1** time, 441 ms wall. Over HTTP on :8032: 8 parallel cold `/me/interactions?days_back=365` → per-request 1087–1090 ms, **wall 1092 ms**, versus 913 ms for a single cold request on the same loaded box (load average 3.5–6.7, several agents active). Confirms the implementer's ~1.3 s and my pass-1 baseline of 7.3 s. The wake-after-store bug they describe finding is real and correctly fixed. |
| 10 | Tests restored | `test_me_page.py:225` is `== 1`; both oracle call sites carry the coupling comment; the hit-timing assertion is best-of-3. |

- **Warm hits over real HTTP on :8032** with the real data copy, best-of-5, all byte-identical across 5 repeats: `/statistics` 208 ms → **0.3 ms**; `/birthdays/all` 3.2 → **0.7 ms**; `/me/interactions?days_back=365` 368 → **2.5 ms**; `/me/timeline?limit=500` (the UI's actual limit, 235 KB, 500 items) 18.6 → **2.5 ms**; `/family/interactions` 273 → **2.2 ms**; `/family/timeline?…&limit=500` 13.7 → **2.5 ms**; `/people?sort=strength&limit=300` 252 → **4.6 ms**. Criterion (<20 ms, identical body) met with a wide margin.
- **End-to-end invalidation, separate connection, after two warm hits:** one row committed into `_review876_probe` in the staging **crm.db** → `/statistics` 0.7 ms → **206 ms** (recompute) → 1.6 ms; `DROP TABLE` (a second commit) → 203 ms. Same for **interactions.db** → `/me/interactions` 2.8 ms → **407 ms** → 3.6 ms, then 401 ms after the drop. *Staging data changed and reverted:* created, inserted one row into, and dropped `_review876_probe` in `data/crm.db` and `data/interactions.db`; `sqlite_master` afterwards shows `[]` `%probe%` tables in both. No CRM row touched.
- **Middleware:** `_ScopedGZipMiddleware` still compresses hits — cold and warm both `content-encoding: gzip`, `content-length: 34427`. `/api/perf/routes` records hits and misses (`/api/crm/statistics` count 13, p50 0.5 ms, max 207.5 ms; `/api/crm/me/timeline` count 8, p50 2.4 ms).
- **Tests (targeted only, no full suite):** `pytest tests/test_crm_api.py tests/test_me_page.py tests/test_me_family_aggregates_oracle.py tests/test_crm_people_list_batching.py tests/test_aggregate_cache.py tests/test_aggregate_cache_routes.py tests/test_route_handlers_no_shared_connections.py tests/test_route_handlers_sync.py -q -n 4` → **120 passed, 2 skipped, 1 failed**. The failure is `test_my_person_id_is_valid_uuid`, which reads `settings.my_person_id` and fails only because my worktree has no `.env`; unrelated, and it fails identically at 6733192.
- **Correction to my pass-1 report:** an intermediate run of mine showed 35 skips in `test_crm_api.py`. That was my own worktree — an app import had recreated `data/` as a real directory after I removed the symlink, so the tests saw an empty database. It reproduced identically at 6733192, and after restoring the symlink (15,391 people rows) the skip count is back to 2. **Not a regression from this PR**; flagging so nobody chases it.

### Nits (not blocking)

- The deep copy roughly doubles warm-hit CPU. In-process (TestClient, so inflated by httpx overhead) `/me/timeline?limit=500` went 4.9 ms → 15.8 ms and `/people?limit=300` 5.3 ms → 10.1 ms between the two heads; over real HTTP the same requests are 2.5 ms and 4.6 ms, so the budget is not at risk. Worth knowing the number exists if the timeline limit ever grows.
- A follower waits `event.wait(timeout=ttl_seconds)` — 300 s by default — holding a threadpool thread. In practice it is released when the leader finishes, but a timeout that large is not a useful safety valve; the leader's own realistic upper bound (a few seconds) would be.
- The leader pops `self._in_flight[key]` inside the lock and calls `event.set()` just outside it. A request arriving in that gap becomes a second leader and recomputes. Harmless, and it self-corrects, but it means single-flight is best-effort rather than exact.
- `mode=rw` requires write permission on the database file and its directory (for `-wal`/`-shm`) even though this cache only ever reads. `mode=ro` would express the intent better; if it was rejected because a read-only open cannot create the `-shm` file for a WAL database, that reasoning is worth a line in `_open_existing_db`'s docstring.

### Not verified

- The full unit suite (targeted files only, per the standing brief's push-lock rule). The PR reports 5834 passed / 8 skipped and 285 browser passed under the lock.
- Whether finding 11's window is reachable through the *specific* write paths the CRM's own edit endpoints use (I proved it at the `AggregateCache` level and with two concurrent requests, not by racing a real `PUT /api/crm/people/{id}` against a dashboard load).
- Long-run memory at the new 7.49 MB cap — the 7× ratio it is derived from is still my single peak-RSS measurement from pass 1, not a soak.

nbramia:
Addressed the new blocker plus all four nits.

**11. [BLOCKER] Store-race after finding 8's generation-stamp redesign** — Fixed exactly as prescribed: the leader now keeps the `versions` it read before calling the handler, and re-checks `self._generation == versions` under the lock at store time; a mismatch skips the store, and the next request for that key misses and recomputes against the new generation. New test `TestGenerationRaceDuringCompute::test_write_during_leader_compute_is_not_served_stale` reproduces the exact shape from the report (leader mid-compute → external commit → concurrent follower → a third request afterward), asserting the follower and the later request both see the post-write value. Mutation-verified: reverting just the re-check (`if self._generation != versions: pass` back to unconditional storage) reproduces your trace precisely — `results["req2"] == {'value': 'BEFORE'}`, i.e. the follower served the leader's stale pre-write result.

**Nit — `mode=rw` → `mode=ro`** — Switched. Verified directly (not just assumed) against a real WAL-mode database: opened read-only, committed a write from a separate connection, confirmed `PRAGMA data_version` still changes on the read-only connection. No functional reason `mode=rw` was needed; docstring updated to note this and that a missing file still raises rather than being created either way.

**Nit — follower wait timeout** — Replaced `event.wait(timeout=ttl_seconds)` (300s) with a dedicated `FOLLOWER_WAIT_TIMEOUT_SECONDS = 30`, with a comment explaining why: a follower should never legitimately wait anywhere near the TTL for a leader that isn't hung, and the existing compute-fallback on timeout already handles this correctly either way.

**Nit — `_in_flight.pop()` ordering** — Swapped so `event.set()` now runs before the pop (inside the same `finally`), narrowing the window where a brand-new request could miss both the cache and the in-flight marker and start a fully redundant computation instead of taking the (already-set, non-blocking) follower path. Still best-effort, as you noted, but the window is now as small as it can be without a heavier mechanism.

**Nit — deep-copy CPU cost** — Documented in the module docstring with your measured numbers (roughly 2x on a warm hit in-process; negligible in absolute terms over real HTTP).

### Verification

Full unit suite (`5835 passed, 8 skipped`) and server-free browser suite (`301 passed`) both green as part of the push under the shared lock, rebased onto the current tip (`5b45c8c`, #909 merged — no conflicts, `web/crm.html` untouched by this PR). `test_aggregate_cache.py` (22 tests) and `test_aggregate_cache_routes.py` (3 tests) all pass, including the new race test run 5x in a row for stability before committing.

Pushed at `5759091` (was `44b6a5d` → `57e8a73` for the rebase-only push → `5759091` for this fix).

nbramia:
## Adversarial review of #917 by Opus — final pass (5759091)

**Verdict:** APPROVE

Finding 11 is fixed, the fix is locked in by a test that actually kills the mutant, and all four nits are addressed. Nothing new found.

### Finding 11 — fixed and mutation-tested

- **My exact reproduction now passes.** Two concurrent identical requests with a `crm.db` commit between them: `request 1: 'BEFORE'` (correct — it read pre-write data), `request 2: 'AFTER'`, `request 3: 'AFTER'`, `STALE-AFTER-WRITE: False`, handler runs 3. The three-endpoint variant (an unrelated request advancing the stamp) likewise returns `'AFTER-THE-WRITE'`, and the control (no interleaved request) is unchanged.
- **The new test kills the mutant.** Replacing `if self._generation != versions:` with `if False:` in `api/services/aggregate_cache.py` makes `TestGenerationRaceDuringCompute::test_write_during_leader_compute_is_not_served_stale` fail with exactly the right message (`follower must not be served the leader's pre-write result: {'value': 'BEFORE'}`) — **1 failed, 21 passed**, and it was the *only* test that failed, three runs out of three. Restored the file afterwards (`git checkout --`, tree clean).
- **Not flaky in the other direction either:** `TestGenerationRaceDuringCompute` + `TestSingleFlight` passed 5/5 consecutive runs on the unmutated code.
- Skipping the store on a generation mismatch degrades correctly rather than hanging: a follower whose leader stored nothing falls through to computing its own result (verified below).

### The four nits

1. **`mode=ro` — verified against WAL, which is what every store here uses.** `PRAGMA data_version` on a read-only watcher tracks external commits in every case I could construct: with a live independent writer (`2 → 3 → 4`); on a cleanly closed database with no `-wal`/`-shm` sidecars present at open time (`2 → 3`); with a hot 12 KB `-wal` and a `-shm` present; and with a hot `-wal` and **no** `-shm` (the crashed-writer shape) — that opened fine too. On the **real staging databases** (`crm.db` and `interactions.db`, both `journal_mode=wal`) the ro watcher observed two successive version changes on each. And it really is read-only: an `INSERT` on that connection raises `OperationalError: attempt to write a readonly database`. Should a hostile filesystem ever make a ro open fail, finding 1's fallback covers it (serve uncached).
2. **`FOLLOWER_WAIT_TIMEOUT_SECONDS = 30`** (vs the 300 s TTL it used to inherit), with a docstring explaining the reasoning. Behaviourally: a leader that raises after 0.3 s releases four waiting threads in **0.60 s** total, each with the real exception and nothing cached — no one waits out the timeout.
3. **`event.set()` now precedes the `_in_flight` pop.** Exercised through the case it exists for — a leader whose result is too big to store: four concurrent threads finished in **0.60 s**, all four got a correct result, `entries=0`, no hang. A request arriving in the (now much smaller) gap takes an already-set event and falls straight into the follower's own compute-it-myself path.
4. **Deep-copy cost documented** in the store comment with the measured numbers.

I also re-checked the follower path for a second-order hole: a follower that registered *before* a write can wake to an entry the leader stamped under the pre-write generation. That entry is stamped V1, so the next request's version read clears it and recomputes — and the follower's own request predates the write anyway, so it gets the same answer the leader legitimately returns. Not a defect.

### Re-verified end to end

- **Targeted tests** (`test_aggregate_cache`, `test_aggregate_cache_routes`, `test_route_handlers_no_shared_connections`, `test_route_handlers_sync`, `test_crm_api`, `test_me_page`, `test_me_family_aggregates_oracle`, `test_crm_people_list_batching`, `-q -n 4`): **121 passed, 2 skipped, 1 failed** — the failure is the environment-only `test_my_person_id_is_valid_uuid` (no `.env` in my worktree; fails identically on every head I have tested).
- **Warm hits over HTTP on :8032** with the real data copy, best-of-5, all byte-identical across five repeats: `/statistics` 213 ms → **0.4 ms**; `/birthdays/all` 3.7 → **0.7**; `/me/interactions?days_back=365` 389 → **2.5**; `/me/timeline?limit=500` 20.1 → **2.6**; `/family/interactions` 290 → **2.3**; `/family/timeline?…&limit=500` 19.2 → **2.9**; `/people?sort=strength&limit=300` 273 → **4.6**.
- **Invalidation cycle:** one row committed from a separate connection into `_review876_probe` in the staging **crm.db** → `/statistics` 0.6 ms → **220 ms** (recompute) → 1.2 ms; `DROP TABLE` → 212 ms. **interactions.db** → `/me/interactions` 3.5 ms → **443 ms** → 5.5 ms; drop → 431 ms. Both probe tables dropped; `sqlite_master` shows no `%probe%` tables in either file afterwards. Also dropped a `_review917_ro_probe` used for the WAL check — likewise verified gone.
- **Single-flight over HTTP:** 8 parallel cold `/me/interactions?days_back=365` → per-request 1168–1170 ms, **wall 1173 ms**, against 1021 ms for a single cold request on the same busy box. One computation, as designed.
- **gzip** still applied on hits, `content-length: 34435` identical cold and warm.

### Optional, not blocking

`architecture.md` describes the generation stamp but not the store-time re-check that makes it safe. One clause — "a result computed against a generation that has since changed is not cached" — would leave the next reader with the whole invariant rather than half of it. The module docstring already covers it, so this is a preference, not a gap.

### Housekeeping
Staging on :8032 stopped, `.env` removed, worktree clean apart from the untracked `data` symlink (left in place deliberately — deleting it lets an app import recreate `data/` as an empty real directory, which silently turns 33 CRM tests into skips).

Issue #6, opened 2026-02-28
Auto-repoint stale merged person_ids in interactions


nbramia:
## Problem

\`InteractionStore.add()\` (\`api/services/interaction_store.py:266-314\`) follows the merge chain via \`get_canonical_id()\` for new interactions (line 284-287), but pre-existing interactions created before a merge still reference old (merged) person_ids. The merge script (\`scripts/merge_people.py:287-320\`) updates known interactions, but edge cases exist (e.g., interactions added between merge steps). \`test_all_person_ids_exist\` (\`tests/test_p91_data_integrity.py:66-80\`) catches this; \`scripts/cleanup_interactions.py\` pass C fixes it — but both are reactive.

**Merge chain storage:** \`data/merged_person_ids.json\`
**Chain resolution:** \`api/services/person_entity.py:724-736\` (\`get_canonical_id()\`)

## Impact

Stale merged IDs cause orphaned-looking interactions that fail data integrity tests.

## Proposed Direction

Add \`repoint_merged_ids()\` to InteractionStore that runs as part of the \`person_stats_full\` phase (\`scripts/sync_person_stats.py\`):
1. Load merged_ids mapping from \`data/merged_person_ids.json\`
2. Query interactions table for any \`person_id\` in the merged_ids keys
3. UPDATE those rows to point to the canonical target in a single batch

This makes the nightly sync self-healing for stale merged IDs.

## Acceptance Criteria

- [ ] After \`person_stats_full\`, zero interactions reference merged person_ids
- [ ] Runs automatically in nightly pipeline (called from \`sync_person_stats.py\`)
- [ ] \`cleanup_interactions.py\` pass C finds zero repoint candidates after sync

## Key Files

- \`api/services/interaction_store.py\` — add() merge chain logic (284-287), new \`repoint_merged_ids()\` goes here
- \`api/services/person_entity.py:724-745\` — \`get_canonical_id()\`, \`_load_merged_ids()\`
- \`data/merged_person_ids.json\` — merge chain mapping
- \`scripts/sync_person_stats.py:66\` — calls \`verify_person_stats()\`, integration point for repoint
- \`scripts/cleanup_interactions.py\` — pass C (lines 83-143) currently does this manually

## Dependencies

None — can be done independently.

PR #901, opened 2026-09-04
fix(agents): report only real assignment save errors; host picker as a registry dropdown (#883)

nbramia:
## Summary

Fixes two defects in the board's assignment pickers (issue #883):

- A false red "Couldn't save assignment:" toast appeared on **every successful** model/effort/host/engine change, because the picker's success path called the same error callback the failure path uses (`showError('')` → `onError('')`), which `board.js` turns unconditionally into a toast. Every save was already succeeding (the operator confirmed 200s against the live API); the toast was purely a lie.
- The host field was a free-text `<input>`, requiring the operator to remember and type a registry host name exactly, with no indication of whether that machine was reachable.

## Changes

**Backend**
- New `api/services/agent_worker/host_catalog.py` — `HostCatalog`, an injectable-everything, TTL-cached catalog builder mirroring `model_catalog.py`'s pattern (status_runner + clock test seams, `probe_call_count` for cache-hit assertions), kept deliberately simpler: one probe instead of four, a 30-second TTL as a module constant (`_HOST_CATALOG_TTL_SECONDS`) rather than a new setting.
  - The list is always the API host (`remote_spawn.api_host_name()`, `is_api_host: true`, `online: true` unconditionally) plus every entry in `settings.agent_hosts`, deduplicated by name — a registry entry naming the API host merges into that one row rather than duplicating it.
  - Reachability comes from `tailscale status --json` via an injectable `status_runner`, run inside `asyncio.to_thread`, bounded at `_TAILSCALE_TIMEOUT_SECONDS = 1.5` (comfortably under the endpoint's 2-second budget). A registry host is matched case-insensitively against every Tailscale peer (including `Self`) by comparing the registry name and the ssh_target (stripped of `user@`/`:port`) against the peer's `HostName` and the first label of its `DNSName`.
  - `online` is `true`/`false` only when the probe succeeded AND matched a peer; `null` for every inconclusive case — `tailscale` missing, the probe failing/timing out/returning bad JSON, or a clean run that simply found no matching peer (a host can be reachable over plain LAN ssh with none of that).
  - A cache MISS always re-reads `settings.agent_hosts` fresh (unlike the cached probe result) since the registry is local config, not a network call.
[...]

PR #923, opened 2026-09-04
feat(agents): restrict human moves on agent-owned cards, add Cancel

### LINKED ISSUE #881: feat(agents): restrict human moves on agent-owned cards and add Cancel
## Problem

The board lets the operator move agent-owned cards in ways that desync the card from the agent working on it. A running or waiting-on-answer agent card can be dragged to Unassigned, which strips its assignee while the worker keeps running; dragged to Assigned under a different engine, which relabels a live task; or dragged to Human queue, which marks it blocked with no question attached. An unclaimed agent card can be dragged straight to Done or Human queue, silently closing or re-routing work that was handed to an agent. The drawer's assignee picker permits the same edits. There is no explicit way to cancel an agent card; the only route is Kill on the session, and an unclaimed card cannot be cancelled at all. The morning briefing also names the Human queue section "Waiting on You", a term used nowhere else.

The intent is simple: agent-owned cards are managed by agents. A human may assign, reassign, unassign, or cancel before the worker claims a card; after the claim, the only human actions are Answer, Kill, and Accept.

## Solution

Enforce one rule set for every write path that can change an agent-owned card's lane, assignee, or status, and add a Cancel action.

- **Before claim** (assignee is an agent, no running or blocked tag): the operator may move the card between Unassigned and Assigned, change the assignee, or Cancel. Moves to In progress, Human queue, Review, Scheduled, or Done are refused.
- **After claim** (running or blocked-on-answer tag present): every drag and every assignee, model, effort, or host change is refused with an explanation. Answer, Kill, and Accept remain the only actions.
- **Cancel** is a card action available on any agent-assigned card that is not in Review. If a live session exists it performs the same teardown as Kill, cascading to descendants; then it marks the task cancelled so the card lands in Done, shown under "include cancelled". Cancelling an unclaimed card writes the cancelled status only.
- **Kill** keeps its current outcome: the worker marks the task failed and cancelled, the card lands in Done as cancelled.
- The same rules apply to the lane endpoint, the assignee picker, and the field pickers, through one shared decision function, so the drawer and the board cannot disagree.
- The morning-briefing section heading becomes "Human queue".

## Acceptance Criteria

- [ ] Given a card with assignee `claude` and no running or blocked tag, `PUT /api/agents/board/cards/{id}/lane` to `in_progress`, `human_queue`, or `done` returns 409 with a reason; to `unassigned` or `assigned` with a new assignee succeeds.
- [ ] Given a card tagged `agent-running` or `agent-blocked`, every lane move returns 409, and `PUT /api/tasks/{id}` requests that change the assignee tag or the `model`, `effort`, or `host` fields through the board's write path return 409; the drawer disables the pickers and shows the reason.
- [ ] `POST /api/agents/board/cards/{id}/cancel` on an unclaimed agent card sets status `cancelled` and no session is touched; on a claimed card with a live session it kills that session and every descendant, then sets status `cancelled`; on a Review card it returns 409 "accept or reject the review".
- [ ] After Cancel or Kill, the card appears in Done only when "include cancelled" is on; the vault line carries the cancelled checkbox.
- [ ] Given a card assigned to `me` or with no assignee, all existing human moves keep working exactly as today.
- [ ] The lane endpoint, the assignee picker, and the field pickers call one shared decision function, and a unit test table enumerates every (state, action) pair with its allowed or refused outcome.
- [ ] The morning-briefing prompt's Human queue section heading reads "Human queue"; the existing briefing test asserts the new heading.
- [ ] The server-free browser test covers a refused drag with its toast, disabled pickers on a claimed card, and Cancel on an unclaimed card.

## Verification

- **Automated tests:** unit tests for the decision table; route tests for lane, cancel, and the guarded task update; browser test additions.
- **Manual checks:** on the live board, try to drag a running agent card anywhere and confirm the refusal; cancel an unclaimed test card and confirm the vault line.
- **Existing test suite:** green.

## Scope

**In scope:** the decision function, lane and task-update guards, the Cancel endpoint and drawer button, picker disabling, the briefing heading, product and technical agent-viz spec updates.

**Out of scope:** changing what Kill writes to the vault; the graph view; retry-after-kill behaviour, which stays "cancelled in Done" per the operator's choice.

## Technical Context

- **Key files:**
  - `api/services/agent_board.py` — `plan_lane_move` (:126–222) already refuses In progress and Done for worker-owned cards and In progress for agent-assigned cards; extend it into the shared decision function covering Unassigned, Assigned, Human queue, and field edits. `derive_assignee`, `AGENT_ASSIGNEES`, `RUNNING_TAG`, `BLOCKED_TAG`, `COMPLETED_TAG`, `ACCEPTED_TAG`.
  - `api/routes/agents.py` — lane endpoint (~:870–920), accept endpoint, board card helpers; add `POST /board/cards/{id}/cancel` next to them. Kill mechanics: `api/services/agent_worker/inter_agent.py` `teardown_session` (:705) and the kill endpoint's descendant walk (`agents.py` `/sessions/{id}/kill`).
  - `api/routes/tasks.py` `PUT /{task_id}` — the board's picker writes go through here with `fields.assigned_by: "board"`; guard on that marker so agent-side and vault-side writes are unaffected.
  - `web/agents/board.js` — `renderDrawerActions` and drag handling; `web/agents/assignment.js` — pickers; disable and explain rather than hide.
  - `scripts/seed_proactive_reminders.py` `MORNING_BRIEFING_PROMPT` and `tests/test_audit_e2e.py::test_morning_briefing_surfaces_human_queue`.
  - Worker outcome after a kill: `api/services/agent_worker/worker.py` around :2830–2900 writes `#agent-failed` plus `cancelled`; Cancel reuses that by killing and then writing `cancelled` itself for the unclaimed case.
- **Patterns to follow:** the existing 409 reasons in `plan_lane_move`; the accept endpoint's shape for cancel; the board's toast-on-error pattern.
- **Constraints:** the operator's already-seeded Morning Briefing entry in the vault carries the old heading; the change to the seed prompt does not rewrite it, so note in the PR that the top-level operator updates the live entry.

## References

- `docs/specs/product/agent-viz.md` §Lanes and §Drawer; `docs/specs/technical/agent-viz.md`
- `docs/specs/product/agent-worker.md` §Tag lifecycle
- Follows #850 (PR #856), #851 (PR #858), #852 (PR #857)

nbramia:
## Summary

Closes #881.

The board let a human desync an agent-owned card from the worker actively running it: a claimed card (`agent-running`/`agent-blocked`) could be dragged to Unassigned (stripping its assignee mid-run), Assigned (relabeling a live task), or Human queue (marking it blocked with no question attached); an unclaimed agent card could be dragged straight to Done or Human queue, silently closing or re-routing work handed to an agent. There was also no explicit way to cancel an agent card short of Kill (which needs a live session — an unclaimed card had no route at all).

This adds one shared decision function, `agent_board.evaluate_card_action`, that every write path touching an agent-owned card's lane, assignee, or status now consults, plus a `POST /api/agents/board/cards/{id}/cancel` endpoint.

- **Before claim** (agent-assigned, no `agent-running`/`agent-blocked`, and no live CLI session opened on it): reassign, unassign, or Cancel remain allowed; a drag to In progress, Human queue, or Done is refused.
- **After claim**: every drag and every assignee/model/effort/host/tags change is refused; Answer, Kill, Accept, and Cancel remain.
- **Cancel**: available on any agent-assigned card not in Review and not already finished. Kills the linked session (and its whole subtree, via a helper extracted from the kill endpoint so the teardown mechanics live in exactly one place) if one exists, then marks the task `cancelled`, landing it in Done behind "include cancelled". Refuses with `"accept or reject the review"` on a Review card, `"cancel is only available for agent-assigned cards"` on a `#me`/unassigned card, `"this card is already finished — nothing to cancel"` on an already-done/cancelled one. Idempotent.
- Every task card in `GET /api/agents/board` now carries a server-computed `policy` block; the drawer's assignee select, the Tags field, the model/effort/host pickers, the drag handler, and the Resolve/Cancel buttons all read it instead of re-deriving the rules in JS.
- `PUT /api/tasks/{id}` enforces the same claimed-card guard when the request carries the board's `fields.assigned_by: "board"` marker (the assignment pickers' write path) — agent-side and vault-side writes are unaffected.
- The morning briefing's Human-queue section heading changed from "Waiting on You" to "Human queue" for consistency with the rest of the product.

## Acceptance criteria

- [x] Agent-assigned, unclaimed card: lane PUT to `in_progress`/`human_queue`/`done` → 409 with a reason; to `unassigned`/`assigned` with a new assignee → succeeds.
- [x] Claimed card (`agent-running`/`agent-blocked`, or a live CLI session opened via Open): every lane move → 409; `PUT /api/tasks/{id}` changing the assignee-tag set, the claim-tag set, or `model`/`effort`/`host` through the board's write path → 409; drawer disables the assignee select, the Tags field, and the pickers, and shows the reason.
- [x] `POST /api/agents/board/cards/{id}/cancel`: unclaimed → `status=cancelled`, no session touched; claimed with a live session → kills that session + every descendant, then `status=cancelled`; Review card → 409 `"accept or reject the review"`; already-finished card → 409 `"this card is already finished — nothing to cancel"`.
- [x] After Cancel or Kill, the card shows in Done only behind "include cancelled"; the vault line carries the cancelled checkbox (unchanged Kill behavior; Cancel reuses `TaskManager`'s existing `status="cancelled"` write path).
- [x] Cards assigned to `me` or unassigned: every existing human move keeps working byte-identically — proven by an explicit parametrized decision table plus the pre-existing lane-move tests, all still green.
- [x] The lane endpoint, the assignee/field pickers' guard, and the card's `policy` block all call `evaluate_card_action` — one shared decision function, no duplicated rule logic (including in JS, which reads `card.policy`).
- [x] Morning briefing heading is "Human queue"; `test_morning_briefing_surfaces_human_queue` asserts it.
- [x] Server-free browser tests cover a refused drag + toast, disabled pickers + visible reason on a claimed card, and Cancel on an unclaimed card landing it in Done.

**Note for the operator:** the already-seeded Morning Briefing vault entry still carries the old "Waiting on You" heading — the seed script's prompt constant changed, but re-running it with `--force` does not update an existing entry (see `docs/guides/human-queue.md`). Update the live entry by hand, or delete and re-seed it.

## Test plan

- [x] `tests/test_agent_board.py` — exhaustive (assignee × claim-state × action) decision table for `evaluate_card_action`, including a `cli_opened` condition, plus the existing lane-move suite updated for the rule extensions.
- [x] `tests/test_agents_board_api.py` — extended lane-refusal route tests, a `TestCancelBoardCard` class (unclaimed, claimed + cascade, review, already-finished, idempotent, missing card, the 200-row-window bypass, an untorn-down CLI session, ownership-before-idempotence, cache invalidation, terminal-session skip), and `policy`-block shape tests (including the refused-only shape) on `GET /board`.
- [x] `tests/test_tasks_api.py` — the `assigned_by: "board"` guard, including the tag-SET comparison that closes AR2/AR3.
- [x] `tests/test_audit_e2e.py` — updated heading assertion.
- [x] `tests/test_agents_board_ui_browser.py` / `tests/test_agents_assignment_ui_browser.py` — refused drag + toast, claimed-card drawer disabling the assignee select, Tags field, and pickers with visible reasons in a dedicated neutral element, a refused drag on Scheduled with zero requests, Cancel landing an unclaimed card in Done and re-rendering the open drawer, Cancel's untorn-down-CLI-session toast, a Resolve button correctly offered when its lane is absent from the refused-only policy, and the engine select gated by policy too.
- [x] Full server-free browser set (`browser and not requires_server`) green.
- [x] Pre-push hook's full suite green (see push output).

## Round 2 — adversarial review response

Three independent round-1 reviews (general, hostile-driving-the-real-app, and a cross-model permission-matrix pass) found 7 Action-Required issues and 9 Recommended fixes; all are addressed in this round, each with a mutation-proof test (revert the production hunk → confirm the new/changed test fails → restore → confirm green).

**Action Required:**
- **AR1** — Cancel's session lookup was bounded by the 200-row, most-recently-started snapshot window used for board display, so a still-running session older than that window was silently skipped. Now a direct `SessionStore.get(task_id)` primary-key read (unbounded).
- **AR2** — A second assignee tag added alongside an existing one slipped past the claimed-card guard because `derive_assignee` (first-match-wins) still resolved to the same value. The guard now compares the normalized assignee-tag *set*.
- **AR3** — The drawer's Tags field let a human strip a claimed card's `agent-running`/`agent-blocked` tag (freeing every restriction in one blur) or grant one to a `me` card by typing it. The field now never shows a lifecycle tag as an editable token, rejects one if typed, always preserves the card's own lifecycle tags on save, and is disabled-with-reason on a claimed card exactly like the Assignee select. The server-side guard was tightened to match (tag-SET comparison closes this too).
- **AR4** — A card opened via the board's Open button (CLI session spawned, `status="in_progress"`) but not yet claimed by the worker (`agent-running` not yet added) stayed fully editable. `is_claimed` now also treats that state as claimed, carefully excluding a pending Review card so the accept-by-drag Done carve-out doesn't regress.
- **AR5** — Cancel silently reported success (`killed: []`, no error) on a live cc:/cx: CLI session, which it can't tear down (that's a separate, deferred issue). It now detects a live CLI session via the `cli_sessions` table and reports it under `failures`, toasted as a warning.
- **AR6** — Dropping a card on Scheduled always round-tripped to the server and 400'd (Scheduled isn't a `policy.lanes` key at all — `TASK_LANES` excludes it). Now refused client-side with zero requests, via the same `DIRECT_LANE_IDS` set that already gates the composer.
- **AR7** — Cancel didn't re-render the open drawer, so a stale Open button on a card that just moved to Done would 409 on the next click. Cancel now explicitly re-renders, mirroring the assignee handler's existing workaround for the same focus-retention issue.

**Recommended (accepted):**
- **RC1** `evaluate_card_action` now raises on an unrecognized action instead of falling through to an implicit allow.
- **RC2** Cancel's ownership/review check now runs *before* its idempotence short-circuit, and it refuses an already-**finished** card (`status` `"done"` too, not just `"cancelled"`) — **this narrows the issue's literal "any agent-assigned card not in Review" wording to exclude already-finished cards; flagging for an explicit operator decision on whether that's the intended scope.**
- **RC3** Added a production-shaped Cancel-forbidden fixture (a `me` card with a real `policy.cancel = {allowed:false,...}` block, not "no policy at all").
- **RC4** Added coverage for the terminal-session teardown skip and the board-cache invalidation.
- **RC5** `policy.lanes` now lists only refused lanes (absence = allowed), cutting the per-card board-stream payload.
- **RC6** The disabled-fields explanation now renders in its own neutral element, not the element reserved for a failed save.
- **RC7** The Resolve gate now guards the (now-possibly-absent) `.done` leaf RC5 exposed.
- **RC8** `assignment.js`'s engine select is now gated by policy too.
- **RC9** The decision table gained a `cli_opened` condition covering AR4's production state.

**Deferred (not addressed here, filed separately or left as-is per the original review):**
- An operator-filed `#human` card that also carries an agent assignee loses its Resolve action and can only be Cancelled — this follows correctly from the settled agent-owned rule, but is a real behavior change from before this PR that deserves an explicit operator decision rather than arriving as a side effect.
- Lane-move TOCTOU (the route plans against one snapshot and writes later).
- `docs/specs/product/api-reference.md`'s pre-existing line-count overage.
- The 1280×800 lane-strip layout overflow.
- Actually killing a live `cc:`/`cx:` CLI session (Kill has the same gap Cancel now reports honestly instead of hiding).

## Round 3 — round-2 review response

Round 2's three reviewers (hostile-driving-the-real-app, codex cross-model, and general) found 5 Action-Required issues and 6 Recommended fixes; the general reviewer's relaunch after a session-limit kill added 2 more Action-Required plus 4 Recommended. All are addressed here, each guard's test proven with a mutation (revert the production hunk, confirm the new/changed test fails, restore, confirm green).

**Action Required:**
- **AR1** — Round 1's tags guard was gated on the `assigned_by: "board"` marker, but the drawer's Tags field sends a bare `{"tags": [...]}` PUT with no `fields` key at all, so the marker-gated guard could never fire for the request the product actually sends — a tags edit could silently strip a claimed card's claim tag or grant one to an unclaimed card. The guard now keys on the card's own current claim state, independent of any marker, and runs on every request carrying a `tags` key. The drawer's Tags blur handler also now reads the card's current tags fresh from live board state instead of the stale render-time closure the focus-retention workaround left it bound to.
- **AR2** — The Tags field's lifecycle-tag exclusion was a prefix regex (`/^(agent(-[\w-]+)?|accepted)$/i`) that also caught the bare `#agent` queue tag and any operator label starting with `agent-`, making them silently unremovable. Replaced with an explicit allowlist matching the worker's actual tag constants.
- **AR3** — The status-derived claim (`status == "in_progress"` on an agent-owned card with no claim tag) fired on any card whose status merely read `in_progress`, including a plain vault edit or a board reassignment that leaves a stale `in_progress` status behind — freezing cards with no live session at all. `evaluate_card_action` now takes a `has_live_session` input; every call site passes `SessionStore.has_live_session`'s real answer, computed lazily only when the tag-claim is already false.
- **AR4** (general reviewer's relaunch) — The worker selects candidates by the bare `#agent` tag and its claim swap never touches assignee tags, so a claimed card with no engine-specific assignee at all lost every action including Cancel — the one recovery action such a card needs, with no assignee tag left to edit it back to a workable state. Per the operator's ruling: agent-owned now also means "carries a worker claim tag with no assignee tag," Cancel/Kill/Answer apply to any claimed card regardless of assignee, and no control on a claimed card is ever hidden when refused — Cancel included, which now renders disabled-with-reason instead of disappearing.
- **AR5** (general reviewer's relaunch) — The round-1 tests certifying the tags guard all hand-added the `assigned_by` marker to a request shape the drawer never sends, so re-gating the guard behind the marker would have left the suite green. New tests send the exact unmarked shape, with the positive case (an ordinary tags edit on a claimed card that changes nothing sensitive) covered too.

**Recommended (accepted):**
- **RC1** Kill now disables itself with an explanation for a CLI-backed session instead of offering a button that 404s (the actual teardown is a separate, deferred issue).
- **RC2** Cancel tears the drawer's session panel down through its own cleanup path before re-rendering, so no request aborts visibly mid-teardown.
- **RC3** The disabled-field reason text now uses a token that clears WCAG AA contrast.
- **RC4** `policy.lanes` now emits `scheduled` (and `review`) as refused entries alongside every other lane, so a consumer following "absence means allowed" never has to special-case either one.
- **R7** (general reviewer's relaunch) A genuine subtree-teardown failure (a new `SubtreeTeardownError`, distinct from an individual managed-agent teardown failure that's folded into `failures` instead) now stops Cancel from ever marking the task cancelled, reporting which sessions were already stopped and inviting a retry; a write conflict on the final status update *after* a successful teardown says the session is already stopped and invites a retry, instead of the generic conflict text.
- **R8** (general reviewer's relaunch) The live-CLI-session lookup now runs before the already-cancelled short-circuit, so a repeat Cancel call on a card whose CLI session is still open reports that failure every time, not just the first.
- **M5/M6** (general reviewer's relaunch) The claimed-card field_edit guard now also covers a board-marked raw `status` patch (not just model/effort/host); the bare-tags-PUT lifecycle guard now also covers `agent-completed`/`accepted`, not just the two claim tags.
- **M7** (general reviewer's relaunch) Corrected three technical-spec statements that no longer matched the code: the status-derived claim requires a live session, not status alone; `scheduled` does appear in `policy.lanes`; the tags guard pays for the extra task read on every patch carrying a `tags` key, not only when the marker is present. The product spec's Cancel scope and lifecycle-tag list now match the AR4 ruling and the current allowlist.

Also rewrote every `docs/AGENTS.md` "Current behavior only" narration hit this branch's changes introduced (issue-number citations, round/finding references, before/after framing) into present-tense prose, per the newly-merged `tests/test_no_change_narration.py` gate.

**Operator note carried over from round 2:** the `#human` + agent-assignee Resolve-loss remains a real behavior change from before this PR and still needs an explicit operator decision; it's unaffected by this round's fixes. A CLI session whose `session_end` hook never fires stays non-`ended`, so the live-session claim (AR3) keys on it and Cancel still recovers the card — noted for the CLI-kill follow-up issue.

Full pre-push gate green: 7179 unit tests + 443 server-free browser tests.

## Round 4 — round-3 review response (converged)

Round 3 (general + hostile, page-driven at 1280x800) found no reachable behaviour defect. The remaining accepted findings were coverage and documentation:

- **Positive tests for three accepted fixes that previously reverted green**, each with a fail→restore→pass mutation proof: `#agent` and operator labels stay editable in the drawer's Tags box on a `me` card (mutation: the prefix-regex lifecycle matcher); Kill renders disabled with a "close it manually" reason on a card whose live session is a CLI pane (mutation: drop the CLI branch); the Tags blur handler reads the card's current lifecycle tags from fresh board state so a claim that lands while the box is focused is preserved in the PUT (mutation: read from the render-time closure).
- **Won't-fix, recorded with evidence:** a "no aborted `/stream` request after Cancel's drawer re-render" assertion is not achievable — closing an `EventSource` is always recorded by the browser as `net::ERR_ABORTED`, and the same abort occurs on plain drawer close on the base branch. No operator-visible symptom. The panel-teardown-before-re-render code stays as hygiene; verify's zero-failed-requests check excludes `ERR_ABORTED` on `/sessions/*/stream`.
- **Docs/comments:** the technical spec's agent-owned sentence states that the claim-tag branch applies regardless of any assignee tag (a `#me` card the worker has claimed is agent-owned and offers Cancel); the product spec gives Kill the same CLI carve-out clause Cancel has; a stale helper name in a `tasks.py` comment is corrected; the server's protected tag set carries a comment explaining why it differs from the client's lifecycle set.

Rebased onto `feat/agents-board-polish` at `c31efbc` (module-scoped Playwright fixtures); full server-free browser set and the narration gate re-run green on the rebased head.

### Follow-ups filed
#927 (Kill/Cancel cannot tear down board-opened CLI sessions) · #928 (re-validate board decisions against the state they write to) · #929 (seven-lane board at 1280px) · #930 (split `api-reference.md`).


nbramia:
## Adversarial review — round 1 (consolidated)

Three independent reviewers: **general** (Opus, diff + truth table + scratch-clone reproductions), **hostile** (Opus, drove the real app at 1280x800 over a temp vault on port 8137 with real pointer drags), and **codex** (cross-model, permission-matrix focus). Findings below are the referee's decisions, not a transcript.

The core is right. The general reviewer built the (state x action x lane) truth table independently and ran all 7,920 combinations of the old vs new `plan_lane_move`: **there are no false refusals on `me` or unassigned cards** — the only newly-refused combinations are cards carrying `agent-running`/`agent-blocked`, which is the operator's "after the claim" clause. The hostile reviewer confirmed the same on the real page, plus zero uncaught console exceptions on load and after every allowed interaction. The Review→Done accept-by-drag carve-out survives, the kill-endpoint extraction is behaviourally identical, and the cancel cascade genuinely tears down descendants.

Seven things block.

### Action Required

**AR1 — Cancel silently skips the session teardown for any session outside a 200-row window.** (general A1)
`_find_latest_session_for_task` (`api/routes/agents.py:1508-1516`, used at `:1018`) scans `_build_snapshot()["sessions"]`, which is `list_sessions(limit=200)` ordered `started_at DESC`. A still-running session that started before the 200 most recent is not found, so Cancel writes `status=cancelled` while the worker keeps running and spending. Reproduced: 200 newer sessions + one old running session → `killed == []`, session still `running`, task marked cancelled. The `sessions` table has `task_id` as PRIMARY KEY, so the direct per-task lookup returns it unbounded and without 200 transcript reads.

**AR2 — A second assignee tag slips past the `tasks.py` guard on a claimed card.** (general A2)
`api/routes/tasks.py:545-551` compares `derive_assignee(old)` with `derive_assignee(new)`; `derive_assignee` returns the first match in `ASSIGNEE_TAGS` order, so `["claude"]` and `["claude","codex"]` both derive `"claude"`. Reproduced: `PUT` with `{"tags":["claude","agent-running","codex"],"fields":{"assigned_by":"board"}}` on a claimed card returns **200**. Compare the normalized assignee-tag *set*, not the derived value.

**AR3 — The drawer's Tags field defeats every guard this PR adds, in both directions.** (hostile A1, general A3, codex A3 — all three, independently)
The Tags input renders `agent-running`/`agent-blocked` as ordinary editable tokens and saves via `putTask(card.id, {tags: [...]})` with **no** `assigned_by: "board"` marker (`web/agents/board.js:929-955`), so the new guard never runs. Reproduced entirely through the real page: open the claimed card whose drawer correctly shows a disabled assignee select and disabled model/effort/host selects, clear the Tags box, blur → **200**, tags go `['claude','agent-blocked']` → `['claude']`, lane `human_queue` → `assigned`, and `policy.claimed` flips to `false`. Every restriction is gone in one blur and the live session is orphaned. The converse also reproduces and breaks the rule we were told to protect hardest: typing `agent-running` into a **`me`** card's Tags field writes it, after which every lane move on that `me` card is refused. The marked path does not save it either — a marked tags patch that leaves the derived assignee unchanged also strips `agent-running` with a 200 (same root cause as AR2), and `test_board_marker_tags_unchanged_assignee_skips_assignee_check` currently sanctions it.

**AR4 — A CLI-opened card is live but not "claimed", so it stays fully editable.** (codex A1 — missed by both Opus reviewers)
`cli_session_event` (`api/routes/agents.py:1656-1661`) moves an opened `#claude`/`#codex` card to `status="in_progress"` but never adds `agent-running`; `is_claimed` is tag-only (`api/services/agent_board.py:145`). So after clicking **Open** and the hook registering a live `cx:` session, the card's policy still reports `claimed: false` and permits reassignment, unassignment and model/effort/host edits — recreating exactly the live-agent desync this PR exists to prevent. Verified in source. Note the ordering constraint: a status-derived claim must not capture a Review card, or the accept-by-drag carve-out regresses.

**AR5 — Cancel reports success on a live CLI session it cannot kill.** (codex A2)
CLI sessions live in the separate `cli_sessions` table, so the `SessionStore` lookup returns nothing and Cancel writes `status=cancelled` with `killed: []` while the `cc:`/`cx:` process keeps running. (Kill has the same blind spot today — actually killing CLI processes is a separate issue — but Cancel must not silently claim a teardown it did not perform.)

**AR6 — Dropping on Scheduled always round-trips, 400s, and logs a console error.** (hostile R1)
`_card_policy` builds `lanes` from `TASK_LANES`, which excludes `scheduled`, but the board renders Scheduled as a real drop target. Reproduced: dragging the claimed card onto Scheduled issues the PUT, gets **400**, and logs `Failed to load resource: ... 400 (Bad Request)`. Every other lane on that card is refused client-side with zero requests. The milestone bar is zero failed network requests after each interaction.

**AR7 — The drawer is not re-rendered after Cancel, so its stale Open button 409s.** (hostile R2)
Reproduced: Cancel an unclaimed card with the drawer open → toast fires, the card moves to Done, but the drawer still lists `['Open','Cancel']`; clicking Open POSTs and gets **409** plus a console error. The assignee handler at `web/agents/board.js:901-921` already re-renders explicitly for this exact staleness and documents why `fetchBoard()` alone is not enough.

### Recommended (accepted — please fix, all small)

- **RC1** `evaluate_card_action` fails **open** on an unrecognised action (`agent_board.py:212`); `CARD_ACTIONS` is defined but never used to validate. Raise instead of silently allowing.
- **RC2** Cancel's idempotence short-circuit precedes the permission check, so a cancelled `me` card or a cancelled Review card returns 200 instead of its 409; and `policy.cancel.allowed` stays true on an already-cancelled card, so the button is still offered and a second click toasts "Cancelled." again. Refuse cancel on a terminal-status card in the decision function and check ownership before any short-circuit. **Note this narrows the operator's literal wording** ("any agent-assigned card not in Review") to exclude already-finished cards — call it out in the PR body for the operator to veto.
- **RC3** `test_cancel_button_absent_when_policy_forbids_it` exercises a card with *no* policy block at all, which never occurs in production — add the production-shaped fixture (a `me` card carrying `policy.cancel = {allowed:false, reason:...}`).
- **RC4** No test covers the cancel cache invalidation or the terminal-session skip.
- **RC5** Emit only *refused* lanes in `policy.lanes` (absent = allowed, which the client already assumes). This fixes AR6 uniformly and removes most of the 555–1261 bytes per card that `GET /board/stream` re-serializes every 0.5s to compute its change signature. Pair it with a test of the **positive** case, not only absences.
- **RC6** The fields refusal reason is written into `.assignment-error`, the same element failed saves use, so a normal explanation paints red. Give it the neutral `.drawer-field-reason` treatment the assignee reason already has.
- **RC7** `card.policy.lanes.done.allowed` is read without guarding the leaf (`board.js:1036`); every other policy read in the file guards it.
- **RC8** Gate `assignment.js`'s engine select by policy too — unreachable today only because `board.js` hides that row.
- **RC9** The decision table has no row for the production state that makes AR4 possible (agent assignee, `status=in_progress`, no `agent-running`). Add it as AR4's regression test.

### Deferred (not for this PR)

- An operator-filed `#human` card that also carries an agent assignee loses its Resolve action and can only be cancelled. This follows correctly from the settled agent-owned rule, but it is a real behaviour change that deserves an explicit operator decision rather than arriving as a side effect. Flagged in the PR body.
- Lane-move TOCTOU: the route plans against one snapshot and writes later, so a worker claim landing in between can still be overwritten. `TaskConflictError` covers the write itself; filing separately.
- `docs/specs/product/api-reference.md` is 1,171 lines against an 800-line Product Spec max — pre-existing (1,137 before this PR); splitting it inside a surgical change would be the wrong call. Filing separately.
- At 1280x800 with all seven lanes shown the lane strip is 1484px, so Unassigned and Done cannot be on screen together. Pre-existing layout. Filing separately.
- Actually killing `cc:`/`cx:` CLI sessions (Kill has the same gap as AR5). Filing separately.

### Mutation proofs

Accepted. Every new/modified test has a row, each shows a positive assertion flipping pass→fail→pass, and the general reviewer independently reproduced rows B and I in a scratch clone. Row J is worth calling out as the process working: the mutation caught the implementer's own briefing-heading assertion being vacuous (the bare substring also matched body text), which was then fixed and re-proved. Round 2 needs the same treatment for every new and changed test.

nbramia:
## Adversarial review — round 2 (consolidated)

Head `06905bb` (rebased onto `feat/agents-board-polish` @ 5b3e11c). Reviewers this round: **hostile** (Opus, drove the real app at 1280x800 over a temp vault with real pointer drags, eleven driver runs), **codex** (cross-model, permission-matrix focus), and **general** (Opus — killed by a session limit mid-run, relaunched; its findings fold into round 3).

Most of round 1's fixes hold on the real page: AR4's positive cases (a `me`/unassigned card in In progress keeps every move; a Review card with status `in_progress` still accepts-by-drag), AR6 (Scheduled refused client-side, zero requests), AR7 (no stale Open after Cancel), AR5's warning toast, RC6/RC8, the unclaimed-agent Unassigned↔Assigned path, zero uncaught exceptions, and 442/442 server-free browser tests. The referee's own round-1 direction on AR3 was the weak link: guarding the server on a marker the tags box never sends left the blocker open through a different door. Three things block; all three are the shapes this milestone keeps producing — a guard that can be bypassed and an over-correction.

### Action Required

**AR1 — Round 1's blocker is still open: a tags write can strip a live claim, and the server backstop cannot fire.** (hostile A1 + A2, codex A2)
Two reproductions on the real page. (a) *The stale-drawer race:* open an unclaimed `#claude` card, click into **Tags** and type a label; from another process the worker claims the card (`PUT {"tags":["claude","agent-running"]}` — exactly what `worker.py` does); the board visibly moves the card to In progress but `updateOpenDrawer` skips the drawer re-render while the box holds focus (`web/agents/board.js:178-187`, the pre-existing focus workaround), so the Tags box stays enabled and bound to a stale `card` closure; blur → **200**, server tags `['claude','needs-design']`, lane back to `assigned`, `policy.claimed` false, live session orphaned. Zero console errors, zero failed requests — the page reports total success. (b) *The backstop:* `api/routes/tasks.py:531` wraps the entire AR2/AR3 tags check inside `if fields.assigned_by == "board"`, but the drawer's tags handler calls `putTask(card.id, {tags})` with **no `fields` key** (`board.js:993`). Back to back on the same claimed card: `PUT {"tags":["claude"]}` → **200**, tags become `['claude']`; the identical PUT with the marker → **409**. The two tests certifying AR3 (`tests/test_tasks_api.py:471`, `:490`) hand-add the marker to a request shape no client emits. The same unmarked shape injects `agent-running` onto a `me` card and freezes it — the false refusal the semantics forbid.
**Fix:** guard on the card's own claim state, not on the marker. Any `PUT /api/tasks/{id}` whose `tags` patch would remove a claim tag from a currently-claimed card, or change the assignee-tag set on a claimed card, is refused with the worker-owned 409 — marker or no marker. Adding a claim tag via `PUT` (the `me`-freeze vector): claim tags are written only by the worker through `/swap-tag`; refuse that too unless a grep of services and tests shows a legitimate HTTP caller doing it, in which case say so in the report and rely on the client-side reject. Also make the blur handler read the card's current lifecycle tags from fresh board state (`findCard(card.id)`) rather than the render-time closure. Retire the marker-only tests in favour of unmarked ones that match what the product sends.

**AR2 — Over-correction: `#agent` can no longer be removed from any card through the drawer.** (hostile A3, codex R2)
`LIFECYCLE_TAG = /^(agent(-[\w-]+)?|accepted)$/i` (`board.js:39`) matches the bare `agent` **queue** tag and any operator label starting `agent-`. The Tags box hides them (`:842`), rejects them (`:977`) and force-re-appends them on save (`:991`). Reproduced on a `me` card with tags `['me','agent','notes']`: the box shows `notes`; saving exactly what is displayed returns `['me','agent','notes']`, silently. Same on an unassigned card, where Cancel is not even offered. A capability the base branch had, removed silently, on the card classes the semantics say must be untouched.
**Fix:** an explicit allowlist of the worker's lifecycle tags — `agent-running`, `agent-blocked`, `agent-completed`, `accepted`, plus the worker's terminal tags (`agent-failed`, `agent-budget-exceeded`; confirm names against `worker.py`) — not a prefix regex. Bare `agent` and operator labels stay editable. Positive test: removing `#agent` from a `me` card via the drawer works.

**AR3 — The status-derived claim is too broad: it freezes agent cards with no live session, and the board itself can create that state.** (hostile R1, codex A1)
`is_claimed` treats `status == "in_progress"` on an agent-owned non-review card as claimed (`api/services/agent_board.py:151-175`). Two producers other than CLI-open reach that state: a vault edit from `- [ ]` to `- [/]` on a `#claude` line (reproduced on the page: every drag refused with "answer or kill the session first" — a session that does not exist — assignee and Tags disabled, only Cancel left), and, sharper, a **board reassignment**: an in-progress `me`/unassigned/`#human` card changed to `codex` via the drawer is allowed (the assigned-lane planner rewrites tags only, `:316`), and the very next policy read reports it worker-owned. The operator's semantics define the claim by the tags; the reason AR4 was accepted is that a CLI-opened card has a *live session*. That is the signal to key on.
**Fix:** give `evaluate_card_action` a `has_live_session: bool = False` input (keeps it pure) and define the status-derived claim as agent-owned ∧ not-review ∧ `status == "in_progress"` ∧ `has_live_session`. Every call site — the lane endpoint, cancel, the `tasks.py` guard, and `_card_policy` — passes the real value: a non-terminal `sessions` row for the task OR a non-`ended` `cli_sessions` row. Compute it lazily (only when the tag-claim is false and the status/assignee make it relevant). Tests: codex's transition scenario (in-progress `me` card reassigned to `codex` stays movable), the vault-edited `#claude` `- [/]` card with no session stays movable, a CLI-opened card with a live `cli_sessions` row is locked, and the decision table gains the live-session dimension. The refusal reason must never name a session that does not exist.

### Recommended (accepted — please fix)

- **RC1** (hostile R2) After AR5's "close it manually" toast, the drawer still offers **Kill** for that same `cc:`/`cx:` session and clicking it 404s with a console error. Pre-existing, but this PR now points the operator straight at it. Disable Kill for a CLI-backed session with the same explanation rather than offering a button that cannot work; the real teardown is #927.
- **RC2** (hostile M2) Cancel's explicit drawer re-render tears the session panel down mid-flight, surfacing two `net::ERR_ABORTED` requests (`/stream`, `/summary`). Tear the panel down through its existing cleanup path before re-rendering so a "zero failed requests" assertion stays honest.
- **RC3** (hostile M3) `.drawer-field-reason` at `--text-dim` on the drawer background measures 3.03:1, below WCAG AA, and it is now the only explanation the operator gets for a refused action. Use a token that meets 4.5:1.
- **RC4** (codex R1) `policy.lanes` is documented as "absent means allowed" but omits `scheduled`, so an API consumer following the contract would try the move and 400. Emit `scheduled` (and keep `review`) as refused entries; `DIRECT_LANE_IDS` on the client stays as the fast path.
- **RC5 — docs standard (gate-enforced).** Main merged `tests/test_no_change_narration.py` (`docs/AGENTS.md` "Current behavior only"): docs, comments and docstrings must describe current behaviour, never narrate change — no `(#881)`/`(#881 AR3)` citations in prose, no "round-1 finding", "used to", "now", "no longer", "previously". This PR introduces **47 new hits** (listed with line text in the addresser brief). Rewrite them in plain present tense; the test must pass before every push once the base carries it.
- **RC6** (codex M1) Add the transition test — reassigning an already in-progress human/unassigned card to an agent — so the claim rule is pinned by a scenario the board can actually produce, not only by a synthetic `cli_opened` state.

### Deferred (unchanged from round 1, plus)

- `#human` + agent-assignee cards: AR3's live-session keying relieves the extra lock-down hostile R3 observed; the underlying Resolve-loss remains the operator's call.
- The generic "Can't move card to Scheduled." toast for a claimed card (hostile M1) is honest and correct — Scheduled is never a target for a task card in any state. Declined.
- Lane-strip width at 1280px: #929. CLI kill: #927. TOCTOU: #928. api-reference split: #930.

### Mutation proofs

Round 1's table is accepted. Round 3 must carry a row for every test touched this round, using the commit-first procedure, and for every guard added here the POSITIVE case (something still allowed) must be tested as hard as the refusal.

nbramia:
## Round 2 — addendum (general reviewer, relaunched)

The general reviewer finished after the consolidated comment above. Two more blockers, folded into the same address round.

### Action Required (continued)

**AR4 — A `#agent` task with no assignee, or `#agent #me`, becomes completely un-actionable the moment the worker claims it.** The worker selects candidates by the `#agent` tag alone and `_claim` swaps `#agent` → `#agent-running` without touching assignee tags (`worker.py:1497`, `:1717`). The claim makes `is_claimed` true, so every lane and every picker is refused — but Cancel is refused too, because the card has no agent assignee tag, and the Tags box is disabled. Zero controls, no reason shown, and the only recovery the base branch had (editing the tag) is gone; the worker's own operator notice says "re-tag manually". **Ruling:** *agent-owned* means (has an agent assignee tag) **or** (carries a worker claim tag). The after-the-claim lock-down stays as is; **Cancel, Kill and Answer apply to any claimed card regardless of assignee** — the operator's "after the claim, only Answer, Kill, Accept" applied consistently. And no control on a claimed card is ever hidden — refused controls render disabled with the server's reason (this also covers R3 below). Unit-table rows and a browser test for the no-assignee claimed card.

**AR5 — The marker-independent tags guard has no test that fails without it.** Every tags-guard test still hand-adds `fields.assigned_by`; re-gating the guard behind the marker leaves `tests/test_tasks_api.py` fully green. Add tests sending the exact bare `{"tags":[...]}` patch the drawer emits against a claimed card, asserting 409 and no write, with their mutation proof, plus the positive case (a bare tags patch that changes only ordinary labels still succeeds).

### Recommended (accepted)

- **R7** Cancel tears down, then writes. A raised teardown escapes as a 500 and strands the task with the killed-so-far list lost; a `TaskConflictError` on the final write leaves the session dead and the card still claimed, with a toast that reads like an ordinary refusal. Catch teardown failure (do not write cancelled; return the sessions already stopped in `detail`); on a write conflict after a successful teardown, say so and invite a retry.
- **R8** A second Cancel after the CLI warning returns `failures: []` because the already-cancelled short-circuit runs before the CLI lookup. Look up CLI sessions first.
- **R9** The Cancel button is hidden when refused; disable-and-explain like Kill.
- **R10** The guard now reads the task on every patch carrying `tags` (the set comparison needs it). Accept the cost; correct the technical spec sentence that says an unmarked request never pays for the read.
- **M5/M6** Latent API-surface gaps in the marked guard: a `status` patch bypasses it, and the protected tag set omits `agent-completed`/`accepted`. Close both.
- **M7** Three technical-spec statements are wrong for the code now being written (claim requires a live session; Scheduled does appear in `policy.lanes`; the extra-read sentence) and the product spec's "only actions left" sentence must match the AR4 ruling.

### Deferred
- A CLI session whose `session_end` hook never fires stays non-`ended`, and the live-session claim now keys on it; Cancel still recovers the card. Noted on #927.

### Verified clean by this reviewer
Mutation rows 6 and 9 reproduce exactly as claimed; `list_cli_sessions_for_task` is safe on a fresh DB and its status filter matches the writer; `review` is refused in every one of 180 enumerated states and no `{"allowed": true}` entry can appear; the SSE signature changes on every claim transition; no case/whitespace bypass of the guard; unmarked agent-side and vault-side field writes are unaffected.

### Process
The reviewer saw the shared worktree change under it with uncommitted address work. From round 3 on, the addresser is dispatched only after every reviewer report is in, and reviewers read a committed head.

nbramia:
## Adversarial review — round 3 (consolidated) — CONVERGED

Head `a2f304e`. Reviewers: **hostile** (Opus, real app at 1280x800 over a temp vault, real pointer drags) and **general** (Opus, truth table rebuilt by executing the decision function, mutation spot-checks in a scratch clone). Refereed under the recalibrated bar: only defects reachable on a normal path count.

**No reachable behaviour defect survives.** The hostile pass verified every scenario end to end: unclaimed `#claude` cards move Unassigned↔Assigned and are refused elsewhere with zero writes; claimed cards refuse every drag with every picker disabled-and-explained and Cancel working; the formerly un-actionable `['agent-running']` no-assignee and `['me','agent-running']` cards now render a populated drawer with a working Cancel; fourteen drags across `me`/unassigned/`#human` in-progress cards behave exactly as the base; the in-progress `me` → `codex` transition and the vault-style `- [/]` `#claude` card stay movable; Review→Done still accepts and Cancel on Review is a disabled button with its reason; `#agent` is removable again from the Tags box; Cancel with the drawer open leaves no stale Open, no 404, and Kill on a CLI-backed card is disabled with the "close it manually" reason; reason text measures 5.39:1 in both colour schemes; zero console exceptions throughout; 443/443 server-free browser tests. The general pass confirmed by execution that the broadened `is_agent_owned` moved no lane outcome for any `me`/unassigned/`#human` card, that all four `has_live_session` call sites pass real values, that the marker-independent tags guard blocks no legitimate caller (worker `/swap-tag`, `/complete`, accept endpoint, human-queue resolve, vault sync, MCP), that `operator_kill_session` is behaviourally unchanged, and that the live-session claim, the unmarked tags guard and the `policy.lanes` scope are all genuinely mutation-proven.

### Action Required — coverage, not behaviour (one final address, no round 4)

**AR1** (general A1) — AR2's fix has no test that fails when reverted: the round-2 prefix regex restores with 135/135 browser tests green, and the positive test round 2 explicitly asked for was never written. Add server-free browser assertions on an unclaimed `me` card carrying `['me','agent','agent-notes','notes']`: the Tags box renders `agent agent-notes notes` as editable tokens, and clearing `agent` sends a PUT whose tags no longer contain it.

**AR2** (general R1) — two more accepted fixes revert green: RC1 (Kill disabled for a CLI-backed session — the existing test only checks visibility) and AR1's client half (the Tags blur reads fresh board state). Add one assertion each: a claimed card with a `claude_code` session renders Kill disabled with "close it manually"; with the Tags box focused, push a board frame that adds `agent-running`, then blur — the PUT must carry `agent-running`.

### Won't fix (recorded with evidence)

- **RC2** (round 2) — "no aborted `/stream` request on Cancel's re-render" is not achievable: closing an `EventSource` is always recorded by the browser as `net::ERR_ABORTED`, and the same abort appears on plain drawer close on the base branch (hostile M1, reproduced). No operator-visible symptom. The teardown-before-re-render code stays as hygiene; no test is required for it, and verify's zero-failed-requests check excludes `ERR_ABORTED` on `/sessions/*/stream`.

### Minor (fold into the same address)

- **M1** `docs/specs/technical/agent-viz.md` says the second `is_agent_owned` branch fires "with no assignee tag at all"; the code ORs unconditionally (a `['me','agent-running']` card is agent-owned and gets Cancel — correct under the ruling). Fix the sentence; same phrasing in a test comment at `tests/test_agent_board.py:586`.
- **M2** Product spec lists Kill as "stop a running session" without the CLI carve-out Cancel's sentence already has. Add it.
- **M3** `api/routes/tasks.py:23` comment names a helper that does not exist (`_has_live_session`); it is `SessionStore.has_live_session`.
- **M4** The server's protected tag set omits `agent-failed`/`agent-budget-exceeded` while the client's lifecycle set includes them; neither affects lane or claim, so it is an asymmetry, not a hole — one comment sentence explaining why.

### Deferred (tracked)
#927 CLI kill · #928 TOCTOU · #929 lane-strip width at 1280 (visibly reproduces; already tracked) · #930 api-reference split · the `#human`+agent-assignee Resolve question (operator).

After this address: docs gate → verify → merge into `feat/agents-board-polish`.

nbramia:
## Docs Compliance Gate Round 1 — Consolidated Review & Referee Decisions

Head reviewed: `775c492` (rebased locally onto `feat/agents-board-polish` @ `2b61756` before addressing). Reviewer: **docs** (Opus). Standard: docs describe current behavior only (`tests/test_no_change_narration.py`), and match the code.

### Reviewer Findings

#### Docs
**Action Required**
- A1 `docs/specs/technical/agent-viz.md` ~503: "`operator_kill_session` itself is now just the 404/terminal pre-checks…" — change narration.
- A2 `docs/specs/product/api-reference.md` ~390-396: the `PUT /api/tasks/{id}` guard paragraph is backwards — per `api/routes/tasks.py` the tags guard runs on any patch carrying `tags` (marker or not), adding a lifecycle tag 409s regardless of claim state, the comparison is of assignee-tag and claim-tag sets, and a board-marked raw `status` change trips the field-edit guard.
- A3 `api-reference.md` ~528-530: Cancel's 409 condition is wrong — the route gates on `is_agent_owned` (a claimed bare `#agent` card returns 200), and the already-`done` 409 is missing.
- A4 `api-reference.md` ~484-486: "not just `in_progress`/`done`" is narration; the worker-owned definition omits the live-CLI-session claim path.

**Recommended**
- R1 product spec line 50: "Answer, Kill, Accept, and Cancel are the only actions left" is incomplete (Focus is still offered on a claimed CLI-backed card).
- R2 product spec line 47: "as they always have" is history.
- R3 `docs/guides/human-queue.md` `Last Updated` stale.
- R4 technical spec cancel row: "before deciding anything else" contradicts its own opening sentence.
- R5 `tests/test_no_change_narration.py`: the `now (...)` alternation lacks `just`, which let "is now just" through.

### Referee Decisions

| # | Finding | Reviewer Severity | Concern | Decision | Reasoning / smallest remedy |
|---|---------|-------------------|---------|----------|-----------------------------|
| A1 | kill handler narration | Action Required | Valid | Accept | Rewrite as present-tense current behavior. |
| A2 | tasks PUT guard paragraph backwards | Action Required | Valid (verified against `tasks.py` 536-655) | Accept | Rewrite to the four verified facts. |
| A3 | cancel 409 condition wrong / finished 409 missing | Action Required | Valid (verified against `agents.py` cancel route) | Accept | Document `is_agent_owned` gating and `CANCEL_ALREADY_FINISHED_ERROR`. |
| A4 | lane bullet narration + incomplete claim definition | Action Required | Valid | Accept | Drop "not just", add the live-session claim path. |
| R1 | incomplete control list | Recommended | Valid | Accept | List what the drawer actually still offers. |
| R2 | "as they always have" | Recommended | Valid | Accept | Present-tense rewrite. |
| R3 | stale Last Updated | Recommended | Valid | Accept | Bump touched docs to 2026-09-07. |
| R4 | cancel row self-contradiction | Recommended | Valid | Accept | State the real order. |
| R5 | widen ratchet with `just` | Recommended | Valid | Accept conditionally | Only if no new baseline entries are needed. |

**Findings forwarded to addresser:** 9

**Address result (commit `e29f435`, docs-only):** A1–A4 and R1–R4 fixed. R5 skipped: widening the regex tripped two files outside this PR, so the ratchet is unchanged and the baseline untouched. `tests/test_no_change_narration.py` passes. Mutation proofs not applicable (pure docs).

nbramia:
## Docs Compliance Gate Round 2 — Consolidated Review & Referee Decisions

Head reviewed: `e29f435` (local; push pending). Reviewer: **docs** (Opus). Round-1 items A1–A4, R1–R4 re-verified as fixed against the code; narration ratchet passes (1 passed).

### Reviewer Findings

#### Docs
**Action Required**
- `api/services/agent_board.py:249-250` — `evaluate_card_action` docstring says "Answer, Kill, and Accept are the only actions left"; the function's own `cancel` branch allows Cancel on a claimed card.
- `api/routes/agents.py:1146-1147` — cancel route docstring omits the already-finished refusal (`CANCEL_ALREADY_FINISHED_ERROR` on `status == "done"`).
- `docs/specs/technical/agent-viz.md:505` — Operator kill response documented as `{killed, failures: [{session_id, error}], reason: <input>}`; the handler returns `{killed, failures}` with failure entries `{session_id, reason}`, and `reason` only on the already-terminal early return. Contradicts this PR's own api-reference text.

**Recommended**
- `docs/specs/technical/agent-viz.md:119` — precedence rule 2 says `cancel` is allowed on a claimed card, but the cancel branch evaluates `is_review_pending` first, so a claimed + pending-review card is refused.

**Minor**
- `docs/specs/technical/agent-viz.md:145` — "the assignment pickers (including the engine select)" reads as though a disabled engine select is visible; the drawer hides that row.
- `docs/specs/product/agent-viz.md:341-348` — no back-link to `api-reference.md` (bidirectional-links rule).
- `web/agents/assignment.js` — pre-existing `(#901 round N, finding …)` comments adjacent to this PR's additions; grandfathered in the baseline.

### Referee Decisions

| # | Finding | Reviewer Severity | Concern | Decision | Reasoning / smallest remedy |
|---|---------|-------------------|---------|----------|-----------------------------|
| 1 | `evaluate_card_action` docstring wrong | Action Required | Valid | Accept | One-line docstring fix. |
| 2 | cancel route docstring omits finished refusal | Action Required | Valid | Accept | One-sentence docstring fix. |
| 3 | kill response shape wrong | Action Required | Valid (verified: `operator_kill_session` returns `{killed, failures}`; failure entries carry `reason`) | Accept | Replace the line with the real shape. |
| 4 | precedence inverts cancel vs pending review | Recommended | Valid | Accept | One clause in rule 2. |
| 5 | engine select wording | Minor | Valid | Accept (folded) | One clause. |
| 6 | missing back-link | Minor | Valid | Accept (folded) | Add one Related Documents line. |
| 7 | pre-existing `#901` narration in assignment.js | Minor | Valid but out of scope | Reject | Grandfathered in `tests/narration_baseline.json`; not introduced by this PR. |

**Findings forwarded to addresser:** 6

nbramia:
## Verification — PASS (head `e29f435`)

Verifier: Opus, real routers (`agents`, `agent_assignment`, `tasks`, `scheduler`) mounted from the worktree on `127.0.0.1:8181` over a temp vault and temp data dir; a second server on `:8182` ran the base commit `2b61756` for a differential. Playwright/Chromium at 1280x800 against `/agents`, capturing console, `pageerror`, `requestfailed` and every response >= 400.

| Acceptance criterion | Evidence |
|---|---|
| Unclaimed `#claude`: In progress / Human queue / Done refused; Unassigned + reassign allowed | API: three 409s with the documented reasons, two 200s. UI: each refused drag toasted the server reason and the card stayed in Assigned by DOM assertion after the drop and after a re-fetch; drawer reassign to `codex` persisted. |
| Claimed (`agent-running`, `agent-blocked`): every drag refused, every field write refused, pickers disabled with a visible reason | 10/10 lane moves 409. Board-marked model/effort/host PUTs 409. Bare PUTs adding `agent-running`, swapping the assignee tag, or stripping the claim tag 409; a bare unrelated-tag PUT on a `#me` card 200. Drawer: assignee, Tags, engine, model, effort, host all `disabled` with the reason rendered as text; Kill/Cancel (running) and Kill/Answer/Cancel (blocked) enabled. |
| Cancel | Unclaimed `#claude` -> `cancelled`, no session touched. Claimed with a seeded session + descendant -> both sessions `failed`, then `cancelled`, `killed: [root, child]`. Review card -> 409 "accept or reject the review" (drawer shows Cancel disabled with that reason, not hidden). Claimed bare `#agent` -> 200 cancelled. Already-done accepted -> 409 "this card is already finished — nothing to cancel". `#me` -> 409. |
| Done only with "include cancelled"; vault checkbox | Card absent from the DOM with the filter off, in Done with it on; vault line `- [-] … [cancelled:: 2026-09-07] … #claude`. |
| `#me` / unassigned moves unchanged | 30 (start-state x target-lane) transitions replayed against head and base servers: 0 differences. On the page a `#me` card landed in In progress, Human queue, Done, Unassigned. |
| One shared decision function + matrix test | Lane endpoint, board-marked field PUT, bare tags PUT, cancel route and `_card_policy` all call `agent_board.evaluate_card_action`; `tests/test_agent_board.py` enumerates the matrices. |
| Briefing heading | `scripts/seed_proactive_reminders.py:68` reads `**Human queue**`; `test_morning_briefing_surfaces_human_queue` passes and asserts the old heading is absent. |

Console/network: zero console errors, zero uncaught exceptions, zero failed requests other than expected `net::ERR_ABORTED` on SSE streams during reloads, zero unexpected >= 400 responses.

Suites (under the shared lock, `-n 4`):
- `browser and not requires_server`: 472 passed in 100s.
- `test_agent_board.py test_agents_board_api.py test_tasks_api.py test_audit_e2e.py`: 954 passed.

No defects found. Two non-defects recorded: `human_queue -> assigned` and `done -> unassigned` are tags-only writes that leave the card in place (identical at base, documented); at 1280x800 all seven lanes overflow horizontally so an off-screen synthetic drop never delivers `mouseup` (a Playwright artifact; the harness used the board's lane filter).

nbramia:
## Docs Compliance Gate Round 3 — Consolidated Review & Referee Decisions

Head reviewed: `f49bfb3`. Reviewer: **docs** (Opus). All six round-2 fixes verified against the code (`evaluate_card_action` docstring, cancel route docstring, kill response shape, precedence rule 2, engine-select clause); narration ratchet passes (1 passed).

### Reviewer Findings

#### Docs
**Action Required**
- `docs/specs/product/agent-viz.md` Related Documents — round-2 fix #6 added a second `api-reference.md` entry; the section already carried one (bidirectionality was already satisfied), so the file now lists `api-reference.md` twice with different taglines.

**Recommended** — none.

**Minor**
- `api/services/agent_board.py:250` — the reflowed docstring line runs ~82 chars against a ~72-char wrap elsewhere in the docstring; no lint enforces it.

### Referee Decisions

| # | Finding | Reviewer Severity | Concern | Decision | Reasoning / smallest remedy |
|---|---------|-------------------|---------|----------|-----------------------------|
| 1 | duplicate `api-reference.md` entry | Action Required | Valid (the round-2 referee accepted a mistaken finding) | Accept | Delete one of the two lines, keeping the newer tagline that names the accept and cancel endpoints. |
| 2 | docstring line length | Minor | Valid, cosmetic | Reject | No configured lint; not worth a commit on its own. |

Also in this round: the base branch advanced to `41677e7`; the PR is being rebased onto it and re-gated before merge.

nbramia:
## Verification round 2 — PASS at the exact head `e919c6a`

The branch moved after the round-1 PASS at `e29f435`: two docs/docstring-only commits, a rebase onto the base tip `41677e7` (which brought in the snapshot-cache row-copy fix and the Hermes-reported-model change, both touching `api/routes/agents.py` and `session_store.py`), and one doc line deletion. The verifier rebuilt its real-router harness at `e919c6a` (ports 8181/8182, temp vault, base server from `git archive 41677e7`) and re-ran everything except the suites, which the pre-push gate had already run on this exact head.

- Playwright UI checks at 1280x800: 26/26 passed; console errors 0, page errors 0, failed requests 0 (only the expected SSE `ERR_ABORTED` on reload), responses >= 400: none.
- API acceptance checks: 25/27, the two non-passes being the harness's known sequencing artifact on tags-only lane writes, disproved as a regression by the differential.
- Human-card head-vs-base differential against `41677e7`: 30 transitions, 0 differing.
- Review-lane 409 reasons: 4/4 match `api-reference.md`.

```
verification-head: e919c6a5d1ac05d1a4d2aaab07b662d4f9d8de64
suite-result: pass
suite-command: python -m pytest -m "browser and not requires_server" tests/ -q --tb=line --no-header -p no:cacheprovider   (pre-push gate; /tmp/lifeos-prepush/feat_agent-card-move-rules-753142-browser.log: 472 passed; unit log: 7282 passed, 9 skipped)
suite-executions: 1
suite-exit-status: 0
```

PR standards check: 6/7 PASS, 1 WARN (the description carries per-round review-response sections), 0 FAIL. Docs gate: round 3's single finding (duplicate Related Documents link) fixed by the one-line deletion in `e919c6a`, confirmed by inspection; no further docs round run for a one-line deletion.

PR #924, opened 2026-09-04
chore(crm): final verification follow-ups (docs, tab highlight, birthdays navigation, residual refetches)

nbramia:
## TL;DR

This closes out the final verification pass on the CRM performance milestone. It fixes a batch of documentation inaccuracies the end-to-end verifier found (wrong log-file paths, a description that called a working filter "a no-op," an undocumented config field, a few missing cross-links, an oversized reference doc), plus five small, cheap product bugs the same pass flagged: an unstable sort order in a "who have you been neglecting" list, a tab that highlighted the wrong button when you opened a person's Timeline tab directly, a full page reload when clicking a person from the Birthdays page, three network requests that fired again on a second visit to a page that should have used its cache, and a Tone Evolution chart that quietly kept showing old data after a failed refresh with no visible sign anything went wrong.

Not tied to a single tracked issue — this is the CRM performance milestone's end-to-end verification follow-up pass (docs findings 1–13 and code findings 14–19 from the verifier's report).

## Design

- **Docs**: corrected the slow-request log path (systemd on Linux writes `logs/server.log`, not the macOS launchd path), removed an inaccurate "pinned" claim about a transitive dependency, corrected a stale line-count figure, clarified how a relationship-strength filter actually behaves instead of calling it a no-op, documented an undocumented config field and its effect on avatars, documented a database table's schema, added missing cross-references between related docs, removed changelog-style narration and implementation detail that had crept into consumer-facing specs, corrected a stale parameter list, and split the oversized API reference doc into a new file the same way its CRM section was already split out.
- **Neglected contacts ordering**: added person id as a final tie-breaker so two equally-overdue contacts always render in the same order.
- **Tab highlighting**: scoped the person-detail tab bar's active-tab selector so it can't match the Birthdays page's identically-named tab.
- **Birthdays navigation**: a person row on the Birthdays page now navigates the same client-side way every other CRM link does, instead of reloading the whole page.
- **Session caching**: extended the existing per-session caching pattern to three requests that were still firing on every visit (a family member list, the owner's aggregate stats, and the owner's own profile fetch), and wired their invalidation into the existing merge/split cache-clearing so a mutation that could affect family membership doesn't leave a stale list behind.
- **Tone Evolution failure notice**: when a manual refresh can't recompute a single month and everything falls back to stored data, the card now shows one small banner in addition to the existing per-point "stale" markers.

## Implementation Notes

- `docs/guides/operations.md`, `docs/guides/troubleshooting.md`, `AGENTS.md`: corrected slow-request/error-log paths for Linux systemd vs. macOS launchd.
- `docs/specs/technical/client-surfaces.md`, `api/main.py`: "the installed Starlette (0.52.x)" instead of "pinned Starlette version (0.52.1)".
- `docs/specs/technical/architecture.md`: updated `crm.py`'s LOC figure.
- `docs/specs/product/api-crm.md`, `docs/specs/product/crm-graph.md`: `min_strength` description corrected (it's validated to 0.0–1.0 while `relationship_strength` is 0–100, so only a narrow band of values filters anything).
- `docs/specs/product/api-crm.md`, `docs/specs/product/crm-people.md`: documented `photos_enabled` on `GET /api/crm/config` and that avatars are gated on it alongside `has_profile_photo`.
- `docs/specs/technical/observability.md`, `docs/guides/operations.md`: documented the access-log query-string redaction in `api/services/log_redaction.py`.
- `docs/specs/technical/frontend.md`, `docs/specs/product/crm-graph.md`, `docs/specs/product/crm-ui.md`: added current-state sections for bounded-neighborhood graph loading and client-side dashboard/tab navigation; moved an implementation detail (`history.pushState`) out of the product spec.
- `docs/specs/technical/data-and-sync.md`, `docs/specs/product/crm-analytics.md`: documented the `tone_analysis_results` schema (`api/services/tone_analysis_store.py`) and shrank a duplicated description in the product spec down to a pointer.
- `docs/specs/technical/observability.md`, `docs/guides/configuration.md`, `docs/specs/product/crm-graph.md`, `docs/specs/product/api-crm.md`: added the three missing return links.
- `docs/specs/product/api-reference.md` (was 1159 lines, over the 800-line target) → split its Chat/Search, Google Integration, and Messaging sections into new `docs/specs/product/api-communications.md` (450 lines), following the same pattern as the existing CRM/MCP-tools splits; updated all inbound anchors, including two in `client-surfaces.md`.
- `docs/specs/product/api-crm.md`: removed "in SQL rather than Python" implementation detail; corrected `GET /api/crm/people`'s parameter list against the actual route (`dunbar_circles`, `tags`, `has_interactions`, `min_interactions` were undocumented; `has_pending`/`order` don't exist).
- `docs/specs/product/crm-graph.md`: removed two internal function names (`calculateOptimalEdgeThreshold`, `compute_person_category`) from the product spec.
- `api/routes/crm.py` — `list_neglected_contacts`'s sort key (around the `neglected.sort(...)` call) now ends in `x.person_id` as a tie-break.
- `web/crm.html` — `switchTab()`: `.tab[data-tab="..."]` instead of the unscoped `[data-tab="..."]`. `navigateToPerson()`: calls `navigateTo()` instead of setting `window.location.href`. Added `meStatsCache` (mirrors the existing `meInteractionsCache` pattern) and skip the person-detail cache's background revalidation fetch for the owner's own page. Added `familyMembersLoaded` (a plain `familyMembers.length === 0` check can't tell "not loaded" from "loaded, genuinely zero members") and reset it (alongside `familyMembers`/`familyData`) in `confirmMerge()`/`confirmSplit()`. `renderToneTimeline()` now shows a `.tone-refresh-failed-notice` banner when a `refresh=true` load comes back with every month `status: "stale"`.

### Automated tests

New/extended tests, each verified to fail against the pre-fix code and pass against the fix:

```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_me_family_aggregates_oracle.py tests/test_crm_client_nav_ui_browser.py -q
...
tests/test_me_family_aggregates_oracle.py ................               [ 45%]
tests/test_crm_client_nav_ui_browser.py ...................              [100%]
35 passed in 15.88s
```

- `test_me_family_aggregates_oracle.py::TestMeInteractionsOracle::test_neglected_contacts_tie_broken_by_person_id` — two contacts with an identical sort key except person id now come back in a stable, deterministic order.
- `test_crm_client_nav_ui_browser.py::TestDirectLoadTabHighlight` — a direct load of `/crm/{id}/timeline` highlights the person-detail Timeline tab, not the Birthdays page's same-named tab.
- `test_crm_client_nav_ui_browser.py::TestBirthdaysNavigateToPersonIsClientSide` — clicking a Birthdays timeline person issues zero document loads.
- `test_crm_client_nav_ui_browser.py::TestDashboardStatePersists` (extended) — a Me → Family → Me → Family round trip issues at most one request each for `/me/interactions`, `/me/stats`, the owner's own `/people/{id}` fetch, and `/family/members`.
- `test_crm_client_nav_ui_browser.py::TestToneRefreshFailureNotice` — an all-stale `refresh=true` response shows the failure banner; a normal (non-refresh) load never does.

Full suites, run under the shared push lock exactly as a push would:

```
flock -w 7200 /tmp/lifeos-push.lock ./scripts/test.sh
6390 passed, 8 skipped, 806 warnings in 280.00s
```

```
flock -w 7200 /tmp/lifeos-push.lock <server-free browser scope: browser and not requires_server>
357 passed, 6785 deselected in 275.07s
```

```
flock -w 7200 /tmp/lifeos-push.lock git push ...   # pre-push gate, same two suites, re-run against the pushed tree
Running unit tests... passed (6390 passed, 8 skipped, 805 warnings in 235.15s)
Running server-free browser tests... passed (357 passed, 6785 deselected in 279.89s)
```

Also ran `tests/test_crm_api.py` (integration, real staging data): 57 passed, 2 skipped (pre-existing environment-dependent skips, untouched per the task brief).

### Manual verification

Private staging instance on `127.0.0.1:8036` (venv `uvicorn`, staging `.env` — Telegram/Anthropic/autodeploy/agent-worker/local-LLM autostart all disabled), against the real staging data copy. Playwright script (Chromium, console-error capture on every page, `d3`-proxy noise filtered like the test suite does):

| Page | Result |
|---|---|
| `/me` | No console errors |
| `/family` | No console errors |
| `/birthdays`, then clicking a timeline person row | Navigated client-side to `/crm/{id}` (`window.location.pathname` confirmed, no console errors) |
| `/relationship`, real config (no partner configured on this staging copy) | Renders the no-partner empty state, no console errors |
| `/relationship`, with a synthetic (non-real) partner id and a stubbed all-stale `refresh=true` response | `.tone-refresh-failed-notice` appeared with the exact text "Refresh failed; showing stored scores"; no console errors |
| A person overview | No console errors |
| A direct `/crm/{id}/timeline` load | `.tab[data-tab="timeline"]` carried the `active` class (not the Birthdays tab); no console errors |

Staging instance killed and the copied `.env` deleted afterward.

## Review focus

- The api-reference.md → api-communications.md split touches a lot of line count but no behavior; the two things worth double-checking are (1) every inbound anchor link (especially the two in `client-surfaces.md` pointing at `#get-apichatturn-context`) was updated to the new file, and (2) nothing outside `docs/` referenced an anchor that moved (I grepped the whole repo for this).
- `docs/specs/product/crm-graph.md`'s `min_strength` wording and `docs/specs/product/api-crm.md`'s matching line were reworded to describe the current (narrow-band) behavior accurately without calling it a "no-op." This PR only corrects the description — the underlying scale mismatch (`min_strength` validated to 0.0–1.0 while `relationship_strength` is 0–100) is a separate, already-tracked fix in another PR's scope, not touched here.
- `familyMembersLoaded`, `meStatsCache`: these are new per-session-only caches (no TTL, no invalidation beyond merge/split) — consistent with the existing `meInteractionsCache`/`familyMembers` patterns already in the file, but worth confirming that's still the intended tradeoff for this milestone.
- Three pre-existing findings from the verifier's report were deliberately left untouched per the task brief: the tab-selector collision, the Birthdays hard-navigation, and the neglected-contacts tie-break were all confirmed pre-existing on `main` too (not introduced by this branch) — fixed here anyway since they were cheap and already investigated; three environment-dependent test skips were left alone as instructed.


nbramia:
Addressed all six review nits in `035c17d`:

1. **Tautological tie-break test** — `test_neglected_contacts_tie_broken_by_person_id` now monkeypatches `get_person_julianday_timestamps` to return rows reversed, so the assertion exercises the `x.person_id` sort key instead of coincidentally passing via the covering index's natural person_id-ascending order. Confirmed it fails with the key removed.
2. **Dangling anchors** — all three `#get-apiconversations` references in `api-communications.md` now point to `api-reference.md#get-apiconversations`, where that section actually stayed.
3. **Tautological tone-notice test** — the first-load (non-refresh) fixture is now all-stale too, so "must not show on first load" pins the `toneCacheWasRefreshAttempt` half of the condition rather than the already-false `allStale` half. Confirmed it fails if that half is dropped.
4. **min_strength wording** — `api-crm.md` and `crm-graph.md` now say "the only nodes it can drop are those with a `relationship_strength` below 1.0 (in practice, strength-0 nodes)".
5. **selectedFamilyIds reconciliation** — `confirmMerge()`/`confirmSplit()` now refetch family members immediately after invalidating and drop any id no longer present from `selectedFamilyIds` (only on a successful refetch, so a transient failure doesn't wipe the selection).
6. **invalidatePersonCache on edit** — `saveNotes()`, `addTag()`, `removeTag()`, `addBirthday()`, `removeBirthday()` now invalidate the cached person after a successful save. New browser test (`TestNotesEditInvalidatesPersonCache`) proves it without a timing race: the post-edit refetch is left permanently pending, so a correctly-invalidated cache shows the loading skeleton while an un-invalidated cache would render the stale cached person synchronously — the two outcomes never converge.

**Gate:** `Running unit tests... passed (6390 passed, 8 skipped, 803 warnings in 283.93s)` / `Running server-free browser tests... passed (358 passed, 6785 deselected in 293.47s)`

Rebase wasn't needed — `feat/crm-performance` tip was still `d0591a6`.

Issue #733, opened 2026-08-26
feat(observability): daily rollup tables and a Trends view for agent spend, sync, and latency


nbramia:
## Context

Investigated whether Prometheus + Grafana would close the "no longer-term analytics" gap. Conclusion: **no** — the gap is an aggregation/query gap over data LifeOS already owns, not a missing telemetry pipeline. Full analysis with file:line evidence, a six-option comparison, and resource estimates lives in the investigation doc; the substance is summarized here.

Four SQLite telemetry stores already exist and work — `perf_traces.db` (2,673 chat spans / 6mo), `agent_sessions.db` (226 sessions / 3mo, plus a sparse `daily_spend`), `usage.db` (1,075 chat-API cost rows / 7.5mo), `sync_health.db` (4,293 runs, 200 errors / 7mo). Total footprint **~4.3MB over seven months**. Alerting (CRITICAL/WARNING/INFO tiers, Telegram + email, maintenance suppression) is already implemented and tuned for a single-operator household.

What is genuinely missing is time-bucketed aggregation. `/agents` is a live SSE snapshot by design — its spend chip sums the current filtered view, so there is no history anywhere. Adding Prometheus+Grafana would cost two always-on services (~300-500MB RSS) on a host that has already OOM'd under load, and would violate the project's own bar against anything broken-by-default for a fresh cloner — while replacing working alerting with something that would need re-integrating.

## Scope (first increment)

- Nightly `agent_daily_rollup` and `sync_daily_rollup` tables — plain `GROUP BY` over the existing small tables, written by the nightly pipeline.
- One `GET /api/observability/trends` endpoint serving those rollups.
- One Trends view: agent spend and session outcomes by day/model/routing, sync phase duration and error rate by source, chat latency percentiles bucketed by week.
- No new services, no new dependencies, no config flags — ships like any other LifeOS page and works on a fresh clone.

Backfill the rollups once from existing rows so the first render is not empty.

## Explicitly out of scope

Prometheus, Grafana, VictoriaMetrics, OpenTelemetry, and a Grafana SQLite datasource were all evaluated and rejected for this system's scale. If LifeOS ever runs across multiple hosts, revisit — and note the cardinality rule then: bucket by `model_tier`/`routing`/`source`/`status`, never by `person_id`/`session_id`/`trace_id`.

## Acceptance criteria

- [ ] Rollup tables populate nightly and are backfilled from existing history.
[...]

Issue #611, opened 2026-08-21
feat(chat): a turn in flight should finish server-side when the client goes away


nbramia:
## Problem

Closing the chat page or app mid-response abandons the turn. The work stops where it stopped, and the answer the user was waiting for never exists.

That is the wrong contract for the way this assistant is used. Turns here are not cheap round trips — a tool-using turn can run over a minute, and the mobile case makes interruption routine rather than exceptional: a phone locks, an app is backgrounded, a network switches from wifi to cellular. Each of those currently discards work the user asked for and, on a paid backend, work they have already been billed for.

The half-measure is already in place, which makes the gap sharper. #592 persists whatever assistant content arrived before the stream ended, so a partial answer survives as a message. But *generation* stops at disconnect, so what persists is a truncated reply with nothing to indicate it was cut off — arguably worse than an obvious failure, because it looks like a complete answer that trails away.

## Solution

Decouple completing a turn from streaming it. The turn should run to completion server-side once accepted, with the stream being a live view of it rather than its lifeline, and the completed result readable afterward from the conversation the client reopens.

Design questions to settle rather than assume, and worth recording in the issue before implementing:

- **Which surfaces.** The report says "with any chat interface" — the native path and the Hermes proxy have different shapes (the proxy is a relay whose upstream is another service), and voice adds a third. Scope explicitly.
- **Cancellation must remain possible.** Today closing the page is the cancel gesture. If disconnect no longer cancels, there must be an explicit way to stop a turn — otherwise a runaway tool loop becomes unstoppable, which on a paid backend costs money.
- **Interaction with #592's tee.** Persistence currently finalizes when the relay ends. If the relay outlives the client, finalize must fire on the real end of the turn, exactly once, without double-writing when a client also disconnects.
- **Bounded lifetime.** A turn that survives its client needs a ceiling, or an abandoned turn can run until its upstream timeout with nobody waiting.
- **Reconnection.** Whether a returning client resumes the live stream or simply reads the finished message changes the work substantially; the second is much cheaper and may be sufficient.

## Acceptance Criteria

- [ ] Given a turn in flight, when the client disconnects, then generation shall continue to completion server-side.
- [ ] Given such a turn, when the user reopens the conversation, then the complete assistant reply shall be present.
- [ ] A turn that is genuinely truncated shall be distinguishable from one that completed — no truncated reply shall be presented as if whole.
- [ ] An explicit cancellation path shall exist and shall stop a turn that no client is watching.
[...]

PR #932, opened 2026-09-04
docs: describe current behavior only in comments, docstrings, and specs

nbramia:
## TL;DR

Removes every trace of change-narration (review rounds, findings, reviewers, "used
to"/"now"/"previously", issue/PR numbers cited as history) that the CRM performance
milestone (merged as `116b821`) added to comments, docstrings, tests, and docs, so the
repository describes only current behavior. Adds a repo-wide "current behavior only"
rule (`docs/AGENTS.md`, `docs/specs/standards/python-conventions.md`,
`docs/specs/technical/frontend.md`, `AGENTS.md`, `CLAUDE.md`) plus a fast, ratcheted
regression test (`tests/test_no_change_narration.py` + `tests/narration_baseline.json`)
so this can't silently recur. No behavior changes anywhere — proven mechanically per
file, since only comments, docstrings, and prose were ever touched.

**Updated after an adversarial-review BLOCK** (round 2): the guard's own `#NNN`-as-
history regex clause was unreachable (nested inside a `\b(...)` group it couldn't match
from), so an entire category of milestone-added issue-number citations was invisible to
the first sweep. That's fixed, along with three real bypasses, an over-broad word list,
and a count-keyed (rather than line-keyed) ratchet. Details below.

## Files changed (~70)

- **Rule text:** `AGENTS.md`, `CLAUDE.md`, `docs/AGENTS.md`,
  `docs/specs/standards/python-conventions.md`, `docs/specs/technical/frontend.md`
- **Guard:** `tests/test_no_change_narration.py` (new), `tests/narration_baseline.json`
  (new, line-keyed)
- **Source:** `api/main.py`, `api/routes/crm.py`, `api/routes/perf.py`,
  `api/routes/photos.py`, `api/services/aggregate_cache.py`,
  `api/services/interaction_store.py`, `api/services/log_redaction.py`,
  `api/services/person_entity.py`, `api/services/person_facts.py`,
  `api/services/relationship.py`, `api/services/route_timing.py`,
  `api/services/source_entity.py`, `api/services/tone_analysis_store.py`,
  `config/settings.py`, `scripts/cleanup_orphaned_records.py`,
  `scripts/merge_people.py`, `web/crm.html` (JS/HTML comments only)
- **Docs:** `docs/guides/configuration.md`, `docs/guides/operations.md`,
  `docs/specs/product/api-communications.md`, `docs/specs/product/api-crm.md`,
  `docs/specs/product/api-reference.md`, `docs/specs/product/crm-analytics.md`,
  `docs/specs/product/crm-graph.md`, `docs/specs/product/crm-interactions.md`,
  `docs/specs/product/crm-people.md`, `docs/specs/technical/architecture.md`,
  `docs/specs/technical/client-surfaces.md`, `docs/specs/technical/observability.md`
- **Tests:** 34 files under `tests/` (see `git diff --stat` for the full list)

## The guard: what it does and how round 2 fixed it

`tests/test_no_change_narration.py` scans comments, docstrings, and any bare
string-expression statement — never code that runs — across `docs/` (excluding
`docs/adr`, `docs/archive`, `docs/plans`), `AGENTS.md`, `CLAUDE.md`, `README.md`, every
`.py` file under `api/`, `scripts/`, `tests/`, `mcp_server.py`, the JS/CSS/HTML comments
in `web/*.html`, and every `.js` file under `web/` (including `web/chat/`,
`web/agents/`). It's a ratchet, not a zero-tolerance gate: `narration_baseline.json`
maps each file to the **set** of whitespace-normalized offending-line texts it already
carries, and a line whose normalized text isn't in that set fails the test. Deleting an
offending line never requires a baseline edit.

**Round 2 fixes**, each verified against the reviewer's exact reproduction:

1. **BLOCKER — unreachable `#NNN` regex.** The issue-number clause is now its own
   top-level alternative, `(?<![\w#])#\d{3,5}\b`, outside the `\b(...)` group that made
   it unreachable. `PATTERN.search('fixes #123')`, `'# see #611 for context'` now match;
   `'x#123'` and a `#fff` CSS colour still don't.
2. **MAJOR — over-broad vocabulary.** Removed bare `orchestrator`, `coordinator`,
   `regression`, `reviewer`, `bugfix`, `belt-and` (ordinary domain/testing vocabulary).
   Kept `review(er)? (round|finding)`, `round [0-9]`, `finding [0-9]`.
3. **MAJOR — count-keyed ratchet.** Rebuilt as line-keyed (see above). Verified: deleting
   one already-baselined hit in `api/routes/crm.py` while adding a brand-new unrelated
   narrating comment (same net count) still fails on the new line.
4. **MAJOR — three demonstrated bypasses, all closed:**
   - Every `Expr(Constant(str))` node is now scanned, not just a body's first statement
     — closes the bare mid-function "comment string" bypass (verified: a
     `"""previously..."""` string inserted mid-`__init__`, not a docstring, is now
     caught).
   - The allowlist now masks only its own matched span before `PATTERN` runs, never the
     whole line (verified: `# previously this was slower; timing via time.time()` is
     now caught).
   - `<!-- -->` HTML comments are now scanned everywhere in an HTML file (verified: an
     injected `<!-- previously ... -->` in `web/crm.html` is now caught).
5. **MAJOR — chat/agents SPA uncovered.** The guard now globs every `.js` file under
   `web/` recursively (15 files, including `web/chat/*.js` and `web/agents/*.js`), not
   just top-level `*.html`.
6. **MINOR — guard's own self-violation.** `_js_comments`'s docstring no longer says
   "in this cleanup"; needs no baseline entry.
7. **MINOR — duplicated rule text.** `docs/specs/technical/frontend.md` § Comments now
   links to `python-conventions.md#comments-and-docstrings` instead of restating the
   rule; both docs' Related Documents sections link each other.
8. **MINOR — over-applications in observability.md.** Restored "regressions" (no longer
   pattern-banned); kept the `#733` forward-pointer removed (open work belongs in
   issues, not specs, per `docs/AGENTS.md`).

After the fix, the corrected regex was re-swept across the full contents of every one
of the 80 milestone-touched files (`git diff --name-only 87bc8dc 116b821`), cross-
referenced against the pre-milestone tree (`git archive 87bc8dc`) to separate
genuinely-new citations from pre-existing/relocated ones, and every genuine hit was
cleaned. **Zero genuinely-new `#NNN` citations remain anywhere in the milestone's
files** (verified programmatically, not just by eye).

## Mechanical no-behavior-change proof

- **Every touched Python file**: `ast.dump(ast.parse(...))`, with the leading docstring
  `Expr(Constant(str))` node stripped from every module/class/function body before
  comparing (docstring *content* is free to change — it's invisible to the diff —
  while any real code edit would show up), is byte-identical before vs after. All
  touched `.py` files print `OK` (the new guard file itself has no "before" to compare
  against since it didn't exist before this PR, but is syntax- and lint-clean).
- **`web/crm.html`**: a hand-written character-scanner (tracks single/double-quoted
  strings, template literals with `${}` interpolation, regex-vs-division, and now
  `<!-- -->` comments too) strips comments from both the pre-edit snapshot and the
  current file. The only diff between the two stripped outputs is blank lines left
  where whole comment-only lines were deleted (inert in JS). `node --check` passes on
  the raw concatenated `<script>` content both before and after, at every stage of both
  rounds of edits.
- Docs and rule-text changes are prose-only, no executable content.

## Residual hits after cleanup (full re-sweep with the corrected pattern)

Every remaining hit is one of:
- **Pre-existing** — present verbatim before the milestone (confirmed against
  `87bc8dc`), out of scope.
- **Relocated** — docs content a milestone file split moved without rewriting (e.g.
  `api-communications.md` from the old `api-reference.md`), left untouched per the
  docs-specific scope rule.
- **A genuine false positive** — idiomatic "used to" ("used in order to"), "before/after
  this call" as position-in-flow, "no longer" describing present runtime/domain state,
  a fictional illustrative ID (`api/routes/crm.py`'s docstring example "Message
  #12345"), or a real-code assert-message/fixture string that isn't a comment/docstring
  (touching it would break the AST-invariance proof).
- **This PR's own rule text**, which legitimately discusses the banned phrases as prose
  examples (`AGENTS.md` principle 8, `python-conventions.md`'s BAD/GOOD block) — these
  get baseline entries like any other file; they aren't narration about a change.

<details>
<summary>Full per-line residual table</summary>

**api/routes/crm.py**
- L1583/1597/1607 "were moved"/"no longer have source entities" — phone/email records
  moved during a person merge (domain state at request time), not code history.
- L2670/2915 "used to rank"/"used to select" — "used [in order] to", not narration.
- L4603 "before this point" — position within a chronological list, not "before this
  change".
- L5679 "used to serialize" — same purpose-phrase false positive.
- L1256 `"Message #12345 is linked to Person A"` — a fictional illustrative message ID
  in a docstring example, not a real issue citation.

**api/services/aggregate_cache.py**
- L274/285/392 "used to keep.../before this [call/store] runs" — current control flow,
  not history.

**api/services/interaction_store.py**
- L1335 "Used to size..." — idiom, not history.
- L1810 "...vault notes with improved date extraction logic" — pre-existing.

**api/services/person_entity.py**
- L192/215 "was added"/"already exists" — boolean return semantics, not history.
- L913 "no longer needed" — pre-existing.

**api/main.py, config/settings.py, api/services/{entity_resolver,route_timing,
source_entity}.py** — every `#NNN` citation the corrected regex finds in these files
(dozens) was individually checked against the pre-milestone tree; the ones that were
milestone-added were fixed (see file list above), the rest are pre-existing.

**docs/specs/product/api-communications.md, api-reference.md, crm-analytics.md,
crm-graph.md, api-crm.md, crm-people.md, client-surfaces.md, architecture.md,
docs/guides/{configuration,operations}.md** — genuinely new milestone-added `#NNN`
citations were fixed; everything else remaining is pre-existing/relocated text,
confirmed against `git archive 87bc8dc`.

**tests/** (remaining hits are either real code — not a comment/docstring, so touching
it would fail the AST-invariance proof — or pre-existing)
- test_interaction_store.py:1331, test_crm_api.py, test_entity_resolver.py,
  test_log_redaction.py, test_mcp_server.py — all remaining citations confirmed
  pre-existing against `87bc8dc`.
- test_mcp_server.py:1181,1186, test_person_category_response.py:26-27,
  test_relationship_page_partner_config_browser.py:296,
  test_route_handlers_no_shared_connections.py:158 — assert-message strings or fixture
  data (real code), not narration prose.

**web/crm.html**
- L8082 "How you've improved" — UI copy text, not a comment.
- L9322,9364,18021,18064 — present-tense invariants about current DOM/runtime state.
- L9536,9956 "after this" — temporal reference within the same function, not history.
- L12217,18334 "now that we know.../now that the DOM is rendered" — present-tense
  conditionals.
- L12654 "previously-computed score" — a domain term, not code history.
- L16120,16482 "Previously, filtering was client-side..." — pre-existing.
- L17227 "belt-and-suspenders" — pre-existing.
- L18202 "Double-click is used to navigate" — idiom.
- L19039 "was cached before this edit" — a user note-editing action, not a code change.
- L19091 "Check if we were previously in..." — runtime filter-state comparison, not
  code history.

</details>

### Automated tests

Guard test alone:
```
$ pytest tests/test_no_change_narration.py -m "unit and not slow" -q
1 passed in ~3s
```
Four required static guards:
```
$ pytest tests/test_route_handlers_{sync,no_shared_connections,mutation_lock,concurrency}.py \
  -m "unit and not slow" -n 4 --dist loadscope -q
17 passed
```
All touched test files (multiple batches, `--dist loadscope` per repo convention):
```
270 passed, 3 skipped / 226 passed / 98 passed
```
Full unit suite, under the shared push lock, exactly as the pre-push hook runs it:
```
$ flock -w 7200 /tmp/lifeos-push.lock ./scripts/test.sh
6390 passed, 9 skipped, 0 failed in 222.22s
```
Server-free browser scope, under the same lock, exactly as the pre-push hook runs it:
```
$ flock -w 7200 /tmp/lifeos-push.lock pytest -m "browser and not requires_server" tests/ -q
358 passed, 0 failed
```

### Manual verification

N/A — comments, docstrings, and docs only; no runtime surface to click through. The
mechanical proofs above (AST-invariance, JS-comment-stripped equivalence, `node
--check`) are the applicable verification for this kind of change.

## Review focus

- Whether any rewritten comment/docstring lost an invariant a maintainer actually
  needs, versus losing only narration.
- Whether `tests/narration_baseline.json`'s per-file line sets are honest — spot check
  a few against the actual current file content.
- Whether the corrected `#NNN` regex and the closed bypasses hold up against further
  adversarial probing.


nbramia:
## Adversarial review of #932 by Opus

**Verdict:** BLOCK

The no-behavior-change claim holds — I proved it independently and it is airtight. The
cleanup claim does not: the guard's issue/PR-number rule is a dead regex branch, so an
entire prohibited category was never swept, and milestone-added review-finding citations
are still in files this PR edited.

1. **[BLOCKER] The `#NNN` alternative in the guard's regex is unreachable, so the PR's
   headline claim is false and the ratchet can never catch a recurrence** —
   `tests/test_no_change_narration.py:41-49` — the whole alternation is wrapped in
   `\b( ... )`, and `#[0-9]{3}\b` begins with a non-word character, so it only matches
   when `#` is *immediately preceded by a word character*. Reproduce:
   `PATTERN.search('fixes #123')` → `None`; `PATTERN.search('# see #611 for context')` →
   `None`; `PATTERN.search('x#123')` → `#123`. Separately `{3}` is exact, so `#1234`
   would not match even with the `\b` fixed — LifeOS is at #932, so the branch expires
   within weeks regardless.

   Consequence — milestone-added citations still present at `ac5414e`, in files this PR
   edited:
   - `web/crm.html:12765` — `// (#899 review finding N2).` A literal review-finding
     citation. The *adjacent* block at `12651` had the identical phrase removed, because
     that block also contained "review finding 3" and `finding [0-9]` does match. The
     12765 block contained only "finding N2", so nothing flagged it. This is direct
     evidence the sweep was regex-driven rather than read.
   - `web/crm.html:8713 (#874)`, `8726 (#876)`, `9407 (#876)`, `10388 (#874)`,
     `12113 (#876)`, `12202 (#891)`, `12447 (#873)`.
   - `tests/test_route_timing.py:2` `(#877)` and `:33` `(#877 review)` — in a file where
     the PR removed five sibling `#877 review, finding N` citations.
   - `tests/test_log_redaction.py:107` — `"""#904: ... That's how the CRM people list's
     raw search text (?q=<text>) reached logs/server.log even after the handler's own log
     line stopped naming it."""` — issue number plus explicit before/after narration, in
     a file this PR edited.
   - `tests/test_me_family_aggregates_oracle.py:55,62,384,391,939,956` — "pre-#871",
     "#871's family/timeline acceptance criterion", "cache (#917)" — file edited by this PR.
   - `tests/test_me_page.py:21,76,80,86,209,211` — "#871 replaced the 'load every
     PersonEntity and sum in Python' approach", "#897's review follow-up moved the ..." —
     milestone-touched, untouched here.

   Repo-wide, 3081 comment/docstring/doc lines across 372 files cite `#NNN` and are
   invisible to the guard. Fix: hoist the `\b` into the word-initial alternatives only,
   widen to `#\d{2,6}`, then re-clean the files above rather than baselining them.

2. **[MAJOR] The pattern blacklists ordinary domain and testing vocabulary, so it fails
   correct present-tense writing** — `tests/test_no_change_narration.py:48-49` —
   `orchestrator`, `coordinator`, `regression`, `reviewer`, `bugfix`, `belt-and` are
   narration-free terms. Reproduce: append to `api/services/relationship.py` (baseline 0)
   `# The orchestrator calls this service once per turn.` and
   `# Regression coverage for empty inputs lives in tests/test_relationship.py.` →
   `2 hits, 0 allowed`, gate red. 339 of the 1111 baselined hits (31%) are caused *only*
   by these six terms (`orchestrator` 146, `regression` 140); `AGENTS.md` itself carries a
   baseline of 3 for this reason even though "Orchestrator LLM" is a documented Key
   Concept. The only escapes are the two the test's own docstring forbids — raise the
   baseline, or extend `ALLOWLIST`. Fix: drop those six terms; keep the pattern to
   narration forms.

3. **[MAJOR] The ratchet is count-keyed, not line-keyed, so a file at its baseline can
   swap old hits for brand-new narration and stay green** —
   `tests/test_no_change_narration.py:229-236` — `count > allowed` only. Delete one
   pre-existing hit in `web/crm.html` (15) or `api/routes/crm.py` (7) and you may add a
   fresh narrating comment for free. Fix: key the baseline on `(path, normalized line
   text)` so removing an old hit does not buy a new one.

4. **[MAJOR] Three demonstrated bypasses of the scan** (each probed and reverted):
   - **Multi-line string used as a comment.** A bare string expression that is not a
     body's first statement is neither a docstring (only `body[0]` is collected) nor a
     `#` comment (its `#`s fall inside a Constant span and are skipped). Probe body
     `y = 2` then `"""previously this used to be slower and the old implementation was
     worse."""` → not detected.
   - **`ALLOWLIST` suppresses the entire line.**
     `tests/test_no_change_narration.py:56` — probe `# previously this was slower; timing
     via time.time()` → not detected. Worse, the allowlist buys nothing in exchange: none
     of `datetime\.now\(`, `\bnow\s*=`, `time\.time\(\)` can itself trip `PATTERN` (the
     `now` branch requires a following keyword), so all it does is open holes. Delete it,
     or make each entry a `(trigger, context)` pair that suppresses one match rather than
     a whole line.
   - **HTML comments are never scanned.** `_html_lines_to_check` walks only `<script>`
     and `<style>`. Injecting
     `` into `web/crm.html` left the count at 15.

   Correctly caught, for the record: a trailing comment on a code line, and a docstring
   in a nested function.

5. **[MAJOR] The guard does not cover the chat/agents SPA at all** —
   `tests/test_no_change_narration.py:200` — `(REPO / "web").glob("*.html")` is top-level
   HTML only and matches no `.js`. `web/chat/*.js` (11 files) and `web/agents/*.js` (4)
   are unscanned. `docs/specs/technical/frontend.md`'s new § Comments scopes the rule to
   "`web/*.html`", so the written standard excludes them too — the frontend rule covers
   one file and the guard covers one directory level.

6. **[MINOR] The PR's own new file violates the rule it introduces, and baselines it
   rather than fixing it** — `tests/test_no_change_narration.py:111-112` — `_js_comments`'s
   docstring reads "mirrors the tokenizer used to verify web/crm.html edits **in this
   cleanup**". "in this cleanup" is change narration about this PR's own process.
   `narration_baseline.json` grants this file an allowance of 1 instead. (The baselined
   hit is actually the "used to verify" idiom on line 111; the real narration on 112 is
   not matched at all.) Rewrite the docstring and delete the entry.

7. **[MINOR] The new frontend rule duplicates rather than links, against the
   single-source-of-truth principle this PR reaffirms** —
   `docs/specs/technical/frontend.md:265-281` restates the whole prohibition instead of
   pointing at `docs/specs/standards/python-conventions.md` § Comments and Docstrings,
   which `docs/AGENTS.md` principle 7 (unchanged, and re-emphasised by this PR's new
   principle 8) asks it not to do. It also names "the Python conventions" in prose while
   neither doc's Related Documents section links the other, which docs/AGENTS.md requires
   to be bidirectional. Line counts are fine (python-conventions 271 / Standards target
   200-500; frontend 325 / Technical target 400-800), and both `Last Updated` fields are
   bumped.

8. **[MINOR] Two over-applications in `docs/specs/technical/observability.md`** — "catches
   **regressions** perf tracing can't see" → "catches **slowdowns**" (L73), and
   "(#733 covers persisted rollups this can feed)" dropped (L83). Neither was history:
   "regression" is the standard present-tense name for what the signal detects (and is only
   in the pattern per finding 2), and #733 is a forward pointer to tracked work, not a
   timeline. Both edits look driven by the regex rather than by the rule.

### What I verified and how

- `ast.dump(ast.parse(src))` with every module/class/function leading docstring stripped,
  `origin/main` (116b821) vs PR head (ac5414e), all 36 modified `.py` files → **0 differ**.
  Non-leading string expressions and assert-message strings stay in the AST, so a
  changed one would have shown. **No Python behavior change.**
- `web/crm.html`: hand-written scanner stripping `//`, `/* */`, `<!-- -->`, tracking
  single/double quotes, template literals with `${}` nesting, and regex-vs-division;
  applied to both revisions. `diff -B` on the stripped outputs → **empty**; whitespace-free
  compare → **byte-identical**. Stripper sanity: `https://` in strings, `replace(/.../)`
  regex literals, and 190 `border-radius` CSS lines all survive; zero `// ` residue.
  **No frontend behavior change.**
- `git diff --name-status 116b821 ac5414e` → 48 files, all comments/docstrings/prose plus
  two new test artifacts; nothing outside that class.
- `pytest tests/test_no_change_narration.py -m "unit and not slow" -q` → 1 passed in 3.02s
  (in the `unit and not slow` scope, fast).
- Mutation: `sed -i '1i # previously this used to be slower' api/routes/crm.py` →
  `FAILED — api/routes/crm.py: 8 hits, 7 allowed`, offending line printed first. Reverted.
- Mutation: two blameless comments in `api/services/relationship.py` (baseline 0) →
  `FAILED — 2 hits, 0 allowed` (finding 2). Reverted.
- Bypass probes in `api/services/tone_analysis_store.py` (trailing comment / nested
  docstring / bare multi-line string / `#1234` / `time.time()` line) and an HTML comment
  in `web/crm.html` → results in finding 4. All reverted; `git status` clean.
- Baseline honesty: scanned 87bc8dc, 116b821, and PR head with the PR's own extractor.
  Milestone added 326 hits (1086 → 1412); PR removes 301, leaving 1111. **Baseline equals
  the current count for all 298 files exactly — no padding.** 34 baselined hits post-date
  87bc8dc, of which 9 are the rule text's own deliberate BAD examples.
- Spot-checked 5 `api-communications.md` entries against `git show
  87bc8dc:docs/specs/product/api-reference.md`: `model_override` L70→39, `client_turn_id`
  L72→41, "A disconnect no longer stops the turn (#611)" L93→62, `search_memories`
  L123→92, `claude_intent` L199→168 — **all verbatim pre-existing text relocated by the
  milestone's file split. The PR's "relocated, not authored" claim is honest.**
- Independent full-file sweep with a wider pattern over all 80 milestone-touched files;
  everything it surfaced beyond finding 1 is pre-existing repo narration outside this
  PR's scope (`api/main.py:87/157`, `config/settings.py:243`, `docs/guides/operations.md:28`).
- Rewrite quality: read the rewritten blocks in `api/services/aggregate_cache.py` (module
  docstring, `_default_crm_db_paths`, `_open_existing_db`, `cached()`),
  `api/routes/crm.py` (`_mutation_lock`, `_me_exclude_ids`, `_rendered_edge_weight`,
  `get_network_graph` selection), and `web/crm.html` (`loadPeople` config, `navigateTo`,
  `dispatchRoute`, relationship/family dashboard hiding). **These are good** — the
  invariant and the reason survive in every case; only the citation is gone. The
  numbered "#917 review findings 1..11" list becoming an "Invariants" bullet list is a
  genuine improvement. One cosmetic reflow ragged edge at
  `api/services/aggregate_cache.py:127-128`. No truncation-into-unclarity found.
- Tests, all with `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES=""`,
  `-n 4 --dist loadscope`, `data` symlinked to the staging copy:
  - four static guards (`test_route_handlers_{sync,no_shared_connections,mutation_lock,concurrency}.py`)
    `-m "unit and not slow"` → **17 passed**
  - `tests/test_crm_api.py` → **24 passed, 35 skipped**
  - all 29 PR-touched test files, `-m "not slow"`, two batches → **240 passed / 35 skipped**
    and **176 passed / 3 skipped**

### Not verified

- Full unit suite and the server-free browser scope. Targeted runs only per my brief; the
  PR reports both green and the AST/HTML proofs make a behavioral break impossible.
- `tests/test_crm_api.py`'s 35 skips: the staging copy at
  `/home/nathanramia/orca/workspaces/LifeOS/performance/data` currently holds a 176 KB
  `crm.db` with no people ("No people in database to test" / "large CRM database not
  present"). Not attributable to this PR — the code is AST-identical to main, so the
  same tests skip on either side.
- No browser run: `web/crm.html` is proven byte-identical after comment stripping, so
  there is nothing for a browser to observe.
- I did not re-audit the ~1086 pre-existing baselined hits for whether each is a true
  false positive; I verified the baseline is arithmetically honest and spot-checked the
  files this PR touched.

nbramia:
Addressed all 8 findings. Summary below; full details and updated evidence are in the PR body.

1. **[BLOCKER] Unreachable `#NNN` regex.** Fixed exactly as suggested: the issue-number
   clause is now its own top-level alternative, `(?<![\w#])#\d{3,5}\b`, outside the
   `\b(...)` group. Verified against all four repro cases (`fixes #123`, `# see #611 for
   context`, `x#123`, `#fff` colour). Re-swept the full contents of every one of the 80
   milestone-touched files (`git diff --name-only 87bc8dc 116b821`) with the corrected
   pattern, cross-referenced each new hit against the milestone diff to confirm it was
   actually added by the milestone (not pre-existing), and cleaned every one that was —
   web/crm.html plus ~28 more files the widened regex surfaced beyond the five you named
   (api/main.py, config/settings.py, several api/services/*.py, several docs, and several
   more test files). Residual table with justifications is in the PR body.

2. **[MAJOR] Overbroad vocabulary.** Removed `orchestrator`, `coordinator`, bare
   `regression`, bare `reviewer`, `bugfix`, `belt-and` from the pattern. Kept
   `review(er)? (round|finding)`, `round [0-9]`, `finding [0-9]`. Your two mutation probes
   (`# The orchestrator calls this service...` / `# Regression coverage for empty
   inputs...`) now pass against `api/services/relationship.py`'s baseline of 0.

3. **[MAJOR] Count-keyed ratchet.** Rebuilt as line-keyed: `narration_baseline.json` now
   maps each file to the set of whitespace-normalized offending-line texts, and the test
   fails on any current line whose normalized text isn't in that set. Deleting a line
   never requires a baseline edit (the set is a ceiling); there's no way to swap one
   allowance for a different new violation, since every current hit is checked
   independently. Documented how to tighten it (delete now-unused entries) in the test's
   docstring.

4. **[MAJOR] Three bypasses.**
   - Multi-line/bare string "comments": now every `Expr(Constant(str))` node is scanned,
     not just a body's first statement. Your probe (`y = 2` then a bare narrating
     triple-quoted string) is now caught.
   - Allowlist scope: now masks only its own matched span (replaced with spaces) before
     `PATTERN` runs, never the whole line. Your probe (`# previously this was slower;
     timing via time.time()`) is now caught.
   - HTML comments: `<!-- -->` is now scanned everywhere in an HTML file, independent of
     `<script>`/`<style>`. Your probe is now caught.

5. **[MAJOR] Chat/agents SPA uncovered.** The guard now globs every `.js` file under
   `web/` recursively (`web/chat/*.js`, `web/agents/*.js` included), not just top-level
   `*.html`. `docs/specs/technical/frontend.md`'s § Comments now says "all of `web/`"
   explicitly and points at the Python conventions doc for the rule text (see finding 7).

6. **[MINOR] Self-violation.** Rewrote `_js_comments`'s docstring to drop "in this
   cleanup"; it no longer needs (and no longer has) a baseline entry.

7. **[MINOR] Duplicated rule text.** `frontend.md` § Comments now links to
   `python-conventions.md#comments-and-docstrings` instead of restating the prohibition,
   and both docs' Related Documents sections link each other.

8. **[MINOR] Over-applications in observability.md.** Restored "regressions" (now that
   the word isn't pattern-banned, there's no reason to avoid it); the `#733` forward
   pointer stays removed per docs/AGENTS.md (open work belongs in issues, not specs).

Full residual table, updated AST/JS proofs, and test evidence are in the refreshed PR
body. New head SHA and gate line follow once pushed.

nbramia:
Pushed. New head SHA: `a3b6ca4` (was `ac5414e`).

Gate line (pre-push hook, both phases, this push):
```
Running unit tests... passed (========== 6390 passed, 9 skipped, 806 warnings in 178.36s (0:02:58) ===========)
Running server-free browser tests... passed (=============== 358 passed, 6786 deselected in 275.24s (0:04:35) ===============)
```

PR body updated to reflect round 2's fixes, the full residual table, and the updated test evidence.

nbramia:
## Verification pass on #932 @ `a3b6ca4` by Opus

All eight findings are genuinely fixed, not papered over. I re-derived every claim rather
than reading the implementer's comment: the regex repair is correct, the residual
survivors are all accounted for, the ratchet is really line-keyed, all three bypasses are
closed, `.js` is covered, and no behavior changed anywhere in the enlarged file set.

### Finding-by-finding

1. **BLOCKER — dead `#NNN` branch — FIXED.** `(?<![\w#])#\d{3,5}\b` is now its own
   top-level alternative outside the `\b( ... )` group. My four repro cases:
   `'fixes #123'` → `#123`, `'# see #611 for context'` → `#611`, `'(#595)'` → `#595`,
   `'// (#899 review finding N2).'` → `#899`. 4-and-5-digit numbers now match too
   (`#1234`, `#12345`); `x#123` and `url#900` correctly don't.

   Every survivor I named is gone. `web/crm.html` no longer carries `#874`, `#876`,
   `#891`, `#873`, or the `(#899 review finding N2)` line; `tests/test_route_timing.py`
   and `tests/test_me_family_aggregates_oracle.py` have dropped out of the baseline
   entirely. `tests/test_me_page.py` is down to two `#682` hits and
   `tests/test_log_redaction.py` to one `#519` — both pre-milestone, correctly baselined.

   I re-swept all 80 milestone-touched files with the corrected extractor, diffing each
   file's current hit set against its own text at `87bc8dc`. **39 survivors, zero of them
   genuinely new:**
   - 5 are the new rule text's own deliberate BAD examples (`AGENTS.md:415`,
     `python-conventions.md:215/216/223/224`) — unavoidable, a rule must quote what it bans.
   - 18 are idiomatic/positional false positives already judged last round ("used to
     rank/select/size", "before this store runs", "no longer owns", "previously-computed
     score", `crm-graph.md:135` "was dropped by the filters").
   - 16 are pre-existing text relocated by the milestone's `api-reference.md` split. I
     verified each against the pre-milestone pool programmatically: **13 hit verbatim**;
     the other 3 (`api-communications.md:40`, `api-reference.md:169`,
     `client-surfaces.md:339`) I checked by hand and they differ only in a cross-file link
     anchor rewritten by the split (old `api-reference.md:71/580`, `client-surfaces.md:337`).

2. **MAJOR — vocabulary blacklist — FIXED.** `orchestrator`, `coordinator`, `regression`,
   `bugfix`, `belt-and` are gone; bare `reviewer` is narrowed to
   `review(er)? (round|finding)`. Both blameless `relationship.py` comments from my repro
   now pass, as does `# The coordinator dispatches to a reviewer for a bugfix;
   belt-and-suspenders.` `AGENTS.md` drops 3 → 2 baselined lines and
   `docs/specs/technical/frontend.md` to 0.

3. **MAJOR — count-keyed ratchet — FIXED.** Baseline is now `{path: [normalized line
   texts]}`. Probed on `web/crm.html`: replacing the baselined
   `// was cached before this edit.` with a neutral comment → **passes** (removal needs no
   baseline edit); replacing it with a *different* narrating comment
   (`// this used to recompute on every keystroke.`) → **fails**, naming the line. No
   allowance is freed by deleting an old hit.

4. **MAJOR — three bypasses — ALL CLOSED.** Re-probed each:
   - Bare mid-function string: `_py_lines_to_check` now walks every
     `Expr(Constant(str))`, not just `body[0]` → caught at `tone_analysis_store.py:209`.
   - Whole-line allowlist: `ALLOWLIST.sub` masks only its own matched span, so
     `# previously this was slower; timing via time.time()` → caught at `:215`.
   - HTML comments: `_html_comment_lines()` added → my injected `` caught at `web/crm.html:7369`.
   - Bonus: `# see #1234 and #12345 for the old behavior` → caught, three separate hits.

5. **MAJOR — `.js` coverage — FIXED.** `_gather_targets` uses `(REPO/"web").rglob("*.js")`
   and `_scan_file` dispatches `.js` to `_js_comments`. Probed: a `//` comment in
   `web/chat/main.js` and a `/* */` comment in `web/agents/board.js` are both caught. The
   baseline now carries these files (e.g. `web/chat/main.js` 9 lines,
   `web/agents/board.js` 18), so the SPA's existing debt is pinned rather than invisible.

6. **MINOR — guard's own self-violation — FIXED.** `_js_comments`'s docstring no longer
   says "in this cleanup"; `grep -n "in this cleanup"` returns nothing and
   `tests/test_no_change_narration.py` has **0** baseline entries.

7. **MINOR — duplicated rule text — FIXED.** `frontend.md` § Comments now links to
   `python-conventions.md#comments-and-docstrings` instead of restating the rule, both
   Related Documents sections cross-link, and the scope is widened correctly to "all of
   `web/`". `frontend.md` 325 → 312 lines.

8. **MINOR — observability over-application — FIXED.** "catches **regressions** perf
   tracing can't see" is restored. (The `#733` forward pointer stays dropped — fine;
   `docs/AGENTS.md` says specs aren't for task references.) Bonus cleanup:
   `python-conventions.md:96` lost its `(#868)` / "as of #868" narration, and
   `api/main.py`, `config/settings.py`, `api/services/route_timing.py`,
   `api/routes/perf.py`, `source_entity.py`, `person_facts.py`, `scripts/*` got the same
   treatment — all invariant-preserving.

### Nits (non-blocking)

- **N1.** The line-keyed baseline is a *set*, so an exact duplicate of an
  already-baselined line text added elsewhere in the same file passes. Probed: appending a
  second `// was cached before this edit.` to `web/crm.html` → passes. Realistic exploit is
  only copy-pasting existing narration; a multiset (or `{text: count}`) would close it.
- **N2.** Two mechanical-edit ragged edges: `api/services/route_timing.py:79` (`Handlers
  run in FastAPI's threadpool, so recording must be safe` — short line) and
  `api/main.py:356` (`# Route timing is added last, which Starlette's middleware stack
  makes / # the OUTERMOST ...` — the relative clause now dangles slightly). Cosmetic.
- **N3.** `_html_comment_lines()` scans raw source, so a JS string literal containing
  `<!-- ... -->` would be read as a comment. Over-detection only, never a bypass.

### What I verified and how

- `PATTERN` repro, 12 cases incl. all four originals → results above. `_line_has_violation`
  on the three blameless comments → all `False`. Allowlist span-mask: line with a real
  violation *plus* `time.time()` → `True`; pure-FP line → `False`.
- **No behavior change, enlarged set.** Docstring-stripped `ast.dump(ast.parse())`,
  `origin/main` (116b821) vs `a3b6ca4`, all **49** modified `.py` files → **0 differ**.
  (Strict form: only *leading* docstrings stripped, so an edited mid-function bare string
  would still have shown as a diff. None did.)
- `web/crm.html`: same string/template/regex-aware stripper as round 1, base vs `a3b6ca4`
  → `diff -B` **empty**, whitespace-free compare **byte-identical**; `https://` in strings,
  regex literals and 190 CSS lines all survive the strip.
- **Baseline honesty:** baseline **431 files / 4110 lines**; current hit set **431 files /
  4110 lines**; baseline entries matching no current hit: **0**; current hits with no
  baseline entry: **0**. Exact, no padding, no slack. 3248 of the 4110 are `#NNN`-only —
  the repo-wide citation debt the fixed regex can now see, pinned so it can only shrink. I
  checked for hex-colour false positives among them and found none.
- Ratchet probes (swap → fail, removal → pass, duplicate → pass) and bypass probes
  (bare string / allowlist / HTML comment / `web/chat/*.js` / `web/agents/*.js`) as
  described. All reverted; `git status` shows only the harness's `.worktree-info`.
- `pytest tests/test_no_change_narration.py -m "unit and not slow"` → **1 passed in 3.6s**
  (in scope, still fast despite scanning ~15 more `.js` files).
- Tests, CPU-only env, `-n 4 --dist loadscope`, `data` symlinked:
  - four static guards `-m "unit and not slow"` → **17 passed**
  - `tests/test_crm_api.py` → **24 passed, 35 skipped**
  - all 34 touched test files, `-m "not slow"`, two batches → **247 passed / 35 skipped /
    1 failed** and **274 passed / 3 skipped**
- The one failure, `tests/test_me_page.py::TestMyPersonIdConstant::test_my_person_id_is_valid_uuid`,
  is **not this PR's**: it is `@pytest.mark.integration` (so the `unit and not slow` gate
  never selects it), needs a configured `settings.my_person_id`, and my worktree has no
  `.env`. `git diff 116b821 a3b6ca4 -- tests/test_me_page.py` touches docstrings only, and
  I ran the same test at `116b821` → **identical failure**. Pre-existing and environmental.

### Not verified

- Full unit suite and the server-free browser scope — targeted runs only per my brief; the
  AST and HTML proofs make a behavioral break impossible.
- `test_crm_api.py`'s 35 skips: the staging `data/crm.db` currently holds no people. Same
  on either side of the diff.
- No browser run: `web/crm.html` is byte-identical after comment stripping.
- I did not re-audit all 4110 baselined lines individually; I verified the baseline is
  exact against the current scan and hand-judged every survivor in the 80 milestone files.

PR #939, opened 2026-09-05
fix(test): isolate test.sh's server handling and Playwright's parked event loop across concurrent and co-scheduled runs (#919, #916)

### LINKED ISSUE #916: test(agents): board-stream tests fail when co-scheduled with the agents browser files
## Problem

Two board-stream tests fail reproducibly when run in the same pytest invocation as the agents browser test files, and pass when run alone. The failure is cross-file test pollution, not a product defect, and it is invisible to the repo's own gate.

Failing when co-scheduled:

- `tests/test_agents_board_api.py::TestBoardStream::test_stream_emits_a_second_frame_after_a_task_mutation`
- `tests/test_agents_board_api.py::TestHermesLabelAndCodexStream::test_stream_codex_session_actually_reads_a_rollout_file`

Reproduced by running `tests/test_agents_board_api.py` together with `tests/test_agents_assignment_ui_browser.py`, `tests/test_agents_board_ui_browser.py` and `tests/test_agents_host_filter_ui_browser.py` under `-n 4`. Each file passes on its own.

**This is pre-existing, not caused by any in-flight branch.** It was reproduced identically on `feat/agents-board-polish` at `20df9ba` with no feature changes applied.

The reason nobody has hit it: `scripts/test.sh` and the pre-push hook run the unit suite and the `browser and not requires_server` suite as **separate** invocations, so the two groups are never co-scheduled in CI. Only an agent or developer running an ad-hoc mixed selection sees it — which makes it a recurring waste of investigation time, since the natural first assumption is that the branch under test broke something.

Both failures involve the board's SSE stream, which suggests shared module-level or filesystem state (a stream generator, a cached task manager, a watcher, or a temp vault path) leaking between the board-API tests and the browser tests rather than a genuine race in the stream itself — but the actual cause needs diagnosis, not a guess.

## Solution

Find the shared state and isolate it, so the two tests pass regardless of what else is in the same invocation.

- Diagnose which module-level or on-disk state is shared. `-p no:randomly`, `--dist loadfile` vs `loadscope`, and bisecting the co-scheduled file set will localise it.
- Fix the leak at its source — an autouse fixture that resets the singleton, a per-test temp path, or explicit teardown of the stream/watcher — rather than by reordering tests or marking them.
- Do not fix it by skipping, xfailing, or deleting either test.

## Acceptance Criteria

- [ ] Both named tests pass when `tests/test_agents_board_api.py` is run in the same `-n 4` invocation as the three agents browser files.
- [ ] Both still pass when run alone and under the repo's normal separate unit / browser invocations.
- [ ] The root cause is named in the fix's commit message or a code comment — which state was shared, and what now isolates it.
- [ ] No existing assertion is weakened, skipped, or removed to achieve this.

## Verification

- **Automated tests:** `pytest tests/test_agents_board_api.py tests/test_agents_assignment_ui_browser.py tests/test_agents_board_ui_browser.py tests/test_agents_host_filter_ui_browser.py -n 4` green, and each file green alone.
- **Manual checks:** none needed.
- **Existing test suite:** green — both `./scripts/test.sh` and the full `browser and not requires_server` set.

## Scope

**In scope:** test fixtures/teardown and any module-level state reset needed to isolate the board-stream tests; a conftest fixture if that is the right home for it.

**Out of scope:** the board stream's product behaviour, the `/agents` board UI, and the agents browser tests' own assertions.

## Technical Context

- **Key files:** `tests/test_agents_board_api.py` (the two failing classes), `tests/conftest.py`, and whatever the board stream holds at module scope in `api/routes/agents.py`.
- **Patterns to follow:** existing autouse reset fixtures in `tests/conftest.py` for other module-level singletons.
- **Constraints:** unit tests run under `pytest-xdist` with `-n auto --dist loadscope`; reproduce in that mode, not sequentially. Tests are sacred — the fix must be isolation, never suppression.

## References

- Surfaced by the general adversarial reviewer during #883's round-3 review (PR #901), and confirmed pre-existing on the base commit `20df9ba`, so it was deliberately not folded into that PR.
- Related: `AGENTS.md` § Testing (the separate unit/browser invocations that hide this).

### LINKED ISSUE #919: fix(scripts): give test.sh's server PID file a per-run path so concurrent worktrees cannot collide
## Problem

`scripts/test.sh` writes and reads a fixed PID file at `/tmp/lifeos_test_server.pid`, shared by every worktree, branch, and concurrent agent on the machine:

```
scripts/test.sh:229    if [ -f /tmp/lifeos_test_server.pid ]; then
scripts/test.sh:230        PID=$(cat /tmp/lifeos_test_server.pid)
scripts/test.sh:235        rm /tmp/lifeos_test_server.pid
```

This is the same class of defect as #908 (fixed in #913), in a different script. Several agent worktrees run `test.sh` concurrently on this box by design, so two overlapping runs that start a test server will collide on the file: the second run overwrites the first's PID, and whichever finishes first reads and `rm`s a PID that may belong to the other run — killing a server another run is still using, or leaking one because its PID was clobbered.

Unlike #908 this has not been observed causing a wrong result. It was found while auditing the rest of the tree for the same pattern during #913's review, and is filed so it is not lost.

## Solution

Key the PID file on something unique per run, the way #913 keyed the pre-push log path — a directory under `${TMPDIR:-/tmp}` plus the branch or PID — or hold the PID in a shell variable for the lifetime of the script if it does not genuinely need to survive across invocations (determine which before changing it; the file may exist specifically to clean up after a crashed run).

Whatever the shape, follow the pattern #913 settled on: never adopt, `chmod`, or delete a path the script does not own, and print the resolved path so a stuck server is traceable to its run.

## Acceptance Criteria

- [ ] Two concurrent `test.sh` runs that each start a test server do not share a PID file and cannot kill or leak each other's server.
- [ ] A crashed run's server is still cleaned up by a later run, or the reason that is no longer needed is stated in the script.
- [ ] The resolved path (if a file is still used) appears in the script's output.
- [ ] No stale PID files accumulate without bound.
- [ ] The scope of which tests `test.sh` runs, and its exit semantics, are unchanged.

## Verification

- **Automated tests:** extend `tests/test_test_sh_auto.py` (or a sibling) to assert two different branches/runs resolve to distinct paths, mirroring `tests/test_prepush_gate.py`'s approach.
- **Manual checks:** start two `test.sh` runs concurrently from two worktrees, confirm each manages only its own server process, and confirm no server is left running afterwards.

## Scope

**In scope:** the PID file's path and lifecycle in `scripts/test.sh`, and the message naming it.
**Out of scope:** which tests `test.sh` selects, its plan/`decide_plan` logic, its markers, and its exit codes.

## Technical Context

- **Key file:** `scripts/test.sh` — the hardcoded `/tmp/lifeos_test_server.pid`.
- **Precedent:** #908 / PR #913 did exactly this for `scripts/pre-push`. That PR's log-directory block is a working reference for the ownership rule and the fallback behaviour, and `tests/test_prepush_gate.py` is a reference for testing a shell hook's path resolution without running the real suite.
- **Constraints:** no new dependencies; the script must stay usable from any worktree and on a fresh clone.

## References

- #908 — the same defect in `scripts/pre-push`.
- PR #913 — the fix, and the review that surfaced this one.

nbramia:
## Summary

Two test-infrastructure isolation defects, same class (shared state leaking between concurrent or co-scheduled test runs), adjacent files:

- **#919** — `scripts/test.sh` read, killed, and `rm`'d a fixed, shared `/tmp/lifeos_test_server.pid` from a `stop_test_server()` helper. Nothing in the repo ever wrote that file and the helper was never called (unchanged since the initial commit `96e56a9`), so the fix is removal plus a comment stating why `test.sh` tracks no PID at all: `scripts/server.sh` owns the server lifecycle and identifies the process by port. Two guard tests keep a fixed shared path or a PID-tracking helper from coming back. **What this does not fix:** the server itself is still one shared resource on hardcoded port 8000, and `server.sh start` kills and replaces whatever is listening there, so two `test.sh` runs that both need a server still contend for it. That is `server.sh`'s contract and out of this PR's scope — filed as #944. The comment in `test.sh` and the guard's assertion message say so plainly.
- **#916** — the two `async def` tests in `tests/test_agents_board_api.py` fail when co-scheduled after the agents browser files. Root cause is an **event-loop leak, not board state**: pytest-playwright's session-scoped sync driver parks its asyncio loop inside a greenlet for the whole worker lifetime, leaving the thread's "running loop" marker set, so pytest-asyncio's runner raises `RuntimeError: Runner.run() cannot be called from a running event loop` for any later async test in that worker. `tests/conftest.py` re-scopes the five upstream fixtures (`playwright`, `browser_type`, `browser_context_args`, `launch_browser`, `browser`) from session to module scope, delegating to upstream's own bodies via `.__wrapped__`, so the loop is torn down at the end of each browser file. Two guard tests cover it: one fails if the local overrides are missing or not module-scoped, one fails if a pytest-playwright upgrade reshapes the upstream fixtures (parameters or generator-vs-return).

Docs: `AGENTS.md` § Testing gains the co-scheduling caveat (`--dist loadscope` groups by class for tests inside a class and by module for top-level functions and hands groups out largest-first, so a co-scheduling bug can appear or vanish with the file set — bisect in one process with no `-n`, browser file first) with a pointer to the conftest comment for the mechanism and reproducers; `docs/specs/standards/testing-standards.md` points at it. A stale `scripts/pre-push` comment describing the same mechanism ("silently never awaited") is corrected to the error actually raised. All comments/docs follow the "current behavior only" standard; `tests/test_no_change_narration.py` passes.

## #919 — Problem / Fix

**Problem:** several agent worktrees run `test.sh` concurrently on one machine; a fixed `/tmp` PID path is the same cross-worktree collision class `scripts/pre-push` had (#908 / PR #913).

**Fix:** verified independently that the path had no writer and the helper no caller (`grep -rn lifeos_test_server` and `grep -rn stop_test_server` across `*.sh *.py *.md`, hooks, systemd templates, `docs/` → only the helper's own three lines; `git log -S 'lifeos_test_server' --all` → only `96e56a9`). `server.sh` identifies the server with `get_server_pid()` (`lsof -ti :$PORT`) and its `start` kills/replaces whatever is on the port, so a crashed run's server is cleaned up by the next `start`, not by a PID file. Removed the dead helper; the comment above `start_server_background()` states this and states the remaining port contention. Which tests `test.sh` selects, `decide_plan`, markers, and exit codes are unchanged. A per-run PID path was deliberately NOT introduced: a file nothing writes needs no safer path.

## #916 — Problem / Root cause / Fix

**Reproducers** (the issue's own command never fails — see below):
```
# minimal, deterministic, no xdist
pytest tests/test_voice_mic_block_ui_browser.py tests/test_agents_board_api.py::TestBoardStream -q
# as it shows up in the real suite
pytest tests/test_agents_assignment_ui_browser.py tests/test_agents_board_ui_browser.py \
       tests/test_agents_host_filter_ui_browser.py tests/test_agents_board_api.py -n 4
```
The browser files must be listed **before** `test_agents_board_api.py`. `--dist loadscope`'s `_split_scope` is `nodeid.rsplit("::", 1)[0]`: class for tests in a class, module for top-level functions (27 scope groups across these four files — `test_agents_assignment_ui_browser.py` is one module group; the other three contribute 16, 4, and 6 class groups). Groups are handed to workers largest-first (pytest-xdist's `--loadscope-reorder` default), with command-line order only breaking ties; on a four-worker run that puts `TestBoardStream` on a worker after a parked driver loop with the browser files listed first and not with the board file first — the issue's board-file-first command never triggered the leak (~50 runs, `-n 1/2/4`, `--dist loadfile`, CPU-pinned via `taskset`, all green). Reversed, on the unfixed base: `2 failed, 163 passed` (deterministic in every reviewer's runs; 3 of 5 in the implementer's first batch), always the two named tests, always:
```
RuntimeError: Runner.run() cannot be called from a running event loop
  .../python3.12/asyncio/runners.py:93
FAILED tests/test_agents_board_api.py::TestBoardStream::test_stream_emits_a_second_frame_after_a_task_mutation
FAILED tests/test_agents_board_api.py::TestHermesLabelAndCodexStream::test_stream_codex_session_actually_reads_a_rollout_file
```

**Root cause:** `playwright/sync_api/_context_manager.py` enters `self._loop.run_until_complete(self._connection.run_as_sync())` once inside a greenlet and thereafter parks it via greenlet switches; the call returns only on `pw.stop()`, i.e. the session-scoped `playwright` fixture's teardown. asyncio clears its per-thread running-loop marker only in `run_forever`'s `finally`, so after the first browser test the marker points at Playwright's loop for the rest of the worker's life. `asyncio_mode = "auto"` (pyproject.toml) then runs the next `async def` test through `Runner.run()`, which refuses. The `stores` fixture's per-test resets of `_board_cache`, `_label_cache`, `_session_store`, `_transcript_store` are correct and were not the problem.

**Fix (isolation, not suppression):** module-scoped overrides of the five upstream fixtures in `tests/conftest.py`, each delegating to `pytest_playwright.pytest_playwright.<name>.__wrapped__` so their behaviour always matches the installed plugin and only the scope differs. `context`/`page` are function-scoped upstream and untouched, so per-test isolation is unchanged; only the browser process is recycled per file. All five must move together (pytest raises `ScopeMismatch` otherwise — verified by removing one). No test was reordered, deselected, marked, skipped, or weakened; no timeout was widened. The hostile reviewer also found a co-scheduling combination outside the four named files (`test_voice_mic_block_ui_browser.py test_web_manifest_ui_browser.py test_hermes_proxy.py test_agents_hosts_api.py -n 4`) where the override turns `68 failed, 93 passed` into `161 passed`.

**Cost:** `pytest -m "browser and not requires_server" tests/ -n 4`: 428 passed in 91.63s (session scope, base) → 428 passed in 100.93s (module scope). About +10% for one Chromium launch per file.

### #919
| AC | Evidence |
|----|----------|
| Two concurrent `test.sh` runs do not share a PID file and cannot kill/leak each other's server **through it** | `test_no_fixed_shared_pid_path` (no fixed shared path exists); two concurrent plan-only `test.sh auto` runs from separate directories each produced their own independent result (implementer ×3, hostile reviewer ×5). **Honest caveat:** runs that both need a server still kill each other's server through `server.sh start` → `kill_server` on the single hardcoded port — pre-existing, out of scope, tracked in #944, and stated in the script comment. |
| A crashed run's server is still cleaned up, or the reason it is unnecessary is stated in the script | Stated in the comment above `start_server_background()`: `server.sh start` kills/replaces whatever is on the port |
| The resolved path (if a file is still used) appears in the output | N/A — no file; `test_no_fixed_shared_pid_path` asserts none exists |
| No stale PID files accumulate | No PID file is ever written (hostile reviewer confirmed none created across concurrent runs) |
| Test scope and exit semantics unchanged | `decide_plan` untouched; plan-only `auto` output identical to base across ten inputs (hostile reviewer); existing `test_auto_scope_mapping` (24 cases) and `test_frontend_fixture_paths_exist` green; `bash -n scripts/test.sh` clean |
| `server.sh` owns lifecycle by port (pattern pin, passes on base by design) | `test_server_owns_lifecycle_by_port` |

### #916
| AC | Evidence |
|----|----------|
| Both named tests pass when `test_agents_board_api.py` is in the same `-n 4` invocation as the three browser files | Browser-first reproducer: `165 passed` — implementer 5/5 post-rebase, orchestrator 3/3, hostile reviewer 3/3 (+ `--dist load`, `--dist loadfile`, `-n auto` green); board-file-first order green |
| Both still pass alone and under the normal separate invocations | Each of the four files green alone (45 / 28 / 76 / 16); unit suite green in the pre-push gate (6489 passed, 9 skipped); `browser and not requires_server` 428 passed |
| Root cause named in commit message and code comment | Commit "fix(tests): tear down Playwright's driver loop per-file …" and the block comment above the overrides in `tests/conftest.py` |
| No assertion weakened, skipped, or removed | Diff touches only `tests/conftest.py` (new fixtures), two guard test files, script comments, and docs; no product code |
| Fix has automated regression coverage | `test_local_ui_driver_fixtures_are_module_scoped` (5 cases) fails 5/5 with the overrides removed — verified by both round-2 reviewers |
| Guard against upstream drift | `test_upstream_ui_driver_fixture_signatures_unchanged` (5 cases: parameters + generator-vs-return shape) |

## Mutation proofs (commit-first procedure; one mutation, restored in the same step)

- **`test_no_fixed_shared_pid_path`** — re-added a `stop_test_server()` reading `/tmp/lifeos_test_server.pid` to `scripts/test.sh` → `FAILED ... assert '/tmp/lifeos_test_server' not in ...`. Restored → 29 passed. (Re-run by both round-1 Opus reviewers.)
- **`test_server_owns_lifecycle_by_port`** — renamed `get_server_pid()` in `scripts/server.sh` → `FAILED ... assert 'get_server_pid()' in ...`. Restored → green.
- **#916 fix** — `tests/conftest.py` reverted to the base (overrides removed), browser-first reproducer → `2 failed, 163 passed` with `RuntimeError: Runner.run() cannot be called from a running event loop` on exactly the two named tests; minimal two-file reproducer → `1 failed, 7 passed`. Restored → `165 passed` / `8 passed`. (Re-run by all three round-1 reviewers and the round-2 reviewer.)
- **`test_local_ui_driver_fixtures_are_module_scoped`** — same base-conftest swap → `5 failed, 5 passed`, all five failures this test (`scope 'session', not 'module'`). Restored → 10 passed.
- **`test_upstream_ui_driver_fixture_signatures_unchanged`** — appended a nonexistent parameter to the expected `browser_type` signature → `FAILED ...[pw-engine-type]`; flipped an entry of the generator table → fails. Restored → green.
- All ten guard cases are confirmed **selected** by the unit filter (`--collect-only` ids carry none of the `browser`/`playwright` auto-marking substrings; `0 deselected`).
- Docs/comment-only commits: no mutation proof applicable.

## Verification (post-rebase onto `668fecd`; all foreground; pre-push gate is the authoritative full run)

- Pre-push gate on push: unit `6489 passed, 9 skipped`; `browser and not requires_server` green
- Browser-first reproducer: `165 passed` ×5 (implementer), ×3 (orchestrator), ×3 (hostile reviewer); original order `165 passed`; each file alone `45 / 28 / 76 / 16 passed`
- `tests/test_fixture_scope_isolation.py` + `tests/test_test_sh_auto.py` under the unit filter: all selected, green; `tests/test_prepush_gate.py` 39 passed
- `tests/test_no_change_narration.py`: 1 passed; `bash -n` clean on `scripts/test.sh` and `scripts/pre-push`; plan-only `test.sh auto` paths green; no port 8000 bound, no real server started

## Review
Round 1: Opus general + Opus hostile + Codex (3 Action Required, 4 Recommended accepted; consolidated comment on the PR). Round 2: Opus general on the delta (code delta approved; two Recommended one-liners folded in). Follow-ups filed: #944 (shared port-8000 test server), #945 (two `test_prepush_gate.py` tests auto-marked `browser` by name and never run).

## Docs
- `AGENTS.md` § Testing — co-scheduling caveat, bisection guidance, pointer to the conftest comment
- `docs/specs/standards/testing-standards.md` — pointer to the caveat
- `scripts/pre-push` — one-sentence correction of the mechanism description

## Scope
**In:** `scripts/test.sh` PID handling and its comment; `tests/conftest.py` fixture scope overrides; guard tests; docs; one `scripts/pre-push` comment sentence.
**Out:** the shared port-8000 server contention (#944); which tests `test.sh` selects, `decide_plan`, markers, exit codes; the board stream's product behaviour; the `/agents` UI; the browser tests' own assertions; `api/routes/agents.py` (untouched).

Closes #919
Closes #916


nbramia:
## Review round 1 — consolidated (reviewers: `general` Opus, `hostile` Opus, `codex`)

Head reviewed: `bc7a8e2`. All three reviewers independently reproduced the #916 defect on the unfixed tree (browser-files-first reproducer: `2 failed, 163 passed`, deterministic 3/3 in the hostile pass) and its absence on the head (`165 passed`, 3/3 + `--dist load`, `--dist loadfile`, `-n auto` all green). The hostile pass also found a co-scheduling combination outside the four named files where the override prevents 68 failures. Both root causes and the #919 dead-code diagnosis were confirmed. Verdicts: general REQUEST_CHANGES, hostile REQUEST_CHANGES, codex one Action Required.

### Referee table

| # | Finding | general | hostile | codex | Decision |
|---|---------|---------|---------|-------|----------|
| 1 | `tests/test_fixture_scope_isolation.py` asserts only upstream's shape; deleting all five conftest overrides leaves it (and the whole suite) green — the #916 fix has no automated regression coverage | AR | AR | Rec | **Accepted — Action Required.** Add a parametrized test that asserts the *nearest visible* definition of each of the five fixtures is module-scoped (`request._fixturemanager.getfixturedefs(name, request.node)[-1].scope == "module"`), and assert generator-vs-return shape of each upstream fixture matches how conftest delegates it (`inspect.isgeneratorfunction`). Mutation proof: remove the five overrides → new test fails 5/5; restore → green. |
| 2 | "`--dist loadscope` groups tests by test *class*, not by file" is wrong: `_split_scope` is `nodeid.rsplit("::", 1)[0]`, so top-level functions group by module; `test_agents_assignment_ui_browser.py` (28 module-level tests) is one module group; the four files form 27 groups, not 26; `tests/conftest.py:455` already states it correctly | Rec | AR | Minor | **Accepted — Action Required.** Fix `AGENTS.md` and the conftest block comment to "by class for tests inside a class, by module for top-level functions"; fix the PR body's group count. |
| 3 | `scripts/test.sh` comment and `test_no_fixed_shared_pid_path`'s message imply that removing the PID path stops concurrent `test.sh` runs killing each other's server; they still can, via `server.sh start` → `kill_server` (`lsof -ti :$PORT \| xargs kill -9` + `pkill -9 -f "uvicorn.run.*api.main:app"`) on one hardcoded port | Rec | AR | AR | **Accepted — Action Required (wording + honest AC mapping).** The removal stands; the comment and the assertion message must say plainly that the server itself remains one shared port-8000 resource that `server.sh start` kills and replaces, so two runs that both need a server remain mutually destructive — that is `server.sh`'s contract and out of this PR's scope. The remaining hazard is filed as a follow-up issue (number in the PR body). The PR body's AC table must state this honestly. Fixing the port contention itself is **out of scope** (would touch `server.sh`, `check_server`, e2e base URLs). |
| 4 | Guard checks parameter names but not generator-vs-return shape | — | Rec | — | **Accepted**, folded into #1. |
| 5 | Name the minimal deterministic two-file sequential reproducer (`pytest tests/test_voice_mic_block_ui_browser.py tests/test_agents_board_api.py::TestBoardStream -q`, no xdist) alongside the `-n 4` one | Rec | — | — | **Accepted.** Add to the conftest comment; keep the `-n 4` reproducer as the real-suite example. |
| 6 | Mechanism prose duplicated in full in both `AGENTS.md` and the conftest comment while `AGENTS.md` also says the comment carries it (single source of truth) | Rec | — | — | **Accepted.** `AGENTS.md` keeps the bisection advice (browser files first) and the pointer; the mechanics live only in the conftest comment. |
| 7 | `scripts/pre-push:334-336` pre-existing comment says the parked loop makes a later async test "silently never awaited"; it actually raises `RuntimeError: Runner.run() cannot be called from a running event loop` — after this PR the repo carries two incompatible explanations | — | Rec | — | **Accepted (one-sentence correction).** Same mechanism the PR documents; leaving a contradicting comment in place is worse than the surgical-change cost. |
| 8 | Two unit-marked tests in `tests/test_prepush_gate.py` (~L289, ~L459) have `browser` in their names and are auto-marked/deselected from every unit gate | — | — | Rec | **Out of scope** (pre-existing, sibling file). Filed as a follow-up issue (number in the PR body). |
| 9 | PR body cites pre-rebase SHAs (`86bd19a`, `134109d`, `c4cb0a4`, `5376bfb`, base `5b3e11c`) | Minor | — | — | **Accepted.** Orchestrator updates the PR body after the round. |
| 10 | `scripts/test.sh` "on this box" → machine-flavoured phrasing | Minor | — | — | **Accepted** ("on one machine"). |
| 11 | `test_server_owns_lifecycle_by_port` passes on the base (a pin on `server.sh` internals, not a regression test); placement of the two guards in a file scoped to `auto` plan mapping | Minor | — | — | **Noted, no change.** It is labelled as a pattern pin in the AC table; the placement drift is pre-existing (`test_frontend_fixture_paths_exist`). |
| 12 | `assert "stop_test_server" not in text` is a raw substring check | — | Minor | — | **Noted, no change.** |
| 13 | Module-level `import pytest_playwright.pytest_playwright` in conftest widens a missing plugin's blast radius to collection | — | Minor | — | **Noted, no change.** `pytest-playwright>=0.7.0` is pinned in `requirements.txt`. |

Accepted findings this round: **7** (3 Action Required, 4 Recommended). Codex's Action Required overlapped the Opus reviewers' #3; its Recommended #1 overlapped Action Required #1.

nbramia:
## Review round 2 — consolidated (reviewer: `general` Opus, delta-only per the referee calibration; no hostile pass this round)

Delta reviewed: `f7e11a7` (bc7a8e2..HEAD). Verdict on the code delta: correct and behaviour-free (comments, docstrings, assertion messages, one table, one assertion, one new test). Reviewer verification: the new `test_local_ui_driver_fixtures_are_module_scoped` fails 5/5 with the conftest overrides removed and passes 10/10 with them, 0 deselected; `getfixturedefs` order confirmed outermost-first on pytest 9.0.2; generator table matches pytest-playwright 0.7.2; the minimal two-file reproducer reproduces (`8 passed` head, `1 failed, 7 passed` base conftest); loadscope wording matches `_split_scope`; narration test green; `bash -n` clean on both scripts; `tests/test_prepush_gate.py` 39 passed.

| # | Finding | Decision |
|---|---------|----------|
| 1 | PR body still carries round-1 text: "26 scope groups" (it is 27), class-only loadscope claim, no port-contention caveat in the #919 AC table, no follow-up issue numbers, pre-rebase SHAs, no row for the new scope guard | **Accepted — Action Required (orchestrator).** PR body rewritten in this round. |
| 2 | Scope guard checks upstream's generator shape against a constant but not conftest's own delegation shape; a `yield from`→`return` regression in conftest passes the guard (the browser suite still fails loudly) | **Accepted — Recommended, folded into the convergence commit** (`assert inspect.isgeneratorfunction(defs[-1].func) == _EXPECTED_IS_GENERATOR[...]`). |
| 3 | `scripts/test.sh:58` pre-existing comment "keeps every test in a module on the same worker" contradicts the corrected wording two lines below | **Accepted — Recommended, folded into the convergence commit** (one sentence). |

Surviving accepted findings are Recommended-only with no reachable defect → converged. Proceeding to the docs gate and verify after the convergence commit lands.

nbramia:
## Docs gate — consolidated (reviewer: `docs` Opus)

| # | Finding | Decision |
|---|---------|----------|
| 1 | `AGENTS.md`, the `tests/conftest.py` block comment, and `testing-standards.md` said `--dist loadscope` hands scope groups out **in collection order** and that file order therefore decides which browser group runs before the async test. Against the installed pytest-xdist 3.8.0 that is false: `--loadscope-reorder` defaults to on, so groups are handed out largest-first with command-line order only breaking ties (reviewer confirmed by reading `xdist/scheduler/loadscope.py:374-382` + `xdist/plugin.py:131-135`, an empirical two-file run where the first-listed file ran last, and the computed hand-out order of this PR's four-file reproducer, where a group from the last-listed file precedes one from the second-listed file). | **Accepted — Action Required (factual).** Fixed in a docs-only commit: all three sites state the largest-first rule and point bisection at the deterministic single-process form (`pytest <browser file> <suspect> -q`, no `-n`), where sequential order is exactly command-line order. The reproducer commands and the fix are unchanged; the browser-first `-n 4` reproducer remains the real-suite example (it reproduces deterministically on this machine, as all four reviewers observed), just no longer explained by "collection order". |

Everything else the docs reviewer checked (server.sh `kill_server` claim, the Playwright greenlet mechanism, fixture signatures/shapes, header blocks, Related Documents, narration gate, stale-reference sweep) was found accurate. Converged; proceeding to merge after the push gate.

PR #941, opened 2026-09-05
fix(agents): enforce _CACHE_MAX on the summary cache's disk-hit promotion writes (#902)

### LINKED ISSUE #902: fix(agents): enforce _CACHE_MAX on the summary cache's disk-hit promotion writes
## Problem

`api/services/agent_viz_summary.py`'s in-process `_cache` has four write sites. Two of them — the
paths that promote a disk hit back into memory — assign `_cache[session_id] = (...)` without the
`if len(_cache) >= _CACHE_MAX: _cache.pop(next(iter(_cache)))` eviction that the other two apply. A
disk-hit-heavy workload therefore grows the cache past its declared bound.

Measured on a branch build: 600 disk-hit promotions left `len(_cache) == 600` against
`_CACHE_MAX == 500`. FIFO eviction on the two guarded write sites is intact — 600 writes through
those leave exactly 500 entries.

This is **pre-existing**, not introduced by #863. The same shape is present on `d93b34c`
(`api/services/agent_viz_summary.py:385` and `:466` there). It was found during #863's round-3 review
because that PR touched all four sites, and was deliberately left out of scope.

Impact is bounded — the entries are small summary tuples — but the cap exists to bound them, and a
long-lived API process that serves many cached summaries will exceed it without limit.

## Solution

Apply the same eviction check on the two disk-hit promotion writes, or route all four writes through a
single small `_cache_put(session_id, entry)` helper that enforces `_CACHE_MAX` in one place.

## Acceptance Criteria

- [ ] Given `_CACHE_MAX` consecutive disk-hit promotions plus one more, `len(_cache)` never exceeds
      `_CACHE_MAX`.
- [ ] Eviction on the disk-hit path is FIFO, matching the existing write sites.
- [ ] All four write sites enforce the bound (ideally through one helper, so a fifth cannot forget).
- [ ] Existing summary and prefetch tests continue to pass.

## Verification

- **Automated tests:** a test that drives `_CACHE_MAX + N` disk-hit promotions and asserts the cache
  size and that the oldest keys were evicted; extend the existing FIFO test to cover all write sites.
- **Existing test suite:** green.

## Scope

**In scope:** `_cache` bound enforcement in `api/services/agent_viz_summary.py`.
**Out of scope:** the disk cache, `prune_disk_cache` scheduling, and any change to what is cached.

## Technical Context

- **Key file:** `api/services/agent_viz_summary.py` — `_cache`, `_CACHE_MAX`, the two guarded writes
  and the two unguarded disk-hit promotions.
- **Constraints:** no new dependencies; no behavior change beyond bounding the cache.

## References

- Found during PR #890's round-3 review (issue #863). Both reviewers independently confirmed it is
  pre-existing on `d93b34c`. See PR #890's round-3 consolidated review comment.
- Related: `prune_disk_cache` in the same module currently has no caller anywhere in `api/` or
  `scripts/`, so the disk cache's 90-day cutoff never runs. Worth considering alongside this, though
  it is a separate concern.

nbramia:
## TL;DR

The `/agents` page keeps a small in-memory cache of per-session summaries so it does not re-ask the LLM for sessions it has already summarised. That cache is meant to hold at most 500 entries, but two of the paths that fill it — the ones that copy a summary back into memory from the on-disk cache — skipped the size check, so a long-running server that served many cached summaries could grow the cache without limit. This change makes every path that writes to the cache go through one guarded helper, so the 500-entry cap holds everywhere and the oldest entries are the ones dropped. Nothing about what is summarised, stored on disk, or shown to the user changes.

## Summary

`api/services/agent_viz_summary.py`'s in-process `_cache` had four write sites, but only two applied the `_CACHE_MAX` FIFO eviction. The two disk-hit promotion writes (in `summarize_session` and `get_cached_summary`) assigned `_cache[session_id] = (...)` directly, so a disk-hit-heavy workload grew the cache past its declared bound (600 promotions left 600 entries against a cap of 500).

This PR adds a single `_cache_put(session_id, last_activity_at, result, *, is_error_fallback=False)` helper that enforces the cap with FIFO eviction and routes all four write sites through it, so a future fifth write site cannot forget the bound. No tuple contents, disk writes, return values, or freshness semantics change.

Closes #902.

## Acceptance criteria → tests

| Criterion | Test |
|-----------|------|
| `_CACHE_MAX` consecutive disk-hit promotions plus more never exceed `_CACHE_MAX` | `test_disk_hit_promotion_respects_cache_max[get_cached_summary]`, `[summarize_session]` (asserts `len(_cache) <= cap` after every promotion and `== cap` at the end) |
| Eviction on the disk-hit path is FIFO | same test: `list(_cache) == sids[3:]` (the three oldest evicted, in order) |
| All four write sites enforce the bound through one helper | `test_cache_put_enforces_cache_max` (helper), `test_cache_if_terminal_evicts_at_capacity` (terminal-cache site), the parametrized test above (both promotion sites); the real-summary write site is covered structurally by the helper |
| Existing summary and prefetch tests continue to pass | `tests/test_agent_viz_api.py` 66/66 |

## Mutation proofs (safe commit-first procedure)

- Mutation A — `get_cached_summary` promotion reverted to a bare `_cache[...] =` write: `test_disk_hit_promotion_respects_cache_max[get_cached_summary]` FAILS; restored → passes.
- Mutation A' — `summarize_session` promotion reverted the same way: `test_disk_hit_promotion_respects_cache_max[summarize_session]` FAILS; restored → passes.
- Mutation B — eviction removed from `_cache_put`: all four cases (the two helper-level tests plus both parametrized cases) FAIL; restored → 66/66.

## Who wrote what

- **DeepSeek v4 Flash on Fireworks wrote the production change** (`_cache_put` helper + the four call sites), via the LifeOS agent worker's `#cloud` route pinned to an isolated worktree (`fields.working_dir`). It **hit the worker's default 500k-token cap** after 26 tool calls (≈$0.06) before committing; the draft was committed as-is (first commit on this branch). Its two tests were helper-level only — mutation testing showed they still passed with the actual bug reintroduced at either promotion site.
- **A Sonnet addresser replaced that coverage with call-site tests**: the parametrized disk-hit promotion test that drives both `get_cached_summary` and `summarize_session`, plus `monkeypatch` normalisation and a narration-clean header for the two helper-level tests it kept (second commit).

## Scope

In scope: `_cache` bound enforcement in `api/services/agent_viz_summary.py` and its tests. Out of scope (per the issue): the disk cache, `prune_disk_cache` scheduling, and what is cached.


nbramia:

**Reviewers run:** general (Opus), hostile (Opus, adversarial/execution-driven), codex (cross-model)

**Specialist rationale:** none beyond the milestone's standing roster — the change is a single helper in one module; the hostile reviewer and Codex substitute for specialists by executing the code against the acceptance criteria.

**Head reviewed:** 051cf7f (rebased onto feat/agents-board-polish @ 668fecd; pre-push gate green: 6517 unit + server-free browser).

### Reviewer Findings

#### General (Opus)
**Action Required:** none.
**Recommended:** none.
**Minor:**
- `api/services/agent_viz_summary.py:482-484` — `_cache_put` evicts the FIFO-oldest entry whenever `len(_cache) >= _CACHE_MAX`, including when `session_id` is already present. A dict overwrite does not add a key, so an at-capacity re-write of an existing session drops the cache to `_CACHE_MAX - 1` and discards an unrelated, still-valid entry. Reachable on the newly-routed promotion path: an in-memory error-fallback entry whose TTL has expired falls through to `_disk_get`, which can return a fresh real summary for the same key. Cost is one extra disk read later, never unbounded growth; `if session_id not in _cache and len(_cache) >= _CACHE_MAX:` is the whole fix.
- `tests/test_agent_viz_api.py:1325-1350` — `test_disk_hit_promotion_respects_cache_max` leaves five entries in `avs._cache` on exit while its two neighbours clear in a `finally`. Harmless today (last test in file; every `summary_db` consumer resets), but it adds latent ordering dependence.
- `tests/test_agent_viz_api.py:1275` — `test_cache_put_enforces_cache_max` requests `summary_db` only for its `reset_cache()` side effect.
**Summary:** PR is sound; all four former write sites funnel through one `_cache_put` (exactly one `_cache[` assignment remains, at :484). Criteria 1–2 are bound by the parametrized disk-hit test whose in-loop assertion fires on the sixth iteration if either promotion hunk is reverted; removing the helper's eviction breaks all four new cases. Criterion 4: 66 passed against 051cf7f. Tuple contents, `time.time()` stamping, disk writes, return values, freshness semantics unchanged.

#### Hostile (Opus)
Verified first: at the real `_CACHE_MAX = 500`, 600 promotions through `get_cached_summary` leave exactly 500 entries in FIFO order `p100..p599`; same through `summarize_session`; interleaving with `_cache_if_terminal` stays at 500. Single-assignment invariant confirmed across `api/`, `scripts/`, `tests/`. 66/66 green.
**Action Required:** none — could not make `len(_cache)` exceed `_CACHE_MAX` through any entry point.
**Recommended:**
1. `api/services/agent_viz_summary.py:482-484` — `_cache_put` evicts an unrelated live entry when the key is already present at capacity. Reproduced end-to-end through production `get_cached_summary` with no monkeypatching of the module under test (`_CACHE_MAX = 5`, four healthy entries + an in-process error fallback for `hot`, disk holding a fresh real summary for `hot`, live session whose activity advanced): before `['other-0','other-1','other-2','other-3','hot']` → after `['other-1','other-2','other-3','hot']` (len 4, `other-0` lost); pre-change code keeps all five. Each lost entry is one extra Gemma summarization later; the loss does not compound (cache sits at MAX−1, further re-puts are no-ops). Second, cosmetic consequence: when the rewritten key IS the head, `pop(next(iter(_cache)))` pops it and reinsertion moves it to the tail (`['f0','f1','f2']` re-put `f0` → `['f1','f2','f0']`), so FIFO order depends on whether the rewritten key was at the head. Smallest fix: `if session_id not in _cache and len(_cache) >= _CACHE_MAX:` plus one test asserting a re-put of a present key at capacity leaves `len == _CACHE_MAX` with no unrelated key dropped.
**Minor:**
2. `tests/test_agent_viz_api.py:1300-1321` — `test_cache_if_terminal_evicts_at_capacity` passes with the entire production change reverted in a scratch copy (it asserts pre-existing behavior at an already-guarded site). Honest "sanity" test; guards the routing if that site is later un-routed; do not count it as coverage of the fix.
3. `api/services/agent_viz_summary.py:617` — the real-summary tail of `summarize_session` routes through `_cache_put` but no test drives it; reaching it needs a stubbed `generate_text`. Flagged against AC 3 ("all four write sites"), not as a blocker.
4. `tests/test_agent_viz_api.py:1325-1350` — the disk-hit test has no `try/finally: avs._cache.clear()` unlike its siblings.
**Summary:** Fix is correct on its stated goal; nothing blocks merge. The one substantive finding is that `_cache_put` conflates insert with overwrite at capacity.

#### Codex (cross-model)
**Action Required:** none.
**Recommended:**
- `api/services/agent_viz_summary.py:478` — `_cache_put` evicts before checking whether `session_id` is already present. At capacity, refreshing an existing non-oldest entry evicts the oldest unrelated entry and leaves `_cache` at `_CACHE_MAX - 1`; the pre-change disk-hit assignment only refreshed that tuple. Semantic drift on the disk-hit path. Use `if session_id not in _cache and len(_cache) >= _CACHE_MAX:` unless intentional and tested.
- `tests/test_agent_viz_api.py:1316` — the disk-hit promotion test covers both disk-hit call sites, but the real post-LLM summary write site is not capacity-tested; if it were changed back to a bare `_cache[...] = ...`, the new tests would still pass. Add a fake-`generate_text` capacity test driving the real-summary branch with distinct session ids, asserting cap/FIFO.
**Minor:** could not run the focused pytest locally (its sandbox failed on `bwrap: loopback`); reviewed via the diff and file contents.
**Summary:** the change centralizes the four write sites; the sequential distinct-disk-hit path is bounded and FIFO; main gap is overwrite-at-capacity behavior.

**Cross-model overlap:** Codex's two Recommended findings were independently raised by the Opus hostile reviewer (REC-1 and MINOR-3) and the first also by the Opus general reviewer — full overlap, no Codex-only findings.

### Referee Decisions

| 1 | `_cache_put` evicts on size alone; a re-put of a present key at capacity drops an unrelated entry (cache → MAX−1) and shuffles FIFO when the key is the head | codex, general, hostile | Recommended / Minor / Recommended | Valid — reproduced end-to-end via production `get_cached_summary`; violates the issue's "no behavior change beyond bounding the cache" on the promotion path | **Accept** | Add `session_id not in _cache and` to the guard; add one positive test: at capacity, re-put a present key → `len(_cache) == _CACHE_MAX`, no unrelated key lost, insertion order unchanged. Mutation proof: drop the membership clause → test fails. |
| 2 | Real-summary write site (`summarize_session` tail, :617) has no capacity test; reverting it to a bare write leaves all tests green | codex, hostile | Recommended / Minor | Valid — AC 3 names all four sites and the milestone's mutation rule requires each routed hunk to be pinned | **Accept** | One test with a stubbed async `generate_text` (pattern at `tests/test_agent_viz_api.py:833`) driving `summarize_session` with `_CACHE_MAX + 3` distinct sids and non-empty events; assert cap and FIFO. Mutation proof: revert :617 to a bare write → fails. |
| 3 | Disk-hit test leaks cache state (no `finally: clear`) | general, hostile | Minor | Valid but harmless today | **Accept (ride-along Minor)** | Match the sibling `try/finally: avs._cache.clear()` pattern while the file is open. Cannot keep the loop open by itself. |
| 4 | `test_cache_put_enforces_cache_max` uses `summary_db` only for reset | general | Minor | Unproven as a defect | **Reject** | Stylistic; `summary_db` is the file's standard isolation fixture and also points `_DB_PATH` at a temp DB, which is prudent for any test touching this module. |
| 5 | `test_cache_if_terminal_evicts_at_capacity` passes with the whole change reverted | hostile | Minor | Valid observation, not a defect | **Reject** | It pins the bound at write site 1 regardless of implementation and fails under the helper-eviction mutation; keeping it satisfies AC 3 at that site. Not counted as coverage of the disk-hit fix (that is the parametrized test). |
| 6 | Codex could not execute tests in its sandbox | codex | Minor | n/a | **Record** | The Opus reviewers and the orchestrator executed the suite and mutation proofs on 051cf7f. |

**Findings forwarded to addresser:** 3 (2 Recommended + 1 ride-along Minor)

nbramia:

**Reviewers run:** general (Opus), hostile (Opus, execution-driven), codex (cross-model)

**Specialist rationale:** none — same standing roster as round 1; both Opus reviewers were briefed to attack the round-1 fix delta (051cf7f → 289b899) as the suspected source of new defects.

**Head reviewed:** 289b899 (locally; the push was still in its pre-push gate while reviewers ran — content identical).

### Reviewer Findings

#### General (Opus)
**Action Required:** none.
**Recommended:**
- `tests/test_agent_viz_api.py:1300-1315` — `test_cache_put_repeat_key_at_capacity_evicts_nothing` does not exercise the behavior its docstring claims and does not check that a repeat put still refreshes the entry. (1) Both re-puts target the key that is currently *oldest* (`"a"` in `[a,b,c]`, then `"b"` in `[b,c,a]`); under the pre-fix guard each pops the key itself, so nothing unrelated is lost — simulated: result `['c','a','b']`, `len == 3`, `set == {a,b,c}`; the `len`/`set` assertions still pass and only the ordering assertion kills the mutant. Re-putting a **non-oldest** key is the actual loss case (`[a,b,c]` + `_cache_put("c")` under the pre-fix guard → `['b','c']`, unrelated `"a"` destroyed). Adding one `avs._cache_put("c", 1.0, ...)` before the existing re-puts makes the test fail on all three assertions. (2) The refreshed value is never asserted: a mis-implementation `if session_id in _cache: return` leaves `len`, `set`, and order intact while silently discarding the refreshed summary and `last_activity_at`; one assertion such as `assert avs._cache["a"][2].short_label == "a2"` and `assert avs._cache["a"][0] == 1.0` closes it.
**Minor:**
- `api/services/agent_viz_summary.py:473-474` — the `_cache_put` docstring still reads "When at capacity, evicts the oldest-inserted entry (dict insertion order) and then writes the new one," omitting the repeat-key condition at :482 that is the whole point of the round-1 fix; a future editor could "simplify" the guard back to size-only.
**Summary:** the round-1 delta is sound and all three accepted round-1 findings are genuinely resolved; `test_real_summary_write_site_respects_cache_max` really reaches the tail write site (module-level `generate_text` import at :24 so the monkeypatch binds; the events yield non-empty context). `_cache[...] =` appears exactly once in the module; 68 passed in 11.5s. No regression from the delta.

#### Hostile (Opus)
Everything executed, not read off the diff. **Action Required:** none — could not break the round-1 fix. Evidence: reverting the r1 hunk to `if len(_cache) >= _CACHE_MAX:` fails `test_cache_put_repeat_key_at_capacity_evicts_nothing` (order becomes `['c','a','b']`); a bare assignment at the `summarize_session` tail (:617) fails `test_real_summary_write_site_respects_cache_max` ("exceeded at real-summary-5: 6") and an instrumented `generate_text` was called 8/8 times; bare assignments at the two disk-hit sites (:550, :642) fail both parametrizations of the disk-hit test. A 4000-step mixed workload at `_CACHE_MAX = 500` with 40% repeat keys randomly interleaving all five entry points never exceeded 500. AC 1 verbatim at `_CACHE_MAX = 50`: 51 promotions → `len == 50`, survivors `dh-1 … dh-50` (FIFO). `_cache_put` has no `await` and every production caller (`/api/agents/snapshot`, the SSE stream, the prefetch task) runs on the single event loop; an 8-thread × 3000-insert hammer found 0 breaches. `created_at` semantics are byte-for-byte equivalent to pre-PR (all four sites already stamped `time.time()` at write time). Exactly one assignment to the dict in `api/`. 68/68.
**Recommended:**
1. `tests/test_agent_viz_api.py:1385-1410` — `test_real_summary_write_site_respects_cache_max` never asserts it took the branch it is named for. Against the shipped code vs. the tail-site mutant: as written → pass / **fail** (correct); with `generate_text` raising → pass / pass; with `events=[]` → pass / pass. Both fallback branches route through `_cache_if_terminal → _cache_put`, which is bounded independently, so the assertions hold even with the site under test broken. Today the test exercises the right branch, but it would silently stop doing so if the event-shape contract `_extract_context` reads ever drifts. Smallest fix: have `_succeed` append to a `calls` list and `assert len(calls) == len(sids)`.
**Minor:**
1. `api/services/agent_viz_summary.py:470-481` — the `_cache_put` docstring does not state the re-put position rule the new test pins ("a re-put refreshes the value and `created_at` but not the insertion position"). Right call (the issue asks for FIFO, not LRU) and it matches the pre-PR bare-assignment sites; worth one clause so the next reader does not "fix" it.
2. Bounding the cache introduces a sharp cliff once the snapshot fan-out exceeds 500 nodes: `_build_snapshot` calls `get_cached_summary` for every node in a fixed order every 2s per SSE client — the FIFO worst case. Measured with 600 seeded sessions and cap 500: 600 disk reads per tick (0% hits, ~20 ms) vs 0 reads (~0.2 ms) when the cap exceeds the node count. This regime did not exist pre-PR (promotion was unbounded) but is the direct consequence of the bound the issue asked for. Headroom today: one live tick labels 224 nodes against a cap of 500. No change needed now.
**Summary:** the round-1 fix is correct; no new defect. One test-quality item worth fixing before merge (the real-summary test should assert it reached the LLM branch); two minor notes (docstring; the >500-node cliff).

#### Codex (cross-model)
**Action Required (process, not code):** exact local-diff review was blocked — its sandbox shell failed at setup (`bwrap: loopback: Failed RTM_NEWADDR`) and GitHub still showed head 051cf7f (push in gate), without the two round-1 tests.
**Recommended / Minor:** none from the described delta.
**Summary:** based on the described fixes, the round-1 findings are conceptually resolved (membership guard preserves size for existing-key re-puts without changing `created_at`/grace/TTL semantics; a non-empty-event `generate_text` stub test exercises the LLM-success branch and fails if the tail were reverted). Could not independently verify local files.

**Cross-model overlap:** Codex produced no independent findings this round (tooling failure). The two Opus reviewers' findings are disjoint but complementary (each hardened one of the two round-1 tests) and both flagged the same docstring gap.

### Referee Decisions

| 1 | Repeat-key test re-puts only the head key (loss case never exercised) and never asserts the refreshed value | general | Recommended | Valid — simulated against the pre-fix guard: only the order assertion kills the mutant; a `return`-on-present mis-implementation passes | **Accept** | Inside the existing test: re-put a non-oldest key (`"c"`) before the existing re-puts; assert `avs._cache["a"][2].short_label == "a2"` and `avs._cache["a"][0] == 1.0`. Mutation proof: pre-fix guard → all three assertions fail; `if session_id in _cache: return` variant → value assertion fails. |
| 2 | Real-summary test never asserts the LLM branch was taken; passes on both fallback branches with the site broken | hostile | Recommended | Valid — demonstrated with `generate_text` raising and with `events=[]` | **Accept** | Count `_succeed` invocations; `assert len(calls) == len(sids)`. Mutation proof: `events=[]` (or a stub that raises) → the count assertion fails. |
| 3 | `_cache_put` docstring omits the repeat-key rule the fix and test pin | general, hostile | Minor | Valid | **Accept (ride-along Minor)** | One clause describing current behavior only (a re-put updates in place and keeps its FIFO position; eviction happens only for a genuinely new key at capacity). Must stay narration-clean (no issue numbers / "now"). |
| 4 | FIFO cap becomes a 0%-hit cliff if snapshot fan-out exceeds 500 nodes (224 today) | hostile | Minor | Valid observation, not a defect | **Reject for this PR** | Inherent to the bound the issue requests; sizing `_CACHE_MAX` is out of scope. Recorded as an operator follow-up in the final report. |
| 5 | Codex could not read local files / GitHub lagged | codex | (process) | n/a | **Record** | The two Opus reviewers executed the delta directly; the orchestrator independently re-ran every mutation proof on 289b899. |

**Findings forwarded to addresser:** 3 (2 Recommended + 1 ride-along Minor)

### Convergence audit (after two non-clean rounds)

| Sub-problem | Round-1 findings | Round-2 findings | Density | Status |
|-------------|------------------|------------------|---------|--------|
| `_cache_put` eviction semantics (production) | 1 (evict-on-size; accepted, fixed in 289b899) | 0 production findings; 1 docstring clause | 2 | Resolved in code; docstring catch-up only |
| Coverage of the four write sites | 1 (real-summary site untested; accepted, fixed) | 0 | 1 | Resolved |
| Robustness of the two round-1 tests | — | 2 (assertions strengthened) | 2 | Refinements of review-introduced tests; each is 1–3 lines inside an existing test, no new fixture/helper |
| Test hygiene | 1 minor (finally-clear; fixed) | 0 | 1 | Resolved |

Converging: no open production defect; the acceptance criteria hold under execution (hostile round 2); every remaining item is an assertion or docstring clause inside code this PR already contains. The equal accepted count (3 → 3) is severity-deflating (2 Recommended + 1 Minor both rounds, but round 2's Recommended items are test-assertion hardening rather than behavior). Proceeding to a round-3 address + review with the expectation of a clean exit; if round 3 surfaces anything beyond mechanical follow-through, the loop escalates.

nbramia:
## End-to-End Verification — PR #941

### Verdict: PASS (with two process flags, no code defects)

```
verification-head: e7ac88a53e92d841b5dddca270e6b376dc14b825
suite-result: not-executed
suite-command: ./scripts/test.sh
suite-executions: 0
suite-exit-status: n/a
```

**Flag 1 — head mismatch.** `gh pr view 941 --json headRefOid` returns `289b899b07f5b26c5bf7e52b41312ed55abb389a`; the local branch head is `e7ac88a` (clean worktree, one unpushed commit on top). I verified `e7ac88a`. The production delta between the two is a docstring clause only, so runtime behavior is identical; the difference is the two hardened test assertions from `e7ac88a`. **Merging what GitHub currently has would ship without them** — push before merging.

**Flag 2 — authoritative suite not executed.** No complete exact-head evidence record exists (PR comments carry gate evidence for `051cf7f` and `289b899`, without the five fields). I did not execute it: the orchestrating agent's brief explicitly forbade a full-suite run, and at execution time two full gates were already running on this host (a `flock`-held browser gate plus a `pytest -m "unit and not slow" -n auto` run) with 6 GB of 7 GB swap consumed. Memory records 3–4 concurrent gates OOM-rebooting this box, which hosts the live API. Gap noted, proceeding per the skill.

### System Flow Verified

`GET /api/agents/snapshot` → `_build_snapshot` → `_session_to_dict` → `_short_label_for_snapshot` → `get_cached_summary` → SQLite disk cache hit → `_cache_put` promotion → `short_label` in the JSON response. Also the async `summarize_session` path the prefetch loop drives.

### Evidence

All scenarios ran the real module at the real `_CACHE_MAX = 500`, imported with `PYTHONPATH` at the PR head, disk cache and stores pointed at temp dirs. No live service, real vault, or `data/` touched.

| Scenario | Result |
|---|---|
| A — 600 genuine disk-hit promotions via `get_cached_summary` | max and final `len(_cache)` = 500; survivors `gcs-0100…gcs-0599`, FIFO order exact |
| B — 600 genuine disk-hit promotions via `summarize_session` | max and final = 500; FIFO exact; **0 LLM invocations** (stub raises if called) |
| C — real router, `TestClient(api_main.app)`, 600 seeded sessions | tick 1 (cold): 200, 600 rows, 0 short_labels. tick 2: 200, **600 of 600 short_labels**, `len(_cache)` = 500, FIFO survivors = last 500 promoted in tick order, 0 LLM calls. tick 3: still 500 |
| D — baseline, same workload against `origin/feat/agents-board-polish` | **600 entries against a declared cap of 500** — the issue's bug reproduced, so the fix is proven, not assumed |
| E — 4000-step interleaved workload, 40% repeat keys, all write sites | max `len(_cache)` = 500, never breached |
| E2 — re-put a present non-oldest key at exact capacity | size 500→500, insertion order unchanged, oldest survives, value and `last_activity_at` refreshed |
| F — failure mode, disk cache unreadable mid-promotion | returns `None`, no exception, cache untouched |

**Regressions (real endpoints, same run):** `GET /sessions/{id}/summary` → 200 with the correct cached label; `GET /search?q=...` → 200, 1 match on `short_label`; `GET /board` → 200. Cache still at 500 afterwards.

**Test files:** `tests/test_agent_viz_api.py` 68 passed; `tests/test_no_change_narration.py` 1 passed; `tests/test_llm_client.py` 108 passed (it imports the module). Prefetch tests live inside `test_agent_viz_api.py`.

**Mutation proofs** (isolated scratch copy; the PR worktree was never modified). Baseline 5 passed, restored 5 passed.

| Mutation | Caught by |
|---|---|
| `_cache_put` guard reverted to size-only | `test_cache_put_repeat_key_at_capacity_evicts_nothing` (order assertion, :1317) |
| `_cache_put` early-returns on a present key | same test (value assertion, :1318) — the assertion `e7ac88a` adds |
| `summarize_session` real-summary tail un-routed | `test_real_summary_write_site_respects_cache_max` (:1411) |
| Both disk-hit sites un-routed | both `test_disk_hit_promotion_respects_cache_max` parametrizations |
| Eviction removed from `_cache_put` | 4 of the 5 selected tests |

**Invariant:** exactly one `_cache[...] =` assignment remains across `api/`, `scripts/`, `tests/` — line 486, inside `_cache_put`. The prefetch service never writes `_cache` directly.

### Issues Found

None. All four acceptance criteria hold under execution against the real code paths.

### Holistic Assessment

The system is correct as a whole after this change, and the acceptance criteria are met at the real cap through both promotion sites and through the real HTTP handler. One observed consequence of the bound, already recorded as an operator follow-up and not a defect: with 600 nodes and a cap of 500, tick 3 still returned all 600 labels but re-read the 100 evicted entries from disk each tick. No label is lost; the cost is extra SQLite reads once snapshot fan-out exceeds 500 nodes, against ~224 today.

nbramia:
## Docs Compliance Gate Round 1 — Consolidated Review & Referee Decisions

**Reviewer:** docs (Opus). **Head reviewed:** e7ac88a.

**Code-review loop status:** converged after round 2 — round 2 surfaced no reachable production defect (only test-assertion hardening and a docstring clause, all applied in e7ac88a). A round-3 general + hostile pass had already been dispatched before the convergence call and both returned clean (Action Required / Recommended: none; hostile additionally re-ran every production mutation and confirmed each is caught by exactly the intended test, 68/68 ×3 with no flakes); they are recorded here as informational, not as a loop round.

### Reviewer Findings

**Action Required:** none.
**Recommended:**
- `docs/specs/technical/agent-viz.md:328` — the "Snapshot caching" section enumerates each cache with its key and bound (`_snapshot_cache`: keyed by `(projects_dir, lookback_days)`, TTL 30s; `_label_cache`: keyed by session id, capped at 500 entries), but the third cache's three paragraphs never state the in-process half's size bound or eviction policy; the only bound named for it is `_FAILURE_FALLBACK_TTL_SECONDS`, which covers the error-fallback path alone, so a reader is left to infer the in-process cache is otherwise unbounded. This PR is what makes the declared `_CACHE_MAX` hold at every write site, and these paragraphs already document private helpers by name (`_cache_if_terminal`, `_is_frozen`, `_NO_CONTENT_SUMMARY`). Smallest fix: one clause appended to the paragraph at line 328, e.g. "The in-process half is keyed by session id and capped at `_CACHE_MAX` (500) entries with FIFO eviction of the oldest-inserted key; every write goes through `_cache_put`, so the bound holds on the real-summary, terminal-fallback, and disk-hit promotion paths alike." Present-tense, no issue citation, narration-clean.
**Minor:**
- `api/services/agent_viz_summary.py:48-56` — the module-level comment above `_cache` does not mention the cap or the single write funnel; `_cache_put`'s docstring covers both, so pointer-only.
- `docs/specs/standards/python-conventions.md:163` lists `agent_viz_summary.py` among stores still using the bare `sqlite3.connect` form, to be left alone "unless a change already touches that file"; this PR touches the file but no SQLite call site, so nothing goes stale and no migration is implied.
**Summary:** no documentation is stale or contradicted; the one substantive gap is an omission — the caching section documents every other cache's bound but not this one's, and this is the PR that makes that bound real. `tests/test_no_change_narration.py` passes on e7ac88a.

### Referee Decisions

| # | Finding | Reviewer Severity | Concern | Decision | Reasoning / smallest remedy |
|---|---------|-------------------|---------|----------|-----------------------------|
| 1 | `agent-viz.md` caching section omits the in-process summary cache's bound/eviction while stating every other cache's | Recommended | Valid — the section's own pattern is "each cache with its key and bound"; the omission misleads toward "unbounded" | **Accept** | One sentence appended to the paragraph at `docs/specs/technical/agent-viz.md:328`, present tense, no issue number; the narration gate must stay green. |
| 2 | Module-level `_cache` comment does not mention the cap | Minor | Valid but covered by `_cache_put`'s docstring two screens below | **Reject** | Duplicating the docstring in the declaration comment adds nothing; the helper is the single write path and documents the rule where it is enforced. |
| 3 | `python-conventions.md:163` sqlite note | Minor | n/a — reviewer confirmed nothing goes stale | **Record** | No action. |

nbramia:
## Docs Compliance Gate Round 2 — Consolidated Review & Referee Decisions

**Reviewer:** docs (Opus). **Head reviewed:** 46e2f35 (one sentence appended at `docs/specs/technical/agent-viz.md:328`; `Last Updated` bumped to 2026-09-05).

### Reviewer Findings

**Action Required:** none. **Recommended:** none. **Minor:** none.

**Summary (verbatim):** Round-1's single accepted finding is resolved correctly and introduced nothing new. The appended sentence at `docs/specs/technical/agent-viz.md:328` is accurate against the PR head: `_cache` is `dict[str, ...]` keyed by session id (`api/services/agent_viz_summary.py:55`), `_CACHE_MAX = 500` (line 56), and `_cache_put` (line 468) pops `next(iter(_cache))` — FIFO by insertion order — only under `if session_id not in _cache and len(_cache) >= _CACHE_MAX`, so a repeat key updates in place and evicts nothing. All four write sites route through `_cache_put` (lines 505, 552, 619, 644 — terminal/error fallback via `_cache_if_terminal`, both disk-hit promotions, and the real-summary tail); the only direct `_cache[...] =` assignment is inside `_cache_put` itself, so the paragraph's "real-summary, terminal-fallback, and disk-hit promotion paths alike" covers the full set. The sentence is present-tense with no issue numbers and none of the banned history markers, satisfying `docs/AGENTS.md` rule 8 ("Current behavior only"); `tests/test_no_change_narration.py` passes (1 passed in 3.55s). The header block's `Last Updated` is bumped to 2026-09-05 as `docs/AGENTS.md:360` requires, `Status`/`Owner` are unchanged, and Related Documents is untouched with no new cross-links needing a back-link. No other doc references `_CACHE_MAX`, `_cache_put`, or the summary cache's size behavior, so nothing else is stale or contradicted; the rejected round-1 items are not reopened. Documentation is accurate and complete.

### Referee Decisions

| # | Finding | Reviewer Severity | Concern | Decision | Reasoning / smallest remedy |
|---|---------|-------------------|---------|----------|-----------------------------|
| — | (no findings) | — | — | — | Orchestrator independently re-ran `tests/test_no_change_narration.py` on 46e2f35: 1 passed. |

**Result:** no actionable findings — proceeding to verification. (Verification already ran on e7ac88a and returned PASS with real-router evidence; the only delta since is this docs-only commit, for which end-to-end verification is N/A.)

nbramia:
## Merge Readiness — PR Standards Check & Exact-Head Evidence

### PR Standards Check

| # | Check | Result | Notes |
|---|-------|--------|-------|
| 1 | Branch naming | PASS | `fix/summary-cache-max-promotion` |
| 2 | PR title | WARN | Conventional `fix(agents): …`, imperative, no trailing period; 88 characters (over the 72 guideline) — kept for searchability of the issue reference |
| 3 | TL;DR | PASS | Plain-language `## TL;DR` added before the mechanism section |
| 4 | Test evidence | PASS | Acceptance-criteria → test table; mutation proofs listed; pre-push gate results recorded below |
| 5 | Commit messages | FAIL (mitigated) | One safety commit is titled `wip(agents): DeepSeek draft …` — it preserves the DeepSeek draft exactly as the worker left it (uncommitted at its token cap). Rewording it requires a force-push whose unchanged tree makes the pre-push hook fall back to a full unit run behind a 10-deep lock queue, solely to rename a commit the squash merge discards. Mitigation: the squash merge is given an explicit conventional subject and body, so nothing named `wip` reaches `feat/agents-board-polish`. |
| 6 | References | PASS | `Closes #902` — single issue, fully addressed; the base is the integration branch, so GitHub closes #902 when that branch reaches `main` (the closing keyword is carried in the squash body) |
| 7 | Altitude layering | PASS | TL;DR is identifier-free; identifiers appear from `## Summary` downward; provenance section is factual, not review chronology |

**Result: 5/7 passing, 1 warning, 1 failure (mitigated as described)**

**Scope (advisory):** cohesive — one helper, four call sites, their tests, and one documentation sentence.

### Exact-head verification evidence

```
verification-head: 46e2f35596ef8af799ed766ee06eb6c0e3b8d03c
suite-result: pass
suite-command: ./scripts/test.sh (pre-push gate plan: unit + server-free browser), executed by the repository's pre-push hook on the code head e7ac88a
suite-executions: 1 (unit: 6519 passed, 9 skipped, 747.71s; server-free browser: 428 passed, 302.55s)
suite-exit-status: 0
```

- The delta between the gated code head `e7ac88a` and the PR head `46e2f35` is `docs/specs/technical/agent-viz.md` only (2 insertions, 2 deletions — one sentence plus the `Last Updated` date); the pre-push hook classified that push as docs-only and skipped the suites (`not-required` for the delta; `tests/test_no_change_narration.py` passes on 46e2f35).
- End-to-end verification (comment above) returned **PASS** on `e7ac88a` against the real `/api/agents/snapshot` router path, including a baseline reproduction of the bug on `feat/agents-board-polish` (600 entries against a cap of 500) that the head no longer exhibits.
- Repository convention for this milestone: the pre-push gate is the authoritative suite; no additional full suite is run during merge readiness.

**VERIFIED_SHA:** 46e2f35 — merging with `--match-head-commit`.

PR #948, opened 2026-09-05
feat(crm): make the compact tone-analysis endpoint a thin persisted wrapper

### LINKED ISSUE #911: chore(crm): decide the fate of the caller-less synchronous /relationship/tone-analysis endpoint
## Problem

`POST /api/crm/relationship/tone-analysis` (the non-detailed variant) has no caller anywhere in the web app, MCP server, scripts, or tests, yet it still performs a whole-window synchronous LLM call inside the request and returns a 500 on failure. That is exactly the shape #873 removed from the detailed variant. Today it is unreachable, so it is not a live hazard, but anything that wires it up later reintroduces the old behaviour.

## Solution

Either remove the endpoint (public API change: ask first) or make it a thin wrapper over the persisted detailed analysis so both variants share the same storage, freshness, and failure semantics.

## Acceptance Criteria

- [ ] No route in the CRM router performs a synchronous whole-window LLM call inside the request handler without persistence.
- [ ] If removed, docs and the API reference no longer list it; if kept, it returns the same status markers as the detailed variant and never a 500 on LLM failure.

## Technical Context

- **Key files:** `api/routes/crm.py` (the non-detailed handler near the detailed one), `docs/specs/product/api-crm.md`.

## History

Deferred from PR #899 (2026-09-04) by agreement with the reviewer; needs an operator decision because it is a public endpoint.

nbramia:
## TL;DR

`POST /api/crm/relationship/tone-analysis` (the plain, non-detailed tone endpoint) had no caller anywhere in the app and still ran a synchronous, whole-window LLM call inside the request, returning a 500 on failure. It's now a thin wrapper over the same persisted store, chunked-LLM pipeline, and per-person lock the detailed endpoint already uses: by default it only reads what's already stored (no LLM call at all), and an explicit `compute=true` runs the same recompute pass the detailed endpoint runs, never returning a 500 for an LLM failure. A new compact "Tone" card on the person-overview page (for anyone who isn't the owner and has ever had iMessage history) uses this endpoint to show a monthly tone summary with an explicit "Analyze tone" action, so the endpoint now has a real caller with the same safety guarantees.

Closes #911

**Operator decision on #911 (2026-09-05):** keep the endpoint and give it a caller. The response model is reshaped for that role: per-month `status`, top-level `average`/`analyzed_through`, and a top-level `status` distinguishing no analyzable messages, computation in progress, and failure. The endpoint has no other consumer in the repository, the web app, scripts, or the MCP server.

## Design

- The compact endpoint and the detailed endpoint share one implementation of the freshness-check → optional-recompute → assemble pipeline (`_run_tone_analysis`). The compact endpoint adapts the shared detailed shape into a smaller one (month + combined score + status, trend, average, "analyzed through", top-level `status`). `monthly_tones` carries the same months, in the same order, as the detailed endpoint -- including a month with no stored result at all, which gets a placeholder score and `status="error"`, exactly like the detailed endpoint's own convention.
- `compute` defaults to `false`: the endpoint only ever calls the LLM when a caller explicitly asks it to, preserving the "no route performs an unbounded synchronous LLM call" invariant everywhere in the CRM router.
- The top-level `status` field names the request's own outcome: `"ok"` (at least one real score, or nothing attempted yet), `"no-messages"` (no analyzable iMessage interactions anywhere in the window), `"in-progress"` (`compute=true` only -- another request already held this person's lock and this one timed out waiting for it), `"failed"` (`compute=true` only -- the lock was acquired but nothing real was ever produced or found).
- The person-overview Tone card is gated on `person.sources` containing `'imessage'` specifically -- not on message count -- since the pipeline only ever reads `source_type="imessage"`, so a WhatsApp/Slack/phone-only person can never have anything for it to show. It loads with the cheap `compute=false` path on every page render, and only sends `compute=true` when the person clicks "Analyze tone" / the refresh control. While `status` is `"in-progress"` it re-reads with `compute=false` every 5 seconds (up to ~5 minutes) until a real score appears or the poll budget runs out, without further user action. It's independent of the Relationship page's own Tone Evolution chart, which is unchanged.

## Implementation Notes

- `api/routes/crm.py`: `_run_tone_analysis(target_id, months, refresh, compute)` now returns `(response, lock_contended)` so `analyze_relationship_tone` (the compact handler) can distinguish "another request holds the lock" from "nothing was ever attempted" and from "the LLM call itself failed". `analyze_relationship_tone_detailed` unpacks and discards the second element -- its own behavior is unchanged. `ToneDataPoint`/`ToneAnalysisResponse` carry `status="error"` placeholders and a nullable `average`/`analyzed_through`.
- `web/crm.html`: `updatePersonToneCard` gates on `person.sources`; `renderPersonToneCard` branches on the top-level `status` (no-messages / in-progress / failed / ok) before falling into the existing analyzed/not-analyzed rendering; `schedulePersonTonePoll`/`stopPersonTonePolling` implement the in-progress re-read loop; `buildPersonToneSparkline` positions points by calendar-month distance (`toneMonthToAbsolute`) so a month with no interactions leaves genuine empty space instead of being compressed out, and renders an `"error"` month as its own off-line marker distinct from a dimmed `"stale"` point on the line. The refresh button carries `aria-label="Recompute tone analysis"`; the failure notice's border now uses `var(--border)` (a defined token) instead of the undefined `--border-subtle` custom property used elsewhere in this file.
- `docs/specs/product/api-crm.md`, `crm-analytics.md`, `crm-people.md` updated for the `status` field, the iMessage-only gate, and the five card states (no-messages / not-analyzed / analyzed / in-progress / failed).

### Automated tests

```
~/.venvs/lifeos/bin/pytest \
  tests/test_tone_analysis_summary_handler.py tests/test_person_tone_card_ui_browser.py \
  tests/test_tone_analysis_handler.py \
  tests/test_route_handlers_sync.py tests/test_route_handlers_no_shared_connections.py \
  tests/test_route_handlers_mutation_lock.py tests/test_route_handlers_concurrency.py \
  tests/test_no_change_narration.py -q
```
All green (12 + 11 + 26 + static guards).

Mutation-tested the two review-mandated regression tests directly against the implementation:
- Removed the `compute` gate (`if compute and stale_months:` → `if stale_months:`) in `_run_tone_analysis`: 3 of the `TestComputeFalseNeverCallsLLM` tests in `test_tone_analysis_summary_handler.py` failed (`calls == []` assertion caught the LLM factory being invoked). Restored, green again.
- Disabled the `loadPersonToneSummary` identity guard (`if (personToneController !== controller || selectedPersonId !== personId)`) in `web/crm.html`: the new browser test `test_ignored_abort_never_lets_a_superseded_response_render_over_the_new_person` failed (person A's stale response rendered over person B's card). Restored, green again.

Full unit suite + server-free browser scope, run under the shared lock and again by the pre-push hook at push time (both green):
```
flock -w 7200 /tmp/lifeos-push.lock ./scripts/test.sh unit
# 6437 passed, 8 skipped, 326s

flock -w 7200 /tmp/lifeos-push.lock <venv>/python -m pytest -m "browser and not requires_server" tests/ -q
# 369 passed, 6832 deselected, 281s

# at push time, the pre-push hook re-ran both scopes from scratch:
python -m pytest -m "unit and not slow" tests/ -q                # 6437 passed, 8 skipped, 261s
python -m pytest -m "browser and not requires_server" tests/ -q  # 369 passed, 6832 deselected, 275s
```

### Manual verification

Private staging instance on `:8042` (`uvicorn api.main:app` directly, scoped to an isolated worktree, staging `.env` + the real staging `data/` symlink, no `ANTHROPIC_API_KEY` configured so the tone-analysis LLM path falls back to this host's real local llama-server at `:8080`). Drove it with Playwright (chromium) directly against the running server, on real staging data:

- Owner page (`/me`): `#personToneCard` hidden, zero fetches to the tone endpoint, zero console errors.
- A WhatsApp-only person (`sources` has no `'imessage'`): `#personToneCard` hidden, zero fetches, zero console errors.
- An iMessage person whose window has no interactions: card visible, `status: "no-messages"`, explainer text only, no button.
- An iMessage person, never analyzed: card visible, "No tone analysis yet..." + "Analyze tone"; clicking it drove a real `compute=true` request against the local llama-server. Each of the four ~4-month chunks logged `WARNING: Tone analysis failed for a 4-month chunk: timed out` (this host's local model under load from other concurrent sessions) and the request completed in ~270s with a 200 (no 500); the card rendered the `"failed"` state ("Analysis failed; showing what is available" + explainer + re-enabled button) correctly, zero console errors.
- Contention: fired a second `compute=true` request for the same person ~1s after the first (still mid-LLM-call); it returned `status: "in-progress"` in ~5s (the configured lock-wait timeout) while the first request continued in the background -- matching the unit-level contention test.
- `/relationship` (no partner configured in this staging config): loads unchanged, zero console errors.

Zero application console errors were observed across every page/state checked.

## Review focus (resolved)

All ten findings from the prior review round are addressed -- see the review-response PR comment for a per-finding breakdown of what changed and where.

Claude-Session: https://claude.ai/code/session_01MqMhBfLUGZ5cKuqyWe5iqD


nbramia:
## Adversarial review of #948 by Opus

**Verdict:** BLOCK

The server-side half is genuinely good: `_run_tone_analysis` is the only implementation, the compact endpoint is a faithful adapter over it, the `compute=false` cost invariant holds on real data, and the per-person lock, staleness, and never-a-500 guarantees all reproduce. Three things block: the card is shown to ~5x more people than can ever get a result and tells them the analysis "failed" when there was simply nothing to analyze; a second compute attempt while one is running renders a permanent false failure; and the single test guarding the PR's headline cost invariant is inert (mutation-proved).

1. **[BLOCKER] The Tone card's visibility gate counts messages the tone pipeline never reads, so for most people the "Analyze tone" button can only ever end in "Analysis failed".** — `web/crm.html:12515` (`const hasMessages = (person.message_count || 0) > 0;`) vs `api/routes/crm.py:5874` (`source_type="imessage"`). `message_count` is `imessage + whatsapp + phone + slack` (`api/services/person_stats.py:253`), but `_run_tone_analysis` only ever queries `source_type="imessage"`. On the staging copy of the real data: **736 people get the card; 519 have any iMessage ever; only 114 have iMessage inside the default 12-month window.** Reproduced end-to-end in Chromium against a staging instance for a WhatsApp-only person and a Slack-only person — card visible, one `compute=false` request, then clicking the button issues `compute=true`, the server returns in ~4 ms with the empty shape, and the card renders:
   `Analysis failed; showing what is available` / `No tone analysis yet for this person's messages.` / `Analyze tone`.
   Nothing failed; there is simply no iMessage. The button is permanently useless for those people and the notice is actively false. Suggested fix: gate on iMessage specifically — `person.sources` is already on the person payload with no extra request and already carries `'imessage'` (verified: a WhatsApp-only person returns `sources: ['calendar','whatsapp']`) — and separately stop treating "the window contained no analyzable messages" as a failure (the endpoint currently collapses that case and a real LLM failure into the same empty/`not-analyzed` response, so the client cannot tell them apart; consider surfacing whether the window had any analyzable months).

2. **[MAJOR] A second `compute=true` while one is already running renders a permanent false "Analysis failed", and the real result never appears without a page reload.** — `web/crm.html:12639` (`const failed = personToneWasComputeAttempt && (tones.length === 0 || tones.every(...))`) + `api/routes/crm.py:5924` (`lock.acquire(timeout=TONE_LOCK_TIMEOUT_SECONDS)`). The lock-timeout fall-through returns the same empty shape as "nothing to analyze", which the card reads as failure. Measured directly against staging with two concurrent `compute=true` requests for one person: `computeA status=200 secs=38.95 months=2`, `computeB status=200 secs=5.01 months=0` (exactly `TONE_LOCK_TIMEOUT_SECONDS`). Reproduced in the browser on a person with real iMessage history — click Analyze, then invoke it again 800 ms later:
   - after second click → `Analysis failed; showing what is available / No tone analysis yet…`
   - 60 s later (first request long since finished and persisted) → still the same wrong text
   - after `page.reload()` → `Stable Neutral · Avg 58 · through Aug 2026`
   This is reachable from a single tab (the button re-enables as soon as the *client* fetch is aborted or resolves, while the abandoned server-side job keeps the lock for minutes) and from two tabs. Suggested fix: don't treat an empty response to a `compute=true` as failure when the request was contended — either have the endpoint distinguish "another computation is in progress" from "nothing to analyze", or have the card fall back to a `compute=false` re-read after a short delay instead of latching the notice.

3. **[BLOCKER] The one test that guards the PR's headline invariant ("`compute=false` never acquires an LLM client") cannot fail.** — `tests/test_tone_analysis_summary_handler.py:87` (`_fail_if_llm_requested`). The fake raises `AssertionError`, but `get_anthropic_llm()` is called *inside* the chunk loop's broad `except Exception` (`api/routes/crm.py:5956`), which swallows it and sets `chunk_results = {}`. Mutation test: replacing `if compute and stale_months:` with `if stale_months:` (i.e. removing the compute gate entirely) leaves the file **green — 10 passed**. Verified the mechanism two ways: (a) a handler-shaped `try/except Exception` swallows the AssertionError while the factory records 1 invocation; (b) with the gate mutated away and a *recording* (non-raising) factory, the same request produces `status: 200 / monthly_tones: [] / trend: not-analyzed` — byte-identical to what every assertion in `TestComputeFalseNeverCallsLLM` checks — with `LLM FACTORY CALLS ON compute=false: 1`. Suggested fix: patch `get_anthropic_llm` with a call *recorder* that returns a stub, and assert `recorder == []` after the response, so the check lives outside the code under test's own error handling.

4. **[MAJOR] The stale-response guard the PR's own "Review focus" names is untested.** — `web/crm.html:12560` and `:12566`. Replacing `if (personToneController !== controller || selectedPersonId !== personId)` with `if (false)` leaves all 7 browser tests green. `TestPersonSwitchCancelsStaleRequest` only exercises the *abort* path (mutating the abort away is correctly caught), so the identity/person check — the defence for a response that resolves before the abort lands — has no coverage. Suggested fix: add a case where the superseded response resolves normally (mock fetch that ignores the abort signal for person A) and assert person B's card content survives.

5. **[MINOR] `--border-subtle` is not defined anywhere in `web/`, so the failure notice's border silently does not render.** — `web/crm.html:12642`. `crm.html` defines only `--border`; `--border-subtle` had 5 existing uses on `main` and this PR adds a 6th. An undefined custom property makes the whole `border` shorthand invalid at computed-value time. Pre-existing pattern, but worth using `var(--border)` in the new code rather than propagating it.

6. **[MINOR] The refresh control's accessible name is the glyph, not its label.** — `web/crm.html:12659`. Measured in the browser: `{btn_tag: "BUTTON", btn_text: "↻", btn_title: "Recompute tone analysis", btn_aria_label: null}`. `title` is only an accessible-name fallback when the element has no text content, so screen readers announce "↻". Add `aria-label="Recompute tone analysis"`. (The element is a real focusable `<button>`, which is right.)

7. **[MINOR] Docs and the HTML comment misdescribe the gate on both sides.** — `docs/specs/product/crm-people.md` ("has message (iMessage/WhatsApp) interactions") and `web/crm.html:8242`. The count also includes Slack and phone, and the analysis is iMessage-only — neither doc states that the card only ever has data to show for iMessage history. Fix alongside finding 1 and state the actual invariant.

8. **[MINOR] The public response model was reshaped without a recorded operator decision, and one acceptance criterion is only partially met.** — `api/routes/crm.py:5336-5367`. `ToneDataPoint` drops `tone` and `sample_count` and gains `status`; `ToneAnalysisResponse` gains `average`/`analyzed_through` and changes the `trend` vocabulary. AGENTS.md § Boundaries lists "Public API changes (new or modified endpoints)" as ask-first, and issue #911 explicitly says the endpoint needs an operator decision because it is public (the issue has 0 comments). Its second criterion — "if kept, it returns the same status markers as the detailed variant" — is not literally met: the compact endpoint never emits `status="error"`, it drops those months. The divergence is deliberate and documented and nothing else in the repo consumes the endpoint (verified: no caller in `api/`, `web/`, `scripts/`, or `mcp_server.py` besides the new card), so the risk is low — but it should be the operator's call, not the PR's.

9. **[MINOR] The sparkline hides gaps.** — `web/crm.html:12607`. Months with no stored result are dropped, then the remaining points are laid out at equal x-spacing, so a missing month reads as continuous. Observed on real data: a person whose stored months are 2025-12, 2026-01, 2026-02, **2026-04**, 2026-05, 2026-06 renders as six evenly spaced points with no indication that 2026-03 is absent. Consider positioning by month index rather than array index.

10. **[MINOR] New small text sits below WCAG AA contrast, and the empty response carries a fabricated average.** — `web/crm.html:12662` renders `Avg 55 · through Jun 2026` at `0.7rem` in `var(--text-muted)`; measured `rgb(106,106,138)` on `rgb(22,33,62)` = **3.06:1** (AA wants 4.5:1 for non-large text). It uses an existing token, so this is the palette rather than the PR, but it is new small text. Separately, an empty response returns `average: 50.0` — a made-up number in a public payload; `null` would be honest (the card never displays it, so this is cosmetic).

### What I verified and how

- `git fetch origin feat/person-tone-summary && git checkout -B review-948 origin/feat/person-tone-summary` → head `63b6fc3`, base `main`, 7 files, `web/crm.html` diff is **purely additive (0 removed/modified lines, 174 added)**.
- **Cost invariant, real data, LLM factory patched to raise:** staging on :8041 with the real `data/` copy and `get_anthropic_llm` wrapped in a tripwire that logs a stack trace and raises. Six `compute=false` requests across the heaviest person in the DB (76,705 messages / 20,297 recent iMessage), a WhatsApp-only person, a Slack-only person, the owner, and a person with stored results → all **HTTP 200 in 0.005–0.040 s, tripwire hits: 0**. Same for `person_id` omitted with no partner configured, and for an unknown id → both `200` with `monthly_tones: []`, `trend: "not-analyzed"`, `analyzed_through: null`. Also `compute=true` on the WhatsApp-only and Slack-only people → `200` in ~4 ms, tripwire hits: 0.
- **One implementation:** `analyze_relationship_tone` and `analyze_relationship_tone_detailed` both delegate to `_run_tone_analysis`; no second pipeline exists (grep for the endpoint across `api/`, `web/`, `scripts/`, `mcp_server.py`). Empirically, for two people, every month present in both responses matched exactly on `score == combined_score` and on `status` (0 mismatches), and the months the compact endpoint drops are exactly the `status="error"` ones.
- **Rows written by both endpoints:** ran `tone-analysis-detailed` for one person (9 rows persisted) and `tone-analysis?compute=true` for another (1 row persisted), then compared `tone_analysis_results` — same table, same `interaction_count`, same result keys `['combined_score','partner_sample_count','partner_score','user_sample_count','user_score']`, same `model` string. Identical write path.
- **Lock reuse and read-during-compute:** two concurrent `compute=true` for one person → `computeA 200 in 38.95 s (2 months)`, `computeB 200 in 5.01 s (0 months)` = the 5 s `TONE_LOCK_TIMEOUT_SECONDS` fall-through. A `compute=false` read for the same person while the compute held the lock returned in **0.01 s** — never blocked.
- **Never a 500:** `grep -c "status=5" staging.log` → `0` across the whole session, including a 34.8 s and an 88.9 s real LLM run and both lock-timeout fall-throughs.
- **Server responsiveness during a compute:** polled `/health` once a second for the duration of a 34.8 s `compute=true` → 34 samples, all 200, worst latency 0.0 s. (Earlier `/health` stalls I saw were the overview page's own fact-extraction LLM job, not this endpoint.)
- **Browser, real staging data, Chromium:** owner page `/me` → card hidden, **zero** tone requests, zero console errors. WhatsApp-only and Slack-only → card visible (finding 1), exactly 1 request on load and exactly 1 more on click. Person with stored results → `Stable Neutral / ↻ / Avg 55 · through Jun 2026`, sparkline present, 1 request, zero console errors.
- **Rapid person switching (the specific ask):** three people selected within 200 ms while a tone request was pending, with a fetch shim that honours `AbortSignal`. Instrumented log: `FETCH A → ABORT A → FETCH B → ABORT B → FETCH C → render C months=6`; `selectedPersonId` = C, card content = C's, exactly one request per person, no leak, no stale render, zero console errors. Clean.
- **Relationship page unchanged:** served the PR's `crm.html` and `origin/main`'s `crm.html` from identical stubbed harnesses, loaded `/relationship` with a synthetic configured partner, and diffed. Request lists **identical** (7 requests each, including `/relationship/tone-analysis-detailed`); `toneTimelineViz`, `toneRefreshBtn`, `toneViewSelect` outerHTML **byte-identical** (2624 / 292 / 690 chars); zero console errors on both. No identifier collisions (none of the new `personTone*` symbols exist on `main`).
- **Layout shift:** `.quick-facts-section` bounding box measured before and after the tone card's content renders — **identical** (`x:344 y:888.15625 w:448 h:790.140625`); the card lives in the right grid column.
- **Mutation testing (10 mutants):** caught — owner check dropped, `message_count` gate dropped, Analyze button removed, computing state not disabled, failure notice suppressed, sparkline removed, abort removed, `analyzed_through` label dropped. **Survived — the compute gate (finding 3) and the stale-response guard (finding 4).**
- **Tests:** `pytest tests/test_tone_analysis_summary_handler.py tests/test_tone_analysis_handler.py tests/test_tone_analysis_store.py tests/test_no_change_narration.py tests/test_route_handlers_sync.py tests/test_route_handlers_no_shared_connections.py tests/test_route_handlers_mutation_lock.py -q` → **67 passed**. `tests/test_person_tone_card_ui_browser.py -q` → **7 passed**. `tests/test_route_handlers_concurrency.py tests/test_anthropic_api_guard.py tests/test_no_change_narration.py -q` → **8 passed**. No test deleted, skipped, or weakened by this PR.
- **Narration:** `test_no_change_narration.py` passes; scanning added lines for `previously / now / no longer / used to / improved / issue-or-PR-as-history` turns up nothing — the "unlike the detailed endpoint" phrasing contrasts two current endpoints, not two versions.
- **Privacy:** tests and fixtures use obviously synthetic ids and names only; the failure log line records `len(chunk)` and the exception, no prompt or message content; the staging log's four tone lines contain no names, no message text, and the route template rather than the `person_id`-bearing path; PR body carries no personal data. Clean.
- **Docs:** all three anchors used by the new prose resolve (`api-crm.md#post-apicrmrelationshiptone-analysis`, `crm-analytics.md#tone-analysis-apis`, `crm-analytics.md#communication-visualizations`); grepped `docs/` for stale descriptions of the compact endpoint elsewhere — none. Except for finding 7, the doc text matches observed behaviour.

### Not verified

- The full unit suite and the full `browser and not requires_server` scope — I ran the targeted files the change touches plus every static/route guard, per my standing brief (targeted tests only). The PR reports both scopes green twice.
- Behaviour with an Anthropic key present: staging is keyless, so every `compute=true` above exercised the ADR-025 local-llama fallback, not the Anthropic path.
- `months` values other than 12 (the card never sends one) beyond the `months=1/2/4` probes used to keep LLM runs cheap.
- Whether the reshaped public response breaks any consumer outside this repo — I could only confirm there is none inside it.

nbramia:
## Review response (all 10 findings addressed)

Commit 7319dd9 (pushed on top of 63b6fc3) resolves every finding from the Opus adversarial review. Per finding:

**1. [BLOCKER] Tone card visibility gate counted messages the pipeline never reads.**
Fixed. `updatePersonToneCard` now gates on `Array.isArray(person.sources) && person.sources.includes('imessage')` (`web/crm.html:12541`) instead of `message_count`. The endpoint also gained a top-level `status` field (`ok` / `no-messages` / `in-progress` / `failed`, `api/routes/crm.py:5370`) so "the window has no analyzable iMessage" (`status: "no-messages"`) is never conflated with a real failure — the card renders distinct text with no button for `no-messages` (`web/crm.html:12769`), separate from the not-analyzed state's button. Verified live against real staging data: a WhatsApp-only person now gets no card at all, and an iMessage person with an empty window shows "No iMessage in the last 12 months." with no button.

**2. [MAJOR] A second concurrent `compute=true` rendered a permanent false "Analysis failed".**
Fixed, both ways the review suggested. The endpoint now distinguishes contention from failure via `status: "in-progress"` (returned when this request's own lock-acquire attempt timed out because another request already held it, `api/routes/crm.py:5561-5578`), and the card polls `compute=false` every 5s for up to ~5 minutes while in that state (`schedulePersonTonePoll`/`stopPersonTonePolling`, `web/crm.html:12751-12767`) rather than latching a failure notice. New unit test `TestContention` (`tests/test_tone_analysis_summary_handler.py`) drives two real concurrent `compute=true` requests against one person with a slow fake LLM client and asserts the second returns `in-progress` while a later `compute=false` read sees the first's persisted result. Verified live: fired a second `compute=true` ~1s after the first (still mid-call) — it returned `status: "in-progress"` in ~5s (the configured lock timeout) while the first continued in the background.

**3. [BLOCKER] The compute=false cost-invariant test was inert (mutation-proved).**
Fixed. `_llm_call_recorder` (`tests/test_tone_analysis_summary_handler.py`) replaces the raising fake with a factory that records every invocation and returns a working stub, and every `TestComputeFalseNeverCallsLLM`/`TestNoMessagesStatus`/`TestPartnerDefault`/`TestUnknownPersonId` test now asserts `calls == []` from outside the handler's own `except Exception`, so a swallowed assertion can no longer hide. **Mutation-tested**: reverted `if compute and stale_months:` to `if stale_months:` (removing the compute gate) — 3 tests failed (`assert [1] == []`) with the LLM factory demonstrably invoked. Restored, green again (12/12).

**4. [MAJOR] The stale-response identity guard was untested.**
Fixed. New test `test_ignored_abort_never_lets_a_superseded_response_render_over_the_new_person` (`tests/test_person_tone_card_ui_browser.py`) extends the fetch mock with an `ignoreAbort` rule that still records the abort request but resolves person A's response anyway after its delay, then switches to person B and asserts B's card is never overwritten by A's stale data. **Mutation-tested**: replaced `if (personToneController !== controller || selectedPersonId !== personId)` with `if (false)` in `loadPersonToneSummary` (`web/crm.html:12591`) — the new test failed, rendering A's "stable positive / through Jan 2026" over B's card. Restored, green again (11/11 in the file).

**5. [MINOR] `--border-subtle` is undefined, silently breaking the failure notice's border.**
Fixed. The new failure-notice `<div>` uses `var(--border)` (`web/crm.html:12778`), a token that is actually defined in `:root`. (The four pre-existing uses of `--border-subtle` elsewhere in the file are untouched, per "surgical changes" — not part of this PR's scope.)

**6. [MINOR] The refresh button's accessible name was the glyph, not its label.**
Fixed: `aria-label="Recompute tone analysis"` added alongside the existing `title` (`web/crm.html:12796`).

**7. [MINOR] Docs and the HTML comment misdescribed the gate.**
Fixed. `docs/specs/product/crm-people.md`, `api-crm.md`, and `crm-analytics.md` now state the iMessage-only invariant explicitly, and the HTML comment above `#personToneCard` (`web/crm.html:8238`) says the same.

**8. [MINOR] Reshaped public response model with no recorded operator decision; `status="error"` criterion only partially met.**
The operator decision is recorded on issue #911 (2026-09-05: keep the endpoint, give it a caller, reshape the response for that role) and is now also quoted in the PR body. The second half is fixed in code: `monthly_tones` now carries the same months, in the same order, as the detailed endpoint, including a `status="error"` placeholder for a month nothing was ever stored for (`api/routes/crm.py:5585-5588` in the assembly, response shape in `ToneDataPoint`/`ToneAnalysisResponse` at `api/routes/crm.py:5358-5372`) — full parity with the detailed endpoint's own convention, not a divergence.

**9. [MINOR] The sparkline hid gaps (equal x-spacing regardless of calendar distance).**
Fixed. `buildPersonToneSparkline` now positions every point by absolute calendar-month distance (`toneMonthToAbsolute`, `web/crm.html:12669`) rather than array index, so a month with no interactions leaves genuine empty space and the connecting line breaks across it. An `"error"` month gets its own small off-line marker at its correct calendar position rather than being silently dropped.

**10. [MINOR] Sub-AA-contrast text and a fabricated `average: 50.0` on an empty response.**
Both fixed. `average` is now `null` (not a fabricated `50.0`) whenever no month has a real score (`api/routes/crm.py:5567` for the no-messages case, `5590` for the general case), and the card renders an em dash for that case (`web/crm.html:12791`). For contrast: measured `var(--text-secondary)` (`#a0a0a0`) against the card's `var(--bg-secondary)` background (`#16213e`) — **6.07:1**, comfortably over the 4.5:1 AA threshold for normal text — so the small "Avg … · through …" line now uses that token (`web/crm.html:12792`) instead of the failing `var(--text-muted)` (3.06:1). Bumped to `0.8rem` alongside the token swap for extra headroom, though the token change alone already clears AA.

### Verification
- Targeted suite: `tests/test_tone_analysis_summary_handler.py` (12), `tests/test_person_tone_card_ui_browser.py` (11), `tests/test_tone_analysis_handler.py` (26), plus all four static guards (`test_route_handlers_sync.py`, `test_route_handlers_no_shared_connections.py`, `test_route_handlers_mutation_lock.py`, `test_route_handlers_concurrency.py`) and `test_no_change_narration.py` — all green.
- Mutation-tested findings 3 and 4 directly (see above) — both catch the regression, both restored clean.
- Full unit suite (6437 passed, 8 skipped) and the server-free browser scope (369 passed, 6832 deselected) green both standalone under the shared lock and again at push time via the pre-push hook.
- Manual Playwright verification against a private staging instance on real data: iMessage not-analyzed → Analyze → graceful local-model-fallback failure handling (LLM timeouts on this shared host, no 500), a WhatsApp-only person (no card), the owner (no card), a live two-request contention case (`in-progress` then real result), and the Relationship page (unchanged, zero console errors) — see PR body for the full write-up.

Claude-Session: https://claude.ai/code/session_01MqMhBfLUGZ5cKuqyWe5iqD

nbramia:
## Adversarial review of #948 by Opus

All ten findings from the previous round are genuinely fixed, and I mutation-proved each one where the fix is guarded by code that a test could reach. The two blockers are dead: the card now gates on `person.sources` containing `'imessage'` (521 of 14,422 staging people, down from the 736 the `message_count` gate admitted), and the `compute=false` cost invariant is now enforced by a call *recorder* whose assertion lives outside the handler's own `except Exception` — removing the gate fails 3 tests with `assert [1] == []`, and a raising tripwire on `get_anthropic_llm` in a real staging process recorded **0 invocations** across nine `compute=false`/non-iMessage requests. The contention case now returns `status="in-progress"` at exactly the 5 s lock timeout on real data, and the card polls instead of latching a failure. What's left are minors: three of the cosmetic fixes (aria-label, sparkline gap positioning, the null-average em dash) have no regression test — I mutated each away and all 11 browser tests stayed green — and a contended recompute against a person who already has scored months reports `ok` rather than `in-progress`, so the card shows the partial data and stops polling. None of that should block.

1. **[MINOR] Three of the round-1 fixes have no regression test — mutation-proved.** — `web/crm.html:12796` (`aria-label`), `:12669`/`buildPersonToneSparkline` (calendar positioning), `:12791` (`data.average != null ? … : '—'`). Each is correct in the code and I verified all three live in the browser (below), but each survives mutation: removing `aria-label="Recompute tone analysis"`, replacing `monthlyTones.map(t => toneMonthToAbsolute(t.month))` with `map((t, i) => i)`, and replacing the em-dash fallback with `Math.round(data.average || 50)` each leave `tests/test_person_tone_card_ui_browser.py` at **11 passed**. Contrast (finding 10's other half) likewise has no automated floor. Suggested fix: one browser assertion on `get_attribute("aria-label")`, one on the sparkline's `path`/`circle cx` values for a month-gapped `monthly_tones` fixture, and one on the card text for `average: null`. Cheap, and they're the only thing standing between these and a future refactor. (Findings 3 and 4 are properly covered — see below.)

2. **[MINOR] A contended `compute=true` for a person who already has scored months returns `ok`, not `in-progress`, so the card renders the partial result and never picks up the running computation's new months.** — `api/routes/crm.py:5578-5585` (`if scored: status = "ok"` is checked before `elif lock_contended:`). The `in-progress` branch is only reachable when *nothing* in the window has a real score. In the common case — a person with a few analyzed months whose remaining months are being recomputed by another tab — the second request falls through to `ok`, `renderPersonToneCard` takes the ordinary analyzed path, `personToneInProgressMode` is never set, and the months the other request is about to persist don't appear until the next page load. Not a false failure (the round-1 defect is gone) and not a wrong number, just a missed poll. Suggested fix: if `lock_contended` and any month is still `"error"`/`"stale"`, keep the poll running alongside the partial render.

3. **[MINOR] `personToneWasComputeAttempt` survives the in-progress poll, so a poll that lands all-`stale` months shows "Analysis failed".** — `web/crm.html:12775` (`personToneWasComputeAttempt && scored.length > 0 && scored.every(t => t.status === 'stale')`). The flag is only cleared in `updatePersonToneCard` (person switch/reload), so after a `compute=true` → `in-progress` → poll sequence where the other request's result is itself stale, the card reports failure for a computation that is merely stale, not failed. Narrow, and strictly better than the round-1 behaviour.

4. **[NIT] A single scored month headlines the card as "Insufficient Data" next to a concrete average.** — `_derive_trend` returns `insufficient-data` for `len(scores) < 2` (`api/routes/crm.py:5652`), so the analyzed state renders `Insufficient Data / Avg 73 · through Sep 2026`. Observed on real staging data for a person with one scored month among eleven `"error"` months. Honest — the sparkline does draw the eleven off-line markers — but the trend label and the average read as contradicting each other. Consider suppressing the trend label (or the average) when fewer than two months are scored.

5. **[NIT] The in-progress poll keeps firing after the user navigates away from the person detail.** — `web/crm.html:12751` (`schedulePersonTonePoll`'s guard is `selectedPersonId !== personId`). `selectedPersonId` isn't cleared by the Me/Family/Birthdays nav, only by `selectPerson`/hide, so leaving a person page mid-poll leaves up to ~5 minutes of 5-second `compute=false` re-reads running against a hidden card. Each is 1–40 ms server-side and writes into a hidden element, so the cost is real but tiny.

6. **[NIT] `api/routes/crm_models/models.py:651-663` still carries the old `ToneDataPoint`/`ToneAnalysisResponse` (`tone`, `sample_count`, no `status`), which this PR pushes further out of sync.** Leaving it is the surgical choice — `docs/specs/technical/architecture.md:243` documents the package as not wired into the app, and only `is_family_member` is contract-bound — so this is a pointer, not a request to change it.

### What I verified and how

- Fetched `feat/person-tone-summary`, checked out **7319dd9** detached; base `main` at `0864123`. `web/crm.html`'s diff against `main` is **purely additive** (0 removed or modified lines, 329 added). Read the full diff, both new test files, `_run_tone_analysis` and both handlers end to end.
- **Finding 1 — iMessage-only gate.** Code gates on `Array.isArray(person.sources) && person.sources.includes('imessage')`. Mutating it back to `(person.message_count || 0) > 0` **fails** `test_hidden_for_a_whatsapp_only_person_no_request_at_all`; mutating `if (data.status === 'no-messages')` to `if (false)` **fails** the no-messages test. On the staging copy of the real data: **521** people carry `imessage` (vs 736 under the old `message_count` gate). In Chromium against a private staging instance on :8043, on real data: a strict WhatsApp-only person, a strict Slack-only person (neither has `imessage`), and the owner's `/me` → **card hidden, zero tone requests, zero console errors** in all three.
- **Finding 2 — contention, on real data.** Two genuinely concurrent `compute=true` requests for one person against the live local llama-server: `SECOND → 200 in 5.007 s, status="in-progress"` (exactly `TONE_LOCK_TIMEOUT_SECONDS = 5`), `FIRST → 200 in 45.907 s, status="ok", 1 scored month`, and a `compute=false` read fired *while* the compute held the lock returned in **0.005 s** — never blocked. A later `compute=false` read returned the first request's persisted month (`status="ok"`, `average=72.5`, `analyzed_through=2026-09`). Zero 5xx in the staging log for the whole session. Mutating `lock_contended = not lock_acquired` to `= False` **fails** `TestContention`; mutating `status = "failed"` to `"ok"` **fails** the failure test. In the browser, with `in-progress` served for `compute=true`: card shows "Another analysis is already running… " at 1.2 s, **still no "Analysis failed"** after the first 5 s poll, and resolves to `Stable Positive / Avg 62 / through Jan 2026` on a later poll with **no further user action** — 4 tone requests total (initial load, `compute=true`, two 5 s polls), zero console errors. Mutating the `data.status === 'in-progress'` branch away **fails** that test.
- **Finding 3 — `compute=false` never reaches the LLM.** Mutating `if compute and stale_months:` back to `if stale_months:` **fails 3 tests** in `TestComputeFalseNeverCallsLLM` with `assert [1] == []` — the recorder assertion genuinely bites where the round-1 raising fake did not. Independently, on real data with `get_anthropic_llm` wrapped in a **raising** tripwire in the staging process: the heaviest iMessage person in the DB (76,705 messages, 12 window months), a person with 6 stored months, a person with 9 stored months, a WhatsApp-only person, a Slack-only person, the owner, and an unknown id — plus `compute=true` on the two non-iMessage people — → **all HTTP 200 in 0.001–0.034 s, tripwire invocations: 0**.
- **Finding 4 — superseded-response identity guard.** Replacing `if (personToneController !== controller || selectedPersonId !== personId)` with `if (false)` **fails** `test_ignored_abort_never_lets_a_superseded_response_render_over_the_new_person`. Live rapid switching on real data (three `selectPerson` calls 60 ms apart while a tone request was pending) left the final person's card with the final person's data, correct `selectedPersonId`, zero console errors.
- **Findings 5–7, 9, 10 — measured in the browser on real data.** Refresh control: `aria-label="Recompute tone analysis"`, computed accessible name `"Recompute tone analysis"` (was the glyph). Failure notice uses `var(--border)`; the five pre-existing `--border-subtle` uses in the file are untouched. Small text: computed `rgb(160,160,160)` on `rgb(22,33,62)` at `12.8px` → **6.078:1**, clears AA's 4.5:1. `average` is `null` with no scored months (mutating it to a fabricated `50.0` **fails 2 tests**) and the card renders an em dash. Sparkline for a person whose stored months are 2025-12, 2026-01, 2026-02, **2026-04**, 2026-05, 2026-06: measured path `M 0.0 18.0 L 36.7 11.0 L 73.3 18.0 **M** 146.7 18.0 L 183.3 17.3 L 220.0 18.0` — points at their true calendar offsets, the missing 2026-03 leaves a genuine two-step gap, and the line breaks across it. HTML comment above `#personToneCard` and `crm-people.md` / `api-crm.md` / `crm-analytics.md` all state the iMessage-only invariant; every cross-doc anchor the new prose uses resolves.
- **Finding 8 — parity with the detailed endpoint, on real data.** For three people with stored results, the compact and detailed responses returned **identical month lists, identical per-month `status` values, and identical scores** (`score == combined_score`), 0 mismatches. `"error"` placeholders are really emitted: one person's response is 11 × `("error", 50.0)` plus one real `(None, 72.5)`. Mutating the assembly to drop `"error"` months **fails 3 tests**; mutating `analyzed_through` to use the last month regardless of status **fails 2**. The operator decision is on issue #911 (2026-09-05) and quoted in the PR body. No consumer of the compact endpoint exists anywhere in `api/`, `web/`, `scripts/`, `mcp_server.py`, or `docs/` besides the new card.
- **Static guards and narration:** `tests/test_route_handlers_sync.py`, `test_route_handlers_no_shared_connections.py`, `test_route_handlers_mutation_lock.py`, `test_route_handlers_concurrency.py`, `test_no_change_narration.py` → all pass (64 passed with the tone files). Scanned every added line for change narration (previously/now/no longer/used to/review rounds/issue-or-PR numbers as history) — nothing.
- **Relationship page unchanged.** Served the PR's `crm.html` and `origin/main`'s `crm.html` from the **same** staging instance via document interception, twice: with no partner configured (4 API requests each) and with a partner stubbed into `/api/crm/config` so the page actually fetches `tone-analysis-detailed` (7 API requests each). Request lists **identical** both times; `toneTimelineViz`, `toneRefreshBtn`, `toneViewSelect` `outerHTML` **byte-identical** in all four loads; **zero console errors and zero failed requests** on every load. No `personTone*` identifier exists on `main`.
- **Console errors:** zero application console errors and zero failed requests across every state — not-analyzed (real data), analyzing, analyzed (real data), no-messages, in-progress plus two polls, failed, owner `/me`, WhatsApp-only, Slack-only, rapid person switching, and `/relationship` × 4.
- **Tests:** `tests/test_tone_analysis_summary_handler.py` (12), `tests/test_person_tone_card_ui_browser.py` (11), `tests/test_tone_analysis_handler.py` (26), `tests/test_tone_analysis_store.py` (7), the four static guards, `test_no_change_narration.py` → all green. **13 mutants run in total; 10 killed, 3 survived** (the three in finding 1 above). No test was deleted, skipped, weakened, or rewritten by this PR.
- **Privacy:** the new tests use obviously synthetic ids and names only; the failure log line records `len(chunk)` and the exception, never prompt or message content; nothing personal in the PR body, the docs, or this comment.

### Not verified

- The full unit suite and the full `browser and not requires_server` scope — targeted files only, per my standing brief. The PR reports both green twice (standalone and at push time).
- Behaviour with an `ANTHROPIC_API_KEY` present: staging is keyless, so every real `compute=true` above exercised the ADR-025 local-llama fallback.
- `tests/test_anthropic_api_guard.py::test_guard_blocks_async_anthropic_call` fails when it runs in the same process after **any** Playwright browser test file — I reproduced it with `tests/test_voice_mic_block_ui_browser.py` from `main`, so it is a pre-existing cross-file interaction, not this PR, and the real gate runs the unit and browser scopes separately.
- `months` values other than 12 beyond the `months=1`/`months=2` probes I used to keep real LLM runs cheap; the card never sends one.
- **Side effect to note:** the real-data contention probe persisted one `tone_analysis_results` row (person `cb93e7bd…`, month `2026-09`) into the shared staging `crm.db` — the same write the card itself makes. Nothing else was written; no merges, hides, or syncs were run.

Claude-Session: https://claude.ai/code/session_01MqMhBfLUGZ5cKuqyWe5iqD

Issue #885, opened 2026-09-04
feat(agents): delete cards from the drawer with confirmation and kill for claimed tasks


nbramia:
## Problem

A card cannot be deleted from the board. Removing a task or a schedule means editing the vault or calling the API. When a card is claimed by an agent, deleting it without stopping the agent would leave a session running against a task that no longer exists.

## Solution

Add Delete to the drawer for task and scheduled cards, behind a confirmation. For a task card with a live agent session, the confirmation states that the session and its subagents will be killed first, and confirming performs the kill cascade before the delete.

## Acceptance Criteria

- [ ] The drawer shows a Delete action for task cards and scheduled cards; clicking it opens a confirmation naming the card.
- [ ] For a task card tagged running or blocked with a live session, the confirmation text includes "kill the running session and its subagents", and confirming calls the kill endpoint, waits for it, then `DELETE /api/tasks/{id}`; the card disappears and the vault line and its notes body are removed.
- [ ] For an unclaimed task card, confirming calls `DELETE /api/tasks/{id}` only; for a scheduled card, `DELETE /api/scheduler/{id}`.
- [ ] If the kill fails, no delete happens and the error is shown; cancelling the confirmation changes nothing.
- [ ] The server-free browser test covers all three paths and the cancel.

## Verification

- **Automated tests:** board browser test additions with stubbed kill and delete endpoints.
- **Manual checks:** delete a scratch card on the live board and confirm the vault.
- **Existing test suite:** green.

## Scope

**In scope:** the Delete action, confirmation, kill-then-delete sequencing, spec update.

**Out of scope:** Cancel semantics (#881); bulk delete.

## Technical Context

- **Key files:** `web/agents/board.js` (`renderDrawerActions` ~:694–746; the Kill call already there), `api/routes/tasks.py` (`DELETE /{task_id}` removes line plus notes body since #854), `api/routes/scheduler.py` (`DELETE /{id}`), `api/routes/agents.py` kill endpoint (cascades descendants).
- **Patterns to follow:** the graph panel's kill confirmation modal in `web/agents/graph.js` (~:41–61) for the confirmation and cascade preview.
- **Constraints:** delete only after a successful kill; never delete a Review card without the same confirmation.

## References

- `docs/specs/product/agent-viz.md` §Drawer; #881 for the movement rules and Cancel
- Follows #850 (PR #856)

nbramia:
[...]

Issue #912, opened 2026-09-04
feat(agents): show a card's assigned host on the board, and align the host filter with what it filters


nbramia:
## Problem

The board gives no sign of which machine a card is assigned to, and the host filter added by #890 cannot see assigned hosts at all.

Two different fields carry a host on the `/agents` board, and only one of them is surfaced:

- `fields.host` — the **assignment**: where the card *will* run. Written by the drawer's host dropdown (#883).
- `card.session.host` — the **observation**: where a session *did* run. Read by `board.js`'s card chips and by `#board-filter-host`'s option list.

So a card the operator has just assigned to `studio-box` shows only its engine and effort chips, with nothing indicating the host, and `studio-box` never appears in the "all hosts" filter dropdown until a session has actually run there. On a board with no sessions yet, the filter offers only `all` even though several cards carry a `host` field.

Reproduced by driving the real board at 1280×800: card `alpha` with `fields.host = laptop` rendered chips `claude` / `high` only, and the host filter's options were `['all']`.

This is an affordance collision rather than a functional break — the two controls coexist cleanly (one `#board-filter-host` in the toolbar, one `.assignment-host` in the drawer, different ids, different selectors, no conflicting state, no console errors). It surfaced when #883's host dropdown landed on top of #890's host filter.

## Solution

Make the assignment visible on the board, and make the filter's label honest about what it filters.

- A card whose `fields.host` names a machine other than the API host gets its own chip, visually distinguishable from the "ran on" information the session carries — so an operator can see at a glance which cards are pointed at another machine.
- The filter either learns about assigned hosts (its option list unions `fields.host` with `card.session.host`, and a card matches on either) or is relabelled to say it filters by where sessions ran. Pick one and say which in the spec; do not leave the label ambiguous.

## Acceptance Criteria

- [ ] A card with `fields.host` set to a non-API host renders a host indicator on the card face; a card with no `host` field renders none.
- [ ] A card whose `session.host` differs from its `fields.host` does not conflate the two — the assignment and the observation stay distinguishable to the reader.
- [ ] The host filter's behaviour matches its label: either its options include hosts that only appear as assignments (and selecting one shows those cards), or the control is relabelled to name the session-host semantics explicitly.
- [ ] Filtering by a host still hides non-matching cards and leaves lane layout intact at 1280×800.
- [ ] A server-free browser test covers both the chip and the filter behaviour, and fails without the change.

## Verification

[...]

PR #951, opened 2026-09-08
feat(agents): delete cards from the drawer with confirmation and kill for claimed tasks (#885)

nbramia:
## TL;DR
The board's card drawer carries a Delete action, always behind a confirmation naming the card, for both task cards (every lane, including Review) and scheduled cards. For a task card whose linked session is live and can actually be stopped, the confirmation warns that deleting will kill the running session and its subagents first, and confirming does that kill before removing the card; if the kill fails, the card stays and the error is shown. A live Claude Code or Codex CLI session can't be torn down this way, so its confirmation deletes the card without attempting a kill and says to close the pane by hand.

Refs #885

## Design
Delete renders last in the drawer's action row, mirroring the existing Kill/Cancel modal pattern (`panel.js`'s `openKillModal`) rather than introducing a new dialog shape. The drawer's schedule branch calls `renderDrawerActions` too, with a matching `data-field="actions"` container in its template, so a scheduled card can offer Delete. Deleting is sequenced client-side: kill first (only when the session is live and not CLI-backed), then delete, so a failed kill never leaves an orphaned delete. A 2xx kill response with a non-empty `failures` list counts as a failed kill — no delete happens.

Both the confirmation's wording and the kill decision come from the card's live board state rather than the drawer's last render. This matters because `updateOpenDrawer` skips rebuilding the drawer while focus is inside it, so a card the worker claims while the operator's caret sits in the Notes field would otherwise be deleted against a session-less snapshot — with no kill and no warning. If a kill becomes necessary in the window between the confirmation opening and the operator confirming, the note rewrites itself to the kill wording and the confirm button re-enables rather than killing something the operator was never told about; a second confirm proceeds.

## Implementation Notes
- `web/agents/board.js`: `renderDrawerActions` always pushes a `Delete` button (styled `drawer-action danger`) as the last entry, wired to `openDeleteModal(card)`. The button loop carries an `opts.danger` branch for the extra class. A `killDecision(card)` helper maps a card to `{needsKill, noteHtml}` — one of three confirmation bodies — and runs both when the modal opens and again at confirm time against `findCard(card.id)`. On confirm the handler awaits `POST /api/agents/sessions/{id}/kill` (when needed) before `DELETE /api/tasks/{id}` or `DELETE /api/scheduler/{id}` depending on `card.kind`, then closes the modal, closes the drawer (`closeDrawer()`, which tears the session panel down via `panel.close()`), toasts "Deleted.", and calls `fetchBoard()`. A `pending` flag makes the backdrop-click and Cancel handlers no-ops while a request is in flight, so a failed kill always finds its own modal still mounted.
- `web/agents.html`: `.drawer-action.danger` CSS, since the drawer's danger-button styling had no rule of its own.
- `tests/test_agents_board_ui_browser.py`: `_stub_routes`/`_open_board` carry `kill_calls`/`kill_status_code`/`kill_failures`, `task_deletes`/`schedule_deletes`, and a shared `call_log` for order assertions, plus a `_remove_card_from_state` helper mirroring the existing `_move_card_in_state`. A `TestDeleteCard` class covers all three delete paths, kill-then-delete ordering, both kill-failure shapes, cancel leaving everything unchanged, a Review-lane card, the CLI-backed no-kill path, session-panel teardown, the claimed-while-focused case, and redisclosure.
- `docs/specs/product/agent-viz.md` describes Delete in the drawer actions list and the Scheduled column; `docs/specs/technical/agent-viz.md` records its internals alongside the other drawer-action policies.

### Automated tests
`tests/test_agents_board_ui_browser.py` (101 tests, including 10 in `TestDeleteCard`) passes in full, run repeatedly with no flakiness:
```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" LIFEOS_AGENT_DEFAULT_ROUTE= \
  pytest tests/test_agents_board_ui_browser.py -q
========================== 101 passed in 44.27s ==========================
```
Every new test carries a mutation proof: a targeted revert of the corresponding production hunk breaks the test's positive assertion, and restoring the code makes it pass again.

### Manual verification
Driven end to end at 1280x800 with Playwright against a FastAPI app mounting the real `agents`, `agent_assignment`, `tasks` and `scheduler` routers over a temp vault on an ad-hoc high port, with real `TaskManager`/`SchedulerStore`/`SessionStore`/`TranscriptStore`:
- Claimed card with a live session: the confirmation names the card and carries the kill wording; confirming kills the session tree (root and its subagent both terminal in the store), then deletes; the task line and its indented `> ` notes body disappear from the vault markdown on disk.
- Unclaimed card: one `DELETE /api/tasks/{id}`, zero kill calls.
- Scheduled card: one `DELETE /api/scheduler/{id}`, never the task endpoint; the entry and its message line leave `Scheduler/Inbox.md`.
- Cancel: zero HTTP requests of any kind, drawer still open on the same card, vault file byte-identical by sha256.
- Failed kill, both shapes (non-2xx and 2xx with a non-empty `failures` list): no DELETE, a visible error toast, card still present, confirmation still open with the confirm button enabled and labelled "Delete".
- CLI-backed live session: no kill call, and the confirmation says to close the pane by hand.
- Zero uncaught exceptions and zero unexpected failed network requests on load and after every interaction.

## Review focus
- The three confirmation bodies in `killDecision` — plain delete, kill-then-delete, and CLI-backed-can't-kill — and that the wording matches what Kill and Cancel already say for the same session shapes.
- Treating a 2xx kill response with non-empty `failures` as a failed kill rather than a partial success, matching "delete only after a successful kill".
- The redisclosure path: whether requiring a second confirm when the kill decision changes under an open modal is the right trade against silently doing the safer thing.


nbramia:
## Review record

Three review rounds ran against this PR before merge. Reviewers: an Opus general reviewer (diff plus surrounding code), an Opus hostile reviewer that drove the real page with Playwright at 1280x800 against the real routers, and Codex (`codex exec`). Findings were refereed against a fixed bar: accept a defect reachable on a path a real operator takes, with a concrete reproduction, or a contract violation of an acceptance criterion; reject hardening for contrived inputs, cosmetic nits, and anything without a repro on a normal path.

### Round 1 — general + hostile + codex, on `9a9529e`

| # | Finding | Raised by | Severity | Decision | Rationale |
|---|---|---|---|---|---|
| F1 | `openDeleteModal` captured `hasLiveSession`/`isCliSession`/`needsKill` from a stale card snapshot. `updateOpenDrawer` skips rebuilding the drawer while focus is inside it, so a card the worker claims while the operator's caret sits in the Notes textarea keeps a session-less snapshot and Delete removes the task with **no kill**, stranding a live session. | general, hostile, codex | Action Required | **ACCEPT** | Breaks AC2's whole safety property; hostile reproduced it 3/3 against real routers with a clean no-focus control proving focus was the trigger. |
| F2 | The "2xx kill with a non-empty `failures` list is a failed kill" branch had no test — deleting those lines left all 98 tests green. | general, hostile, codex | Recommended | **ACCEPT** | An untested branch on a destructive path that the issue's constraints name explicitly. |
| F3 | The confirmation was dismissable while the kill/delete was in flight, so the catch re-enabled a detached button and "the modal stays open with confirm re-enabled" silently didn't hold. | codex | Recommended | **ACCEPT** | Contract violation; no data-integrity harm, and the guard is two lines in the same function. |
| F4 | No test pinned the session panel / SSE teardown after a live-session delete; a mutation deferring teardown to `fetchBoard()` would pass. | codex | Recommended | **ACCEPT** | Cheap positive assertion for a requirement the issue states. |
| F5 | The spec's Escape sentence enumerated only the composer and the Answer prompt; the delete confirmation is a third stacked modal Escape correctly ignores. | general, hostile, codex | Recommended | **ACCEPT** | The doc misdescribed current behavior. |
| F6 | The Delete prose sat between the Cancel list entry and the Cancel follow-on sentence, so the Cancel CLI-pane caveat read as if it described Delete. | general | Recommended | **ACCEPT** | Real readability defect in a consumer-facing spec; cheap. |
| F7 | A failed kill surfaced the raw response body in the toast; the DELETE path three lines below already parsed `detail`. | general | Minor | **ACCEPT** | Trivial, same function, already being edited. |
| F8 | Add a `blocked` live-session test because AC2 names "running or blocked". | codex | Recommended | **REJECT** | The code branches on `card.session && !TERMINAL.has(status)`, never on tags, so a blocked card exercises the identical path; hostile drove a blocked card live and it behaved correctly. Duplicate coverage with no distinct mutation. |
| F9 | Double-clicking Delete stacks two identical confirmations with duplicate ids. | general | Minor | **REJECT** | Not reproduced — hostile drove it and it fails safe: the second click lands on the first modal's backdrop, 0 modals, zero API calls. |
| F10 | A card deleted out-of-band while the confirmation is open leaves an orphaned modal over a closed drawer. | hostile | Minor | **REJECT** | Cosmetic and recoverable (Cancel dismisses cleanly, the error is surfaced), and pre-existing to every board modal rather than introduced here. |

Cleared by all three reviewers, no change needed: session panel and SSE teardown already worked via `closeDrawer()` → `panel.close()`; Escape and outside-click stacking already behaved correctly in code; the board emits only `task` and `schedule` kinds so no third kind could misroute to the scheduler endpoint; `renderDrawerActions` cannot throw on a schedule card; AC2's vault-line-and-notes-body removal is backend behavior already covered by `tests/test_task_manager.py` and `tests/test_tasks_api.py`; no hardcoded personal values.

Addressed in `d183ebe`.

### Round 2 — hostile + codex, on the fix delta `d183ebe`

| # | Finding | Raised by | Severity | Decision | Rationale |
|---|---|---|---|---|---|
| G1 | Round 1 fixed the *action* but left the confirmation *text* on the stale snapshot, so the operator could read "this session can't be killed from here — close its pane manually" and a kill would fire anyway. | hostile (Recommended, "a real AC2 violation"), codex (Action Required) | Action Required | **ACCEPT** | Contract violation of AC2 with a 3/3 deterministic repro on the same path already accepted for F1; one line in the same function. |
| G2 | The focused-drawer regression test asserted the request sequence but never the confirmation text, so it permitted G1. | codex | Recommended | **ACCEPT** | Folds into G1's fix as its positive assertion. |
| G3 | The `pending` guard is mutation-survivable — removing it leaves all 100 tests green. | hostile | Minor | **REJECT** | Behavior driven and verified correct across all five failure exits including two where `fetch` rejects; a two-line guard that cannot cause a wrong delete. Converge rather than chase coverage. |
| G4 | The kill-failure `detail` parse is mutation-survivable. | hostile | Minor | **REJECT** | Cosmetic toast wording, verified correct live. |

Round 2 also confirmed F1 genuinely closed (kill fires against the new session id, session terminal in the store afterwards), the `pending` flag resetting on every driven failure exit, the inverse case safe (snapshot promised a kill, live session since terminal → no kill, clean delete), the vanished-card fallback sane, and the F1 regression test real and non-flaky — removing its `notes.click()` makes it pass under the mutation, so the focus step is load-bearing.

Addressed in `e15cc9a`, which resolves the live card at modal-open time and, when the kill decision changes while the confirmation is open, rewrites the note and requires a second confirm rather than killing silently.

### Round 3 — codex, on the redisclosure delta `e15cc9a`

**Zero findings at every severity, with an explicit "the loop should stop".** Verified: the redisclosure branch cannot loop (`needsKill` is a closure `let` mutated to `true` before the return, so the second confirm proceeds); the `outerHTML` note rewrite invalidates no held reference (confirm button, cancel button, backdrop node, click guards, `pending` closure all remain valid); `pending` is untouched on that return path; the inverse still deletes on the first confirm; `killDecision` faithfully preserves all three note branches including `sourceLabelFor` and escaping.

### Docs gate

`docs/specs/product/agent-viz.md` tightened from ~180 to ~115 words with implementation detail moved out and the duplicated scheduler-endpoint sentence dropped so it lives in one place; `docs/specs/technical/agent-viz.md` gained Delete's internals in the existing frontend-policy paragraph. `api-reference.md`, `task-management.md`, `scheduler.md`, `agent-worker.md` and `agents-go-to.md` were each checked and judged not stale, with reasons. Current-behavior-only standard enforced. Committed as `69f1844`.

## Verify verdict: PASS

Opus verifier, sync Playwright chromium at **1280x800**, against a FastAPI app mounting the **real** `api/routes/agents.py`, `agent_assignment.py`, `tasks.py` and `scheduler.py` routers with real `TaskManager`/`SchedulerStore`/`SessionStore`/`TranscriptStore` over a temp vault, uvicorn on **127.0.0.1:8151**. Stubbed only `_claude_code_snapshot`/`_codex_snapshot` (no real CLI transcript tree exists for a synthetic session) and, in the kill-failure scenario alone, the kill route to produce the two failure shapes.

- **Claimed card with a live session and a live subagent** — confirmation names the card and carries "kill the running session and its subagents"; ordered calls `POST /sessions/{id}/kill` then `DELETE /api/tasks/{id}`; root **and** descendant both `failed` in the session store; the task line and both indented `> ` notes lines gone from `Tasks/Work.md` on disk, the neighbouring subtask untouched.
- **Unclaimed card** — one `DELETE /api/tasks/{id}`, zero kill calls, line and notes body gone.
- **Scheduled card** — one `DELETE /api/scheduler/{id}`, never the task endpoint; entry and its message line gone from `Scheduler/Inbox.md`.
- **Cancel** — zero HTTP requests of any kind, card still on the board, drawer still open on the same card, vault file byte-identical (matching sha256 `85aefa28…`).
- **Failed kill, both shapes** (503, and 200 with a non-empty `failures` list) — no DELETE, a visible `toast error` naming the failure, card still on the board and still in the vault, confirmation open with the confirm button enabled and relabelled "Delete".
- **CLI-backed live session** — zero kill calls, "close its pane manually" wording, one task DELETE.
- **Redisclosure** — first Confirm rewrites the note and makes zero network calls; second Confirm kills then deletes.

**Console errors: zero.** Zero uncaught exceptions and zero page errors on load and after every interaction, in every scenario. Zero failed network requests other than ones deliberately caused: the intercepted 503 in the kill-failure scenario, and one `net::ERR_ABORTED` per drawer teardown from the session transcript stream being closed by `panel.close()`.

### Suites

| Suite | Result |
|---|---|
| `tests/test_agents_board_ui_browser.py` (pre-rebase, run 4x) | 101 passed, no flakiness |
| `tests/test_agents_board_ui_browser.py` (post-rebase onto `3338ab0`) | 104 passed |
| `tests/test_no_change_narration.py` | 1 passed |
| Pre-push gate, unit (rebased tree) | 7283 passed, 9 skipped |
| Pre-push gate, `browser and not requires_server` (rebased tree) | 485 passed, 7679 deselected |

All 10 tests in `TestDeleteCard` carry a mutation proof: a targeted revert of the corresponding production hunk breaks a positive assertion, and restoring the code makes it pass again.

Rebased onto `3338ab0` after #912 landed — clean, no conflicts — and re-gated in full before merge.

PR #953, opened 2026-09-08
feat(agents): show a card's assigned host on the board, and align the host filter with what it filters

nbramia:
## Summary

Makes a card's host assignment visible on the `/agents` board, and makes the host filter match what its label claims.

Two different fields carry a host on the board: `fields.host` (the assignment — where the card *will* run, written by the drawer's host dropdown) and `card.session.host` (the observation — where a session *did* run). Only the observation was surfaced.

- **Assignment chip.** A card whose `fields.host` names a machine other than the API host renders a `.board-chip-assigned-host` chip, styled distinctly from the ran-on chip (dashed border, different accent) and titled `assigned host`. A card assigned to the API host itself shows none, matching the drawer's "this machine" empty-choice semantics. A card with no `host` field shows none.
- **Ran-on chip** is unchanged, keeps `.board-chip-host`, and gains a `ran on` title. A card whose assignment and observation differ renders both, distinguishably.
- **Host filter** takes the issue's preferred option: its option list unions `fields.host` with `session.host`, and a card matches a selected host when *either* names it. A host that has only ever been an assignment is now reachable through the filter.
- **`GET /api/agents/board`** adds `api_host` to its payload so the client can tell "this machine" apart from an assignment elsewhere without a second round trip. `_build_board()` is the single construction point, so the SSE stream's payload carries it too.

## Testing

- New `TestHostAssignmentChipAndFilter` in `tests/test_agents_board_ui_browser.py` (server-free, no `requires_server`): chip presence and absence across four card shapes, the filter's option list containing an assignment-only host, and filter matching on either source with lane count unchanged at 1280x800.
- New `test_board_carries_api_host` in `tests/test_agents_board_api.py`.
- Mutation proofs, each making a positive assertion fail:
  - Reverting `web/agents/board.js` to the base makes the chip test fail on `to_have_text("build-box-2")` (element not found).
  - Restoring the option list to session-host-only makes the option test fail (`values == ['all', 'build-box-4']`).
  - Restoring the match predicate to session-host-only makes card `t8` invisible under the `build-box-2` selection.
  - Dropping `api_host` from `_build_board` makes the API test fail with `KeyError: 'api_host'`.
- Pre-push gate green: 7283 unit passed / 9 skipped; 475 server-free browser passed.

## Spec

`docs/specs/product/agent-viz.md` states the two distinct host-chip meanings on the card face and states that the host filter unions assignment and session hosts and matches on either.

## Closes

Part of #912.


nbramia:
## Review record (posted after merge for the audit trail)

This PR ran the full adversarial lifecycle; the consolidated per-round comments were not posted at the time, so the record is reconstructed here from the round reports.

### Round 1 — Opus general + Opus hostile (page-driving, 1280x800) + Codex
10 findings, 6 accepted / 4 rejected.

| # | Reviewer(s) | Finding | Decision |
|---|---|---|---|
| 1 | general + hostile (independent) | The card printed the same hostname twice (assignment chip + ran-on chip) whenever a card was dispatched to its assigned host — the steady state of every remote card, contradicting this PR's own spec sentence | **Accept** — contract violation on the primary path |
| 2 | Codex + general | `api-reference.md` documented the board payload as `{lanes, generated_at}` while the route also returns `api_host`, which the client depends on | **Accept** |
| 3 | general | `agent-viz.md` frontmatter `Last Updated` not bumped | Accept (fold) |
| 4 | general | Card-contents sentence read "a host chip" unconditionally | Accept (fold) |
| 5 | general | A lane-count assertion in the browser test was structurally guaranteed and could not fail | **Accept** — vacuous assertion violates the mutation-proof standard |
| 6 | Codex | The board-stream unit test parsed the SSE frame without asserting `api_host` | Accept (contract guard) |
| 7 | hostile | Clearing the last card on the selected host resets the host filter to "all" | Reject — pre-existing behaviour of an unchanged guard, no broken state |
| 8 | hostile | A `#me` card with a stale `fields.host` shows an assignment chip though its drawer offers no host control | Reject — how `fields.host` is written is out of scope; the chip truthfully reflects a set field |
| 9 | hostile | The API host appears as a filter option though a card assigned to it renders no chip | Reject — cosmetic, no reachable broken state |
| 10 | Codex | An unrelated resume-here browser test fails as a full file, passes in isolation | Reject (out of scope) — pre-existing; the authoritative gate passed |

Hostile pass recorded 0 console messages, 0 page errors, 0 failed requests and 0 responses >= 400 across load, four filter selections, a stream tick, two drawer opens, two Escapes and a drawer host save.

### Round 2 — hostile (page-driving) on the round-1 fix
Zero Action Required, zero Recommended. One Minor, explicitly not asked to be fixed: in the suppressed case the card face no longer distinguishes "pointed at that box" from "pointed at that box and it already ran there". The fact is not lost — the drawer's session panel still carries the ran-on host, and run state is still shown by the live dot, the lifecycle tags and the lane. #912's criteria constrain only the differing case, and the equal case is documented. Converged.

### Docs gate
One Action Required, factual: the spec stated the ran-on suppression rule as "the assignment names it" while the code suppresses on "an assignment chip was rendered", which diverge for a card assigned to the API host itself. Corrected, plus the payload-contract and frontmatter items from round 1.

### Verification — PASS
Real `agents`, `agent_assignment`, `tasks` and `scheduler` routers on 127.0.0.1:8137 over a temp vault, page served the way `api/main.py` serves it, Playwright at 1280x800. Port 8000 was never bound and the operator's live service answered 200 after teardown. Six synthetic vault cards covered every host combination: assigned-not-run, no host field, assigned to the API host, assigned-and-ran-elsewhere, assigned-and-ran-same, and no-assignment-but-ran. Chip distinctness was checked by computed style, the lane layout stayed six columns with no horizontal overflow, and a maximally crowded card wrapped without any chip escaping the card box.

Authoritative pre-push gate on the merged head: unit and server-free browser suites green.

PR #954, opened 2026-09-08
fix(agents): keep a saved model absent from the catalog on assignment saves (#906)

### LINKED ISSUE #906: fix(agents): a saved model absent from the catalog is silently wiped by the next assignment save
## Problem

A card's saved `model` is silently erased by the next unrelated assignment save whenever that model id is absent from what `GET /api/agents/models` returns for the card's engine.

`web/agents/assignment.js`'s `populateModelOptions()` builds the model `<select>` purely from the catalog. If the saved model isn't in that list, no option carries it, so `modelEl.value` collapses to `''`. `save()` then runs `if (catalogReady) fields.model = modelEl.value || null;` and writes `model: null`, wiping a deliberate operator choice — on a save the operator made about something else entirely (effort, host, or engine).

`#861` added the `catalogReady` flag, which correctly defends the window *before* the catalog resolves. It does not defend the case where the catalog resolves **without** that model id. That happens routinely:

- the Anthropic key is unset or the provider call fails, so `claude: []` comes back (`model_catalog.py` returns an empty list for an unconfigured engine — by design, not an error);
- the operator pinned a dated snapshot or a model the provider has since retired from `models.list()`;
- a Codex install whose `~/.codex/models_cache.json` is missing and which has no `LIFEOS_OPENAI_API_KEY`, so `codex: []`.

Reproduced against a real app (all routers, throwaway vault) with Playwright: a card carrying `fields.model = "claude-opus-5"` on a box whose catalog returns `claude: []`. Changing **effort** produced

```
PUT {"fields":{"effort":"max","host":"studio-box","assigned_by":"board","model":null}}
```

and the card's `model` field was gone from the vault afterwards. The drawer shows "engine default" from the moment it opens, so the operator has no signal that their pinned model is about to be discarded.

The same class of bug in the host picker was fixed in #883 by seeding the saved value as a selected option before the catalog is fetched; the model picker never got that treatment.

## Solution

Make a saved model round-trip even when the catalog does not list it.

- `populateModelOptions()` keeps the currently-saved model as a selected option when the catalog has no entry for it, flagged as unknown (mirroring the host select's `data-unknown="true"` option from #883), so `modelEl.value` never silently becomes `''` for a card that has a model.
- Seed that option synchronously at render time, before `GET /api/agents/models` is requested, so the value is correct on the very first save of a page load — the same fix shape #883 used for `host`. Whether the `catalogReady` flag can then be retired or must stay is an implementation detail for the fix to decide; behaviour must not regress #861's guarantee.
- Clearing the model stays possible: choosing "engine default" explicitly still writes `model: null`.

## Acceptance Criteria

- [ ] With a card whose `fields.model` is absent from the engine's returned catalog, changing effort or host PUTs the card's existing `model` value unchanged — never `model: null`.
- [ ] That model renders as the selected option, visibly flagged unknown, and its `value` is the bare model id (no marker text leaking into what is saved).
- [ ] Explicitly selecting "engine default" still writes `model: null`.
- [ ] A saved model that IS in the catalog renders with its normal label and carries no unknown flag once the catalog resolves.
- [ ] The saved-model option survives the catalog arriving and an engine change back and forth.
- [ ] #861's guarantee is preserved: no save before the catalog resolves may write `model: null` over a saved model.

## Verification

- **Automated tests:** additions to `tests/test_agents_assignment_ui_browser.py` — a card with a model absent from the stubbed catalog, an effort change, and an assertion on the captured PUT body. Each new test needs a mutation proof (revert the production hunk, confirm the test fails on a POSITIVE assertion, restore).
- **Manual checks:** on a board with no Anthropic key configured, open a `#claude` card that has a `model` field, change its effort, and confirm the model survives in the vault task's inline fields.
- **Existing test suite:** green, including the full `browser and not requires_server` set.

## Scope

**In scope:** `web/agents/assignment.js`'s model-option population and its interaction with `save()`; the browser tests; any doc line that describes the model picker's behaviour.

**Out of scope:** the host picker (fixed in #883), the model catalog's own sourcing/caching (`model_catalog.py`), and the assignment save-error reporting (fixed in #883).

## Technical Context

- **Key files:** `web/agents/assignment.js` — `populateModelOptions()` (~:167-175), `catalogReady` (~:167), `save()`'s `if (catalogReady) fields.model = modelEl.value || null;` (~:255).
- **Patterns to follow:** the host select's synchronous seed plus `data-unknown="true"` flagged option, landed in #883 in the same file — the model fix should read as its mirror image.
- **Constraints:** no build step, plain DOM, everything escaped through the module's existing `escapeHtml`; the new server-free browser tests must stay free of the `requires_server` marker.

## References

- Found by the hostile page-driving reviewer during #883's round-1 review (PR #901); the code is identical on the pre-#883 base `a346ef8`, so this is pre-existing rather than a #901 regression, and was deliberately not folded into that PR's scope.
- Follows #861 (PR #867), which introduced the `catalogReady` guard this extends.
- Related: #883 (PR #901), the host-picker equivalent.

nbramia:
## Problem

A card's saved `model` is silently erased by the next unrelated assignment save whenever that model id is absent from what `GET /api/agents/models` returns for the card's engine. The model select builds its options purely from the catalog, so a saved model the catalog doesn't list collapses the select's value to empty, and the next effort/host/engine change writes `model: null` over the operator's pinned choice.

Closes #906.

## Solution

The saved model is seeded into the select as a selected option synchronously, before the catalog fetch is made, mirroring the host picker. A save fired while that fetch is in flight therefore carries the saved model instead of clearing it.

When the catalog resolves, `populateModelOptions` rebuilds the option list around the LIVE select value and classifies it three ways:

| The live selection is | Rendered as |
|---|---|
| listed by the card's current engine | its ordinary catalog label, selected |
| listed only by some other engine's catalog | dropped to "engine default" |
| listed by no engine's catalog | selected, flagged `data-unknown="true"`, labelled `<id> (unknown)`, bare id as the option's `value` |

That decision runs on every mount, including the remount the board performs after its own Assignee select changes a card's engine. The middle case matters because a model id is valid only for the engine whose catalog carries it: the executors hand `fields.model` straight to the CLI, so a Codex model left on a Claude card spawns `claude --model gpt-…`, which the CLI rejects.

The `catalogReady` flag is retired. The synchronous seed makes the select's value correct from the first render, so `model` is always sent, exactly like `effort` and `host`. #861's guarantee is preserved and tightened: the pre-catalog save carries the saved model forward rather than merely omitting the key. Choosing "engine default" still writes `model: null`.

## Tests

`tests/test_agents_assignment_ui_browser.py` 40 passed, `tests/test_agents_board_ui_browser.py` 106 passed, full `browser and not requires_server` set 483 passed. All server-free.

The board-driven tests are the ones that matter for the engine-ownership rule: a test living only in the assignment file cannot catch that class of defect, because `board.js` hides the module's own engine row and changes engines through its Assignee select instead.

Every new or modified test carries a mutation proof using a targeted mutation that makes a POSITIVE assertion fail, applied to a committed tree and restored immediately.

| Test | Mutation | Failure |
|---|---|---|
| `test_effort_change_before_catalog_resolves_does_not_clear_model` (modified) | removed the synchronous seed | `assert None == 'claude-sonnet-5'` |
| `test_unknown_model_survives_effort_change_before_catalog_resolves_on_first_paint` | removed the synchronous seed | `assert None == 'claude-legacy-9'` |
| `test_host_change_before_catalog_resolves_does_not_clear_model` | removed the synchronous seed | model absent from the PUT |
| `test_unknown_saved_model_resolved_catalog_without_it` | corrupted the flagged option's `value` | option `value` mismatch |
| `test_unknown_flagged_model_option_survives_engine_change` | removed the flagged-option append | flagged count 0, expected 1 |
| `test_selecting_engine_default_model_puts_null` | `model: sentModel \|\| 'claude-sonnet-5'` | `assert 'claude-sonnet-5' is None` |
| `test_known_saved_model_renders_normal_label_no_unknown_option` | flagged every catalog option | label assertion fails |
| `test_model_change_before_catalog_resolves_survives_catalog_landing` | live-value read replaced by the render-time snapshot | resurrected value |
| `test_rejected_model_save_reverts_then_next_save_sends_reverted_value` | deleted both unconditional revert lines | rejected value rides along |
| `test_foreign_model_dropped_to_engine_default_even_via_a_hidden_model_row` | drop branch disabled | `'claude-sonnet-5' == ''` |
| `test_board_assignee_change_to_another_engine_drops_the_foreign_model` (board-driven) | drop branch disabled | PUT body carries the foreign model |
| `test_board_assignee_change_keeps_a_model_no_engine_lists` (board-driven) | drop made unconditional | flagged option missing |

The modified existing test's assertion changed from `assert "model" not in fields` to the stronger `assert fields["model"] == "claude-sonnet-5"`. The guarantee it protects, never writing `model: null` over a saved model, is preserved and tightened. No other existing assertion is weakened.

## Verification

Driven against a real app on port 8137: the actual `agents.py`, `agent_assignment.py` and `tasks.py` routers over a throwaway vault, the real `ModelCatalog` fed synthetic data, Chromium at 1280x800, vault file read back after every step. All six acceptance criteria confirmed with observed PUT bodies and task lines. Zero console errors, zero uncaught exceptions, zero failed network requests other than one self-inflicted SSE abort from a deliberate page reload.

## Docs

`docs/specs/product/agent-viz.md` — the drawer-pickers clause states the three-way classification, that it applies on every drawer open as well as after an engine change, and that choosing "engine default" clears the model.

## Review

Three rounds. Round 1: Opus general, Opus page-driving hostile, and Codex; one Action Required (an engine change carried the outgoing engine's model onto the incoming one) plus five accepted test-coverage findings. Round 2: the first fix proved unreachable from the real board, because it lived in a control `board.js` hides — found independently by the hostile reviewer and Codex. Round 3: no findings; the hostile reviewer proved harness sensitivity by replaying the round-2 code and reproducing the old defect.

Two follow-up issues filed for defects refereed out of scope: #955 (the model catalog cannot distinguish an unavailable fetch from a genuinely empty engine) and #956 (an engine change from the drawer leaves the card's model field stale in the vault, matching pre-existing behaviour).


nbramia:
## Review record (posted after merge for the audit trail)

Three adversarial rounds ran on this PR; the consolidated comments were not posted at the time, so the record is reconstructed here from the round reports.

### Round 1 — Opus general + Opus hostile (real app, temp vault, 1280x800) + Codex
6 accepted, 3 rejected.

- **A1 (Action Required)** — an engine change carried the *outgoing* engine's model id onto the incoming engine and wrote it to the vault: a `#codex` card at a GPT model switched to `#claude` kept that model, which the executor passes straight to `claude --model`, and it is rejected. The base branch cleared the field on the same action, so this was a regression, found independently by two reviewers, one against a real vault.
- **A2, A3 (Recommended)** — two behaviours the PR touched survived deletion with the suite green: the live-value read in the option population, and the two failure-path lines. Both reachable, both now pinned.
- **A4, A5, A6 (Recommended, Codex)** — the unknown option's own `value` attribute was never asserted; the "effort **or host**" arm of AC1 was unpinned; and the engine-change test used a model absent from *both* catalogs, so it would pass even if repopulation were broken. That last gap is exactly what let A1 through.
- **Rejected** — a revert-proof objection already answered by targeted single-hunk mutation proofs; the catalog's inability to distinguish "fetch failed" from "engine has no models" (real, but the issue puts catalog sourcing out of scope, filed as a follow-up); and a one-round-trip `(unknown)` flash, cosmetic and consistent with the host picker.

### Round 2 — hostile + Codex on the delta
**One Action Required, and it invalidated round 1's fix.** The board hides the assignment module's own engine row, so the module's engine-change handler — where the entire outgoing-catalog clear lived — can never fire from the drawer. The board changes the engine through its own Assignee select, which remounts the pickers and discards the retained state. The round-1 defect reproduced verbatim in both directions plus a third path, and the docs sentence the fix added was false for the drawer it described. Round 1's tests passed only because they called the module directly with the engine row visible.

Recorded referee lesson: the diagnosis was right and the *placement* was wrong; the decision must survive a full remount, because that is what the board actually does.

### Round 3 — hostile + Codex
Converged: no Action Required, no Recommended, no new Minor. The reviewer first proved harness sensitivity by replaying the same attack against round 2's code and reproducing the round-2 defect, so the clean result is meaningful. All three round-2 reproductions now match base byte for byte in the picker, the PUT body and the vault file, and four over-correction checks pass, including that a keyless box keeps its model rather than dropping it.

Two Codex findings were rejected with base comparisons as the deciding evidence: a bare assignee change leaving a stale field until the next save (byte-identical on base, pre-existing, and fixing it changes an unrelated contract), and a cold-catalog window that is the direct intended consequence of the "carry the saved model before the catalog resolves" criterion, measured against a 1500 ms forced delay versus a real 50 ms fetch with identical file state.

### Docs gate and verification
Docs gate clean. Verify **PASS**: every acceptance criterion reproduced by driving the real FastAPI app with the real agents, assignment and tasks routers over a temp vault, Playwright at 1280x800. Authoritative pre-push gate green on the merged head.

Issue #884, opened 2026-09-04
feat(agents): full schedule editing in the scheduled card drawer


nbramia:
## Problem

A scheduled card's drawer edits only the name, message, and enabled flag. Everything that defines a schedule, the trigger type, its cron expression or one-off time, the timezone, the action, the executor for agent actions, and the delivery bot, can only be changed in Obsidian or through the API. The board presents schedules as cards but cannot manage them.

## Solution

Give scheduled cards a complete drawer backed by the existing scheduler update endpoint: trigger type with cron or one-off datetime, timezone, action, executor for agent actions, delivery bot for non-agent actions, plus a human-readable next-fire preview and the last run's outcome.

## Acceptance Criteria

- [ ] The scheduled card drawer shows and saves: name, message, enabled, schedule type (`cron` or `once`), schedule value, timezone, action (`notify`, `prompt`, `endpoint`, `agent`), executor when action is `agent`, bot when action is not `agent`, through `PUT /api/scheduler/{id}`.
- [ ] Changing the cron expression updates a next-fire preview from the server response; an invalid expression shows the 422 detail inline and saves nothing.
- [ ] The bot select lists only names the API accepts; picking an unaccepted name is impossible.
- [ ] The drawer shows the last run's time, outcome, and snippet, and a Trigger now button that calls the test-fire endpoint.
- [ ] The server-free browser test covers each field's save, the invalid-cron path, and Trigger now.

## Verification

- **Automated tests:** board browser test additions with a stubbed scheduler API; route tests already exist for the endpoint.
- **Manual checks:** edit the Morning Briefing entry's time from the board and confirm the vault line changes.
- **Existing test suite:** green.

## Scope

**In scope:** the drawer fields, preview, trigger button, product spec update.

**Out of scope:** creating schedules from the board composer; changes to the scheduler store.

## Technical Context

- **Key files:** `web/agents/board.js` (scheduled drawer, ~:284 and the `SCHEDULE_FIELDS` list at ~:33), `api/routes/scheduler.py` (`PUT /{id}`, `POST /{id}/trigger`, bot validation), `api/services/scheduler_store.py` (`ScheduleEntry` fields), `docs/guides/scheduler.md`.
- **Patterns to follow:** the task drawer's field save pattern; `docs/guides/scheduler.md` for accepted values.
- **Constraints:** every write goes through the API, never the vault file directly.

## References

[...]

Issue #958, opened 2026-09-08
test(agents): wait for the snapshot before asserting the graph host filter's options


nbramia:
## Problem

`tests/test_agents_host_filter_ui_browser.py` fails intermittently under the pre-push gate's parallel browser run while passing every time in isolation. Two tests fail together:

```
FAILED tests/test_agents_host_filter_ui_browser.py::TestHostFilter::test_filter_host_lists_every_host_from_the_snapshot[chromium]
  AssertionError: assert 'laptop-a' in ['all']
FAILED tests/test_agents_host_filter_ui_browser.py::TestHostFilter::test_selecting_a_host_hides_sessions_from_other_hosts[chromium]
  playwright TimeoutError: Page.select_option: Timeout 30000ms exceeded
    waiting for locator("#filter-host") - did not find some options
```

The cause is a missing wait, not a product bug. The file's `_open_agents` helper ends with:

```python
page.click('[data-tab="graph"]')
page.wait_for_selector("#filter-host")
```

`#filter-host` exists in the static markup carrying only its `all` option. The per-host options are appended after `GET /api/agents/snapshot` resolves and the graph initializes. Waiting for the element to exist therefore proves nothing about whether the snapshot has been applied. On an unloaded machine the fetch wins the race and the tests pass; when the box is busy (four xdist workers plus a sibling agent's suite) the assertion runs first and sees `['all']`.

Observed on `feat/agents-board-polish`: the gate's browser phase took 401s under contention and these two failed; a re-run of the identical selection under the test lock finished in 106s with 498 passed. The same file passes 26/26 in isolation on both the clean integration tip and a feature branch.

This costs a full 22-minute gate cycle every time it fires, and it trains readers to dismiss red as noise.

## Solution

Make `_open_agents` wait for the state the tests actually depend on rather than for an element that exists before the data arrives. Wait for the host select to contain more than its `all` option, or for whichever node/graph render signals that the snapshot has been applied. Prefer waiting on observable output over a fixed timeout.

While in the file, the two `page.wait_for_timeout(400)` calls in `TestHostFilter` and `TestEndedStatus` are the same hazard with a longer fuse; converting them to waits on the rendered result would remove the remaining timing assumptions.

## Acceptance Criteria

- [ ] `_open_agents` returns only once the snapshot-derived host options are present, with no fixed sleep.
[...]

PR #949, opened 2026-09-08
feat(agents): redesign the graph's encodings, labels, hover card, and controls (#864)

nbramia:
Redesigns the `/agents` graph tab as a session map: the node label is the linked card title, then the session's real label, then the prompt preview, never the model badge; fill colour is the board lane with a lane legend; shape is the engine with five glyphs and a correct engine legend; size is active seconds with a tool-call ring; badges for pending questions and error counts; labelled host columns with collapsible subagent trees; an HTML hover card; single-click drawer with no delay; double-click focus for CLI nodes; zoom-to-fit and reset. Every snapshot and stream row carries a derived `lane` and `pending_question` field so the graph and the board share one lane source.

Closes #864 except the shared-drawer parity item, which follows as a separate PR. Also closes #936, #942, #943.

## Acceptance criteria → tests

| Criterion | Test |
|---|---|
| Label precedence (card title > label > prompt preview; model badge never the label) | `test_agents_graph_redesign_ui_browser.py::TestLabelPrecedence` (linked card, unlinked fallback); `test_agents_graph_encoding_browser.py::TestNodeLabel` (custom label wins, raw ids skipped, `model_label` never the label, no literal `?`) |
| Lane colour + lane legend | `TestLegends::test_lane_legend_lists_every_lane`; `TestLaneColor` (distinct colours, unknown lane falls back) |
| Five engine shapes + engine legend; routing change updates shape | `TestLegends::test_engine_legend_names_all_five`, `TestFiveEngineShapes::test_five_distinct_shapes_render`, `TestFiveEngineShapes::test_routing_change_on_existing_node_updates_its_shape`; `TestEngineOf` (precedence), `TestShapes` (five glyphs, legend, shape tag) |
| Size monotonic in `total_active_seconds`, floored and capped, independent of cache tokens | `TestRadius` (floor, monotonic, cap, bounds), `TestRingWidth`; `TestStreamTickUpdatesNodeSize::test_stream_delivered_active_seconds_renders_as_size` |
| Question badge, error count badge, no literal `?` label | `TestBadges` (pending question present/absent, error count) |
| Labelled host columns with name + count | `TestHostColumns::test_host_column_headers_show_name_and_count` |
| Collapsed subagent trees with numeric badge; click expands/collapses | `TestSubagentCollapse` (hidden by default, badge count, click toggles) |
[...]

PR #959, opened 2026-09-08
feat(agents): full schedule editing in the scheduled card drawer

nbramia:
## Summary

A scheduled card's drawer edited only the name, message, and enabled flag. Everything that defines a schedule had to be changed in Obsidian or through the API by hand. This gives the drawer the full set of controls, all writing through the scheduler API.

**Drawer** (`web/agents/board.js`, the schedule branch of `renderDrawer`)

- Schedule type (`cron` or `once`) and its value, with a type-aware label and placeholder.
- Timezone, action (`notify`, `prompt`, `endpoint`, `agent`), an executor select shown only for the `agent` action, and a bot select shown only for the other actions.
- The bot select is populated from the API, so a name the API would reject cannot be picked. If the fetch fails the select disables itself and shows the reason as visible text, rather than silently looking like nothing is configured.
- A next-fire preview, refreshed from the `next_trigger_at` in each save's own response.
- The last run's time, outcome, and snippet, or a clear line when the entry has not run.
- A **Trigger now** button calling the test-fire endpoint.
- A failed save shows the server's `detail` inline next to the offending field and snaps that control back to its last-saved value.

Switching the action shows and hides the executor and bot controls immediately.

**API** (`api/routes/scheduler.py`)

- `PUT /api/scheduler/{id}` validates before writing anything. An invalid cron expression or one-off datetime returns 422 with a detail naming the problem and stores nothing. Previously the store logged the bad expression, saved it anyway, and set the next fire time to null, so a typo silently retired a schedule. An unknown timezone is rejected the same way instead of raising out of the one-off branch. `schedule_type` and `action` are validated with the status the create endpoint already uses.
- When only `schedule_value` changes, the effective type comes from the stored entry rather than being assumed.
- New `GET /api/scheduler/bots` returns the registry names the API accepts. It is declared ahead of `GET /{schedule_id}` so the capture route does not swallow it.

**Board payload** (`api/routes/agents.py`)

`_schedule_card` carries `schedule_type`, `schedule_value`, `timezone`, `action`, `executor`, and `bot`, so the drawer renders current values without a second round trip. Existing keys are unchanged.

## Testing

- `tests/test_scheduled_card_drawer_ui_browser.py`, 13 server-free browser tests: each field's save and its PUT body, the invalid-cron path showing the detail inline and leaving the value unsaved, the bot select's contents, the executor/bot swap, the preview updating from the response, and Trigger now.
- `tests/test_scheduler_api.py`: 11 validation tests and 2 for the bots endpoint.
- `tests/test_agents_board_api.py`: the scheduled card carries the six added fields.
- Every new test carries a mutation proof: the production hunk was reverted, a positive assertion was confirmed to fail, and the hunk restored. Three tests were strengthened after their first proof showed them passing vacuously.
- Pre-push gate green: 7297 unit passed / 9 skipped; 498 server-free browser passed.

## Docs

`docs/specs/product/agent-viz.md` § Scheduled column, `docs/specs/product/api-reference.md`, and `docs/guides/scheduler.md`.

## Out of scope

No `endpoint_config` editor (no acceptance criterion covers it), no schedule creation from the board composer, and no change to the scheduler store.

## Closes

Part of #884.


nbramia:
## Review round 1 — Opus general + Opus hostile (page-driving, 1280x800) + Codex

10 findings. **8 accepted, 4 rejected** under the milestone's referee bar (a finding is accepted only when it is a defect reachable on a path a real operator takes, with a concrete reproduction; contract violations of an acceptance criterion also count).

Codex was available and ran successfully, and this was the strongest cross-model overlap of the milestone: it independently reproduced all three top findings, and reproduced the most serious one **against a real `SchedulerStore`** rather than by reading the diff.

| # | Reviewer(s) | Finding | Decision |
|---|---|---|---|
| 1 | general + **codex** (hostile saw it, judged it benign) | Changing only the schedule type PUTs `{schedule_type}` alone. `_validate_update_fields` returns early when `schedule_value` is absent, so the mismatch is written: a cron string stored under a `once` schedule. `compute_next_trigger` swallows the parse error, `next_trigger_at` becomes null, and `is_schedule_active` buckets the card into Done — a live recurring schedule stops firing. Codex's repro: PUT 200, `schedule_type=once`, `schedule_value=0 9 * * *`, `next_trigger_at=None`. | **ACCEPT** — reachable by changing the type and closing the drawer, and it defeats the "save nothing on an invalid schedule" guarantee this PR adds, through the other door. The hostile reviewer's "the preview says Not scheduled to fire again" is feedback, not prevention: the schedule is already retired. |
| 2 | general + hostile + **codex** (all three) | The executor select has no empty option, so an `agent` schedule with `executor: ""` displays `local`. Untagged is a distinct documented state meaning the default route, not local, and the drawer can never restore it. | **ACCEPT** — contract miss on AC 1, and it misreports routing, which can silently bill the API. |
| 3 | general + hostile + **codex** (all three) | A stored `bot` absent from `GET /bots` (a renamed bot) leaves the select at `selectedIndex = -1`, rendering blank. The operator loses the only signal that the entry names an unresolvable bot. `web/agents/assignment.js` sets the precedent: keep it as a flagged `(unknown)` option. Not a clobber — each field PUTs only its own key. | **ACCEPT** |
| 4 | general | Trigger now permanently consumes a `once` schedule: `_fire_entry` calls `mark_triggered`, which disables the entry and clears the next fire. This button is the first UI surface for that endpoint, so the effect is one undisclosed click away, and the spec called it a "test-fire endpoint". | **ACCEPT** |
| 5 | general + hostile | The bot select renders `primary` twice — a hardcoded empty option labelled `primary`, plus `valid_bot_names()`'s own. The new test asserted the duplicate. | **ACCEPT** |
| 6 | general | `expect(bot_select).to_be_disabled()` is vacuous: the select ships `disabled` in the template, so it passes against code that never runs the failure branch. | **ACCEPT** — violates the mutation-proof standard. |
| 7 | general | The next-fire preview goes stale when Enabled is toggled; the store clears the next fire on disable but the handler does not refresh the preview. | **ACCEPT** |
| 8 | general + hostile | The Scheduled-column spec paragraph is a ~250-word run-on; and a failed bots fetch drops the stored name from the drawer entirely. | **ACCEPT** |
| — | hostile, self-rejected | Attribute quoting/escaping, two saves in flight, a bots fetch resolving after the drawer closes, clearing the timezone to empty. | **REJECT** — no real-operator path. |

### Acceptance criteria the hostile reviewer tried and failed to break
Controls populated from the entry rather than defaults; per-field PUT bodies never clobbering an untouched field; repeated action swapping never sending a bot for an agent action or an executor for a non-agent one; the invalid-cron path showing the 422 detail inline in the field's own parent, snapping back, and saving nothing; the preview moving from the PUT response rather than being recomputed client-side; Trigger now surfacing a 500's detail; the bots-fetch failure disabling the control with a visible reason; drawer layout at 1280x800 with no clipping and no horizontal overflow; and coexistence with the base branch's Delete, including deleting a schedule end to end.

**Console/network:** zero uncaught exceptions, zero failed requests, zero unexpected 4xx/5xx across seven page loads and every interaction.

nbramia:
## Review round 2 — Opus hostile (page-driving) on the round-1 fix, plus the docs gate

Round 1's fix added a guard and changed an interaction, so round 2's hostile reviewer was briefed to assume the guard over-corrected and to prove it by driving the page. Its stub reimplemented the route's real validation order and recomputed next-fire times with `croniter` / `fromisoformat`, so a "successful" conversion in the browser was checked against a real stored entry.

### Hostile pass — zero Action Required

**The headline question was whether an operator can still convert a schedule. They can, in both directions.**

| Path | Result |
|---|---|
| `cron` to `once` (valid ISO, blur) | ONE PUT carrying both fields. Stored: type `once`, matching value, a real next fire — not null. Preview and value label updated. |
| `once` to `cron` | ONE PUT with both fields, real next fire. |
| Type changed, value blurred unedited | The pair is sent, the server rejects it 422, the detail shows inline verbatim, **nothing is written**, and the type select, value field, label and placeholder all snap back. |
| Type changed, then Escape | Zero PUTs, entry unchanged, schedule still fires. **This is the exact path that retired a live schedule before the fix.** |
| Type changed, then click outside | Zero PUTs, entry unchanged. |
| Type changed, then changed back, then an unrelated save | No stale pending type carried onto the unrelated save. |
| Rejected conversion, then a correct one | The drawer is not poisoned; the following conversion succeeded. |

Executor, bot (known name, orphan name, and registry failure), Trigger now, Enabled, Delete coexistence, and layout at 1280x800 were all re-checked and intact.

**Accepted findings: 2.**

| # | Finding | Decision |
|---|---|---|
| 1 | The Trigger now label is baked in at render time and never follows an in-drawer conversion. Convert a recurring schedule to a one-off and click the button: at click time it still reads `Trigger now`, so no warning appears at the moment it matters, and the label flips only after the schedule is consumed. The mirror case shows a false warning on a freshly recurring schedule. `updateOpenDrawer` skips its rebuild while a drawer element holds focus, so no board tick repairs it. | **ACCEPT** — reachable with a repro, one line in the value-blur success path. |
| 2 | A registry-fetch failure labels a perfectly valid stored bot name `(unknown)`. The name is not unknown; the registry is unreachable, and the adjacent reason text already says so. | **ACCEPT** — kept the `(unknown)` label for the genuine orphan case only. |

**Console/network:** zero uncaught exceptions and zero failed requests across all 22 scenarios. The only responses at or above 400 were the four the harness deliberately returned.

### Docs gate — verdict FAIL, two Action Required

| # | Finding | Decision |
|---|---|---|
| 1 | `docs/guides/scheduler.md` claimed `POST /api/scheduler` already rejects the same inputs. It validates only `schedule_type`, `action` and `bot`; it never parses `schedule_value` and never resolves `timezone`, so a bad cron expression POSTs fine and then never fires. | **ACCEPT** |
| 2 | Both the guide and `api-reference.md` claimed "only the fields present in the request are validated". False for a bare `schedule_type` change, which is checked against the **stored** value — the one new rule most worth documenting. | **ACCEPT** |
| 3 | `docs/specs/technical/architecture.md` route row went stale: `scheduler.py` grew from 243 lines and 7 routes to 332 and 8. | **ACCEPT** |
| 4 | The validation paragraph sat inside § Notification bot, splitting two bot paragraphs that read as one thought. | **ACCEPT** — moved. |
| 5-8 | The `/bots` description dropped the token requirement; "picking a name the API would reject is impossible" was contradicted by the orphan option on a live select; "disable-and-explain" mislabelled a mechanism that sets error text and reverts; and the type+value save was described as requiring the value to be edited, which it does not. | **ACCEPT** — all one-clause corrections. |
| 9 | `api-reference.md` at 803 lines against an 800-line guideline. | **REJECT** — informational, not this PR's to fix. |

All of the above are fixed. The narration gate (`tests/test_no_change_narration.py`) passed throughout, and the added lines were swept by eye for review-finding labels the ratchet cannot see.

**Convergence:** round 2's page-driving pass produced no Action Required and the surviving items were a one-line label fix and wording corrections, so the loop converged here.

nbramia:
## Verification — PASS

Run against the **real routers**, not stubs. A FastAPI app mounted the worktree's actual `api/routes/scheduler.py` and `api/routes/agents.py` over a throwaway temp vault, served `web/` the way `api/main.py` does, on uvicorn at `127.0.0.1:8137`. Port 8000 was never bound and was confirmed healthy afterwards. Exactly one external call was stubbed — `send_message_async`, replaced by a recorder. All request validation, store writes, vault markdown rewriting, next-fire computation, `mark_triggered` and `record_run` ran for real. Playwright, sync API, viewport 1280x800. All names synthetic.

The issue's named manual check is "edit a schedule from the board and confirm the vault line changes", so every check below is evidenced by the temp vault's markdown read back off disk.

| Check | Evidence |
|---|---|
| Edit the cron expression | PUT `{"schedule_value":"15 6 * * 1-5"}`. Vault line went from `[cron:: 0 7 * * *]` to `[cron:: 15 6 * * 1-5]`; preview refreshed from the response. |
| Convert to a one-off | Selecting the type alone issued **no** PUT. Entering the datetime and blurring sent ONE PUT with both fields. Vault key changed from `[cron:: …]` to `[at:: 2027-05-06T08:45:00]`, `next_trigger_at=2027-05-06T12:45:00+00:00` — a real next fire. Converting back restored `[cron:: 15 6 * * 1-5]`. |
| Invalid cron | 422 `Invalid cron expression 'not a cron': Exactly 5, 6 or 7 columns…`, shown inline, input snapped back, **vault byte-identical**. |
| Incompatible type change | 422 `Invalid ISO datetime '15 6 * * 1-5'`, inline, type select snapped back, vault unchanged — and a correct conversion still worked afterwards. |
| Executor | Empty "default route" option was the selected one for a stored empty executor. Picking `cloud-sonnet` added `#cloud-sonnet` to the vault line; picking "default route" removed it and stored `executor=''`. |
| Bot | Options came from the real `GET /api/scheduler/bots` (`primary`, `sample-notifier`, `sample-digest`) plus the empty default. Picking one added `[bot:: sample-digest]`. A direct PUT of an unlisted name returned 422 and changed nothing. |
| Trigger now | `POST /api/scheduler/{id}/trigger` fired the real entry; the Telegram recorder saw the message routed to the bot just saved; the last-run line and `[last:: …]` both updated. On the one-off the button reads `Trigger now (disables this one-off)`, and clicking it flipped the vault checkbox to `- [x]` with `enabled=False`, `next_trigger_at=None` — exactly as the label warns. |
| Timezone and enabled | `[tz:: Europe/Berlin]` moved the computed fire time; an unknown zone returned 422 inline and changed nothing; toggling enabled flipped the vault checkbox and updated the preview. |
| Board cards and lanes | The renamed title, the new next-fire, the recurring/one-off chip, the last-run summary, and the consumed one-off's move to Done all showed on the cards. |

**Console and network:** zero uncaught exceptions and zero failed network requests. Exactly four responses at or above 400, all deliberate 422s provoked by the checks above, each producing one Chromium resource-load log line and nothing else. No other console messages.

**VERIFY: PASS.**

### Full suite
Authoritative pre-push gate on the merged head: **7319 unit passed / 9 skipped; 593 `browser and not requires_server` passed.**

### Gate failures investigated rather than waved away
The gate failed twice on things this change did not cause; each was confirmed against a clean baseline before proceeding.

1. Four failures in `tests/test_merge_people.py` and `tests/test_pair_strength_migration.py`, all `sqlite3.OperationalError: no such table: relationships` on one xdist worker. They pass on a clean baseline and passed on the next run. The underlying cause is the relationship-store singleton isolation fixed on main and brought in by the main sync, not this change.
2. Two failures in `tests/test_agents_host_filter_ui_browser.py`, the Graph tab's filter, untouched here. Both pass in isolation on this branch and the baseline, and the full server-free browser selection passed 498/498 at `-n 4` under the lock in 106s against the gate's contended 401s. Root cause: the file's helper waits for `#filter-host` to exist, but that select ships in static markup with only its `all` option and the per-host options arrive after the snapshot fetch, so the assertion loses a race under load. **Filed as #958.**

### Rebases
The tip moved four times during this lifecycle and the branch was rebased each time: `f385f6d` (#951, a real conflict in `board.js` and the spec, resolved keeping both the schedule controls and the Delete affordance), `a7113b1` (#949), `424daa6` (#954, re-gated in full because it touches `assignment.js` and a shared test file), and finally `c54cda31` (the main sync), which is the squash-merge parent. That last one changes only CRM files, with zero overlap against this PR's eleven, verified by comparing changed-file lists.

Issue #864, opened 2026-09-04
feat(agents): redesign the graph encodings, labels, hover card, and shared drawer


nbramia:
## Problem

The graph tab on `/agents` is hard to use even when its data is correct. Node text shows a model badge instead of the session's name, so many nodes read the same word. Nine statuses collapse into five colours with no colour legend, and the two states an operator most needs to tell apart, waiting on you versus idle, are near-identical ambers. Four different routings share one circle whose legend entry names only Claude. Node size tracks cache-read tokens, so it measures caching, not work, and every remote or subagent row sits at the minimum size. The horizontal position is an unlabelled recency gradient that saturates at one day while the default window is a week, so most nodes pile into one column. Labels render around eight pixels high. Every click waits 220 ms for a double-click check that applies to a minority of nodes. The only hover affordance is a native tooltip. The graph panel and the board drawer share the transcript component but offer different action sets, so Answer, Accept, Resolve, and Open exist only on the board while Resume and Rename exist only on the graph.

## Solution

Redesign the graph as a session map that reads at a glance and shares one drawer with the board.

- **Label:** the node shows the linked card's title when one exists, else the session's human label, else the prompt preview; model, engine, host, and effort render as small chips, never as the label.
- **Colour** is the board lane, using the board's palette, with a visible legend. **Shape** is the engine with five distinct glyphs and a correct legend: Claude Code, Codex, Hermes/DeepSeek, Gemma local, Anthropic API. **Size** is active seconds with tool-call count as a secondary ring. **Badges:** a question ring only when a pending question exists for the operator, an error count, and the existing live pulse.
- **Layout:** labelled columns by host, with a lane sub-order inside each column; subagent trees collapse into their parent with a count badge and expand on click. Recency stays as a filter, not a position.
- **Interaction:** an HTML hover card with name, host, branch or directory, model and effort, cost, duration, last event; single click opens the drawer with no artificial delay; double-click on a CLI node focuses its terminal.
- **One drawer:** the board drawer and graph panel render their actions from one shared component, so both offer Open, Focus, Resume, Kill with cascade preview, Rename, Answer, Accept, and Resolve where applicable.
- **Controls:** zoom-to-fit and reset buttons; the simulation restarts when node sizes change, not only when the visible set changes.

## Acceptance Criteria

- [ ] Given a node linked to a card, its label is the card title; given an unlinked node, its label is the snapshot `label`; the model badge never appears as the label text.
[...]

PR #961, opened 2026-09-08
feat(agents): render one shared action set in the board drawer and the graph panel (#864)

nbramia:
## Summary

PR B of the `/agents` Kanban milestone. Three items:

**(a) One shared drawer/panel action component.** Extracts `web/agents/session_actions.js`: a pure `decideActions(session, card)` decision function plus `renderActionRow`, the shared DOM builder both the Board tab's card drawer (`board.js`) and the Graph tab's side panel (`panel.js`'s `SessionPanel`) call to render Open, Go To, Resume, Kill, Answer, Accept, Resolve, Cancel, and Delete — same order, same enabled/disabled state, same reason text, on both surfaces. Kill's confirmation modal (with its cascade preview of non-terminal descendants), Resume's host select + copy-command box, and Go To's pane lookup all move into `session_actions.js` so neither surface reimplements them — previously the board drawer's own action row and the embedded session panel's header independently built overlapping button sets (board's Kill fired immediately with no confirmation/cascade preview; the panel's Kill/Resume/Go To never showed on the board drawer's own row at all). The Board drawer's embedded `SessionPanel` is now constructed with `showActions: false` so the drawer's own row is the one place these actions render there.

**(b) Go To on a routing-derived subagent.** `isSubagentSession(s)` (`session_actions.js`) tests `is_subagent || parent_session_id`, matching `graph.js`'s existing node-rendering check — both `showResumeFor` and Go To's eligibility now use it, so a session with `parent_session_id` set but no `is_subagent` flag is refused Resume and Go To on both surfaces, and `graph.js`'s own dblclick handler now calls the same shared predicate instead of inlining the check a second time.

**(c) Node label legibility at zoom-to-fit.** Chose the hide-below-threshold approach: `updateLabelLegibility(k)` (`graph.js`) hides every `.node-label` when the zoom transform's scale drops below `0.5` (a plain zoom-scale check, not a computed on-screen-pixel estimate — simpler and robust regardless of window size), relying on the hover card (already independent of the node's own DOM label) to name a node instead. The identity transform (`k = 1`, what Reset restores) never crosses the threshold, so labels are always shown at the resting zoom level.

## Acceptance criteria

- [x] Board drawer and Graph panel render identical action sets for the same session — proven directly: a browser test renders the Board drawer for a card+session, separately mounts a second `SessionPanel` with the same card+session, and asserts the extracted `{id, label, enabled, reason}` sequences are byte-identical.
- [x] Every existing behaviour preserved: disable-and-explain pattern, Delete's confirmation + kill-first-for-a-killable-session path, Kill's cascade preview, Cancel's `failures` warning toast, Resume's host select + copy-command box, Go To's "Locating…" state + pane-missing toast, the click-to-rename flow (unchanged — already single-sourced in `SessionPanel`).
- [x] Routing-derived subagent refused Resume/Go To on both surfaces — proven via `decideActions` unit tests and via `graph.js`'s dblclick handler using the same predicate.
- [x] Node labels legible at any zoom — proven via a 100-node zoom-to-fit browser test: labels present at default zoom and after Reset, absent at Fit, hover card still names a node at Fit.

## Test plan

- `tests/test_agents_action_parity_browser.py` (new, 16 tests): `TestDecideActions` drives `decideActions` directly (no DOM) covering every branch — bare session, routing-derived and flagged subagents, running/terminal/non-CLI sessions, every card lane, cancel's disabled reason, delete's danger flag, and the full canonical order with every action present at once. `TestActionParity` renders the same session+card on the real Board drawer and a directly-mounted `SessionPanel` and asserts identical rendered action rows, for two lane scenarios (assigned-lane rich scenario; review-lane Accept).
- `tests/test_agents_graph_label_legibility_browser.py` (new, 4 tests): stubs a 100-node snapshot concentrated into one host/lane cell (so the collision force spreads the layout wide enough to force `k` well below the threshold), asserts labels visible at default zoom, hidden at Fit, reappear at Reset, and the hover card still names a node at Fit.
- Full existing coverage re-run and green: `test_agents_board_ui_browser.py` (105), `test_agents_resume_here_ui_browser.py` (18 — including the "Resume/select stay mounted, hidden, and reveal in place on a status change while the drawer holds focus" tests, which the new `renderActionRow`'s `visible` field, distinct from `enabled`, was specifically designed to preserve), `test_agents_graph_redesign_ui_browser.py` (33), `test_agents_graph_encoding_browser.py` (41), `test_agents_host_filter_ui_browser.py` (26), `test_agents_assignment_ui_browser.py` (31), `test_agent_threads_ui_browser.py`, `test_no_change_narration.py`.
- Every new/modified test has a recorded mutation proof (production hunk mutated → test fails on the positive assertion → restored → passes) in the implementation report.

## Notes

- Chose **"Go To"** as the canonical label for the shared focus action (not "Focus") — the codebase's existing documentation (`docs/guides/agents-go-to.md`, the product/technical agent-viz specs' "Operator controls — resume and Go To" sections) already established "Go To" as the feature's name; unifying on it is a smaller, lower-risk doc diff than renaming that established surface. The internal action id/CSS hooks stay `focus`/`canFocusFor`/`.panel-focus`, matching the pre-existing `/focus` endpoint naming on both surfaces already.
- `docs/specs/product/agent-viz.md` and `docs/specs/technical/agent-viz.md` updated for the shared action row, the module split (`session_actions.js`), the subagent-check fix, and the label-legibility rule. `docs/guides/agents-go-to.md` updated to drop the now-stale "board drawer labels it Focus" note.


nbramia:
# PR #961 (PR B, #864) — review round 1 referee table

Reviewers: Opus general, Opus hostile (real routers on :8123 over a temp vault, 1280x800), Codex.

| # | Finding | Reviewers | Verdict | Severity |
|---|---|---|---|---|
| 1 | The Graph tab calls `panel.open(s)` / `updateMeta(s)` with no card, so `decideActions` skips Open, Accept, Resolve, Cancel and Delete. The two surfaces render different sets for the same session; a review-lane session's graph panel offers nothing at all. The parity test mounts its own panel and passes a card, a call the product never makes. | general, hostile, codex | ACCEPT | Action Required |
| 2 | `board.js` passes `getDescendants: () => []`, so Kill's cascade preview is always empty on the board while the graph lists descendants. The endpoint does cascade, so the operator confirms a 3-session kill believing it is one. | hostile, codex (general as Recommended) | ACCEPT | Action Required |
| 3 | Label legibility is not fixed. Labels are 12 px in SVG user space against a 1600x1100 viewBox rendered into a 920 px element, so on-screen size is 12 x k x 0.575. Measured 6.90 px at rest and after Reset, 4.16 px after Fit at 72 nodes; the `k < 0.5` rule corresponds to 3.45 px and never fires at realistic node counts. The new test asserts `display` only, and asserts labels visible at k=1, pinning the defect. | general, hostile, codex | ACCEPT | Action Required |
| 4 | Rename is not part of the shared action set; it stays a separate clickable label in the panel. The acceptance criteria name Rename in the identical set. | codex | ACCEPT | Action Required |
| 5 | `panel.js` calls `renderActionRow` without handlers, and the shared module only wires the card-only actions when handlers are supplied, so a card-aware panel would render enabled dead buttons. Must be resolved as part of finding 1. | codex | ACCEPT | Action Required |
| 6 | The disabled-Kill reason div is appended into `.panel-header-actions`, which this PR floats right, so on the graph panel the reason wraps against the buttons and displaces the cwd and branch lines. Measured buttons x1159-1256 y234-258 against reason x942-1264 y238-266. New in this PR. | hostile | ACCEPT | Action Required |
| 7 | The action-row signature omits the pending-question id, so an answered-then-replaced question leaves Answer posting to the old id. | hostile (Minor), codex (Action Required) | ACCEPT | Recommended |
| 8 | The parity comparison iterates a hard-coded id list with `querySelector`, so it cannot catch wrong DOM order, duplicates, or an unexpected extra action. | codex | ACCEPT | Recommended |
| 9 | No test drives the real Graph tab and compares its action row against the board drawer for the same session. | codex | ACCEPT | Recommended |
| 10 | The label test must assert the rendered on-screen size via the screen CTM, not `display` alone. | general, hostile, codex | ACCEPT | Recommended |
| 11 | Both spec files describe behaviour that does not ship: the board Kill cascade preview and the label-legibility rationale. | general (Minor), hostile (Recommended) | ACCEPT | Recommended |
| 12 | The parity test leaves `/api/agents/board/stream` unresolved rather than fulfilling an empty event stream. | codex | REJECT — test hygiene, no defect on any path | Minor |

Cross-model overlap: all three reviewers independently found findings 1 and 3. Findings 4 and 5 are Codex-only. Finding 6 is hostile-only and came from driving the real page. Finding 2 was raised by all three but only the page-driving passes established it as a real operator hazard.

nbramia:
# PR #961 (PR B, #864) — review round 2 referee table

Reviewers: Opus general (drove the real page at 1280x800 and three other viewports, with measurements — this was the round's page-driving pass) and Codex. A separate hostile pass was not run: the general reviewer's pass already drove the page and measured, which is what the hostile brief asks for.

| # | Finding | Reviewers | Verdict | Severity |
|---|---|---|---|---|
| 1 | The counter-scale raises `font-size` but leaves the label layout metrics at their 12 px tuning — character width, line height, the fixed user-space `dy`, and the collision radius. At 1280x800 the boost is always on at rest (12 px becomes 19.13 px), so every multi-line label draws each line 7.53 px over the previous one, and neighbouring labels collide: 2 overlapping pairs at 8 nodes, 4 at 20, 10 at 40, against 0 at the 12 px baseline. An always-on regression introduced by the round-1 fix, invisible to the new tests because they measure font size only. | general, codex (as Recommended) | ACCEPT | Action Required |
| 2 | `svgScreenScale()` computes `rect.width / VIEW_W` and ignores `preserveAspectRatio="xMidYMid meet"`, whose real scale is the smaller of the width and height ratios. The 11 px floor therefore holds only above a 763.5 px page viewport. Measured: 10.70 px with the panel at its minimum width, 9.03 px at 1280x650, 8.43 px at 1440x700, 4.03 px at 1920x500. A maximized browser on a 1280x800 screen sits below the break-even. The test viewport of 800 px is 36 px above it, which is why every reported measurement read exactly 11.00. | general, codex | ACCEPT | Action Required |
| 3 | Nothing tests that the graph panel's card-only buttons are actually wired. Deleting the entire round-1 finding-5 fix leaves 50 tests green, because the extraction helper never reads `onclick`. An accepted fix with no test that fails without it. | general | ACCEPT | Action Required (quality bar: every criterion maps to a test that fails without the change) |
| 4 | The board's descendant lookup returns an empty list on fetch failure, so a failed lookup is indistinguishable from a session with no descendants, and the operator confirms a cascading kill shown as a single-session kill. The confirm button is also enabled before the preview resolves. | codex | ACCEPT (the failure-versus-none conflation as Action Required; gating the confirm on resolution folded in) | Action Required |
| 5 | Graph-tab Delete is constructed without `findCard`, so unlike the board drawer it decides its confirmation text and its kill-first decision from the card captured when the panel opened rather than the live one. | general, codex | ACCEPT | Recommended |
| 6 | The Resume row is rebuilt on a status flip because `visible` is part of the rebuild signature, discarding a chosen resume host or a visible copy-command box, while the code comment claims the elements are mounted once and merely revealed. | general | ACCEPT | Minor — fold into the same address |
| 7 | The product spec says labels are hidden rather than shrunk further; the code grows them and hides only past the boost cap. Docs must match what ships. | general | ACCEPT | Recommended |
| 8 | The cascade preview reads `/api/agents/snapshot`, which starts from `list_sessions(limit=200)`, while the kill endpoint walks the full subtree, so a preview could omit sessions the kill will terminate. | codex | REJECT for this PR — the fix is a new backend endpoint, beyond this PR's scope, and it needs a subtree deeper than the 200 most recent sessions to bite. Filed as a follow-up issue. | Recommended |

Cross-model overlap: the general reviewer and Codex independently found the `preserveAspectRatio` defect with different measurements (9.03 px at 1280x650 versus 4.03 px at 1920x500) and both flagged the stale collision metrics; the general reviewer alone found the line-overlap regression and the untested handler wiring, and Codex alone found the failure-versus-none conflation in the descendant lookup.

**Not converged.** Four Action Required findings; findings 1 and 2 are coupled and must be fixed together, since correcting the scale makes the boost larger.

nbramia:
# PR #961 (PR B, #864) — review round 3 referee table

Reviewer: Opus hostile, driving the real page against the real routers over a temp vault, with A/B servers serving the pre-PR tree and the round-1 tree against identical data. Codex was attempted and returned "Your workspace is out of credits" on every call, so the second independent Opus page-driving pass stands in for it, as the standing brief provides.

| # | Finding | Verdict | Severity |
|---|---|---|---|
| 1 | Zooming re-runs the force layout. The label legibility update is called from the zoom handler and reheats the simulation whenever the boost factor moves, and the collision radius now depends on that boost, so a settled graph re-lays-out on every zoom. Measured max node displacement 2-6 s after one wheel notch on a settled graph: 106.2 user units at 40 nodes and 524.2 at 100, against 0.0 on both the pre-PR and round-1 trees. Fit then invalidates its own transform: 5 s after clicking Fit at 40 nodes and 1280x800, 3 of 31 node centres sit outside the rendered SVG box with label text clipped, where the pre-PR tree leaves 0 of 31 outside. The graph also slides under the cursor for seconds after each notch, and the reheat produces transient label pile-ups that at-rest measurements never show. New in round 2. | ACCEPT | Action Required |
| 2 | The boost-scaled collision radius inflates the resting layout, so more of the graph starts outside the default view. Measured at 1280x800 on identical data: bounding box 1480x1339 with 7 of 31 nodes off canvas at 40 nodes, against 1204x984 and 2 of 31 pre-PR; at 100 nodes 1959x2443 with 47 of 78 off canvas, against 1454x1855 and 27 of 78. Fit compensates by zooming further out, k 0.825 to 0.654 at 40 nodes and 0.476 to 0.422 at 100, which crosses the boost cap and hides all 78 labels where the pre-PR tree showed them. Same root cause as finding 1. | ACCEPT | Action Required |
| 3 | Delete's confirmation builds its target line from the card captured when the panel opened, so a card renamed while the panel is open shows the stale title. The kill-first decision itself is correctly re-resolved through the live card. | ACCEPT — fold into the same address | Minor |
| 4 | While the descendant lookup is pending, Kill's confirm is disabled with an empty preview area and no explanation, so a slow lookup reads as a dead button. Cancel keeps working and a genuinely failed lookup discloses itself correctly. | ACCEPT — fold into the same address | Minor |

Verified as genuinely fixed by this round's measurements: criterion (c) holds at 11.00 px in every configuration where labels are shown, across 8, 20, 40 and 100 nodes by 1280x800, 1280x650 and 1920x500, at rest, after Fit and after Reset, against 6.90 px pre-PR; the line pitch ratio is 1.083 boosted, identical to the unboosted and pre-PR baselines, against 0.68 in round 1; Kill's gating behaves correctly on a hung, aborted, empty and populated lookup; criterion (a) parity holds in set, order, enabled state and reason text across five card shapes, with every graph-panel click firing the right request; criterion (b) holds; Resume survives a status flip in place with its chosen host; and console and network are clean across twelve label configurations and every action run.

**Not converged.** Both Action Required findings share one root cause: zoom-dependent label geometry feeding back into the force layout.

PR #966, opened 2026-09-08
test(agents): wait for the snapshot before asserting the graph host filter's options

### LINKED ISSUE #958: test(agents): wait for the snapshot before asserting the graph host filter's options

nbramia:
Closes #958 (issue stays open for the milestone orchestrator).

## What this does

`tests/test_agents_host_filter_ui_browser.py` asserts on state that only exists after `GET /api/agents/snapshot` resolves and `web/agents/graph.js` applies it, but its open helpers wait only for elements that ship in the static markup. `#filter-host` is present with just its `all` option before any data arrives, so waiting for the element proves nothing about the data. Under parallel load the assertion wins the race and two tests fail spuriously, costing a full gate cycle.

Every helper in the file now waits on observable output instead:

- `_open_agents` waits for `#filter-host` to carry more than its `all` option, which is exactly the snapshot-derived state its tests read.
- `_open_agents2`, `_open_agents3` and `_open_agents4` wait for the first `.node` to render.
- In-test fixed sleeps are gone. Where the following assertion is a Playwright `expect(...)`, the sleep was dead weight and is removed; where the following read is non-retrying (`page.evaluate`), it is replaced by an `expect(...).to_have_count(...)` on the nodes that read depends on. No `wait_for_timeout` remains in the file.

`tests/test_agents_graph_redesign_ui_browser.py`'s module-level `_open_agents` carries the same shape, with a fixed 500 ms sleep in front of non-retrying `page.evaluate` reads, so it gets the same node wait. Nothing else in that file changes.

No product code changes. The file stays server-free with no `requires_server` marker.

## Proof the waits are load-bearing

- Stubbing the snapshot to return no sessions makes both `TestHostFilter` tests fail on the wait rather than pass vacuously or hang: `2 failed in 31.62s`, each with `TimeoutError: Page.wait_for_function`. Restoring the file returns `2 passed in 2.34s`.
- Serving a genuinely slow `/api/agents/snapshot` from the fixture's own HTTP server reproduces the issue's exact failure signature against the pre-change file (`assert 'laptop-a' in ['all']` and `Page.select_option: Timeout 30000ms exceeded ... did not find some options`), and the changed file passes the same scenario.

## Verification

- `tests/test_agents_host_filter_ui_browser.py`: 26 passed, at `-n 4`, three runs before and three after.
- `tests/test_agents_graph_redesign_ui_browser.py`: 33 passed.
- Pre-push gate: 7330 unit passed, and the full `browser and not requires_server` selection 604 passed.

### DIFF
--- a/tests/test_agents_graph_redesign_ui_browser.py
+++ b/tests/test_agents_graph_redesign_ui_browser.py
@@ -176,7 +176,7 @@ def _open_agents(page: Page, base_url, snapshot=None, focus_calls=None, search_m
     page.locator("#filter-terminal").check()
     # Nodes render synchronously once the filters above are applied — wait
     # for the first one rather than for a fixed delay.
-    page.wait_for_selector(".node", timeout=8000)
+    page.wait_for_selector(".node")
     return errors
 
 
--- a/tests/test_agents_graph_redesign_ui_browser.py
+++ b/tests/test_agents_graph_redesign_ui_browser.py
@@ -174,7 +174,9 @@ def _open_agents(page: Page, base_url, snapshot=None, focus_calls=None, search_m
     page.wait_for_selector("#filter-route")
     page.select_option("#filter-recency", "all")
     page.locator("#filter-terminal").check()
-    page.wait_for_timeout(500)
+    # Nodes render synchronously once the filters above are applied — wait
+    # for the first one rather than for a fixed delay.
+    page.wait_for_selector(".node", timeout=8000)
     return errors
 
 
--- a/tests/test_agents_host_filter_ui_browser.py
+++ b/tests/test_agents_host_filter_ui_browser.py
@@ -165,8 +165,7 @@ def _open_agents(page: Page, base_url):
     # applies it, so wait for those options rather than for the select
     # itself, which exists in the markup before any data arrives.
     page.wait_for_function(
-        "document.querySelector('#filter-host').options.length > 1",
-        timeout=8000,
+        "document.querySelector('#filter-host').options.length > 1"
     )
 
 
@@ -443,7 +442,7 @@ def _open_agents2(page: Page, base_url):
     page.locator("#filter-terminal").check()
     # Nodes render synchronously once the filters above are applied — wait
     # for the first one rather than for a fixed delay.
-    page.wait_for_selector(".node", timeout=8000)
+    page.wait_for_selector(".node")
 
 
 class TestRecentChipAndRouteFilterAndNodeLabels:
@@ -682,7 +681,7 @@ def _open_agents3(page: Page, base_url):
     page.locator("#filter-terminal").check()
     # Nodes render synchronously once the filters above are applied — wait
     # for the first one rather than for a fixed delay.
-    page.wait_for_selector(".node", timeout=8000)
+    page.wait_for_selector(".node")
 
 
 class TestSearchDropdownRawIdGuard:
@@ -826,7 +825,7 @@ def _open_agents4(page: Page, base_url, search_matches=None):
     page.locator("#filter-terminal").check()
     # Nodes render synchronously once the filters above are applied — wait
     # for the first one rather than for a fixed delay.
-    page.wait_for_selector(".node", timeout=8000)
+    page.wait_for_selector(".node")
 
 
 class TestSearchDropdownLabelSourcePrecedence:
--- a/tests/test_agents_host_filter_ui_browser.py
+++ b/tests/test_agents_host_filter_ui_browser.py
@@ -160,7 +160,14 @@ def _open_agents(page: Page, base_url):
     # #850: the board is the default tab; the graph (and its filters) live
     # behind the Graph tab and only initialize once it's opened.
     page.click('[data-tab="graph"]')
-    page.wait_for_selector("#filter-host")
+    # The host select ships with only its `all` option; the per-host options
+    # are appended once `GET /api/agents/snapshot` resolves and the graph
+    # applies it, so wait for those options rather than for the select
+    # itself, which exists in the markup before any data arrives.
+    page.wait_for_function(
+        "document.querySelector('#filter-host').options.length > 1",
+        timeout=8000,
+    )
 
 
 class TestHostFilter:
@@ -180,7 +187,6 @@ class TestHostFilter:
         # widen recency to "all time" first.
         page.select_option("#filter-recency", "all")
         page.select_option("#filter-host", "laptop-a")
-        page.wait_for_timeout(400)  # let the filtered re-render settle
         nodes = page.locator(".node")
         expect(nodes).to_have_count(1)
 
@@ -194,14 +200,12 @@ class TestEndedStatus:
     def test_ended_session_hidden_by_default(self, page: Page, agents_base_url):
         _open_agents(page, agents_base_url)
         page.select_option("#filter-recency", "all")
-        page.wait_for_timeout(400)
         expect(page.locator(".node")).to_have_count(2)
 
     def test_ended_session_shown_when_include_finished_is_checked(self, page: Page, agents_base_url):
         _open_agents(page, agents_base_url)
         page.select_option("#filter-recency", "all")
         page.locator("#filter-terminal").check()
-        page.wait_for_timeout(400)
         expect(page.locator(".node")).to_have_count(3)
 
 
@@ -437,7 +441,9 @@ def _open_agents2(page: Page, base_url):
     page.wait_for_selector("#filter-route")
     page.select_option("#filter-recency", "all")
     page.locator("#filter-terminal").check()
-    page.wait_for_timeout(400)
+    # Nodes render synchronously once the filters above are applied — wait
+    # for the first one rather than for a fixed delay.
+    page.wait_for_selector(".node", timeout=8000)
 
 
 class TestRecentChipAndRouteFilterAndNodeLabels:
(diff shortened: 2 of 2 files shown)


nbramia:
## Review round 1 — three reviewers, one accepted finding, converged

Reviewers: Opus general, Opus hostile (mutation-driven), Codex. Codex began the pass but died partway through with `Your workspace is out of credits`, so the hostile reviewer stands in for it, per the milestone's substitution rule.

| Reviewer | Action Required | Recommended | Minor | Accepted |
|---|---|---|---|---|
| general (opus) | 1 | 0 | 0 | 1 |
| hostile (opus) | 0 | 0 | 1 | 0 |
| codex | died out of credits | — | — | — |

### Accepted — the explicit `timeout=8000` shrank the tolerance the old code already had

The first commit put `timeout=8000` on every new wait. The general reviewer showed that this is a regression against the very failure the PR fixes. The pre-change helper ended at `wait_for_selector("#filter-host")`, so a late snapshot was absorbed by `select_option`, which retries for Playwright's default 30 s. The issue's own transcript shows the host options absent for more than 30 seconds under gate load, so an 8 s ceiling fails hard where the old code passed.

Reproduction the reviewer ran, serving a genuinely slow `/api/agents/snapshot` from the fixture's own HTTP server at a 15 s delay:

```
SNAPSHOT_DELAY=15
  test_before (wait_for_selector)      PASSED
  test_after  (wait_for_function 8s)   FAILED  TimeoutError: Page.wait_for_function: Timeout 8000ms exceeded
1 failed, 1 passed in 24.69s
```

Fixed in `88b75e6a`. The explicit timeout is dropped from all five new waits so they inherit the standard 30 s budget. Confirmed first that nothing in `tests/conftest.py`, `pyproject.toml`, or any pytest config lowers that default. The same probe at a 15 s delay now passes at 15.2 s where the 8 s version failed at 8.2 s.

The mutation proof was re-run after the change: stubbing the snapshot to return no sessions makes both `TestHostFilter` tests fail on the wait at 30 s rather than pass vacuously, and restoring the file returns `2 passed in 2.16s`.

### Recorded, not actioned

The hostile reviewer noted that a genuine product regression now takes the full 30 s per test to go red instead of about 1 s. Measured: four tests fail in 124 s under a mutated `updateHostOptions`. That is the deliberate consequence of the accepted finding above, and the tests still fail, which is the property that matters. No change.

### What the hostile reviewer could not break

Every mutation was applied to the product code in a throwaway worktree and reverted immediately after.

| Attack | Result |
|---|---|
| `updateHostOptions` never appends per-host options | 4 failed — not masked |
| `renderGraph` renders nodes with no label content | 6 failed across both files — not masked |
| `applyFilters` stops filtering by host | 1 failed on `to_have_count(1)` — the assertion that replaced a removed sleep catches it |
| search-dropdown raw-id guard removed | 6 failed — not masked |
| stubbed `/api/agents/search` delayed 3 s | 10 passed — no half-rendered read |
| transient-true `to_have_count(N)` | no normal-path sequence found; each render is one synchronous burst and the stream stub is empty |
| `wait_for_selector(".node")` visibility on an SVG `g` | resolves at all five helper sites and every call site |
| zero-node hang at a helper call site | none of the 32 call sites can render an empty visible set |
| explicit reverse test order | 26 passed |
| loaded box, 24 CPU burners plus a concurrent full browser suite, loadavg 11.6 | 26 passed in 16.97s |

The decisive one is the slow-snapshot comparison. With a 3 s delay injected on `/api/agents/snapshot`:

```
head 88b75e6a   59 passed in 219.73s
base 04cb92f6   46 failed, 13 passed in 153.51s
```

The old code collapses under exactly the condition the PR claims to fix, and the new waits cover it.

Full reviewer reports: `/home/nathanramia/.claude/lifeos-milestone/reports/958/`.

nbramia:
## Verify — PASS

Opus verifier, head `88b75e6a`, base `04cb92f6`. Every run at `-n 4` through the two-slot gate semaphore with CPU-only embeddings.

### The file alone at `-n 4`, three runs each side

| Tree | Result | pytest time |
|---|---|---|
| after | 26 passed | 6.51s |
| after | 26 passed | 8.94s |
| after | 26 passed | 11.07s |
| before | 26 passed | 9.54s |
| before | 26 passed | 9.61s |
| before | 26 passed | 9.46s |

### The file at `-n 4` with a second suite resident in the other gate slot

The load generator was the repo's own `browser and not requires_server` selection at `-n 4`, so two four-worker suites were in flight throughout.

| Tree | Result | pytest time | loadavg at start |
|---|---|---|---|
| before | 26 passed | 9.84s | 3.73 |
| after | 26 passed | 6.69s | 3.93 |
| before | 26 passed | 9.50s | 4.33 |
| after | 26 passed | 6.73s | 4.35 |
| before | 26 passed | 9.85s | 4.80 |
| after | 26 passed | 6.74s | 5.29 |

**Honest non-reproduction.** The pristine base did not fail under this load. Three loaded runs of the unpatched file at loadavg 3.7 to 4.8 all passed. This box was not slow enough today to trip the base's race, so these runs are a no-regression result rather than an independent reproduction of the flake. No artificial delay was injected here to manufacture a failure. The reproduction evidence is the review round's slow-snapshot comparison, where the base fails 46 tests and the head passes 59.

### Full server-free browser selection at `-n 4` on the head

604 passed twice, 137.55s and 125.58s, exit 0. Both named tests reported individually:

```
[gw2] PASSED tests/test_agents_host_filter_ui_browser.py::TestHostFilter::test_filter_host_lists_every_host_from_the_snapshot[chromium]
[gw2] PASSED tests/test_agents_host_filter_ui_browser.py::TestHostFilter::test_selecting_a_host_hides_sessions_from_other_hosts[chromium]
```

### Per criterion

1. **Waits on snapshot-derived state, no fixed sleep — PASS.** `options.length > 1` cannot be true before the snapshot lands, because the select ships with only its `all` option.
2. **Passes under the full selection at `-n 4` on a loaded machine — PASS.** 604 passed twice, plus twelve measured runs of the file with zero failures.
3. **No fixed sleep left where a wait on output would do — PASS.** `wait_for_timeout` count in the file goes 14 to 0.
4. **Still fails when the options never populate — PASS.** Stripping every `host` field from the fixture gives `1 failed in 31.55s` with `TimeoutError: Page.wait_for_function: Timeout 30000ms exceeded`, not a silent pass.

The verifier removed its scratch worktree and mutation file and confirmed both gate slots released.

Issue #975, opened 2026-09-09
feat(journal): log under LifeOS/Log/Journal and file work instead of doing it


nbramia:
## Problem

The journal bot writes the day's capture log into the personal-notes area of the vault, next to handwritten material, instead of with the rest of LifeOS's own files. That makes capture look like a personal notebook rather than an operational log, and it sits beside a different, generated daily journal that this bot must never touch.

Separately, journal mode is only a capture-and-file surface in intent, but it still has every LifeOS action available. A fragment like "create a calendar event to call mom tomorrow at 3" can become a real calendar event rather than an Inbox task. "Remind me to call mom tomorrow at 3" can be answered as a chat clarification instead of a Telegram reminder at that time. The bot can also run the work itself — spawn an agent, send mail, complete a task — rather than only filing it. Tags the operator named, including ones that later get a card picked up, should be honored; tags the bot invents so it can "help" should not. The operator sent a note, not a request to finish the work.

## Solution

Capture writes each fragment into the LifeOS-owned daily log (`LifeOS/Log/Journal/YYYY-MM-DD.md`), with the same verbatim bullet shape and frontmatter as today. Existing day files under the old personal-notes log path are moved once into the new folder so history stays in one place. The generated daily journal is still off-limits.

On a journal turn the model may only capture (already done in code), file, ask one clarifying question, or answer a real question with read tools. Filing has two destinations, chosen by what the note is asking for:

- A **to-do** (including "create a calendar event to call mom tomorrow at 3") → a normal Inbox task. Optional due date when a day is named. Unassigned unless the operator explicitly wrote a tag or assignment in that message.
- A **timed ping** ("remind me to call mom tomorrow at 3") → a one-shot Scheduler Inbox entry whose action is notify, so Telegram sends the reminder at that time. It does not create a calendar event and does not run an agent.

Tags that appear in the user message — including `#claude`, `#agent`, and any other pickup tag — are copied onto the item that was filed. Tags and assignment fields the model invents are not. Journal never orchestrates: no calendar create/update/delete, no mail, no agent spawn, no task complete, no schedule whose action is prompt, endpoint, or agent.

Native journal turns skip the generic "task vs reminder" and Claude-Code interceptors so "remind me…" reaches filing instead of the main-bot clarification. The native tool list for journal is an allow-list: search/read, task create/list, and notify-only schedule create/list. Hermes still receives the rewritten preamble; a cross-repo MCP allow-list is not part of this change.

## Acceptance Criteria

[...]

PR #1051, opened 2026-09-11
feat(journal): capture under LifeOS/Log/Journal and file work instead of doing it

nbramia:
## TL;DR

The journal bot now writes each captured fragment into LifeOS's own log directory (`LifeOS/Log/Journal/`) instead of sitting next to handwritten notes, and any day files left over from before this change are moved into the new location automatically. Separately, journal is now strictly a filing surface: on a journal turn it can capture, file a to-do or a timed reminder, ask one clarifying question, or answer with read-only tools — but it can no longer touch a calendar, send mail, spawn an agent, complete a task, or create anything but a plain notification schedule. Tags the operator actually typed in their message are copied onto what gets filed; anything the model invents is stripped.

Part of #975

## Design

**Path choice.** `LifeOS/Log/` is a container directory whose sibling `LifeOS/Log/Pebble/` is already owned by the Pebble producer, so capture gets its own `LifeOS/Log/Journal/` subdirectory rather than writing into the container itself.

**Migration scope — the safest superset of what the issue's Acceptance Criteria and Scope sections each say.** The Acceptance Criteria talk about moving "that day file" (today's); the Scope section talks about moving "leftover day files" generally. Rather than picking one, capture now migrates *every* leftover legacy day file, not just today's: `_migrate_other_legacy_day_files` sweeps all of them in an idempotent, race-tolerant way (a plain atomic `os.replace`, skipped if a destination already exists), and today's own file is migrated a second, more specific way — inside the exact same `flock` the append itself takes — because that's the one file genuinely at risk of two fragments racing on the first capture of the day. Neither path ever copies, clobbers an existing destination, or leaves a duplicate behind.

**Tool-catalog narrowing over prompt-only enforcement.** The issue notes that a prompt instruction alone isn't enough — the static system prompt still advertises calendar/email tools. So the native agent loop now takes a `persona_id`, builds a persona-scoped tool list (`tools_for_persona`) that drops calendar/email/memory/workout/human-queue tools for `journal`, and enforces the same boundary again at execution time (`_journal_tool_gate`) in case the model names an excluded tool or a disallowed `manage_tasks`/`manage_schedules` action anyway. The engine-handoff directive and the generic task-vs-reminder classifier are both skipped for journal turns so "remind me to…" reaches journal's own filing policy instead of being intercepted upstream.

## Implementation Notes

- `api/services/journal_capture.py` — `_LOG_DIR_PARTS` is now `("LifeOS", "Log", "Journal")`, generalized to a full-tuple join in `log_path_for()` and the containment guard rather than an indexed `[0]`/`[1]` pair. `_migrate_other_legacy_day_files` and the in-flock `_take_legacy_content` handle migration as described above.
- `api/routes/vault.py` — the `Personal/Journal/` reservation's error text and docstring now name `LifeOS/Log/Journal/` as the capture target.
- `config/personas/journal.md` — rewritten: capture target, "file the work, never do it", the notify-schedule-vs-Inbox-task split with the two examples the issue specifies, and explicit-tag-only passthrough.
- `api/routes/chat.py` — `parse_engine_directive` and `classify_action_intent` are both skipped when `_effective_pid == JOURNAL_PERSONA_ID`; `persona_id` and the raw `user_message` are threaded into `run_agent_loop`.
- `api/services/agent_loop.py` — `run_agent_loop` takes `persona_id`/`user_message`, uses `tools_for_persona(persona_id)` instead of the raw `TOOL_DEFINITIONS`, and threads both through to `execute_tool_parallel`.
- `api/services/agent_tools.py` — `tools_for_persona` (advertised-tool filter, preserving the cache-control breakpoint on whatever tool ends up last), `_journal_tool_gate` + `_journal_filter_task_create_input` (execution-time rejection of `manage_tasks` complete/update and non-notify `manage_schedules`, plus tag/field attestation against the user's own message), and `bot`/`timezone` added to the native `manage_schedules` schema and threaded into `SchedulerStore.create`/`update` (parity with `POST /api/scheduler`).
- Docs: `docs/guides/personas.md`, `docs/guides/journal-ring-ingest.md`, `docs/guides/telegram-setup.md`, `docs/guides/configuration.md`, `docs/specs/technical/client-surfaces.md` updated to the new path; `personas.md` also corrects the now-inaccurate "all six share the full tool suite" claim.

### Automated tests
`HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" python3 -m pytest tests/test_journal_capture.py tests/test_journal_persona.py tests/test_journal_tool_gating.py tests/test_journal_native_interceptor_skip.py tests/test_journal_ring_ingest.py tests/test_vault_write_route.py tests/test_hermes_proxy.py tests/test_no_change_narration.py -q`
→ 235 passed.

Also ran the broader adjacent suites for regressions (`test_agent_loop_*`, `test_agent_tools_*`, `test_chat_api.py`, `test_persona_api.py`, `test_scheduler_*`) — 978 + 257 collected across two runs, all passed.

New coverage: `tests/test_journal_tool_gating.py` (advertised-tool filtering, execution-time rejection of disallowed actions, tag/field attestation, `bot`/`timezone` persistence), `tests/test_journal_native_interceptor_skip.py` (engine-directive and intent-classifier skip on journal turns), plus new `TestLegacyMigration` cases in `tests/test_journal_capture.py` (today's-file move, other-day sweep, no-clobber, concurrent-race safety).

### Manual verification
N/A — no runnable artifacts outside the automated suite above; this is a feature-branch PR and does not touch the deployed server.

## Review focus

- The migration-scope decision (superset of both AC and Scope readings) in `api/services/journal_capture.py` — flag if a narrower scope was actually intended.
- Whether `_journal_tool_gate`'s tag/field attestation (`api/services/agent_tools.py`) is strict enough — it currently requires a literal `#tag` or verbatim field-value substring match against the raw user message.

https://claude.ai/code/session_01Eja1zfh1oR8VbBzk39TTHf


nbramia:

**Reviewers run:** general, correctness, testing

**Specialist rationale:** the change carries three of the six repository risk triggers. `review-correctness` for irreversible data loss (a migration moving real vault files) and concurrency (flock/rename interaction); `review-testing` because the diff adds two new test files and rewrites a third for a data-loss-bearing path, so test strategy is itself the risk. All three executed code rather than only reasoning; every defect below arrived with a reproducer.

### Scope decision — the automatic migration is removed

The operator has directed that the one-time migration be cut. It existed to relocate two legacy day files totalling 168 bytes, and it accounted for **seven** of the fourteen findings in this round, including all four data-loss/privacy defects. Deleting `_take_legacy_content`, `_migrate_other_legacy_day_files`, `_LEGACY_LOG_DIR_PARTS`, and `_DAY_FILE_NAME` resolves them at the root rather than hardening ~100 lines of code that would run once and then be permanently dead. The operator moves the two files by hand.

Findings dissolved by that removal (no fix required, no follow-up filed):

| Finding | Reviewer |
|---|---|
| Empty legacy day file yields a day file with no frontmatter, permanently | correctness |
| Legacy file unlinked before the replacement is durable — crash in the window destroys the day | correctness |
| Sweep `os.replace` over a day file another capture holds open+locked, losing that fragment while reporting success | correctness, testing |
| Non-UTF-8 legacy file raises `UnicodeDecodeError` past both handlers — 500, and the traceback frame carries the fragment text | correctness |
| `_DAY_FILE_NAME` guard untested — removing it moves every file out of `Personal/Log/`, 29 tests still green | testing |
| Migration failure aborts the capture entirely; the "a migration problem never costs the fragment" claim is false as written | testing |
| Migration `OSError` branch untested | testing |

### Accepted findings

| # | Finding | Reviewer | Kind | Decision | Smallest remedy |
|---|---|---|---|---|---|
| 1 | `manage_schedules` `update`/`delete` ungated on journal turns — journal can delete or retarget the operator's existing `agent`/`prompt` schedules. Both reviewers proved it against the real handler: `action=delete` removed a schedule; `action=update` repointed an agent job's `message_content` and `executor`. The gate inspects only `schedule_action`, never the `action` selector (`create\|list\|update\|delete`). | general, correctness | Defect | **Accept** | Reject `manage_schedules` whose `action` is not in `("create","list")`, mirroring the existing `manage_tasks` complete/update rejection. |
| 2 | Tag/field attestation matches bare substrings, so invented pickup tags survive the strip. `"#agent" in low` is true for a message containing `#agentic`; `{"executor":"local","model":"opus"}` both survive a message mentioning "the local vendor" and "the opus recording". A task the operator never tagged then gets picked up and executed. | general | Defect | **Accept** | Match tags against `re.findall(r"#[\w-]+", low)`; require field values to match on a token boundary. |
| 3 | AC8 has no implementation: nothing ever sets `bot=journal`, so a journal-filed reminder pings on the primary bot rather than the journal surface. The PR adds the capability (schema + store plumbing) but no producer — neither the persona body nor `filing_rules()` mentions `bot`, and the gate does not default it. | general, correctness, testing | Defect | **Accept** | Default `bot` to `JOURNAL_PERSONA_ID` in the gate's `manage_schedules` branch when journal is a configured scheduler bot; leave empty otherwise. |
| 4 | The journal wiring is entirely unpinned. Mutation-proven twice, independently: deleting the `persona_id`/`user_message` kwargs at `chat.py:1259-1260`, or reverting `agent_loop.py` to `tools = TOOL_DEFINITIONS` and dropping the kwargs at the `execute_tool_parallel` call, leaves **all 70 journal tests green** — plus 54 more across the adjacent suites this PR cites as regression evidence. In production either mutation restores calendar and email access to the journal bot with a fully passing suite. | general, testing | Missing Test | **Accept** | Pin the route→loop kwargs, and add one end-to-end journal turn with a fake model emitting `create_calendar_event`, asserting the advertised tools exclude it and the call is refused. Cover the raw-preamble (Telegram) path too. |
| 5 | `test_excluded_tool_is_refused_on_a_journal_turn` calls the real `save_memory` with no store isolation — its only protection against writing the operator's live `~/.lifeos/memories.json` is the production gate it is testing. It wrote a real entry and fired a live LLM call during this review. | testing | Defect | **Accept** | Isolate the store in this test. The repo-wide gap is filed separately as #1065. |

### Rejected findings

| # | Finding | Reviewer | Decision | Reasoning |
|---|---|---|---|---|
| 6 | AC11's second half — a rejected `prompt`/`endpoint`/`agent` schedule should "become an Inbox task instead" — is neither implemented nor tested. | testing | **Reject — deliberate narrowing, recorded** | The gate returns a refusal the model can act on, and the persona already instructs filing; silently converting a refused orchestration request into a task is less predictable than refusing it. Narrowing #975's AC11 to a refusal is the intended behavior, not a gap. |
| 7 | `tests/narration_baseline.json:1440` still carries the now-deleted `Personal/Log/` docstring line. | general | **Reject as a finding; fold into the edit** | The ratchet's contract permits unused slack, so this is not a defect. The docstring changes again in this round anyway; tidy it as housekeeping. |

**Findings forwarded to addresser:** 5 (plus the migration removal)

nbramia:
## Round 1 findings addressed

| 0 | Remove the automatic legacy migration | Applied | Deleted `_take_legacy_content`, `_migrate_other_legacy_day_files`, `_LEGACY_LOG_DIR_PARTS`, `_DAY_FILE_NAME`, the sweep call, and `legacy_path`. `capture_fragment`'s `size == 0` branch is back to its pre-PR shape (`_frontmatter(day) + _bullet(...)`, `created = True` unconditionally). Removed the migration paragraphs from both docstrings. Deleted `TestLegacyMigration` and its `_write_legacy_day_file` helper (5 tests) from `tests/test_journal_capture.py`. Tidied the stale `narration_baseline.json` entry for `journal_capture.py` (the file's docstring no longer carries any narration, so the whole key was removed, not just the `Personal/Log/` line). No replacement migration, shim, or fallback of any kind remains. |
| 1 | `manage_schedules` update/delete ungated | Applied | `_journal_tool_gate` now rejects any `manage_schedules` call whose `action` isn't `create`/`list`, mirroring `manage_tasks`'s complete/update rejection message style. Added tests proving a journal-turn `delete` and `update` are both refused and the pre-existing entry is unchanged. |
| 2 | Tag/field attestation matches bare substrings | Applied | Tags are now attested against the message's own `#tag` tokens (`re.findall(r"#[\w-]+", ...)`), not a substring check — `#agentic`/`#claude-code` no longer attest a bare `agent`/`claude` model tag. Fields require an explicit `key:value` annotation or a quoted span (`_field_attested`), not a bare substring — "the local vendor... the opus recording" no longer attests `executor:local`/`model:opus`. Added the two proven-false-positive regression tests plus one keeping a genuinely-written `#claude` tag. |
| 3 | AC8 unimplemented — nothing sets `bot=journal` | Applied | The gate's `manage_schedules` create branch now defaults `bot` to `JOURNAL_PERSONA_ID` when journal is a configured scheduler bot (checked via `valid_bot_names()`, the same source `GET /api/scheduler/bots` uses), leaving it empty otherwise. Added tests for both the configured and not-configured cases. |
| 4 | Journal wiring entirely unpinned | Applied | Added `TestJournalWiringIsNotUnpinned` in `test_journal_native_interceptor_skip.py`: (a) captures the kwargs `chat.py` passes to `run_agent_loop` and asserts `persona_id == "journal"` / `user_message == request.question`, for both the `persona_id`-selected path and the raw-preamble (Telegram bot) path; (b) drives the REAL `run_agent_loop` (only the underlying model client faked) through a journal turn that tries to call `create_calendar_event`, asserting the advertised tool list excludes it and the real gate refuses the call (a stubbed handler proves it's never reached) — again for both paths. See mutation-check results below. |
| 5 | Test can write the operator's real memories.json | Applied | Added a `mem` fixture (same pattern as `tm`/`sched`) that monkeypatches `api.services.memory_store.get_memory_store` onto a `tmp_path`-backed store, and used it in `test_excluded_tool_is_refused_on_a_journal_turn`, plus an assertion that nothing was written. |

**Mutation-check results (item 4):**
- Dropping `persona_id=`/`user_message=` at `chat.py`'s `run_agent_loop` call site → all 4 new `TestJournalWiringIsNotUnpinned` tests failed (the two kwargs-capture tests directly; the two tool-exclusion/refusal tests transitively, since `run_agent_loop` never received `persona_id="journal"`).
- Reverting `agent_loop.py` to `tools = TOOL_DEFINITIONS` and dropping the `persona_id`/`user_message` kwargs at its `execute_tool_parallel` call → the two tool-exclusion/refusal tests failed (the kwargs-capture tests still passed, correctly — that mutation doesn't touch the chat.py→run_agent_loop boundary).
- Both mutations reverted afterward; `git diff --stat` confirms `agent_loop.py` and `chat.py` are clean.

**Focused test evidence:**
- `tests/test_journal_capture.py`, `tests/test_journal_tool_gating.py`, `tests/test_journal_native_interceptor_skip.py`, `tests/test_journal_persona.py`, `tests/test_hermes_proxy.py`, `tests/test_journal_ring_ingest.py`, `tests/test_vault_write_route.py`, `tests/test_no_change_narration.py`: all pass (241 tests).
- Broader sweep for regressions: `tests/test_agent_tools*.py tests/test_agent_loop*.py tests/test_chat_api.py`: 755 passed.
- `ruff check` on all changed files: clean.

**No migration code, shim, or fallback remains anywhere in the diff** — confirmed by re-reading the full `journal_capture.py` diff and grepping the whole PR diff (code + docs) for `migrat`/`legacy`/`Personal/Log` outside of the pre-existing, unrelated hits (`config/settings.py`'s mtime-trust docs, other stores' schema migrations, etc.).

nbramia:

**Reviewers run:** general

**Specialist rationale:** none — the round-1 scope decision deleted the migration, removing the data-loss and concurrency surfaces that justified the correctness and testing specialists. The remaining delta is a tool gate, an attestation function, and tests, which the general review covers.

**Scope:** the round-1 fix delta only (`ff91624e..2500bf5a`).

### Reviewer Findings

#### General

**Defects**

`manage_reminders` bypasses the AC8 `bot=journal` default. The round-1 fix defaults `bot` only inside `_journal_tool_gate`'s `manage_schedules` branch (`api/services/agent_tools.py:1098-1101`), but `manage_reminders` is absent from `JOURNAL_EXCLUDED_TOOLS` (`:26-35`) so `tools_for_persona("journal")` still advertises it, and the gate has no branch for it — the call reaches `_reminder_create` untouched. `api/services/reminder_store.py` is a re-export shim over `scheduler_store`, so the handler writes a live schedule into the *same* store with `bot` never set. Executed at PR head: a journal-persona `manage_reminders` create returned `Reminder created: "call mom"` and persisted `action == "notify"`, `bot == ""` — the ping fires on the primary bot, exactly what AC8 exists to prevent. Reachable on the ordinary "remind me to X at 3" fragment: the persona names `lifeos_schedule_create`, which is not a native tool name, so the model maps that instruction onto its advertised catalog, where `manage_reminders` ("Manage timed reminders: create or list") is a direct match; the `DEPRECATED` prefix discourages but does not prevent it. No orchestration risk on this path (`message_type="static"` forces `action="notify"`); the defect is confined to the missing `bot`.

**Missing Tests**

`_field_attested`'s quoted-span branch is entirely unpinned (`api/services/agent_tools.py:1016-1030`). The function attests a field two ways — an explicit `key:value` pair, or the value appearing inside quotes — but every field test in `tests/test_journal_tool_gating.py` exercises only the first (`test_manage_tasks_create_keeps_a_field_value_attested_in_the_message` uses `key:callmom`; the two stripping tests assert removal). Deleting the quoted-span `return` and returning `False` leaves the suite green.

**Status**

Focused tests at head `2500bf5a` in a throwaway worktree (removed; trees clean), GPU disabled on every run: journal-adjacent suites → 121 passed; narration ratchet + Hermes proxy → 120 passed (the ratchet accepts the trimmed `narration_baseline.json`); adjacent regression spot-check run. Authoritative broad suite not run — verification owns it.

### Referee Decisions

|---|---|---|---|---|---|---|
| 1 | `manage_reminders` ungated on journal turns, bypassing the `bot=journal` default | general | Defect | Valid — independently confirmed | **Accept** | Verified: `manage_reminders` is absent from `JOURNAL_EXCLUDED_TOOLS` and the gate has only `manage_tasks`/`manage_schedules` branches. It is a second write path into the same scheduler store that skips the fix just landed, on the exact fragment AC8 targets. Remedy: add `"manage_reminders"` to `JOURNAL_EXCLUDED_TOOLS` — journal already has `manage_schedules` for the identical job and the tool is deprecated. |
| 2 | `_field_attested`'s quoted-span branch unpinned | general | Missing Test | Valid | **Accept** | Half of a brand-new attestation function has no falsifying test; deleting the branch leaves the suite green. Remedy: one case asserting a quoted value survives. |

Both findings are small and reachable; neither is edge-case hardening. Round 1 forwarded 5 accepted findings, round 2 forwards 2 — converging.

**Findings forwarded to addresser:** 2

nbramia:
Addressed round-2 findings at 2d4c8ac2.

**1. Defect — `manage_reminders` ungated on journal turns.** Added `"manage_reminders"` to `JOURNAL_EXCLUDED_TOOLS` in `api/services/agent_tools.py`. Journal already has `manage_schedules` for the same job; `manage_reminders` is a deprecated alias with no `bot`-default branch. Added tests: `manage_reminders` absent from `tools_for_persona("journal")` but present for `"primary"`, and a journal-turn `manage_reminders` call is refused by `_journal_tool_gate` with nothing written to the scheduler store (patched `api.services.reminder_store.get_reminder_store`, since that's what `_reminder_create` resolves at call time).

**2. Missing Test — `_field_attested` quoted-span branch unpinned.** Added `test_manage_tasks_create_keeps_a_field_value_attested_by_a_quoted_span`, exercising `use the "local" executor` / `fields={"executor": "local"}`. Mutation check: deleted the quoted-span `return` locally, confirmed the new test failed (`assert {} == {'executor': 'local'}`), then reverted — tree confirmed clean before committing.

No doc changes needed — checked `docs/guides/personas.md`, `docs/specs/technical/architecture.md`, `docs/guides/scheduler.md`, `docs/specs/technical/scheduler.md`, and `config/personas/journal.md`; none claim journal can use `manage_reminders` (the journal persona doc already describes only `lifeos_task_create`/`lifeos_schedule_create`).

Focused tests: `tests/test_journal_tool_gating.py` (28 passed), `tests/test_scheduler_api.py` (42 passed), `tests/test_journal_native_interceptor_skip.py` (8 passed), `tests/test_no_change_narration.py` (1 passed). Ruff clean on touched files.

PR #1082, opened 2026-09-15
feat(home): pause and resume household internet access via eero

### LINKED ISSUE #1081: feat(home): pause and resume household internet access via eero
## Problem

There is no way to cut a single household device off the internet from anything other than the eero phone app. Pausing a kid's iPad requires unlocking a phone, opening the app, and tapping through to the right profile. That makes the action unavailable to every automated or conversational surface in the house — a Telegram message, a persona, a scheduled routine, or a one-tap phone shortcut.

The gap matters most for ad-hoc overrides, which are the common case. Recurring bedtime cutoffs are already handled well by the router's own per-profile schedules. What is missing is "off now, for the next hour," issued from whatever surface is closest to hand.

A second problem is failure visibility. Any integration here depends on a vendor cloud service with no supported public API, so it will eventually break. A pause that silently fails while the requester believes it succeeded is worse than having no automation at all — and a timed pause whose resume silently fails is the worst case, because the device stays offline indefinitely.

## Solution

Add a `home` integration namespace to LifeOS, with eero as its first provider. The namespace anticipates further home-automation providers, which slot in beside eero rather than reshaping the layout.

An eero service module owns the vendor session and translates friendly target names into vendor identifiers. Each configured target declares whether it is a profile or a device; profiles are the recommended kind, because they are the abstraction the vendor's own UI is built on and therefore the most stable surface.

A `/api/home/eero` route group exposes status, pause, and resume. All three are idempotent and state-reconciling: each reads current state, applies the change, reads it back, and returns the state actually observed rather than the state requested.

Pause accepts an optional duration. When given, the resume is not held in memory — it is written into the existing scheduler as a one-off endpoint action, so the resume survives a restart of whatever process issued the pause. A scheduled resume that fails against the vendor is retried and, if it still fails, alerts loudly.

The session token is seeded from the environment or the login script, and every refreshed token is persisted to a gitignored state file so a refresh survives a restart.

Session death, vendor API drift, and failed scheduled resumes surface loudly: a Telegram notification plus a human-queue card keyed stably so repeated failures update one card instead of accumulating. These operations never fail silently.

The same three operations are exposed as agent tools and over the MCP bridge, so the personas, Hermes, and Claude Code all reach one implementation.

### Session and configuration

- [ ] Given no session token is available (neither the state file nor `LIFEOS_EERO_SESSION_TOKEN`), when any `/api/home/eero/*` endpoint is called, the API shall return 503 with a message naming the environment variable and the login script.
- [ ] The token shall be resolved from the gitignored state file `data/home/eero_session.json` first, falling back to `LIFEOS_EERO_SESSION_TOKEN`.
- [ ] The eero client shall send its session token as a `Cookie: s=<token>` header on every vendor request.
- [ ] Given a session token that the vendor rejects as unauthorized, when a request is made, the client shall attempt one refresh against the vendor refresh endpoint and retry the original request once before treating the session as dead.
- [ ] When a refresh succeeds, the new token shall be written to the state file (mode 0600) so it is used after a restart.
- [ ] If a refresh attempt fails, the system shall send a Telegram notification and create a human-queue card with key `eero-session`, and shall return 502 to the caller — it shall not return a success response.
- [ ] Repeated session failures shall update the existing human-queue card rather than creating additional cards.
- [ ] The target map shall be read from the gitignored `config/home/eero_targets.json`, and a tracked `config/home/eero_targets.example.json` with placeholder values shall exist alongside it.
- [ ] Each target entry shall have a `type` of `profile` or `device`. A `profile` target carries the profile's vendor URL (e.g. `/2.2/networks/<network_id>/profiles/<profile_id>`); a `device` target carries `network_id` and `mac`. An entry with a missing or invalid field shall be skipped with a warning naming the target, not raise.
- [ ] Target names shall match case-insensitively.
- [ ] Given the target map is absent, when the service loads, it shall start successfully and report zero configured targets — it shall not raise.
- [ ] `scripts/eero_login.py` shall be non-interactive and two-step: `--login <email-or-phone>` requests a code and stores the pending token in the state file; `--code <code>` verifies it and stores the persistent token. It shall never print the token.

### Pause and resume

- [ ] When `POST /api/home/eero/{name}/pause` is called with a known target name, the system shall set that target paused and return the paused state read back from the vendor.
- [ ] Given a target that is already paused, when pause is called again, the system shall return success with the paused state — it shall not error.
- [ ] Given a target that is already resumed, when resume is called, the system shall return success with the resumed state.
- [ ] When pause or resume is called with an unknown target name, the system shall return 404 naming the unknown target and listing the configured names, and shall not contact the vendor.
- [ ] When `GET /api/home/eero/status` is called, the system shall return every configured target with its name, type, current paused state, and pending scheduled resume time (or null).
- [ ] Pause and resume responses shall have the shape `{name, type, requested_paused, paused, mismatch, resume_at}`, where `mismatch` is true when the read-back state differs from the requested state.
- [ ] If the vendor rejects a write or returns an unexpected response shape, the API shall return 502 and raise the same loud alert as session failure, using human-queue key `eero-api-error`.
- [ ] Pause and resume shall return 200 on success, so the scheduler's endpoint action treats them as successful.
- [ ] Pause and resume responses shall include a short `scheduler_message` so a scheduler-fired resume posts one readable line to Telegram instead of a JSON dump.

### Duration and auto-resume

- [ ] Pause shall accept an optional JSON body `{"minutes": <int>}`; resume shall accept an empty or absent body.
- [ ] When pause is called with `minutes`, the system shall create a one-off scheduler entry whose endpoint action is `POST /api/home/eero/{name}/resume` with `params` `{"scheduled": true}`, and shall return the scheduled resume time as `resume_at`.
- [ ] The scheduler entry shall carry the operation key `eero-resume:<normalized-name>`. A *pending* resume is an entry with that key that is still enabled; disabled (already fired) entries with that key shall be ignored.
- [ ] When pause is called on a target that already has a pending scheduled resume, the system shall delete the pending entry and create a new one, so exactly one pending entry exists. Pause without `minutes` on such a target shall delete the pending entry (the pause becomes indefinite).
- [ ] When resume is called manually on a target with a pending scheduled resume, the system shall delete that entry.
- [ ] Before creating a new entry, finished (disabled) entries with the same operation key shall be removed, so the Scheduler Inbox does not accumulate one dead entry per timed pause.
- [ ] Given a pause with a duration, when the API process restarts before the resume time, the resume shall still fire at the scheduled time (the scheduler's run-once catch-up covers a resume time that elapsed during downtime).
- [ ] If `minutes` is absent, the pause shall be indefinite and no scheduler entry shall be created.
- [ ] If `minutes` is zero, negative, or exceeds 1440, the API shall return a validation error.
- [ ] When a scheduled resume (`scheduled: true`) fails against the vendor, the resume route shall retry the vendor write up to 3 times with backoff inside the request; if every attempt fails it shall send a Telegram notification and create or update a human-queue card with key `eero-resume-failed:<normalized-name>` stating the target is still paused, and return 502. The scheduler marks one-off entries fired before calling the endpoint and never re-fires them, so the route itself owns the retry and the alert.

### Per-target default duration and silent scheduled calls

- [ ] A target entry may carry an optional `default_minutes` (int, 1–1440). An entry with an invalid `default_minutes` shall be skipped with a warning, like any other invalid field.
- [ ] When pause is called without `minutes` on a target that has `default_minutes`, the pause shall use that duration (scheduler-backed auto-resume, as with an explicit `minutes`). An explicit `minutes` overrides it.
- [ ] Pause shall accept `{"indefinite": true}` to force an indefinite pause regardless of `default_minutes`; sending both `indefinite: true` and `minutes` shall return a validation error. Without `default_minutes`, an absent `minutes` stays indefinite.
- [ ] The agent tool `pause_internet` shall apply the same default, and its description shall say a target's configured default applies when minutes is omitted.
- [ ] Pause shall accept `scheduled: true` in its body. For any call with `scheduled: true` (pause or resume) that succeeds with no mismatch, `scheduler_message` shall be empty, so a scheduler-fired call sends no Telegram message. A mismatch shall still produce a non-empty `scheduler_message`; a failure returns non-200, which the scheduler reports.
- [ ] The guide documents `default_minutes`, `indefinite`, and how to create a recurring scheduled pause (a cron `endpoint` schedule POSTing `{"minutes": N, "scheduled": true}`).

### Agent and MCP surfaces

- [ ] The agent tool registry (`TOOL_DEFINITIONS` and `_TOOL_HANDLERS`) shall expose `pause_internet` (name, optional minutes), `resume_internet` (name), and `internet_status`, each dispatching through the same service the HTTP routes use.
- [ ] The MCP bridge shall expose the same three operations as `lifeos_home_eero_pause`, `lifeos_home_eero_resume`, and `lifeos_home_eero_status` via `CURATED_ENDPOINTS`.
- [ ] When an agent tool is called with an unknown target name, it shall return a message listing the configured target names.

### Documentation

- [ ] `docs/specs/product/api-reference.md` documents the three routes.
- [ ] A guide (`docs/guides/home-eero.md`) covers login, target configuration, the Private Wi-Fi Address requirement, and failure alerts; it is linked from the AGENTS.md navigation table.
- [ ] The MCP tool count in AGENTS.md's Key Files table is updated.

## Verification

- **Automated tests:** All vendor HTTP calls are mocked — no test contacts the real service. Cover: the 503-when-unconfigured gate; token precedence (state file over env); cookie header shape; refresh-and-retry success with the new token persisted; refresh-then-fail path including notification and human-queue side effects; idempotent pause and resume; unknown-target 404 with no vendor call; read-back mismatch flagging; vendor error → 502 plus alert; scheduler entry creation, replacement, cancellation, indefinite pause clearing a pending entry, and fired-entry cleanup (against a temp scheduler store); scheduled-resume retry and failure alert; duration validation bounds; missing and malformed config tolerance; login script two-step flow.
- **Manual checks:** The two-step login produces a working token. A real pause takes the target device offline while leaving it associated to Wi-Fi, and resume restores it. A duration-bounded pause auto-resumes after an API restart.

## Scope

**In scope:**
- The `home` service and route namespace, with eero as the only provider.
- Profile-level and device-level pause and resume, each target declaring its type.
- Idempotent, state-reconciling status, pause, and resume endpoints.
- Duration-bounded pause with scheduler-backed auto-resume, retried and loudly alerted on failure.
- Refreshed-token persistence to a gitignored state file.
- Loud failure surfacing via Telegram and the human queue.
- Agent tool and MCP exposure.
- A two-step, non-interactive login script.
- Gitignored target config with a tracked example.

**Out of scope:**
- A dedicated recurring-schedule feature. Recurring pauses are ordinary scheduler cron entries using the endpoint action; no new schedule type or UI is added.
- Blocking devices from joining the network at all, as distinct from pausing internet access.
- Any other home-automation provider. The namespace anticipates them; this issue adds none.
- New public HTTP routes. The REST endpoints are reachable over Tailscale only, like the rest of the API.
- Automatic device discovery or target-map generation. The map is hand-maintained.

## Technical Context

**Key files:**

| Path | Change |
|------|--------|
| `api/services/home/__init__.py` | New namespace package |
| `api/services/home/eero.py` | New — vendor client, session + token persistence, target resolution, pause/resume/status, pending-resume management |
| `api/routes/home.py` | New — `APIRouter(prefix="/api/home", tags=["home"])` |
| `api/main.py` | Add `home` to the `from api.routes import ...` line; add `app.include_router(home.router)` beside `journal_ingest` |
| `config/settings.py` | Add `eero_session_token` field (alias `LIFEOS_EERO_SESSION_TOKEN`) |
| `config/home/eero_targets.example.json` | New — tracked placeholder map |
| `.gitignore` | Add `config/home/eero_targets.json` (`data/` is already ignored) |
| `scripts/eero_login.py` | New — two-step login writing the state file |
| `api/services/agent_tools.py` | Add to `TOOL_DEFINITIONS` and `_TOOL_HANDLERS` |
| `mcp_server.py` | Add to `CURATED_ENDPOINTS` and tool input schemas |
| `docs/specs/product/api-reference.md`, `docs/guides/home-eero.md`, `AGENTS.md` | Docs |

**Patterns to follow:**

- Token settings field: mirror `health_ingest_token` in `config/settings.py` — `Field(default="", alias="LIFEOS_...", description=...)`, empty value disables the feature.
- Unconfigured-feature gate returning 503: mirror `_check_ingest_auth` in `api/routes/fitness.py`.
- Gitignored-real / tracked-example config pair: mirror `config/family_members.json` against `config/family_members.example.json`.
- Human-queue card creation: `api.services.human_queue.add_card` (see `_human_queue_add` in `api/services/agent_tools.py`).
- Telegram delivery: `send_message` / `send_message_async` in `api/services/telegram.py`.
- One-off endpoint schedules: `api/services/scheduler_store.py`. `ScheduleEntry` carries `schedule_type="once"`, `action="endpoint"`, `endpoint_config` shaped `{endpoint, method, params}`, and `operation_key`. Use `get_scheduler_store()`; `find_by_operation`, `delete`, and `create(..., operation_key=...)`. The source of truth is the vault's `LifeOS/Scheduler/Inbox.md`, which is what makes the resume restart-durable.
- Scheduler endpoint semantics (`SchedulerScheduler._call_endpoint`): non-GET methods are sent as POST with `params` as the JSON body to `http://localhost:<port><endpoint>`; any non-200 status posts "API call failed"; a response `scheduler_message` is sent verbatim. `_fire_entry` calls `mark_triggered` *before* the endpoint, so a one-off entry never re-fires.

**Vendor API (confirmed against the reference implementation, do not re-derive):**

- Base host `https://api-user.e2ro.com`. Base paths are versioned per resource; profile and device operations sit on different versions.
- `POST /2.2/login` with `{"login": "<email-or-phone>"}` returns `data.user_token` (temporary).
- `POST /2.2/login/verify` with `{"code": "<code>"}`, sent with `Cookie: s=<temporary token>`, promotes that token to a persistent session.
- `POST /2.2/login/refresh` with the session cookie returns a new `data.user_token`, which replaces the old one.
- Every authenticated request carries `Cookie: s=<user_token>`. Responses wrap payloads in `{"meta": {...}, "data": {...}}`; an expired session returns HTTP 401.
- Profile pause: `PUT` to the profile's own URL with `{"paused": <bool>}`. Current state reads from the profile's `paused` field (`GET` on the same URL).
- Device pause: `PUT /2.3/networks/{network_id}/devices/{mac}` with `{"paused": <bool>}`; state reads from `GET` on the same URL.

**Constraints:**

- The repository is public. No real profile identifiers, network identifiers, MAC addresses, device nicknames, account email, or session token may be committed. Everything personal goes in the gitignored config file, the gitignored state file, or the environment. Logs name targets by configured name only — never the token or cookie.
- Login cannot be headless. It requires a verification code delivered out of band, so the login script is run once by the operator in two steps (non-interactive, so it works from Claude Code's `!` prefix, which has no TTY).
- Implement the small client directly with `httpx` (already a dependency); do not add `eero-client`.
- Targets must have iOS "Private Wi-Fi Address" disabled for the home network, or their MAC-based identity drifts and device-level operations silently target nothing. Document this in the example config and the guide.
- REST routes are Tailscale-only. The MCP tools inherit the MCP HTTP transport's existing exposure (bearer-token protected, same as `lifeos_gmail_send` and the other write tools); that is accepted, not a new public route.

## References

- Reference implementation for endpoint shapes and auth flow: [schmittx/home-assistant-eero](https://github.com/schmittx/home-assistant-eero) — `custom_components/eero/api/`
- Alternative client library: [eero-client on PyPI](https://pypi.org/project/eero-client/)
- Analogous token-gated ingest endpoint: #333
- Human queue: #852

nbramia:
## TL;DR
Adds a way to pause or resume a household device or profile's internet access from anywhere — Telegram, a persona, a scheduled routine, or the MCP tools — instead of needing the eero phone app. A pause can carry a duration ("off for the next hour"), after which the device resumes automatically; a target can also declare a default duration so an untimed pause still auto-resumes, and an explicit indefinite flag overrides that default. Every failure mode (a dead vendor session, a rejected vendor request, a scheduled resume that couldn't complete) is surfaced loudly through Telegram and a Human-queue card rather than failing silently — except a scheduled call (cron or the auto-resume itself) that succeeds exactly as requested, which now posts nothing at all.

Closes #1081

## Design
A `home` integration namespace holds provider-specific modules, with eero as the first (and currently only) provider — the layout anticipates other home-automation integrations without needing to change shape. Every operation (status/pause/resume) is idempotent and state-reconciling: it sets the requested value, reads it back from the vendor, and reports what's actually there, flagging a mismatch rather than trusting the write blindly.

A duration-bounded pause doesn't hold the resume in memory — it's written into the existing scheduler as a one-off `endpoint` action, so the resume survives a restart of whatever process issued the pause. The scheduler marks a one-off entry fired before calling it and never re-fires it, so the resume route itself owns retrying a scheduled resume (up to 3 attempts with backoff) and alerting if every attempt fails — the worst-case outcome, since the device then stays offline until someone notices.

Session tokens are resolved from a gitignored state file first, falling back to an env var; a refreshed token is persisted back to that file so it survives a restart. Targets (profiles or devices) are hand-maintained in a gitignored config file with a tracked example — no automatic discovery. A target may also carry `default_minutes`, applied when a pause omits an explicit duration; an `indefinite` flag on the request forces an indefinite pause regardless. Any pause or resume marked `scheduled: true` (a recurring cron schedule, or the scheduler's own fire of a timed pause's auto-resume) that succeeds with no mismatch returns an empty `scheduler_message`, which the scheduler's fire loop treats as nothing to send — so an unremarkable scheduled call is silent, while a mismatch or failure still alerts.

## Implementation Notes
- `api/services/home/eero.py` — vendor client (session/token persistence, refresh-and-retry), target resolution (now including `default_minutes`), pause/resume/status, scheduler-backed pending-resume management, and the three alert paths (`eero-session`, `eero-api-error`, `eero-resume-failed:<name>` human-queue keys). `pause()` takes `indefinite`/`scheduled` kwargs; `_result()` blanks `scheduler_message` for a scheduled, non-mismatched success.
- `api/routes/home.py` — `GET/POST /api/home/eero/*`, gated 503 until a session token exists; error mapping (404 unknown target, 502 vendor/session failure, 422 for `indefinite` combined with `minutes`).
- `config/settings.py` — `eero_session_token` (`LIFEOS_EERO_SESSION_TOKEN`), mirroring `health_ingest_token`'s pattern.
- `scripts/eero_login.py` — two-step, non-interactive login (`--login`, then `--code`); never prints the token.
- `config/home/eero_targets.example.json` — tracked placeholder target map, now with a `default_minutes` example; the real file is gitignored.
- `api/services/agent_tools.py` — `pause_internet` / `resume_internet` / `internet_status` tools, added to `TOOL_DEFINITIONS`/`_TOOL_HANDLERS` (cache-control breakpoint moved to the new last entry). `pause_internet` gained an `indefinite` boolean and applies the same target default.
- `mcp_server.py` — `lifeos_home_eero_pause` / `_resume` / `_status` added to `CURATED_ENDPOINTS`; `CURATED_TOOL_COUNT` 59 → 62 (71 total with the 9 inter-agent tools).
- `docs/guides/home-eero.md` (new), `docs/specs/product/api-reference.md`, `docs/specs/product/mcp-tools.md`, `AGENTS.md` — docs, including the updated MCP tool count, `default_minutes`/`indefinite`/silent-scheduled-call behavior, and a recurring scheduled-pause example built from a plain cron `endpoint` schedule.

### Automated tests
All vendor HTTP calls are mocked via `httpx.MockTransport` — no test contacts the real eero service.

```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" \
  ~/.venvs/lifeos/bin/python -m pytest \
  tests/test_home_eero.py tests/test_eero_login_script.py tests/test_agent_tools_home_eero.py \
  tests/test_mcp_server.py tests/test_mcp_human_queue.py tests/test_workout_mcp_route.py \
  tests/test_agent_worker_mcp_exposure.py tests/test_scheduler_store.py tests/test_human_queue.py \
  tests/test_agent_tools_human_queue.py tests/test_journal_tool_gating.py tests/test_settings.py -q
```
323 passed, 1 failed. The one failure (`test_agent_worker_mcp_exposure.py::test_mcp_catalog_count_and_inter_agent_schema_contract`) builds its MCP tool catalog against **the currently-running production server's live `/openapi.json`** (not this branch's code) — that server hasn't been redeployed with this PR, so it still reports the pre-eero route set (68 tools instead of 71). Confirmed this is purely a live-server staleness artifact, not a defect: the equivalent in-process check (`test_mcp_human_queue.py::test_total_tool_count_matches_agents_md`, built from `api.main.app.openapi()` against this checkout) passes at 71. This test will pass once the merge deploys.

Added coverage for the default-duration/indefinite/silent-scheduled-call work:
```
HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" \
  ~/.venvs/lifeos/bin/python -m pytest \
  tests/test_home_eero.py tests/test_eero_login_script.py tests/test_agent_tools_home_eero.py \
  tests/test_mcp_server.py tests/test_mcp_human_queue.py -q
```
155 passed. Covers: `default_minutes` applied when `minutes` is omitted and an explicit `minutes` overriding it; an invalid `default_minutes` skipping the whole target entry; `indefinite` overriding the default and clearing a pending resume; `indefinite` combined with `minutes` rejected pre-vendor-call at both the route (422) and the agent tool; a scheduled pause/resume success returning an empty `scheduler_message` versus a mismatch still reporting one; and an end-to-end check that the scheduler's own `_format_endpoint_result` renders a scheduled-success response as an empty string.

`ruff check` passes on every changed file.

### Manual verification
N/A — no real eero account/session token is available in this environment to exercise a real pause/resume against a physical device (the issue's manual-check step: real device pause/resume, restart-durable auto-resume). Automated coverage substitutes a mocked vendor throughout; the operator should run the two manual checks from the issue after merging and configuring a real target.

## Review focus
- `api/services/home/eero.py`: the refresh-and-retry session logic (`_authed_request`) and the scheduler pending-entry lifecycle (`_delete_pending`/`_delete_finished`, `pause`/`resume`) — these encode several of the issue's more intricate acceptance criteria.
- Alert-key design: `eero-api-error` is a single shared key across all vendor write/read failures (not per-target) per the issue text ("the same loud alert as session failure, using human-queue key `eero-api-error`"), while `eero-resume-failed:<name>` is per-target. Worth double-checking that reading matches intent.
- `scripts/eero_login.py` and `api/services/home/eero.py` share private helpers (`_new_http_client`, `_save_token`, `_load_token`) — deliberate, since the script is a thin non-interactive front end onto the same session-state file the service reads/writes at request time.

https://claude.ai/code/session_018gaULYK7J15ST5cPKb1irN


nbramia:

**Reviewers run:** general, security, correctness

**Specialist rationale:** security — session token handling, state-file persistence, public-repo privacy, and network-control operations on REST/agent/MCP surfaces. correctness — state-reconciling read-back, 401→refresh→retry, scheduled-resume retry, and scheduler pending-entry lifecycle, where the dangerous failure is a pause that sticks.

### Reviewer Findings

#### General
**Defects**
- Transport-level vendor failures (`httpx.HTTPError`: ConnectError, ReadTimeout) escape `_vendor_request`/`_authed_request` and the scheduled-resume retry loop, which only catches `EeroSessionDead`/`EeroAPIError` (`api/services/home/eero.py:494-503`). Result: no retry, no `eero-resume-failed:<name>` card, unmapped 500; scheduler never re-fires, device stays paused. Same gap makes manual pause/resume/status an unalerted 500.
- Agent-tool pause bypasses 1–1440 `minutes` validation and `eero.pause` does the vendor write before using `minutes` (`agent_tools.py:3789`, `eero.py:448-480`): string minutes → paused + TypeError, no resume scheduled; 0/negative → once entry with past time never fires but shows `resume_at`.
- `docs/guides/home-eero.md` Related Documents cites `configuration.md` for `LIFEOS_EERO_SESSION_TOKEN`, which has no row there.

**Missing Tests**
- No test of scheduled resume under a raised `httpx.HTTPError`.
- No route-level test of resume 404 for an unknown target.

**Status:** focused suite 130 passed.

#### Security
**Defects:** none. **Missing Tests:** none.
**Status:** 50 passed; diff scanned for real identifiers — placeholders only. Token only in Cookie header, never logged/returned/alerted; state file via mkstemp 0600 + atomic rename; target names never reach vendor URLs; scheduler endpoint URL-encoded; op key restricted. Cross-referenced the transport-error gap to correctness.

#### Correctness
**Defects**
- (High) Transport error / timeout on scheduled resume bypasses retry and resume-failed alert — reproduced with a scratch MockTransport raising `httpx.ConnectError` (1 vendor call, no card). Same on manual paths (500 not 502).
- (Low) `EeroSessionDead` is retried inside the scheduled-resume loop; each attempt re-runs a doomed refresh and `_alert_session_dead()`, producing ~5 Telegram messages for one failure.

**Missing Tests**
- No transport-exception test for scheduled resume or manual routes.

**Status:** 107 passed (home_eero + scheduler_store). Confirmed OK: own fired entry cleaned on next timed pause; 3s backoff well inside scheduler 30s timeout; newer pause deletes pending entry before stale resume can fire; vendor calls async.

### Referee Decisions

| 1 | Transport errors bypass retry/alert/502 | general, correctness, security (dup) | Defect | Valid | Accept | Verified at `eero.py:160-206,500`. Wrap `httpx.HTTPError` from the vendor call as `EeroAPIError` so the existing retry, alert, and 502 paths apply; keep `_refresh_token` returning None on transport failure. |
| 2 | Session-dead retried → repeated Telegram alerts | correctness | Defect | Valid | Accept | Stop retrying on `EeroSessionDead`; go straight to the resume-failed alert. |
| 3 | Agent-tool pause skips minutes validation; vendor write precedes use | general | Defect | Valid | Accept | Verified. Validate `minutes` (int, not bool, 1–1440) at the top of `eero.pause` before any vendor call; tool returns a readable error. |
| 4 | Guide cites missing configuration.md row | general | Defect | Valid | Accept | Add `LIFEOS_EERO_SESSION_TOKEN` row to `docs/guides/configuration.md` beside the other ingest tokens. |
| 5 | No transport-error test | general, correctness (dup) | Missing Test | Valid | Accept | Scheduled resume with ConnectError → 3 attempts, card, `EeroResumeFailed`; manual route → 502. |
| 6 | No resume-route 404 test | general | Missing Test | Valid | Accept | Issue AC requires 404 on resume; add route test. |

**Findings forwarded to addresser:** 6

nbramia:
Addressed all 6 accepted round-1 findings.

| 1 | Transport errors bypass retry/alert/502 | Applied | `_vendor_request` now catches `httpx.HTTPError` and raises `EeroAPIError` (message carries only the path and exception type, never headers/token). `_refresh_token` catches `EeroAPIError` instead of the raw httpx exception and still returns `None` on any transport failure. Manual pause/resume/status routes already map `EeroAPIError` to 502 + the `eero-api-error` alert; a scheduled resume now retries a transport failure the same as any other vendor failure. |
| 2 | Session-dead retried in scheduled resume → repeated alerts | Applied | The scheduled-resume retry loop now breaks immediately on `EeroSessionDead` (no further attempts), then files `_alert_resume_failed` and raises `EeroResumeFailed` — one `eero-session` alert (from `_authed_request`) plus one `eero-resume-failed` alert, not three. |
| 3 | Agent-tool pause skips minutes validation | Applied | `eero.pause` validates `minutes` (None, or an `int`, not `bool`, in 1..1440) before `_resolve_target`/any vendor call, raising `ValueError` otherwise. `_tool_pause_internet` catches `ValueError` and returns a readable error. The route is unaffected — pydantic still owns its 422. |
| 4 | Missing configuration.md row | Applied | Added a "Home — eero" section with the `LIFEOS_EERO_SESSION_TOKEN` row to `docs/guides/configuration.md`, describing state-file precedence and the 503-when-empty gate. |
| 5 | Missing transport-error tests | Applied | `tests/test_home_eero.py`: scheduled resume against a `MockTransport` raising `httpx.ConnectError` makes 3 attempts and alerts `eero-resume-failed`; a route-level manual pause against the same transport error returns 502; `EeroSessionDead` in scheduled resume makes exactly 1 attempt and alerts `eero-session` once (plus one `eero-resume-failed`). `tests/test_agent_tools_home_eero.py`: minutes as a string or `0` is rejected via the agent tool without resolving a target (i.e. before any vendor call). |
| 6 | Missing resume-route 404 test | Applied | Added `test_resume_404_for_unknown_target` — `POST /api/home/eero/<unknown>/resume` returns 404 naming the configured targets, with a handler that asserts the vendor is never contacted. |

**Tests:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/python -m pytest tests/test_home_eero.py tests/test_eero_login_script.py tests/test_agent_tools_home_eero.py -q` — 72 passed. `ruff check` passes on all changed files.

**Commit:** 683e7feccfb2b162b0d9fe9d27e6f4978211182e

https://claude.ai/code/session_018gaULYK7J15ST5cPKb1irN

nbramia:

**Specialist rationale:** correctness re-invoked — 683e7fec changed the error/retry paths and 4e9956e8 added default durations and silent scheduled calls on the pause/resume/scheduler lifecycle. Security dropped: round 1 clean and the delta doesn't touch token handling, persistence, or exposure.

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none.
**Status:** 155 passed. Round-1 fixes #1–#6 verified; 4e9956e8 satisfies every AC in "Per-target default duration and silent scheduled calls" (default_minutes validation/skip, override, indefinite + 422/ValueError before vendor, agent-tool default, empty scheduler_message on mismatch-free scheduled success traced through `_format_endpoint_result` and the `if message:` guard, guide cron example shape).

#### Correctness
**Defects**
- (Medium) Timed pause can stick with no resume scheduled: `pause()` writes (`eero.py:514`), reads back (`:515`), and only then creates the resume entry (`:527`). A read-back failure after a successful write (transport blip, non-200, bad shape) raises before scheduling. Cron flow: PUT lands, Malea offline, GET blips → 502 "API call failed: 502", no pending resume, device stays paused past 05:00. Reproduced with scratch MockTransport (PUT 200, GET ConnectError): 502, vendor paused, `store.list_all() == []`, no card.
- (Low) Manual transport failures return 502 without the `eero-api-error` alert: `_authed_write`/`_authed_read` (`eero.py:257,266`) call `_authed_request` outside their `try`, so the wrapped transport `EeroAPIError` skips `_alert_api_error()`.

**Missing Tests**
- Write-succeeds / read-back-fails on a timed pause should still leave a pending resume.
- Manual-route transport failure should file the `eero-api-error` card.

**Status:** 148 passed. Scratch end-to-end cron flow (temp store): cron pause silent, exactly one pending entry, fired resume silent and resumes, next cron pause removes the fired entry — passes.

### Referee Decisions

| 1 | Timed pause sticks when read-back fails after write | correctness | Defect | Valid | Accept | Verified at `eero.py:513-527`. Compute effective minutes and replace the pending resume entry before the vendor write; resume is idempotent, so a redundant resume is harmless. |
| 2 | Manual transport failure misses eero-api-error alert | correctness | Defect | Valid | Accept | Verified at `eero.py:253-270`; round-1 #1 promised the alert path. Move `_authed_request` inside the `try` in both helpers. |
| 3 | Test: read-back failure still leaves pending resume | correctness | Missing Test | Valid | Accept | Pins #1. |
| 4 | Test: manual transport failure files eero-api-error card | correctness | Missing Test | Valid | Accept | Pins #2. |

**Findings forwarded to addresser:** 4 (round 1: 6 — converging)

nbramia:
## Round 2 findings addressed

| 1 | Timed pause sticks when read-back fails after a successful write | Applied | `pause()` now creates/clears the pending-resume scheduler entry *before* the vendor `PUT`/`GET`, not after. A write/read failure (which raises) now still leaves a durable resume behind — harmless per resume's idempotency even if the write never actually took effect. Docstring updated to describe the new ordering. |
| 2 | Manual transport failures skip the eero-api-error alert | Applied | `_authed_write`/`_authed_read` now call `_authed_request(...)` inside the `try` block, so a transport failure (`httpx.ConnectError`, timeout, etc. — already wrapped as `EeroAPIError` by `_vendor_request`) triggers `_alert_api_error()` before re-raising, same as a bad status code or response shape. `EeroSessionDead` is a distinct exception type, so the session-dead alert path is unaffected. Confirmed the scheduled-resume retry loop (`resume(scheduled=True)`) still calls `_authed_request` directly, bypassing this alert path, so no duplicate alerts per retry attempt. |
| 3 | Test: read-back failure keeps pending resume | Applied | Added `test_readback_failure_after_successful_write_keeps_pending_resume` — PUT succeeds, GET raises `httpx.ConnectError`; asserts `pause(minutes=60)` raises `EeroAPIError` and the temp store retains exactly one enabled entry keyed `eero-resume:<name>`. |
| 4 | Test: manual transport failure files eero-api-error card | Applied | Added `test_pause_transport_error_files_eero_api_error_card` at the route level — PUT raises `httpx.ConnectError` → 502, and asserts the human-queue mock was called once with `key="eero-api-error"` and the Telegram mock was called once. |

**Tests:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/python -m pytest tests/test_home_eero.py tests/test_eero_login_script.py tests/test_agent_tools_home_eero.py tests/test_mcp_server.py tests/test_mcp_human_queue.py -q` — 157 passed.
`ruff check api/services/home/eero.py tests/test_home_eero.py` — all checks passed.

**Commit:** `fix: address review round 2 — eero pending-resume ordering and transport-failure alerts`

https://claude.ai/code/session_018gaULYK7J15ST5cPKb1irN

nbramia:

**Reviewers run:** general

**Specialist rationale:** none — the 218305d9 delta is a bounded reordering plus two tests; general review covered the regression surface (validation order, resume_at, default/indefinite, silent scheduled success, alert de-duplication, EeroSessionDead separation).

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none.
**Status:** focused suite 86 passed at 218305d9. All 4 round-2 fixes verified, both new tests would fail on revert; no regressions found. Noted trade-off (resume entry left behind when the write itself fails is harmless, resume is idempotent) was already accepted in round 2.

### Referee Decisions

No findings.

nbramia:
## End-to-End Verification — PR #1082

### Verdict: FAIL

```
verification-record: v1
verification-head: 218305d9b1b2c614cd136c86c3f12a7d5c3ee20c
suite-result: fail
suite-command: python scripts/verify_candidate.py pushed-ref --repository <worktree> --sha 218305d9b1b2c614cd136c86c3f12a7d5c3ee20c --base 981251bdd90bc37cf4f70843bfa93735f686b3ee --evidence-root ~/.cache/lifeos/verification-evidence --lanes fast-unit,browser-free --workers 2 --lane-log-dir <scratch>/lanes
suite-executions: 1
suite-exit-status: 1
suite-command-results:
  - command: (as above)
    result: fail
    exit-status: 1
retry-reason: none
```

fast-unit: 3 failed, 8503 passed, 1 skipped (browser-free not reached; verifier reason `executed_failure`).

**Genuine, PR-caused (deterministic, will also fail the hosted gate):**
1. `tests/test_repo_no_maintainer_specific_values.py::test_no_hardcoded_home_directory_paths` — the detector reads `config/home/eero_targets.json`, `data/home/eero_session.json`, `api/services/home/eero...` as `/home/<user>` paths. Hits: .gitignore:75, api/routes/home.py:68, api/services/home/eero.py:73,74,346,546, config/settings.py:1409,1411,1416, mcp_server.py:375-387, scripts/eero_login.py:17. `origin/main` has zero tracked `/home/` occurrences. Fix: rename the `home` path segment or extend the detector allowlist with justification.
2. `tests/test_no_change_narration.py::test_no_change_narration_beyond_baseline` — `(#1081)` issue citations in module docstrings: api/routes/home.py:2, api/services/home/eero.py:2, scripts/eero_login.py:3, tests/test_agent_tools_home_eero.py:3, tests/test_eero_login_script.py:2, tests/test_home_eero.py:2.

**Environment-caused:** `test_mcp_catalog_count_and_inter_agent_schema_contract` (68 == 71; reads live undeployed server on :8000). Passes with `LIFEOS_API_URL=http://127.0.0.1:1` (fallback catalog).

**Live eero scenario (Malea only, temp scheduler store, Telegram/human-queue mocked): PASS** — status 200 Malea paused=false; timed scheduled pause 200 paused=true mismatch=false resume_at set, scheduler_message "", one enabled once-entry → `/api/home/eero/Malea/resume` {scheduled:true}; raw vendor read paused=True; scheduled resume (entry marked triggered as the fire loop does) 200 paused=false, scheduler_message ""; raw read paused=False. Manual pause without minutes → indefinite (Malea has no default_minutes configured), message "Malea: paused"; manual resume "Malea: resumed". 404 unknown target, 422 minutes=0 and indefinite+minutes. Zero alerts sent. Malea ended resumed (raw read False).

nbramia:
Addressed both findings from verification round 1 (fast-unit lane).

1. **Home-path scan false positive** — `_HOME_PATH_RE` in `tests/test_repo_no_maintainer_specific_values.py` matched `/home/` embedded mid-path in relative paths and URL paths (`config/home/eero_targets.json`, `/api/home/eero/...`), not just genuine absolute home-directory paths. Added a negative lookbehind `(?<![A-Za-z0-9_.-])` so a match can't start when `/home/`/`/Users/` is immediately preceded by a path-segment character. Added two tests: one proving the false positive is gone (`config/home/eero_targets.json`, `POST /api/home/eero/status`), one proving real absolute paths (quoted, mid-command, `file://` URL) are still caught.
2. **Issue narration** — removed the `(#1081)` issue references from docstrings in `api/routes/home.py`, `api/services/home/eero.py`, `scripts/eero_login.py`, `config/settings.py`, `tests/test_agent_tools_home_eero.py`, `tests/test_eero_login_script.py`, `tests/test_home_eero.py`.

Focused tests: `tests/test_repo_no_maintainer_specific_values.py tests/test_no_change_narration.py tests/test_home_eero.py tests/test_eero_login_script.py tests/test_agent_tools_home_eero.py` — 108 passed. `ruff check` clean on all changed files.

Commit: eb01fc1f

nbramia:
## End-to-End Verification (re-verification) — PR #1082

### Verdict: FAIL (authoritative record) — caused by the host environment, no code defect found

```
verification-record: v1
verification-head: fb4c4238c9533301b1184bfa17d7825544a8491b
suite-result: fail
suite-command: /home/nathanramia/.venvs/lifeos/bin/python scripts/verify_candidate.py pushed-ref --repository /home/nathanramia/orca/workspaces/LifeOS/eero --sha fb4c4238c9533301b1184bfa17d7825544a8491b --base 40a931fc885183a671cf2c71a500cb8525e8f6a3 --evidence-root /home/nathanramia/.cache/lifeos/verification-evidence --lanes fast-unit,browser-free --workers 2 --lane-log-dir <scratch>/lanes
suite-executions: 1
suite-exit-status: 1
suite-command-results:
  - command: (as above)
    result: fail
    exit-status: 1
retry-reason: none
```

**fast-unit:** 8514 passed, 1 skipped, 1 failed, in 36m17s. `browser-free` never ran, because the run stops at the first failed lane.
- The only failure was `tests/test_agent_worker_mcp_exposure.py::test_mcp_catalog_count_and_inter_agent_schema_contract` (`assert 68 == 71`). This is the known problem on this host only. `verify_candidate.py`'s hermetic environment (`make_hermetic_environment`) drops `LIFEOS_API_URL`, so `mcp_server` loads `/openapi.json` from the live production server on :8000, which doesn't have this code yet. That server returned HTTP 200. CI has no server there.
- Confirmed: `LIFEOS_API_URL=http://127.0.0.1:1 pytest tests/test_agent_worker_mcp_exposure.py` gives 9 passed.
- Both earlier failures (`test_no_hardcoded_home_directory_paths`, `test_no_change_narration_beyond_baseline`) now pass.
- No same-head retry was run. The evidence store recorded this as `failure`, not an infrastructure failure. A retry would also fail the same way unless the production server is stopped, and that is not allowed.

**Read-only live check** (router mounted in a bare FastAPI app, so no app startup ran; scheduler store pointed at a temp dir; nothing was paused):
- `api.main` registers `/api/home/eero/status`, `/{name}/pause` and `/{name}/resume`.
- `GET /api/home/eero/status` → 200 from the live eero session: 3 profile targets (Malea among them), each with `name/paused(bool)/resume_at/type`.
- `POST /api/home/eero/<synthetic-unknown>/pause` → 404, and the message lists the configured targets.
- `POST /api/home/eero/Malea/pause {"minutes":0}` → 422, rejected by validation before any eero call.

The hosted `candidate-verification` check should be the authority for both lanes.

nbramia:
## Merge decision

Local re-verification at fb4c4238 ran fast-unit: 8514 passed, 1 skipped, 1 failed. The two earlier policy-test failures (`test_no_hardcoded_home_directory_paths`, `test_no_change_narration_beyond_baseline`) now pass.

**The remaining failure is caused by this host, not the code.**
- The failing test is `test_agent_worker_mcp_exposure.py::test_mcp_catalog_count_and_inter_agent_schema_contract` (68 vs 71 tools).
- `LifeOSMCPServer` loads the tool catalog from `localhost:8000/openapi.json`, and the production server there has not been redeployed with this code.
- `verify_candidate.py`'s hermetic environment drops `LIFEOS_API_URL`, so the fallback address can't be forced from outside.
- Run alone with an unreachable API URL, the test file passes (9 passed).
- The hosted runner has no server, so there the test uses the in-repo catalog.

Live checks against real eero on the Malea profile passed at 218305d9, and a read-only re-check passed at fb4c4238.

**Next step:** merge through `scripts/candidate_publisher.py`. It waits for the required App-bound `candidate-verification` check (fast-unit + browser-free on the hosted runner). Nothing is bypassed.

Follow-up worth filing: this test's dependence on a live :8000 server makes local lane runs fail on the API host whenever a PR changes the tool count.

PR #1116, opened 2026-09-16
fix(agent-worker): sweep terminal sessions for stranded vault tags

nbramia:
## Problem

A session row can reach a terminal status by a path that never reconciles the backing Obsidian task, stranding a card at `#agent-running` on the `/agents` board indefinitely while its session reads `failed`.

`SessionStore.update_status` fires the `set_status_projector` hook on a terminal write, which reconciles the vault tag. A kill does not go through it: `ExecutorRegistry.cancel_once` calls `mark_cancelled`, a raw status write that flips the row to `failed` alongside the cancellation guard without touching Markdown. When the session is still being actively polled this is harmless — the executor observes the terminal row on its next poll and reconciles through `update_status`. But a kill landing on a session parked at `blocked`, or landing while the worker is down, leaves no poll to do that, and nothing else ever revisits the row.

## Change

`Worker._reconcile_lifecycle_drift`, called once per `tick()`, heals that drift.

It lists tasks by non-terminal lifecycle tag (`GET /api/tasks?tag=agent-running`, then `agent-blocked`) and, for each, looks up the backing session with `session_store.get(task_id)` — a primary-key lookup against local SQLite. A task is healed only when its session exists, is vault-backed (`origin != "operator"` and no `parent_session_id`), and is in `TERMINAL_STATUSES`.

The heal writes through exactly the endpoints `_reconcile_vault_terminal` uses, with the same status-to-action mapping: `POST /api/tasks/{id}/swap-tag` to move the lifecycle tag, then `/complete` or a status-only `PUT`. Swap-tag is the only write the API accepts for a lifecycle tag on a claimed card — a `PUT` carrying a `tags` list is refused by the claim-tag guard in `api/routes/tasks.py`, and every card this sweep looks at is claimed by definition. The swap also gates the rest: it reports `swapped: false` when the tag it was told to replace is absent, so an operator retagging the card between the listing and this write is never overwritten, and the task status is only written for a card whose swap actually landed.

The candidate set is the drift set by construction, which is what keeps this simple:

- A task whose session is still live — including one reopened for a follow-up turn, which flips the tag back to `#agent-running` while leaving the session non-terminal — is excluded by the status check, not by special-casing.
- A session parked at `blocked` is non-terminal, so it is a candidate the sweep skips; bringing a park's tag into line belongs to the path that performs the park. `#agent-blocked` is still listed, because it is a legitimate resting place for a card whose session later goes terminal.
- A task already settled at a terminal tag is never listed in the first place, so an operator's `#accepted` decision cannot be overwritten.
- No candidate bookkeeping is persisted, so there is no state that can go stale, and a failed listing simply contributes no candidates that tick.

Cost is two HTTP listing calls per tick regardless of how many sessions or tasks exist, plus two writes per task actually healed. Heals are rare and self-limiting — once a task is healed its tag is terminal and it stops being listed.

The sweep runs ahead of the spend-cap gate: it reconciles existing state and never starts new work.

## Scope

`session_store.py` is unchanged apart from one line. No schema change, no new indexes, no new dependencies, no changes to the kill path, the executors, or the HTTP client contract.

## Tests

`tests/test_task_lifecycle.py`, 24 tests. The sweep-specific ones pin:

- A kill parked at `blocked` heals; a kill that landed while the worker was down heals.
- The heal drives the real FastAPI task routes end to end, not a stubbed client.
- A concurrent operator edit is not clobbered: a tag that moved between the listing and the write leaves the card alone.
- A live session is left byte-identical at every non-terminal status (`claimed`, `running`, `blocked`).
- A task already settled at a terminal tag is left byte-identical, tags and `updated_at`.
- A tagged task with no backing session row is left alone.
- A live kill already reconciled by the ordinary path is not double-projected.
- A reopen with a new attempt, and a reopen on the same attempt, both heal when re-killed.
- A transient tag-listing failure heals nothing, persists nothing, and recovers on the next tick.
- The sweep is wired into `tick()` and still runs when the spend cap blocks.
- Operator-origin and spawned-child sessions are excluded.


nbramia:

**Reviewers run:** general, correctness, testing

**Specialist rationale:** `correctness` — the change is state-machine logic interacting with an existing projector, a cancellation fence, and CAS markers. `testing` — witness-test falsifiability was the central risk, since the implementation reported its witness tests failing only with `AttributeError`. Architecture, security, and docs were not selected: no new layer or dependency, no trust boundary, and this repository folds documentation relevance into the general review.

### Reviewer Findings

#### General

**Defects.** `list_terminal_unprojected`'s "already projected" check is keyed only on `(task_id, target_status)`, not `attempt_id`, so once a task has been reconciled once to a given terminal status, the sweep can never reconcile a *later*, independent drift on that same task at the same status — exactly the reopen-for-followup interaction the brief called out. Reachable path, reproduced against the worktree code: session A on task T killed while parked at BLOCKED → sweep reconciles (projection row for `(T, "failed")`); T reopened via `begin_new_execution` (new `attempt_id`, tag back to `agent-running`); session B killed while parked → new drift; the `NOT EXISTS` check matches session A's stale applied row and excludes session B. `healed == 0`, tag stuck forever. Minimal fix: add `attempt_id` to the subquery — the column already exists and every writer already records it.

**Missing Tests.** No test covers reopen-then-kill-again ending at the same terminal status. `test_drift_sweep_leaves_a_reopened_followup_session_alone` only proves the sweep ignores a *live* reopened session; it never advances that session to a second terminal drift.

**Status.** `pytest tests/test_task_lifecycle.py -q` → 10 passed. No broad gate run.

#### Correctness

**Defects.** Same defect, independently reproduced: `list_terminal_unprojected` (`session_store.py:1816-1858`) matches by `(task_id, target_status, state='applied')` only, while `event_id` is attempt-scoped everywhere else in the codebase. A task that fails, is re-dispatched, and is killed while parked a second time has its second drift silently skipped — `list_terminal_unprojected(since=0, limit=10)` returns `[]` and the sweep heals 0, leaving the tag at `agent-blocked` while the row is terminal FAILED. Retry-after-failure is a supported, tested flow, so this is plausible rather than exotic.

**Missing Tests.** All four new tests use a single attempt per task; the cross-attempt false negative is untested.

**Status.** `tests/test_task_lifecycle.py` → 10 passed; `tests/test_agent_worker_loop.py` + `tests/test_agent_kill_api.py` → 94 passed. Traced and found sound: the `mark_cancelled` vs `update_status`/hook split, CAS `expected_version` handling, the `origin`/`parent_session_id` NULL parity, the reopened-follow-up guard, and same-tick dispatch ordering.

#### Testing

**Defects.** None — the four witness tests exercise real store/projector state and are not self-referential.

**Missing Tests.** Three, all mutation-proven:
1. **`tick()` wiring is unpinned.** Replacing the `self._reconcile_lifecycle_drift()` call in `tick()` with `pass` leaves **10/10 passing**. Every test calls the method directly on a `Worker.__new__` stub.
2. **The `has_vault_task` SQL exclusion has no test.** Deleting both `origin != 'operator'` and `parent_session_id IS NULL` clauses leaves **10/10 passing**. No fixture ever creates an operator-origin or child session.
3. **Cross-attempt second drift** — same defect the other two reviewers reproduced, confirmed here against unmutated code.

**Status.** Repeated focused runs; all mutations reverted, `git status --short` and `git diff --stat` empty.

### Referee Decisions

Independently verified each finding against the code, and ran the sweep's predicate against the live production database.

| 1 | Sweep rewrites tasks that are not drifted, destroying operator decisions | *referee* | Defect | Valid | **Accept** | The implementation substituted "no applied projection row" for the brief's drift signature ("terminal session **and** non-terminal task tag"). Not equivalent: `lifecycle_projections` postdates most sessions — production holds **0 applied rows** against 150 vault-backed terminal sessions, so every historical session matches. Simulated against production: 111 skipped (task deleted), 35 redundant rewrites, **4 actual rewrites of which only 1 is real drift**. `debc4073` and `2f9a6faa` are `#accepted #agent-completed` / `done` and would be rewritten to `#agent-failed` / `cancelled`; `da54a5a7` is re-queued `#agent` / `done` and would be stamped `#agent-completed`. CAS does not protect — the sweep re-fetches and passes the current `updated_at`, so `expected_version` always matches; CAS guards a fetch/write race, not action on stale session state. Remedy: after `_fetch_task`, project only when the task still carries a non-terminal lifecycle tag. Tags live in Markdown, so this belongs in `_reconcile_lifecycle_drift`, not the SQL. |
| 2 | Projection dedupe not attempt-scoped | general + correctness + testing (deduplicated) | Defect | Valid | **Accept** | Confirmed: `attempt_id` exists on `lifecycle_projections` and the live path records it non-NULL (`worker.py:709-713`). Remedy: scope the `NOT EXISTS` to the attempt, **NULL-safe** — SQLite `IS`, not `=`; `=` against a NULL `attempt_id` yields NULL and the row would look permanently unprojected. Note for the fix: the sweep mints `event_id` with suffix `lifecycle_drift_sweep` while the live path uses `turn_id`, so the two paths produce **different** event ids for the same transition and `begin_projection` dedupe does *not* protect across them — this SQL predicate is load-bearing for the no-double-project property. |
| 3 | Unhealable rows are never resolved and are rescanned forever | *referee* | Defect | Valid | **Accept** | `_fetch_task` returning None hits `continue` with no marker written, so the row matches on every tick permanently. Measured: **111 of 150**. `_fetch_task` is an HTTP GET per row (`worker.py:4135`), so this is real per-tick cost against the live API, and `ORDER BY last_activity_at ASC` means these oldest rows crowd the batch first (unbounded scan: 27 of 50 dead). Remedy: record a resolved/abandoned marker when the task is gone so the row stops matching. |
| 4 | 24h window caps the worker-downtime trigger the PR exists to close | correctness (as context) | Defect | Valid | **Accept** | Raised as a discussion point; refereed as a defect. A kill landing via the HTTP API while the worker is down longer than the window is trigger #2 from the brief and is permanently missed — this host has a documented multi-week outage. Proportionate remedy: widen the window substantially (keep `limit`), **after** #3 lands — widening first would make the dead-row cost acute. Not "remove all bounds". |
| 5 | `tick()` wiring untested | testing | Missing Test | Valid | **Accept** | Mutation-proven: deleting the call leaves the suite green. Add one test driving the real `tick()` entry point. |
| 6 | `has_vault_task` SQL gate untested | testing | Missing Test | Valid | **Accept** | Mutation-proven. Add a fixture with an operator-origin session and one with a `parent_session_id`. |
| 7 | Cross-attempt second drift untested | general + correctness + testing | Missing Test | Valid | **Accept** | Regression test for #2: reopen via `begin_new_execution`, re-drift, assert healed. |
| 8 | No test for the production-shaped false positive | *referee* | Missing Test | Valid | **Accept** | Regression test for #1: a terminal session whose task is already `#accepted #agent-completed` / `done` must be left byte-identical by the sweep. This is the case the empty-database fixtures structurally cannot catch. |

**Findings forwarded to addresser:** 8

**Result:** merge blocked pending round 2.

nbramia:
## Round 1 fixes

| 1 | Sweep rewrites tasks that are NOT drifted | Applied | `_reconcile_lifecycle_drift` now skips a candidate unless the task's *current* vault tag is still `agent-running`/`agent-blocked`. `list_terminal_unprojected`'s "no applied projection" signature was only ever a candidate filter (it also matches every historical terminal session that predates the projector); this closes the gap that would have destroyed an operator's `#accepted`. |
| 2 | Projection dedupe is not attempt-scoped | Applied | `list_terminal_unprojected`'s `NOT EXISTS` subquery now adds `AND p.attempt_id IS s.attempt_id` (NULL-safe `IS`, not `=`), so an earlier attempt's applied projection can no longer mask a genuine second drift on a later attempt of the same task/status. |
| 3 | Rows that can never be healed are rescanned forever | Applied | When `_fetch_task` returns `None` (task deleted from the vault), the sweep now records a resolved projection marker (`begin_projection` + `acknowledge_projection`, reusing the existing table) instead of silently `continue`-ing, so the row stops matching the candidate query on every future tick. |
| 4 | 24h window caps the worker-downtime trigger | Applied | Widened `_LIFECYCLE_DRIFT_SWEEP_WINDOW_S` from 24h to 90 days, now safe because #3 caps the cost of permanently-unhealable rows. 90 days gives comfortable margin over the longest documented worker outage on this host (18 days). |
| 5 | `tick()` wiring untested | Applied | Added `test_tick_wires_in_the_lifecycle_drift_sweep` — drives the real `tick()` entry point with every other sub-step stubbed. Verified by mutation: replacing the `_reconcile_lifecycle_drift()` call in `tick()` with `pass` turns this test red. |
| 6 | `has_vault_task` SQL gate untested | Applied | Added `TestListTerminalUnprojectedHasVaultTaskGate` with fixtures for an operator-origin session and a spawned-child session (plus a control case). Verified by mutation: neutering the `origin != 'operator'` / `parent_session_id IS NULL` clauses turns both exclusion tests red. |
| 7 | Cross-attempt second drift untested | Applied | Added `test_drift_sweep_heals_a_second_drift_after_reopen_with_new_attempt` — heals a first drift, reopens via the real `begin_new_execution` (new `attempt_id`), kills again at the same terminal status, asserts the second drift is healed. Confirmed this test fails on pre-fix code (`0 == 1`). |
| 8 | No test for the production-shaped false positive | Applied | Added `test_drift_sweep_leaves_a_settled_task_alone_when_no_projection_was_ever_recorded` — a terminal, unprojected session backing a task already at `#accepted #agent-completed`/`done` is left byte-identical. Confirmed this test fails on pre-fix code (`1 == 0`). |

**Tests:** `~/.venvs/lifeos/bin/python -m pytest tests/test_task_lifecycle.py tests/test_agent_worker_loop.py tests/test_agent_worker_human_queue.py tests/test_agent_kill_api.py tests/test_agent_worker_stores.py tests/test_agent_worker_executor_lifecycle.py tests/test_agent_worker_self_restart.py tests/test_agent_resume_api.py -q` — 218/218 passed. `test_drift_sweep_does_not_double_project_a_live_kill_already_reconciled` stays green. Ruff clean on all touched files.

**Commit:** `fix: address review round 1 — drift sweep false positives, unbounded rescans`

nbramia:

**Reviewers run:** general, correctness, testing

**Specialist rationale:** the fix delta changed all three surfaces — worker control flow (general), the projection/marker state machine (correctness), and 156 lines of new tests (testing). Reviewers were scoped to `git diff 64c652ed..HEAD` and told not to reopen resolved round-1 findings.

### Reviewer Findings

#### General

**Defects.** None. Specifically searched for siblings of the "broad guard conflates distinct causes" pattern elsewhere in `worker.py` and `session_store.py`; found none.

**Missing Tests.** The finding-#3 fix (`worker.py:2690-2700`, recording a resolved marker when `_fetch_task` returns None) has zero coverage — no test verifies the marker is written or that a later sweep stops rescanning that row.

**Status.** `tests/test_task_lifecycle.py` + `tests/test_agent_worker_loop.py` → 101 passed; `tests/test_agent_kill_api.py` → 9 passed.

**Verified resolved:** #1 settled-tag skip, #2 NULL-safe `attempt_id IS` scoping, #4 90-day window, #5–#8 witness tests.

#### Correctness

**Defects.** The "already-settled" branch (`worker.py:2704-2709`) never records a resolved marker, unlike its sibling "task is None" branch two blocks above. Combined with the 90-day window and `ORDER BY s.last_activity_at ASC LIMIT 50`, an already-settled-but-never-projected session permanently occupies a batch slot — nothing marks it done and it only ages out after ~90 days. Once that backlog reaches 50 it crowds out newer genuine drift entirely. Reproduced against real `SessionStore`/`TaskManager`: 55 seeded settled rows plus 1 genuinely-drifted row with a newer `last_activity_at` → `list_terminal_unprojected(since=0, limit=50)` returns 50 rows never including the drifted task; three consecutive `_reconcile_lifecycle_drift()` calls heal 0 and the tag is never fixed. Suggested fix: mirror the existing `task is None` branch and write a resolved marker.

**Missing Tests.** (a) No test exercises `p.attempt_id IS s.attempt_id` against an actual NULL `attempt_id` row — the column was added by `ALTER TABLE` with no backfill, and `SessionStore.create()` always mints a non-null id. (b) No test asserts the batch is self-advancing when non-actionable rows exceed `_LIFECYCLE_DRIFT_SWEEP_LIMIT`.

**Status.** `tests/test_task_lifecycle.py` → 16 passed, plus two scratch repro scripts against real stores. Traced clean: sweep-vs-dispatch ordering (sweep runs synchronously before dispatch; a cross-tick race is covered by `transition()`'s CAS and yields a benign conflicted marker plus a retry), and the missing-task marker's interaction with `replay_pending()` (filters `state != 'applied'`) and `projection_applied()` (exact `event_id` match).

#### Testing

**Defects.** None — all four round-1 witness tests genuinely falsify what they claim; each targeted mutation reddens exactly the claimed test and nothing else.

**Missing Tests.** Three, all mutation-proven against the 16-test file:
1. The resolved-marker write (`worker.py:2689-2701`) — deleting the whole block leaves all 16 green.
2. The NULL-safe `IS` scoping (`session_store.py:1863`) — reverting to `=` leaves all 16 green, since no fixture ever produces a NULL `attempt_id`.
3. `test_tick_wires_in_the_lifecycle_drift_sweep` pins only that the sweep is called *somewhere* in `tick()`, not the documented "never gated on the spend cap" invariant one line above the call site. Moving the call past the spend-cap gate still passes, because the test hardcodes `can_start_task` to always return `True`.

**Status.** All mutations reverted; `git status` clean.

### Referee Decisions

| 1 | A transient `_fetch_task` failure permanently suppresses healing of real drift | *referee* | Defect | Valid | **Accept** | `_fetch_task` (`worker.py:4133-4149`) returns `None` for a definitive 404 **and** for any exception. The round-1 #3 fix treats every `None` as "deleted" and writes an applied marker, removing the row from the candidate query forever. Reproduced on branch code in a tmp fixture: tick 1 with a simulated blip, tick 2 with a healthy API and the task plainly present → `healed=0` both ticks, row no longer a candidate, task still `#agent-running`. Reachable here: the worker ticks every 60s and auto-deploy restarts the API ~10 min after `main` advances. Remedy: write the marker only on a definitive 404; on "unavailable", skip without a marker so the next tick retries. |
| 2 | The already-settled branch writes no marker, so non-actionable rows permanently hold batch slots | correctness | Defect | Valid | **Accept** | Reproduced by the reviewer. Refereeing the severity against production rather than the seeded repro: of 150 vault-backed terminal sessions, 111 are dead (these *do* get markers under the round-1 fix and drain in ~3 ticks), **38 are settled-with-no-marker**, and 1 is genuine drift. So the batch is not starving today — 38 of 50 slots, with 12 to spare — but the settled class only ever grows, and at 76% of every batch the margin is thin. Latent-but-growing, with a trivially symmetric fix, so it is worth closing now rather than after it bites. Remedy: mirror the `task is None` branch and record a resolved marker on the settled path. |
| 3 | The resolved-marker write has no test | general + testing (deduplicated) | Missing Test | Valid | **Accept** | Mutation-proven: deleting the block leaves all 16 tests green. Test both halves so they are pinned apart rather than collapsed: a definitively-deleted task gets a marker and stops being a candidate; a failed *fetch* does not get a marker and is healed on a later tick. The second half is the regression test for #1. |
| 4 | The NULL-safe `IS` scoping is unpinned | testing + correctness | Missing Test | Valid | **Accept** | Mutation-proven: reverting to `=` leaves all 16 green. I checked reachability before accepting, since pinning an unreachable legacy case would be hardening rather than coverage — it is the opposite: **257 of 264 production sessions have a NULL `attempt_id`**, including 75 of 76 `failed` ones. This covers the overwhelming majority of the real table. Remedy: a fixture with an explicitly NULL `attempt_id` row. |
| 5 | The tick-wiring test does not pin the spend-cap invariant | testing | Missing Test | Valid | **Accept** | Mutation-proven: moving the call past the spend-cap gate still passes. The invariant is documented one line above the call site, so it is behavior the diff claims to deliver. Remedy: same test with `can_start_task` returning `False`, asserting the sweep still heals. |
| 6 | No test that the batch self-advances past non-actionable rows | correctness | Missing Test | Valid | **Accept** | Pairs with #2 — without it the starvation fix is unpinned. Remedy: seed more non-actionable rows than `_LIFECYCLE_DRIFT_SWEEP_LIMIT` with older `last_activity_at` than a genuinely-drifted row, and assert the drift is still reached. |

**Findings forwarded to addresser:** 6 (round 1 forwarded 8 — converging)

**Result:** merge blocked pending round 3.

nbramia:
## Round 2 findings addressed

All 6 accepted findings fixed. Both defects were in `_reconcile_lifecycle_drift` / `_fetch_task`.

**#1 — Transient fetch failure permanently suppressed healing (Defect, Applied).** `_fetch_task` collapsed a definitive 404 and any exception into a single `None`, so the sweep treated a transient API error (e.g. a restart mid-tick) the same as "task deleted" and wrote a resolved marker that never let the row heal. Added `_fetch_task_or_absent(task_id) -> (task, is_absent)`; `is_absent` is `True` only on a confirmed 404. `_fetch_task` now wraps it, unchanged for its other ~14 call sites. The sweep only records a resolved marker when `is_absent` is `True`; an unavailable fetch is skipped with no marker so the next tick retries. Witness test `test_drift_sweep_does_not_heal_or_mark_resolved_on_a_transient_fetch_failure`: mutating the code back to marking-on-any-`None` turns it red (`assert [] != []` fails — no candidate row survives the bad path).

**#2 — Settled branch wrote no marker, starving the batch (Defect, Applied).** The `RUNNING_TAG not in tags and BLOCKED_TAG not in tags` branch recorded nothing, so settled rows stayed candidates forever and crowded the bounded 50-row batch. Added `_mark_drift_row_settled()` (shared by both the deleted-task and settled-task branches) and call it here too. Witness tests `test_drift_sweep_marks_a_settled_row_resolved_so_it_stops_being_a_candidate` and `test_drift_sweep_batch_self_advances_past_non_actionable_rows` both go red on the pre-fix code (`healed_total == 0` instead of `1`, backlog of `_LIFECYCLE_DRIFT_SWEEP_LIMIT + 5` settled rows permanently starves the one genuine drift).

**#3 — Resolved-marker write untested (Missing Test, Applied).** Pinned both halves apart: `test_drift_sweep_marks_a_definitively_deleted_task_resolved` (404 → marker written, row excluded on next sweep) and `test_drift_sweep_does_not_heal_or_mark_resolved_on_a_transient_fetch_failure` (unavailable fetch → no marker, still a candidate, heals once the fetch succeeds — the #1 regression test). Mutation-verified separately: deleting the `is_absent` check turns only the second test red.

**#4 — NULL-safe `IS` scoping untested (Missing Test, Applied).** Added `TestListTerminalUnprojectedNullAttemptScoping`: inserts a session with `attempt_id` nulled directly via SQL, confirms it's a candidate, applies a matching NULL-attempt projection, confirms it's then excluded. Mutation-verified: reverting `IS` to `=` turns it red.

**#5 — Tick-wiring test didn't pin the spend-cap invariant (Missing Test, Applied).** Added `test_tick_runs_the_lifecycle_drift_sweep_even_when_the_spend_cap_blocks` with `can_start_task=lambda estimate: False`. Mutation-verified: moving the `_reconcile_lifecycle_drift()` call after the cap gate turns this new test red while leaving the existing `test_tick_wires_in_the_lifecycle_drift_sweep` green (confirming that test alone couldn't have caught the regression).

**#6 — No test for batch self-advancing past non-actionable rows (Missing Test, Applied).** `test_drift_sweep_batch_self_advances_past_non_actionable_rows` seeds `_LIFECYCLE_DRIFT_SWEEP_LIMIT + 5` settled rows older than one genuinely-drifted row and asserts the drift is healed within 3 ticks. Also serves as the #2 regression test (see above).

All 22 tests in `tests/test_task_lifecycle.py` pass (16 original + 6 new); ruff clean.

nbramia:
## Convergence Audit — after Round 2

Posting this instead of dispatching round 3. The loop is converging by count (8 → 6) but one sub-problem has produced a finding in **every round**, and each fix for it has introduced the next one.

### Finding density by sub-problem

| Sub-problem | R1 | R2 | R3 (referee) | Density |
|---|---|---|---|---|
| **Projection-marker bookkeeping as the candidate filter** | #2 dedupe not attempt-scoped | #2 settled rows hold batch slots forever | settled marker masks a later same-attempt re-drift | **3** |
| Drift signature (tag predicate) | #1 destructive rewrites | — | — | 1 |
| Transient-vs-definitive fetch | — | #1 blip suppresses healing | — | 1 |
| Window / cost bounds | #3, #4 | — | — | 2 |
| Test coverage | #5–#8 | #3–#6 | — | 8 |

Everything outside the first row is resolved and stable. The first row is a trap.

### The round-3 defect

Reproduced against `b0d8fede` with real `SessionStore`/`TaskManager`/`Worker`:

1. A task settled at `#agent-failed` with a terminal session on attempt `A1`.
2. The sweep writes a settled marker for `(task, A1, failed)` — correct at the time.
3. The reopen-for-pending-messages path (`worker.py:3815-3823`) swaps the terminal tag back to `#agent-running` and calls `update_status(..., STATUS_CLAIMED, attempt_id=current.attempt_id)` — **same attempt, no `begin_new_execution`**.
4. The session is killed while parked → terminal again on `A1`, same status.
5. `list_terminal_unprojected` excludes it via the step-2 marker. `healed=0`, tag stuck at `#agent-running` permanently.

The comment added in round 2 justifying the marker states that a terminal vault tag "never flips back to running/blocked on its own." Step 3 is that flip.

### Assessment

The marker bookkeeping is **review-introduced architecture**, not part of the original task. It entered in R1 #3 purely as a cost optimization — to stop unhealable rows costing an HTTP GET per tick — and has since needed attempt-scoping (R1), a second write site (R2), and now correct invalidation across same-attempt reopens (R3). Each fix is locally correct and each one exposes the next gap, because the marker encodes *"this (task, attempt, status) was judged non-actionable once"* while the question the sweep actually needs answered is *"has anything changed since I last looked?"*

Per the lifecycle contract's convergence rule, I am stopping before round 3 and requesting direction rather than forwarding a fourth patch to the same mechanism.

### Recommended bounded simplification

Replace the projection-marker candidate filter with a single `sessions.drift_swept_at` timestamp column:

- The sweep skips any row where `drift_swept_at >= last_activity_at`.
- After examining a row and finding it non-actionable, it stamps `drift_swept_at = now()`.
- Both `update_status` and `mark_cancelled` already bump `last_activity_at` on every state change, so any genuine re-drift re-qualifies the row automatically.

This is immune to all three variants above because it keys on *change*, not on a judgement about an attempt/status tuple. It removes the attempt-scoping predicate, both marker write sites, and the 404-vs-error coupling to marker-writing — while preserving the cost property that motivated the marker. It also stops the sweep writing rows into `lifecycle_projections`, a table owned by real lifecycle transitions, for sessions that are having no transition at all.

Net effect: a smaller diff than what is currently on the branch.

nbramia:
## Round 3 — rework: watermark replaces the projection-marker filter

Per the convergence audit's recommended simplification (approved by the operator): replaced the projection-marker candidate filter with a `sessions.drift_swept_at` watermark, compared against `last_activity_at`.

| 1 | Replace the projection-marker candidate filter with a `drift_swept_at` watermark | Applied (rework) | Added a nullable `drift_swept_at INTEGER` column to `sessions` via the existing `ALTER TABLE` migration pattern. `list_terminal_unprojected` (renamed `list_terminal_unswept`) drops the attempt-scoped `NOT EXISTS (... lifecycle_projections ...)` subquery entirely, replaced with `AND (s.drift_swept_at IS NULL OR s.drift_swept_at < s.last_activity_at)`. Deleted `_mark_drift_row_settled` and both call sites, replaced with `SessionStore.mark_drift_swept(session_id)`. The sweep no longer writes into `lifecycle_projections` for a non-actionable row — that table is now touched only by `LifecycleProjector.transition` for an actual heal, which is unchanged. |
| 2 | Settled marker masks a later same-attempt re-drift | Fixed structurally | No longer reachable: `update_status`/`mark_cancelled`/the CAS update all bump `last_activity_at` on every write, including a same-attempt reopen, so the row automatically re-qualifies as a candidate regardless of any earlier stamp. Deleted the round-2 comment claiming a terminal tag "never flips back to running/blocked on its own." |
| 3 | TRAP — do not stamp the watermark on a failed fetch | Applied | `mark_drift_swept` is called only on a confirmed 404, an already-settled tag, or a transition that returned `applied=True`. A merely-unavailable fetch (`is_absent=False`) or a CAS conflict (`transition` returns `False`) is skipped with no stamp, so the next tick retries. `_fetch_task_or_absent`/`_fetch_task` are unchanged. |
| 4 | Preserve every behavior already pinned by the existing 22 tests | Applied | All 22 tests pass, renamed/adapted where they asserted on `lifecycle_projections` rows as an implementation detail: the live-kill-already-reconciled test's pre-sweep assertion now reflects "still a candidate, but nothing to heal" rather than "already filtered by an applied marker"; the NULL-`attempt_id` scoping test (which pinned a NULL-safety hazard in the removed subquery) is rewritten to pin the equivalent watermark behavior on a NULL-attempt row, since the new predicate never references `attempt_id` at all. |
| 5 | Regression test for the round-3 defect | Applied | `test_drift_sweep_heals_a_reopen_on_the_same_attempt_after_being_marked_settled`: settle → sweep (stamps) → reopen the SAME attempt (tag back to running, `update_status(..., STATUS_CLAIMED, attempt_id=<same>)`) → `mark_cancelled` on the same attempt → sweep again → asserts `healed == 1`. Confirmed failing against `b0d8fede` via a standalone repro using the pre-rework `list_terminal_unprojected`/marker API in a throwaway worktree: `second_healed == 0`, tag stuck at `agent-running`. |
| 6 | Watermark invalidation must be pinned directly | Applied | `TestDriftSweepWatermarkInvalidation`: one test pins the store-level predicate (stamped row excluded while `last_activity_at` unchanged, re-included once it advances past the stamp); the other pins the cheap-tick property at the worker level with a fetch-counting stub (no re-fetch on an unchanged row, a re-fetch once genuine drift occurs). Mutation-verified: reverting the predicate to `drift_swept_at IS NULL` (dropping the `last_activity_at` comparison) reddens both of these plus the round-3 regression test in #5. |

**Tests:** `~/.venvs/lifeos/bin/python -m pytest tests/test_task_lifecycle.py tests/test_agent_worker_loop.py tests/test_agent_kill_api.py tests/test_agent_worker_stores.py tests/test_agent_worker_human_queue.py tests/test_agent_worker_executor_lifecycle.py tests/test_agent_worker_self_restart.py tests/test_agent_resume_api.py -q` — 227/227 passed (25 in `test_task_lifecycle.py`: 22 original + 1 round-3 regression + 2 watermark-invalidation). Ruff clean on all touched files.

**Diff vs `b0d8fede`:** production code net `session_store.py` +26 lines (new column, migration, `mark_drift_swept`, updated docstrings), `worker.py` −1 line (net flat: the sweep's control flow is simpler — one stamp call at each of three exit points instead of a helper plus two call sites — offset by an expanded docstring explaining the watermark's invalidation property). The mechanism itself is smaller: no more attempt-scoped `NOT EXISTS` subquery, no more `_mark_drift_row_settled`, no more writes into `lifecycle_projections` for non-transitions.

**Commit:** `fix: replace projection-marker drift-sweep filter with a watermark`

nbramia:
## Round 4 — rework: invert the sweep, delete the bookkeeping

Operator-directed rework per the convergence audit's recommendation, going one step further than the audit's watermark proposal: instead of iterating sessions and filtering with an HTTP task fetch, the sweep now queries tasks directly by non-terminal lifecycle tag (`GET /api/tasks?tag=agent-running` / `agent-blocked`, no pagination) and checks each backing session locally via `session_store.get(task_id)` (a primary-key lookup). The candidate set IS the drift set now, so no per-row bookkeeping is needed at all.

| 1 | Invert the sweep: query tasks by non-terminal tag, then check the session locally | Applied (rework) | `_reconcile_lifecycle_drift` now loops over `(RUNNING_TAG, BLOCKED_TAG)`, calling a new `_list_tasks_by_tag(tag)` (one `GET /api/tasks?tag=...` each). For each returned task, `session_store.get(task_id)` finds the backing session; skipped when absent, when `has_vault_task` is false (`origin == 'operator'` or `parent_session_id` set), or when `session.status not in TERMINAL_STATUSES` (legitimately live). A genuine drift is healed through the same `LifecycleProjector.transition` as before, using `task.get("updated_at")` from the listing response directly — no second per-task fetch. Cost: exactly 2 HTTP calls per tick, constant regardless of table size. |
| 2 | Delete the bookkeeping the inversion makes unnecessary | Applied | Removed `SessionStore.list_terminal_unswept`, `mark_drift_swept`, the `drift_swept_at` column (dataclass field, `CREATE TABLE` column, row mapping) and its `ALTER TABLE` migration (existing dev DBs keep the column harmlessly, no down-migration), `_LIFECYCLE_DRIFT_SWEEP_WINDOW_S`/`_LIMIT`, and `idx_sessions_status_activity` (grepped — served only the removed query). Also dropped `idx_lifecycle_projections_task`, which grepped out as already dead from the round-3 watermark rework (no remaining query joins on `(task_id, target_status, state)`). `_fetch_task_or_absent`'s `is_absent` split had no other production consumer (grepped) — reverted to a plain `_fetch_task` matching its pre-PR shape. |
| 3 | Preserve every behavior the 25 tests pin | Applied | All acceptance-list behaviors kept, adapted to the new mechanism. Deleted 5 tests outright (see below) whose pinned property no longer exists; nothing was contorted to keep a green count. |
| 4 | Round-3 gaps (stamp-write test, no-write-for-non-transition) are moot | Confirmed | (a) disappeared with the stamp itself — nothing to test. (b) is structurally guaranteed: the sweep's only projector call sites are inside the genuine-drift branch, so there's no code path where a non-transition could write a projection row; not worth a dedicated test for a branch that provably doesn't exist. |
| 5 | Add: transient API failure must corrupt nothing | Applied | `test_drift_sweep_survives_a_transient_tag_listing_failure_and_heals_on_the_next_tick`: makes `_list_tasks_by_tag` raise for one tick, asserts zero heals and byte-identical tags/`updated_at`, then restores it and asserts a normal heal next tick. Mutation-verified: removing the `try/except` around the per-tag call reddens it (`RuntimeError` propagates). |
| 6 | Add: a task tagged non-terminal with a live session is never touched | Applied | `test_drift_sweep_leaves_a_live_session_alone_regardless_of_status`, parametrized over CLAIMED/RUNNING/BLOCKED (subsumes the old single-case reopened-followup test). Mutation-verified: deleting the `session.status not in TERMINAL_STATUSES` guard reddens the RUNNING and BLOCKED cases (2 of 3 fail: `1 == 0`). The CLAIMED case still passes even mutated — `TaskManager.update`'s own `VALID_STATUSES` check independently rejects `status="claimed"` and the projector catches that and returns `False` — an incidental second safety net, not a flaw in the guard being tested; the test still demonstrably catches the regression via the other two parametrizations. |

**Tests deleted outright** (property no longer exists under the new mechanism, not contorted to stay green):
- `test_drift_sweep_marks_a_settled_row_resolved_so_it_stops_being_a_candidate` — pinned the settled-row marker write; there's no marker, and a settled task never carries `RUNNING_TAG`/`BLOCKED_TAG` so it's never even a candidate.
- `test_drift_sweep_marks_a_definitively_deleted_task_resolved` — pinned the 404-marks-resolved branch; there's no more per-row task fetch to 404 on (the task comes from the tag-listing response itself).
- `test_drift_sweep_does_not_heal_or_mark_resolved_on_a_transient_fetch_failure` — pinned the is_absent-vs-unavailable distinction for that same removed per-row fetch; superseded by finding #5's new test at the listing level.
- `test_drift_sweep_batch_self_advances_past_non_actionable_rows` — pinned that a bounded batch doesn't starve on a settled-row backlog; there's no batch limit anymore (the tag-listing endpoint takes no pagination param).
- `TestListTerminalUnsweptNullAttemptRows` / `TestDriftSweepWatermarkInvalidation` — pinned NULL-safety and invalidation of query predicates (`p.attempt_id IS s.attempt_id`, then `drift_swept_at`/`last_activity_at`) that belonged to the removed candidate-filter mechanisms; `attempt_id` is now only ever read to build the outgoing event, same as every other terminal-write call site.

**Net line delta:**
- vs `9b173d6f` (round-3 HEAD): `session_store.py` −87 net (drops back to nearly untouched vs `main`), `worker.py` −41 net (rewritten sweep + new `_list_tasks_by_tag` is smaller than the old candidate query + marker calls + reverted `_fetch_task` split), `tests/test_task_lifecycle.py` −390 net. **Total: 218 insertions / 535 deletions, net −317 lines.**
- vs `main` (PR base): 608 insertions / 3 deletions (net +605), down from round-3's 929 insertions / 7 deletions (net +922) — session_store.py is now within 1 line of untouched.

**Tests:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/python -m pytest tests/test_task_lifecycle.py tests/test_agent_worker_loop.py tests/test_agent_worker_human_queue.py tests/test_agent_kill_api.py tests/test_agent_worker_stores.py tests/test_agent_worker_executor_lifecycle.py tests/test_agent_worker_self_restart.py tests/test_agent_resume_api.py -q` — 223/223 passed. Ruff clean on all touched files.

**Commit:** `fix: address review round 4 — invert the drift sweep, delete the bookkeeping`

nbramia:
## Review Round 5 — Consolidated Review & Referee Decisions

Reviewing commit `2fb0726e`, the state after the operator-directed inversion. **No production defects.** One stale artifact and three test gaps.

**Reviewers run:** general, testing

**Specialist rationale:** correctness was dropped this round — the sweep is now ~30 lines of straight-line logic with no persistent state, so the projection/marker state machine that justified that specialist no longer exists. Testing was retained specifically to audit the five test deletions.

### Reviewer Findings

#### General

**Defects.** The PR's top-level description still documented the deleted mechanism — `list_terminal_unprojected(since, limit)`, the window/limit constants, two indexes, a "24h / 50 rows" bound — none of which exist (`grep` confirms no hits). Its test evidence cited a test name that no longer exists and a 10-passed count against a file that now has 21 tests exercising a different mechanism. Its "Review focus" section pointed reviewers at code that isn't there.

**Missing Tests.** A task carrying `RUNNING_TAG`/`BLOCKED_TAG` with **no backing session row** (operator hand-edited the tag, or the row was pruned) is skipped by `worker.py:2697-2698`, but nothing pins it.

**Status.** 21 passed; 131 passed across four files; ruff clean. Verified independently: the `has_vault_task` gate matches the one used elsewhere in the module, live and reopened sessions are excluded structurally rather than by special-casing, the heal goes through the idempotent projector, all round-4 bookkeeping is deleted with no dangling references, and **no docstring or comment in the diff narrates review history**.

**Cost-claim precision.** "Two HTTP calls per tick" is accurate for the listing step, which is what matters architecturally, but a tick that performs a genuine heal also issues a `PUT` via `task_manager.update`. Not a defect; corrected in the description.

#### Testing

**Defects.** `test_drift_sweep_leaves_a_live_session_alone_regardless_of_status[claimed]` does not pin the guard it claims to. Mutating `session.status not in TERMINAL_STATUSES` to `if False:` leaves `[claimed]` green — `STATUS_CLAIMED` has no projector mapping, so `TaskManager.update` rejects the literal string downstream and `transition` returns `applied=False`. The case is protected by an unrelated validation net, not by the sweep's guard. `[running]` and `[blocked]` redden correctly.

**Missing Tests.** The CAS `expected_version` wiring has no sweep-level coverage: changing it to `None` — removing the guard that prevents clobbering a concurrent operator edit — leaves all 21 tests passing. The existing `test_projection_preserves_operator_edit_on_stale_version` exercises the projector directly, never the sweep's wiring of it.

**Deletion audit.** All five deletions verified legitimate by grepping that the mechanisms they pinned (`drift_swept_at`, `list_terminal_unswept`, `mark_drift_swept`, the window/limit constants, both indexes) are absent from production code — not by trusting the claim, and not based on the removed explanatory comments. Three further tests counted as "deleted" were in fact adapted, preserving their behavioral coverage.

**Status.** Seven mutation trials, each reverted; re-anchored and fully re-run after the worktree moved mid-review; final state clean at `2fb0726e`, 21 passed.

### Referee Decisions

| 1 | Stale PR description documents a deleted mechanism | general | Defect | Valid | **Accept — already fixed** | Whoever merges reading only the description would get a materially wrong picture of what ships. This is PR metadata rather than repository content, so the orchestrator rewrote it directly rather than spending an addressing round: it now describes the inverted sweep, and the cost claim is stated precisely as two listing calls per tick plus one `PUT` per heal. |
| 2 | CAS `expected_version` wiring uncovered | testing | Missing Test | Valid | **Accept** | Mutation-proven, and the original brief named this guard load-bearing for protecting operator edits. Remedy: a stale snapshot whose `updated_at` is behind the live task → assert `healed == 0` and the live edit survives. This also kills the surviving `if applied:` mutant, since the suite currently has no case where the guards pass but `transition` legitimately returns False. |
| 3 | `[claimed]` parametrization is coincidentally green | testing | Missing Test | Valid | **Accept** | Refereed as a missing test, not a defect: the production guard is present and correct; only the test fails to isolate it. Remedy: spy on `lifecycle_projector.transition` and assert it is never invoked for a non-terminal session, so the test pins the intended guard rather than a downstream validation accident. |
| 4 | No test for a tagged task with no session row | general | Missing Test | Valid | **Accept** | Reachable — an operator hand-editing a tag in Markdown is a normal path in this system. The code already handles it; the test is cheap. |

**Findings forwarded to addresser:** 3 (all test-only; #1 handled directly)

**Convergence:** 8 → 6 → 6 → 3, with zero production defects this round. This is the fifth and final round within the convergence bound. The remaining work touches no production logic, so it proceeds to verification rather than a sixth review round.

nbramia:
Round 5 addressed — 2 of 3 test-only findings closed; the third surfaces a real production defect (details below, escalated to the orchestrator rather than fixed in-branch per this round's test-only constraint).

- Finding #2 (live-session guard for `claimed` not pinned): fixed by spying on `lifecycle_projector.transition` in the parametrized test and asserting it's never invoked for a non-terminal session. Mutation-proven: setting the sweep's `session.status not in TERMINAL_STATUSES` guard to `if False:` now reddens all three parametrizations, including `[claimed]`.
- Finding #3 (no test for a tagged task with no backing session row): added `test_drift_sweep_leaves_a_tagged_task_with_no_backing_session_alone`. Mutation-proven: neutralizing the `session is None` guard reddens the test with an `AttributeError`.
- Finding #1 (CAS `expected_version` wiring): while writing the required witness test (stale tag-listing snapshot + an operator edit landing before the heal), discovered the guard as currently wired is non-functional for that exact race — `_reconcile_lifecycle_drift` passes `task=SimpleNamespace(**task)` built from the *same* stale snapshot used for `expected_version`, so `transition()`'s CAS check (`current = task or self.task_manager.get(...)`) never re-fetches and compares the stale value against itself. A live edit landing between the tag listing and the heal is not protected. This is a production-logic gap, not a test gap, so no test was added or committed for it — flagged to the orchestrator for a decision on scope.

nbramia:
## Round 6 — tautological CAS guard, fixed at all three call sites

**Finding #1 (Defect).** All three call sites built a `LifecycleEvent`'s
`expected_version` from a task fetch, then passed that *same* fetch as
`transition()`'s `task=` override. The CAS check compared the snapshot to
itself and could never fire. Fixed by dropping the `task=` override at all
three sites — `_project_session_status` (worker.py), the human-queue wake
replay, and the drift sweep — so `transition()` re-fetches live state via
`task_manager.get()` instead.

**Finding #2 (Defect) — recovery per call site, verified with tests, not assumed:**

- **`_project_session_status` (terminal targets, e.g. completed/failed):**
  recovered by the drift sweep's *next tick*. The sweep's own event_id
  carries suffix `lifecycle_drift_sweep` vs. the hook's `turn_id`, so a
  conflicted hook projection never blocks the sweep from re-attempting under
  a fresh event_id once the session row is terminal. Proven by
  `test_project_session_status_cas_guard_fires_and_terminal_status_self_heals`.
- **The drift sweep itself:** self-recovers on its own next tick — each tick
  re-lists tagged tasks and builds a fresh `expected_version`, and
  `begin_projection` allows retry for any non-`applied` state (including
  `conflicted`). Proven by
  `test_drift_sweep_does_not_clobber_a_concurrent_operator_edit`, which also
  kills the previously-surviving `if applied: healed += 1` mutant (round 5,
  finding #1) — the suite now has a case where the guards pass but
  `transition` legitimately returns `False`.
- **Human-queue wake replay (`_replay_wait_wakeups`, targets STATUS_RUNNING):**
  the drift sweep does **not** cover this path (non-terminal target). Traced
  the actual recovery: on a CAS conflict, the method's existing
  `continue` (guarded by `not projected and not projection_applied(event_id)`)
  already leaves the wait row `resolved` with `wake_consumed=0` — it never
  reaches `complete_wait_wakeup`. `_replay_wait_wakeups` runs unconditionally
  every tick (never gated on the spend cap), and `list_resolved_waits()`
  reselects any unconsumed row, so the next tick re-fetches the task fresh
  and retries the *same* event_id (which `begin_projection` permits for a
  `conflicted` row). This is genuine self-recovery, not a permanent miss —
  proven directly by
  `test_human_queue_wake_cas_guard_fires_and_self_heals_on_next_tick`
  (asserts the rearm does *not* happen on the conflicted tick, then does on
  the next).
- **Non-terminal (BLOCKED) targets at `_project_session_status`:** flagged as
  worth considering separately. `_mark_blocked`'s operator-wait path is safe
  by construction — `record_wait` performs the write first with no
  `expected_version` at all (no CAS), and the hook's later call under the
  same event_id is a no-op dedup, never a real CAS check. The one path that
  does rely solely on the hook's CAS for a BLOCKED target is
  `_handle_cli_interrupted`'s resumable-park case. A conflict there is
  bounded, not permanent: the tag mismatch persists only until that
  session's next status transition (its eventual terminal write goes through
  the same now-real CAS check, and if that also conflicts, the drift sweep
  picks it up once the row is terminal). Documenting this rather than
  building new machinery for it, per the round's "propose the smallest fix
  or say so" instruction — this exposure is not new-and-permanent the way
  the wake-replay path could have been.

**Findings #3–#5 (Missing Test):** all satisfied —
`test_drift_sweep_does_not_clobber_a_concurrent_operator_edit`,
`test_project_session_status_cas_guard_fires_and_terminal_status_self_heals`,
and `test_human_queue_wake_cas_guard_fires_and_self_heals_on_next_tick` are
new; each was mutation-proven by temporarily restoring the `task=`
passthrough at its call site and confirming the test reddens. The two
pre-existing projector tests
(`test_projection_preserves_operator_edit_on_stale_version`,
`test_projection_is_idempotent_and_machine_wait_stays_in_progress`) call
`transition()` directly with no `task=` override, so they were never
inert — both still pass unmodified.

**Tests:** `pytest tests/test_task_lifecycle.py` — 25 passed. Also ran the
full adjacent agent-worker suite (`test_agent_worker_loop.py`,
`test_agent_worker_executor_lifecycle.py`, `test_agent_kill_api.py`,
`test_agent_worker_human_queue.py`, `test_agent_worker_clarifications.py`,
`test_agent_worker_self_restart.py`, `test_agent_worker_inter_agent.py`,
and every other `test_agent_worker_*.py` file) — 578+245 passed, no
regressions from making the guard real.

nbramia:
## Review Round 6 — Consolidated Review & Referee Decisions

Reviewing `git diff ea1428d4..HEAD` — making `LifecycleProjector.transition`'s CAS guard real at three call sites. Scoped to that delta; the drift sweep itself is settled after five prior rounds.

**Reviewers run:** correctness

**Specialist rationale:** this delta changes a hook that fires on every terminal status write, two of the three call sites being pre-existing paths the PR did not otherwise touch. The risk was never the fix — it was that a guard which had never fired would start firing without a retry path, trading a rare clobber for a rare permanent silent miss. That is a state-machine question, so correctness was the right and only specialist.

### Reviewer Findings

**Recovery claims (a), (b), (c) — all independently verified**, by code trace and by mutation-tested witness tests:
- **(a)** `update_status` commits the session row before running the projector as a side effect, so a terminal session with a stale tag is exactly what the sweep picks up next tick, under a different `event_id`.
- **(b)** The sweep self-recovers from a fresh listing; `begin_projection` permits retry of a `conflicted` row for the same `event_id` (`state != 'applied'`, `session_store.py:1218`), confirmed by direct code read.
- **(c)** `list_resolved_waits()` filters on `state='resolved' AND wake_consumed=0`, and `complete_wait_wakeup` is never reached on the conflict branch, so the wait re-surfaces every tick with a fresh `expected_version` until it applies. `enqueue_wait_wakeup` is idempotent, so no duplicate `pending_messages` row.

Mutation-verified by restoring the tautological `task=` override at all three call sites: all three new witness tests reddened, then reverted clean.

**Defects.** Claim (d) — the BLOCKED-target residual — was represented as "documented" but appears nowhere in the code; grep of the diff and of the full current `worker.py`/`lifecycle.py` finds no comment, docstring, or test. Separately, `_handle_cli_interrupted`'s docstring states the vault tag is left at `#agent-running` while the code actually projects `#agent-blocked` via the hook — confirmed empirically against the real interrupted-CLI fixture (`PUT` with `tags: [..., "agent-blocked"]`). And "bounded" overstates the guarantee: it is bounded only if a later status transition occurs; a session parked BLOCKED and never resumed or killed keeps a stale tag indefinitely.

**Minor/informational.** `replay_pending` still rebuilds a retried event from the row's stored `expected_version`. Before this round that was dead code, since no row could reach `conflicted`. Now a genuine conflict can leave a permanently-`conflicted` row that a restart retries forever with a stale version.

### Referee Decisions

| # | Finding | Kind | Concern | Decision | Reasoning / smallest remedy |
|---|---------|------|---------|----------|-----------------------------|
| 1 | BLOCKED-target residual unstated in the code | Defect | Valid | **Accept** | The whole premise of this PR is silently-stranded tags, so shipping a known — if narrow — path to one without a word in the code is the wrong trade. Remedy: one accurate sentence, stating that it self-corrects only if a later transition occurs. No machinery, no TODO. |
| 2 | `_handle_cli_interrupted` docstring contradicts the code | Defect | Valid | **Accept** | Pre-existing, and normally "don't improve adjacent code" would apply. Accepted because it actively misstates the tag-writing behavior this PR is about, and it misled this PR's own root-cause analysis — leaving it means the next reader repeats that. Remedy: correct that one docstring; no logic change, no sweep for other stale prose. |
| 3 | No test for the BLOCKED residual's exposure | Missing Test | Valid | **Reject** | The concern is real and is addressed by #1. A test asserting that a known gap is *not* recovered would cement current behavior as intended and make a future fix look like a regression. A comment is the right artifact for an acknowledged limitation; a test is not. |
| 4 | `replay_pending` retries with a stored stale version | Defect | Valid | **Reject for this PR** | Genuinely valid and correctly flagged as non-gating. It is pre-existing, silent, and causes no functional loss: every in-scope call site has its own verified per-tick recovery. The cost is dead rows accumulating in `lifecycle_projections`. Fixing it means changing replay semantics for every projector consumer — beyond what this PR set out to do, and it would need its own review. Follow-up issue. |

**Findings forwarded to addresser:** 2 (both prose-only; no production logic changes)

nbramia:
## Round 7 — addressed

Both findings were prose-only corrections in `api/services/agent_worker/worker.py`. No production logic changed.

**Finding 1 — BLOCKED-target CAS residual unstated.** Added a paragraph to `_reconcile_lifecycle_drift`'s docstring:

> A BLOCKED target is outside that "non-terminal is legitimately live" scope: if the projector's own CAS write for a BLOCKED transition (`_handle_cli_interrupted`'s resumable-park write, for instance) loses a concurrent-edit race, the session row still lands BLOCKED but the vault tag keeps whatever it held before. Because BLOCKED is non-terminal, this sweep treats that session as legitimately live and never revisits it. The stale tag self-corrects only if a later status transition fires the projector again for that same session — a session parked BLOCKED and never resumed or killed keeps the stale tag indefinitely.

**Finding 2 — `_handle_cli_interrupted` docstring contradicted the code.** Verified directly: `session_store.update_status` fires the status projector for any write landing `STATUS_BLOCKED` (`session_store.py`), which routes through `_project_session_status` → `lifecycle_projector.transition`, and `STATUS_BLOCKED` maps to `BLOCKED_TAG` = `#agent-blocked` (`lifecycle.py`, `agent_board.py`). The docstring's claim that the tag "is left at `#agent-running`" was false for this path. Corrected:

> ...The session row moves to BLOCKED through `session_store.update_status`, which fires the status projector and swaps the vault tag to `#agent-blocked` so `/agents` reflects it.

**Tests:** `pytest tests/test_task_lifecycle.py tests/test_agent_worker_executor_lifecycle.py tests/test_agent_worker_claude_code_dispatch.py tests/test_agent_worker_claude_code_executor.py tests/test_agent_worker_codex_dispatch.py` — 166 passed, unchanged pass count (prose-only edit).

https://claude.ai/code/session_014B1LEZGxQbpEAbq3ZNvwDE

nbramia:
## Round 8 — heal path rerouted to `swap-tag`

Verification was right: every heal was refused with **409**. All five findings addressed.

| # | Finding | Action |
|---|---------|--------|
| 1 | Every heal refused with 409 | Applied — sweep now calls `_swap_tag` + `_complete_task`/`_set_task_status`, the same endpoints `_reconcile_vault_terminal` uses, with the identical status→action mapping. `LifecycleEvent`/`event_id` construction removed from the sweep. |
| 2 | Must swap from the tag the card carries | Applied — the `from` tag is the one whose listing produced the candidate, so a `#agent-blocked` card swaps out of `BLOCKED_TAG`. Pinned by the route-level test. |
| 3 | Preserve concurrency protection via swap-tag semantics | Applied — the swap gates the rest. `swapped: false` short-circuits the card: no count, no status write. Tag swap is ordered first, including for the `completed` target. |
| 4 | Regression test must exercise the real HTTP route | Applied — `test_drift_sweep_heals_through_the_real_task_routes` drives the sweep against the real FastAPI app (`TestClient`, temp vault, temp session DB, `worker._http` bound to the app, projector wired through `_WorkerLifecycleTaskManager` exactly as `Worker.__init__` does). Nothing stubbed between the sweep and the route handlers. |
| 5 | Rework the CAS test, don't delete it | Applied — `test_drift_sweep_does_not_clobber_a_concurrent_operator_edit` now runs route-level: an operator retags the card between the listing and the write, and the sweep leaves it byte-identical (`updated_at` unchanged proves no write landed at all). The other two round-6 `transition()` call sites are untouched. |

### Mutation proof (finding #4)

Restored the pre-fix projector call and ran the whole file:

```
1 failed, 25 passed
FAILED tests/test_task_lifecycle.py::test_drift_sweep_heals_through_the_real_task_routes
assert 0 == 3
WARNING lifecycle projection failed for 2dca95f7: Client error '409 Conflict' for url 'http://testserver/api/tasks/2dca95f7'
WARNING lifecycle projection failed for 17c03223: Client error '409 Conflict' for url 'http://testserver/api/tasks/17c03223'
WARNING lifecycle projection failed for 9e57b2ed: Client error '409 Conflict' for url 'http://testserver/api/tasks/9e57b2ed'
```

The 25 in-process tests pass against the broken code; only the route-level test catches it, with exactly the 409 seen in `logs/agent-worker.log`. Fix restored, all 26 pass.

### Tests

```
tests/test_task_lifecycle.py                                        26 passed
tests/test_agent_worker_loop.py tests/test_agent_kill_api.py
tests/test_agents_board_api.py tests/test_agent_worker_human_queue.py  259 passed
ruff check (worker.py, test_task_lifecycle.py)                      clean
```

Cost per tick is unchanged for the no-drift case: two listing calls, plus two writes per card actually healed.

https://claude.ai/code/session_014B1LEZGxQbpEAbq3ZNvwDE

nbramia:
## Round 8 follow-up — ratchet, and two docstrings corrected

**Narration ratchet.** Both flagged comments reworded (your suggested text). `tests/test_no_change_narration.py` run standalone: **1 passed**. No `ALLOWLIST` or baseline edit. While in there I also split-proofed one of my own new docstring lines (`is no` / `longer there` straddled a line break, so it slipped the line-keyed matcher by luck rather than by design) — now reads `is absent`.

**`_handle_cli_interrupted` docstring reverted.** Round 7's claim that the park "fires the status projector and swaps the vault tag to `#agent-blocked`" is false for the same 409 reason this PR fixes. Restored to what actually happens — *"The vault tag is left at `#agent-running`; only the session row moves to BLOCKED."* — without the CLARIFY/GOAL/PLAN comparison and without referencing the projector's behaviour as a defect.

**Residual paragraph rewritten.** It framed the BLOCKED exposure as losing a CAS race; a BLOCKED projection doesn't land at all, so that was a second inaccurate explanation. It now describes the sweep's own scope: a session parked at BLOCKED is non-terminal, so the sweep only ever skips it, and bringing a park's tag into line belongs to the path performing the park. My fix changes only the sweep's write path, not `_project_session_status`, so that remains true. Added one sentence on why `BLOCKED_TAG` is swept anyway.

### On the two clarifications

Both were already satisfied by the pushed implementation, so no code change was needed:

- **Swap gates, actions mirrored not order** — the code is exactly your snippet: `if not self._swap_tag(...): continue`, then `_complete_task` for completed and `_set_task_status(..., "cancelled")` for failed/budget_exceeded. The swap comes first for *every* target including `completed`, deliberately departing from `_reconcile_vault_terminal`'s complete-then-swap order.
- **The `completed` shape** — covered. The route-level test's three cases are `(running, failed) -> #agent-failed + cancelled`, `(blocked, completed) -> #agent-completed + done`, `(running, budget_exceeded) -> #agent-budget-exceeded + cancelled`. The completion write is asserted, not just the tag.
- **Constructing the BLOCKED candidate** — the test creates the card carrying `BLOCKED_TAG` directly rather than expecting organic state, so it is unaffected by BLOCKED parks not writing that tag today.

### Tests

```
tests/test_no_change_narration.py                                    1 passed
tests/test_task_lifecycle.py + test_agent_worker_loop.py
  + test_agent_kill_api.py                                         120 passed
ruff check (worker.py, test_task_lifecycle.py)                       clean
```

Commit `48089da4`. Note: something ran `git checkout origin/fix/agent-lifecycle-drift-sweep` in the task-dispatch worktree between my two pushes, leaving it on a detached HEAD. The follow-up commit landed on the same SHA the branch pointed at, so reattaching was a verified fast-forward — nothing lost, no force-push.

https://claude.ai/code/session_014B1LEZGxQbpEAbq3ZNvwDE

nbramia:
## End-to-End Verification — PR #1116

### Verdict: PASS

```
verification-record: v1
verification-head: 48089da472f478e98a7d8e3feaf22d35950aedb4
suite-result: pass
suite-command: ~/.venvs/lifeos/bin/python scripts/verify_candidate.py pushed-ref --repository <worktree> --sha 48089da472f478e98a7d8e3feaf22d35950aedb4 --base 234a66b98ae6a95ed8dfc95806484ab59f118a9c --lanes fast-unit --workers 2 --lane-log-dir <lane-log-dir>
suite-executions: 1
suite-exit-status: 0
suite-command-results:
  - command: (as above)
    result: pass
    exit-status: 0
retry-reason: none
```

`8667 passed, 1 skipped, 996 warnings in 2527.25s (0:42:07)`; verifier summary
`{"lanes": ["fast-unit"], "lane_totals": {"fast-unit": 8668}, "reason": "missing", "result": "success", "reused": false}`.
Lane selection follows `.github/workflows/candidate-verification.yml`: no `web/`
file changed, so the server-free browser lane cannot be affected and `fast-unit`
is the lane. Head re-checked after execution and unchanged.

**An earlier lane failure on this branch belongs to the superseded SHA
`60de4022`, not to this candidate.** That run failed the change-narration
ratchet; it passes here, inside the full suite.

### System Flow Verified

`Worker.tick()` → `_reconcile_lifecycle_drift()` → `GET /api/tasks?tag=…` →
`session_store.get()` → `POST /api/tasks/{id}/swap-tag` → `/complete` or a
status write → Markdown on disk → the task API views `/agents` reads.
Exercised through the real FastAPI routes against a real vault, on an isolated
`git archive` export of the exact head.

### Evidence

**End-to-end scenarios: 10/10 pass** (23 checks failed on `60de4022`).
Healing works for all three terminal statuses, including the two that were the
open questions:

- `#agent-blocked` candidate: `healed=1`, tag → `agent-failed`, status →
  `cancelled`, no longer listed under the blocked tag. The swap sends
  `from=agent-blocked`, i.e. the tag the candidate was listed under rather than
  a hardcoded `RUNNING_TAG` — which matters because `swap_tag` returns `False`
  on a missing from-tag instead of raising, so the wrong `from` would no-op
  silently.
- `completed` candidate: `healed=1`, tag → `agent-completed`, **status → `done`**
  — the completed branch writes `done`, not the cancellation status.

Regressions all hold: a live session at `claimed`/`running`/`blocked` is left
byte-identical with `updated_at` unchanged; a task reopened for a follow-up turn
is not swept back; an operator-settled `#accepted` card is untouched;
operator-origin and spawned-child sessions are excluded; a tag-listing outage
heals nothing, writes nothing, and recovers on the next tick.

**Swap gate — control green AND races green.** Recorded outbound calls:

```
control:        POST /swap-tag from=agent-running to=agent-failed   +  PUT {'status': 'cancelled'}
race (running): POST /swap-tag from=agent-running to=agent-failed   +  (no status write)
race (blocked): POST /swap-tag from=agent-blocked to=agent-failed   +  (no status write)
```

In each race an operator retags the card between the sweep's listing and its
write: `healed=0`, the operator's tag choice survives, and no status write
follows a swap that did not apply.

**Both halves are load-bearing, and this is the part a future reader should not
skip: a race passing while the control is red means nothing at all.** On the
pre-fix head all three race checks already passed — vacuously, because nothing
healed at all, which makes "no status write follows a failed swap" trivially
true. The control (`an applied swap still writes the status`) was red there and
is green here. Only the conjunction is evidence.

**Mutation testing of the PR's own tests — all five killed, none survived:**

| Mutation | Result |
|---|---|
| M1 ignore `swapped: false`, count the heal and write status | KILLED (11 tests fail) |
| M2 hardcode the swap's `from` tag to `RUNNING_TAG` | KILLED (1 test) |
| M3 drop the terminal-status check | KILLED (3 tests) |
| M4 completed branch uses the cancellation write | KILLED (1 test) |
| M5 restore the projector call (the original defect) | KILLED (1 test) |

M1 answers whether the concurrent-edit test was weakened to match the new path:
it was not — the guarantee is still pinned.

**M5 is the empirical basis for this rework.** Restoring the projector call
yields `1 failed, 25 passed`: only
`test_drift_sweep_heals_through_the_real_task_routes` reddens, while 25
in-process tests pass against defective code. The in-process tests wire the
projector to a `TaskManager` directly and bypass the HTTP route, so they cannot
observe that `PUT /api/tasks/{id}` refuses a lifecycle-tag write on a claimed
card — `agent_board.is_claimed()` returns `True` on `#agent-running`/
`#agent-blocked` before `has_live_session` is consulted, and every card this
sweep targets carries one by construction.

### Issues Found

None blocking.

One observation recorded rather than acted on: that single route-level test is
the sole detector for M2, M4 **and** M5 — three distinct behaviors, one test.
The behaviors are covered, so this is not a defect. But if that test is ever
weakened, skipped, or has its HTTP wiring replaced with an in-process manager,
all three regress silently while the in-process suite stays green. That is the
configuration this PR's own history demonstrates is hard to notice.

### Holistic Assessment

The system works end-to-end after this change: drift heals through endpoints the
API actually accepts, for all three terminal statuses and both non-terminal tags,
while every state the sweep must not touch is left byte-identical. The heal is
gated on the swap actually applying, so a concurrent operator retag cannot be
followed by a forced status change. The pre-existing projector `PUT` path remains
broken and is tracked separately in #1120; this PR no longer depends on it.

PR #1135, opened 2026-09-17
fix: require an explicit filing request before Pebble files a task

nbramia:
## TL;DR
Pebble now presumes log-only: a capture becomes a task only when the speaker actively asks for one, and in a multi-item capture only the thing actually asked for. That judgment lives in the classifier prompt and the model, not in application code — the previous regex gate is deleted. Application code still proves, from an exact unquoted transcript span, that an executor-tagged task, an agent schedule, or a human-queue card was genuinely authorized.

## Design
Deciding whether a fragment of speech is a request is a semantic question, so the model answers it under prompt guidance (`api/services/journal_filing_policy.py`): log-only default, file only what was actually asked for. Deciding whether a span of text literally appears in the transcript is a mechanical question, so code answers that — and it is the only gate kept for plain tasks' neighbours: delegation, agent schedules, and human-queue cards, which cause real execution rather than a to-do appearing on a list.

An action the code refuses is dropped, never raised: a raise out of `validate_plan` leaves the capture permanently pending and retried, so a refused action must still let the capture log and complete.

## Implementation Notes
- `api/services/pebble_capture.py` (−175 net): removes all plain-task filing logic. Keeps the execution-authority gates (`_explicit_task_delegation`, `_explicit_scheduled_delegation`, `_explicit_operator_decision`, `_explicit_tags`, `_scheduled_executors`, `_is_unquoted_evidence`, `_mask_quoted`, `_positive_clauses`, `_explicit_action_evidence`). Keeps three fixes that stand independently of the approach: index-based pairing in `_validated_classifier_actions` (position pairing produced a spurious "unauthorized task delegation" once an action is dropped), drop-not-raise, and a two-set evidence split so a plain task's evidence can no longer abort an independent agent schedule.
- **Provider routing (privacy boundary).** `LocalOnlyJournalClassifier` → `PebbleJournalClassifier`, which prefers the configured remote provider (`settings.remote_llm_configured`) and falls back to the loopback-validated local llama-server. On a keyless install nothing changes. Where a remote provider is configured, capture transcripts are sent to it, so `AGENTS.md` § "Privacy Is Non-Negotiable" is corrected: inference is local by default and on a keyless install, and going off-box is an explicit operator choice, never a silent default. `docs/guides/pebble-capture.md` documents the routing.
- `scripts/eval_pebble_filing.py` (new): standalone scorer for the filing judgment, deliberately **not** a CI gate — it makes real provider calls. Carries a table section and a held-out section whose cases appear nowhere in the prompt, scored separately, because a case promoted into the prompt's examples stops falsifying.
- `tests/test_pebble_capture.py`: drops the tests that existed only to pin the deleted gate; keeps the consumer-level lifecycle, delegation, agent-schedule, index-pairing and evidence-split coverage.

## Test evidence
Change touches a named risk trigger (authentication/privacy boundaries) per [development-lifecycle.md](docs/specs/standards/development-lifecycle.md) § Risk and review, so boundary evidence is recorded here.

- **Hosted gate, head `e368ce96`:** GH Actions run [35266776631](https://github.com/nbramia/LifeOS/actions/runs/35266776631) — `fast-unit` across 4 parts, 8785/8785 node IDs, all `success`. `browser-free` correctly not selected (no `web/` file changed).
- **Executable end-to-end scenario** against the real `PebbleCaptureConsumer` / `CaptureLedger` / `TaskManager` / `SchedulerStore` in an isolated temp location, classifier stubbed: explicit ask files exactly the expected task and the ledger reaches `complete`; a log-only capture files zero tasks and still reaches `complete` (not stuck pending); a multi-item capture files only the requested item; an unprovable executor tag is dropped while the plain task still files; an unprovable agent schedule creates nothing; a corrected redelivery completes cleanly.
- **Filing judgment vs. the configured provider** (`scripts/eval_pebble_filing.py`): 13/13 on the table and 12/12 on the held-out set — 25/25.
- **Focused suite:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py tests/test_no_change_narration.py -q` → 208 passed.

## Review focus
- The kept execution-authority gates: they are what stops an eager classifier from queueing agent work now that plain-task filing is ungated.
- The provider routing and its privacy consequence, including that a keyless clone stays local.
- The held-out section of the eval: it must stay disjoint from `JOURNAL_BEHAVIOR_CASES` or it stops measuring anything.

No linked issue — operator-directed.

https://claude.ai/code/session_01F98UyxdwoJRuUhb2iGDpLy


nbramia:

**Reviewers run:** general, security, testing

**Specialist rationale:** `review-security` — this file is an authority boundary that turns untrusted local-LLM output derived from voice transcripts into real side effects (tasks, scheduler entries that can execute agents, human-queue cards), and this PR adds a new gate on that boundary. `review-testing` — the diff rewrote the premises of four existing tests and changed 10 of 20 rows in a parametrize table, so test integrity was itself a concrete risk under AGENTS.md § "Tests Are Sacred".

### Reviewer Findings

#### General

**Defects**

- `_TASK_REQUEST_MARKER_RE`'s `have`/`let` alternatives (`api/services/pebble_capture.py:61-62`) match ordinary, non-task sentences, not just delegation-style phrasing, which lets a bare imperative through the new authority gate the PR is meant to install. It borrows the shape from `_explicit_tags`'s `have`/`let` assignment patterns (`pebble_capture.py:554-555`) but drops both constraints that kept those originals narrow: the `^\s*` clause-start anchor and the required fourth trailing word (`(?:this|it|that|the\b|[\w-]+)` after the verb). The result is `\bhave\s+#?[\w-]+\s+(?:to\s+)?[\w-]+\b` / `\blet\s+#?[\w-]+\s+[\w-]+\b`, which match any "have/let + word + word" shape anywhere in a clause, with no executor-validity check (that check only applies later, for delegation, not for the plain-task marker gate). Concretely reproduced against `validate_plan` on this checkout:
  - `validate_plan([{"kind":"task","index":0,"title":"x","action_evidence":"Let the dog out"}], transcript="Let the dog out.", ...)` returns a filed `PlannedAction` — i.e. "Let the dog out.", structurally identical to the operator's own explicit log-only example "Take the dog outside.", passes the gate and files a task.
  - The same reproduction with transcript `"I have a lot to do today."` and matching `action_evidence` also files a task.

  Since the premise of this whole PR is that the local classifier is over-eager and will readily propose `action_evidence` equal to the full clause, this is a live, reachable path through the code the PR added — not a hypothetical. It directly undermines the governing contract ("must be very close to explicit... strong presumption should be log").

**Missing Tests**

- No test exercises the `have`/`let` marker alternatives against a non-executor, non-delegation sentence. `test_plain_task_requires_an_explicit_filing_request` (`tests/test_pebble_capture.py:479-497`) only covers the exact calibration phrases, and every existing `have`/`let` case in `test_task_execution_tags_require_positive_valid_delegation` (`tests/test_pebble_capture.py:428-431`) uses a real executor name as the target, never a generic noun. Nothing pins that a bare imperative shaped like "Let the dog out." or "We have a meeting to plan." stays log-only.

**Status** — Ran `pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py -q` → 208 passed, matching the PR's claim. Independently reproduced the marker over-match against `validate_plan()` via a scratch script. Did not run any broad/authoritative suite.

**Summary** — The core design (drop-not-raise, index-based pairing fixing the `zip` misalignment, fixture/test rewrites, doc updates, and the Journal chat persona's unchanged `allow_clarification=True` path) is sound and verified. However, the calibration surface itself is measurably broader than the "very close to explicit" requirement and is untested against exactly the bare-imperative phrasing called out as the baseline log-only case. This is a real gap in the PR's central deliverable.

#### Security

**Defects**

- `_explicit_task_request()` (`api/services/pebble_capture.py:610-627`), wired into `validate_plan`'s task branch (`:768`), binds a task's `action_evidence` to the same sentence-level `_positive_clauses()` clause as an explicit filing marker — but clause splitting is only on `.!?;\n` (`_positive_clauses`, `:506-535`). For any unpunctuated/run-on transcript — the norm for Whisper output without punctuation restoration — one explicit marker anywhere in the run-on clause authorizes **any** unrelated substring elsewhere in that same clause as a distinct filed task, with no per-task dedup (the `used_delegations`/`used_action_evidence` reuse guards at `:714-753` only apply when a routing tag is claimed; a plain untagged task skips them entirely). Demonstrated end-to-end through `validate_plan`: a run-on transcript beginning "remind me to call mom and then also we should probably repaint the whole garage this weekend and honestly I have been meaning to redo the budget spreadsheet since forever and finally book the dentist appointment", submitted with four task actions whose `action_evidence` spans are "call mom", "repaint the whole garage this weekend", "redo the budget spreadsheet", "book the dentist appointment" — **all four are filed**. Only "call mom" trails the "remind me to" marker; the other three are bystander clauses with no filing request of their own. This defeats the PR's own stated design invariant that "a marker and its action must share one clause" (docstring `:611-618`) — the code's notion of "clause" is far coarser than that claim assumes. Reachable by the local classifier alone on realistic run-on voice captures. Suggested fix: require the marker to be adjacent to (not merely co-occurring anywhere within) the `action_evidence` span — bound the marker search to a window around the evidence span, or require the marker phrase to directly govern the evidence clause, rather than clause-wide co-occurrence.

**Missing Tests**

- No test exercises an unpunctuated/run-on transcript containing one explicit filing marker plus multiple candidate `action_evidence` spans, only some of which trail the marker. The existing multi-action delegation-scoping test (`tests/test_pebble_capture.py:708`) relies on a period to separate the marked and unmarked clauses ("Buy milk. Assign code repair to Codex."), so it does not pin the "one marker, many unrelated substrings in the same clause" boundary.

**Status** — Ran `pytest tests/test_pebble_capture.py -q` → 141 passed. Verified via throwaway probes that negation, conditional, and reported/quoted-speech evasion of the new gate all correctly return `False` (`_positive_clauses` filtering and `_mask_quoted` are applied on the new path, not bypassed). Verified the `_validated_classifier_actions` index-pairing refactor is sound (indexes are globally unique per `validate_plan` call via `seen_indexes`). Confirmed no logging of transcript/`action_evidence` content anywhere in `pebble_capture.py`, and all fixture diffs use only synthetic placeholder text.

**Summary** — The new gate correctly blocks negated, conditional, and quoted/reported-speech injection of a filing request, and the delegated-task/agent-schedule authority gates and the index-pairing refactor are not weakened. However, the gate's clause-level binding is too coarse for real unpunctuated voice transcripts.

#### Testing

**Defects** — None.

**Missing Tests**

- No test exercises the "structural contract miss" repair-round path in `_validated_classifier_actions` (`api/services/pebble_capture.py:940-941`, `raise PebbleCaptureError("classifier proposed a task without action_evidence")`). This is one half of trap 6 and the PR claims it "still triggers the classifier's existing one-shot local repair round." **Mutation-tested: deleted the `if`/`raise` entirely, leaving a bare `continue`, and reran the full focused suite (160 tests) — all 160 still passed.** No test proves a raw task action with missing/empty `action_evidence` causes a raise (vs. silently continuing), and none proves that raise reaches `LocalOnlyJournalClassifier.classify()` and triggers the one-shot repair round with the repaired candidate applied. The three existing repair-round tests are pre-existing and exercise different failure shapes.

- The `ask/have/let <name> ...` and `<name>, please ...` alternatives inside `_TASK_REQUEST_MARKER_RE` (`:60-64`) are only tested with a *valid* executor name. Confirmed empirically that `"Ask Dan to walk the synthetic dog."`, `"Have Dan walk the synthetic dog."`, `"Let Dan walk the synthetic dog."`, and `"Dan, please walk the synthetic dog."` all satisfy `_explicit_task_request` and cause `validate_plan` to file an unassigned task, despite containing no task-specific keyword. This differs from the PR's designed-and-commented exception for `assign/delegate/route ... to <invalid-name>`; no equivalent design rationale or test exists for `ask/have/let/"X, please"` targeting a non-executor name.

**Status** — Ran `pytest tests/test_pebble_capture.py tests/test_journal_persona.py -q` → 160 passed. Performed four targeted mutation experiments against production code (each reverted, tree confirmed clean): (1) `_explicit_task_request` forced to always return `True` → 26 tests failed, confirming the calibration-table, log-only-completion, and delegation-rewrite tests all falsify correctly; (2) index-based pairing reverted to `zip()` → `test_validated_classifier_actions_pairs_by_index_not_position` failed exactly as its docstring predicts; (3) the missing-`action_evidence` raise removed → all 160 still passed (finding above); (4) reported-speech clause exclusion in `_positive_clauses` disabled → `test_contextual_reported_speech_cannot_delegate_a_task` failed correctly, showing that rewritten test still pins a real invariant.

**Summary** — The rewritten tests and new calibration-table tests are well-targeted and empirically falsify against the regressions they guard, and the fixture retexting / `_TASK_EVIDENCE` reuse across the ~20 ledger/idempotency tests does not make any of them vacuous. The real gaps are the untested repair-round path and the untested non-executor `ask/have/let` path.

### Referee Decisions

Independently verified each finding against the branch with my own probes rather than relying on the reviewers' reproductions.

| 1 | `_TASK_REQUEST_MARKER_RE` matches ordinary prose — `have …`, `let …`, `<word>, <word> <word>`, and `ask <non-executor> to` | general + testing (+ orchestrator) | Defect | **Valid — confirmed** | **Accept** | Root cause: these alternatives were lifted from `_explicit_tags`/`_scheduled_executors`, where the captured group is subsequently constrained to a known executor (`if tag in _VALID_TASK_ASSIGNEES`) and `have`/`let` are `^`-anchored. As bare markers those constraints are gone. My probe against the branch: `'The blue mug is on the table, right next to the lamp.'` → `'table, right next'`; `"I have a doctor's appointment tomorrow…"` → `'have a doctor'`; `'We talked about the roof, and the gutters need clearing.'` → `'roof, and the'`; `'I have to pick up groceries…'` → `'have to pick'` — **all four file a task through the real `validate_plan`**. Any clause with a comma followed by two non-pronoun words passes, which is most spoken sentences. Smallest remedy: delete the unconstrained `have`/`let`/`<x>, please`/`ask <x> to` alternatives from the regex and satisfy delegation-shaped requests by calling the existing executor-constrained helpers `_explicit_tags(clause)` / `_scheduled_executors(clause)` instead; keep the plain filing literals and `assign\|delegate\|route … to`. I verified this exact combination scores 17/18 on the calibration table (details in the findings file). |
| 2 | One marker in an unpunctuated run-on clause authorizes unlimited unrelated tasks | security (+ orchestrator) | Defect | **Valid — confirmed** | **Accept** | Independently reproduced: a single `"remind me to"` in one run-on sentence filed **4/4** proposed tasks, three of which had no filing request of their own. Directly reachable for Whisper voice output, which frequently lacks sentence punctuation — and it is precisely the "rambling note spawns a pile of tasks" behavior this PR exists to stop. Smallest remedy: require the marker to *govern* the evidence span (evidence begins at or after the marker and within a bounded window / same conjunct) rather than merely co-occurring in the same `.!?;\n` clause. |
| 3 | `"Let me know how …"` counts as a filing request because `me` is a valid assignee | orchestrator | Defect | **Valid — confirmed** | **Accept** | `_explicit_tags`'s `^let #?(\w+) \w+ …` captures `me`, which is in `_VALID_TASK_ASSIGNEES`. Pre-existing on `main`, so normally out of scope — accepted only because finding 1's remedy routes the marker test through `_explicit_tags`, which makes it newly load-bearing. Smallest remedy: exclude the non-delegating `let me <verb>` shape from counting as a filing request. Do not otherwise refactor `_explicit_tags`. |
| 4 | Repair-round raise for absent `action_evidence` is deletable with zero test failures | testing | Missing Test | **Valid — confirmed by mutation** | **Accept** | Behavior the PR explicitly claims to deliver (trap 6) with no coverage; reviewer proved it by deleting the line and seeing 160/160 still pass. Add a test that a task proposed with missing/empty `action_evidence` raises, and that the raise drives the one-shot repair round in `LocalOnlyJournalClassifier.classify` with the repaired candidate applied. |

**Rejected:** none — every finding cleared the two-kind bar and reproduced.

**Findings forwarded to addresser:** 4

nbramia:
## Round 1 review findings addressed

| 1 | `_TASK_REQUEST_MARKER_RE` matches ordinary prose | Applied | Removed the four unconstrained delegation-shaped alternatives (bare `have`/`let`/`ask`/"X, please Y") from the marker regex (now `_PLAIN_TASK_FILING_RE`, plain-filing literals + `assign/delegate/route ... to` only). Delegation-shaped requests are proven instead by calling the existing executor-constrained helpers (`_explicit_tags`, `_scheduled_executors`), which already gate on `_VALID_TASK_ASSIGNEES`/`_VALID_EXECUTORS`. |
| 2 | One marker authorizes unlimited tasks in a run-on clause | Applied | `_explicit_task_request` now requires the marker to *govern* the evidence span: the marker must resolve within a ±30-character window around each occurrence of the evidence text in the clause, not merely appear anywhere in it. Also extended the existing `used_action_evidence` no-reuse guard (previously delegation-only) to plain untagged tasks, so one evidence span cannot back two filed actions. |
| 3 | `"Let me know how…"` counts as a filing request | Applied | Added a `(?!me\b)` exclusion to the `let` delegation pattern in both `_explicit_tags` and `_scheduled_executors`. `"me"` remains a valid assignee for `"assign ... to me"`; only the idiomatic `"let me <verb>"` shape is excluded. |
| 4 | Repair-round raise for absent `action_evidence` has zero coverage | Applied | Added `test_validated_classifier_actions_raises_for_task_missing_action_evidence` (direct raise) and `test_classifier_repairs_task_missing_action_evidence_before_apply` (drives the one-shot repair round through `LocalOnlyJournalClassifier.classify()` to a filed task). |

**Calibration table (finding 1):** 21/21 correct verdicts (13 log-only + 8 task rows, including every reproduction and reviewer-reported example from the findings, plus the required non-regression case `"Assign the task related to … to me."`). Exceeds the 17/18 baseline.

**Run-on reproduction (finding 2):** `proposed=4 filed=1` — only the action adjacent to `"remind me to"` files; the three bystander spans elsewhere in the unpunctuated clause stay log-only.

**Mutation-falsification:** each new/modified test was verified to fail against a reversion of the production line it guards (regex alternatives restored, the "let me" exclusion removed, the governance window widened to 999, the reuse guard removed, the `action_evidence`-missing raise replaced with a bare `continue`), then the mutation was reverted and the working tree confirmed clean before committing.

**Tests:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py tests/test_no_change_narration.py -q` — 225 passed. `ruff check` on changed files — all checks passed.

One pre-existing test (`test_task_delegation_is_scoped_to_the_named_action_not_the_whole_capture`) and one test that shared identical evidence text across two unrelated actions (`test_multi_action_partial_completion_resumes_only_unfinished_action`) were updated: both previously relied on two different actions sharing the identical `action_evidence` text, which the finding 2 reuse-guard now disqualifies for the second action (or, for the latter, was simply an incidental convenience — updated to give each independent task its own distinct evidence span).

https://claude.ai/code/session_01F98UyxdwoJRuUhb2iGDpLy

nbramia:

**Reviewers run:** general, security, testing

**Specialist rationale:** `review-security` retained — the delta rewrote the authority gate itself (governance window + reuse guard), which is where a fix most easily opens a different hole. `review-testing` retained — the delta modified two more pre-existing tests and added the tests for round 1's proven-unpinned repair-round raise, so the addresser's own mutation claims needed independent verification.

Reviewed against `git diff 8d825226..f4df0d1b`. All four round-1 findings were confirmed fixed by all three reviewers.

### Reviewer Findings

#### General

**Defects**

- Change 4's `used_action_evidence` extension to plain (untagged) tasks lets a low-bar plain-task filing abort an entire otherwise-valid capture — including an independently-legitimate agent schedule — instead of degrading safely. `validate_plan()` now adds *any* filed task's `action_evidence` to the shared set (`:811`), not just delegated ones. That same set is read by the `schedule`/`action == "agent"` authority gate (`:853`), which **raises** `PebbleCaptureError("agent schedule lacks explicit valid delegation")` (`:857`) — it aborts the whole `validate_plan()` call rather than dropping. Reproduced: transcript `"Have #claude call the plumber tomorrow."` with one plain untagged task (index 0, `tags: []`, evidence `"call the plumber tomorrow"`) and one agent schedule (index 1, executor `claude`, same evidence) — the plain task passes `_explicit_task_request` only because its ±30-char window contains `"Have #claude"`, so `_scheduled_executors(window)` is non-empty; filing it consumes the span, and the schedule then raises. The *same* schedule files cleanly on its own or ordered second.

  `PebbleCaptureConsumer.process()` has no try/except around `validate_plan(...)`, so this propagates to `PebbleCaptureWatcher.process_file()`, which logs `"Pebble capture remains pending"` (`pebble_capture_watcher.py:130-132`). The capture is retried on later scans and, since the classifier runs at `temperature=0`, will reproduce the same output and stay stuck — silently losing a fully-authorized agent-schedule delegation instead of following the file's own invariant that under-filing is the safe direction and "the capture must still log and complete" (`:806-810`).

  Smallest fix: stop a plain, non-delegated task's evidence consumption from blocking an independent schedule authority check — e.g. compare only against evidence consumed by delegated tasks and other schedules (the pre-round-2 scope), or drop the later action instead of raising.

**Missing Tests** — a regression test that a plain untagged task and an independently-valid agent schedule sharing one `action_evidence` span do not make `validate_plan()` raise when the task is ordered first.

**Status** — 225 passed on the full focused suite; none exercise the cross-kind collision. Verified via scratch probe that the raise is order-dependent.

**Summary** — The delta correctly delivers all four round-1 findings and is well-scoped. Both modified pre-existing tests are legitimate updates to the now-correct one-span-one-action semantics, not weakenings. `(?!me\b)` in `_scheduled_executors` is a no-op for schedule behavior (`"me"` was never in `_VALID_EXECUTORS`), so nothing round 1 protected was tightened. `docs/guides/journal-ring-ingest.md` is **not** stale — it documents a separate pipeline (`/api/journal/ingest` → the Journal persona's own chat-driven extraction) that never calls `validate_plan()`. The one real problem is the Defect above.

#### Security

**Defects**

- The governance window does not fix round 1's run-on defect — it narrows its reach to ~30 characters, which ordinary "and"-chained speech easily spans. `_explicit_task_request()` (`:622-658`) treats *any* match of `_PLAIN_TASK_FILING_RE`, `_explicit_tags()`, or `_scheduled_executors()` inside the ±30-char window as proof the marker "governs" this span — but none of those verifies that the matched marker's own grammatical object *is* the evidence span. `_explicit_tags`/`_scheduled_executors` were designed to answer "who is this delegated to" and are here repurposed as a generic "is there some marker nearby" signal, decoupled from what the marker refers to. Two reproductions against the real `validate_plan()`:
  - `"assign it to Codex then call the pharmacy for a refill"` with a single **untagged** task, evidence `"call the pharmacy for a refill"` — **files**, purely because `_explicit_tags(window)` matches the unrelated `"assign it to Codex"` (whose object is `"it"`).
  - `"add a task to buy milk and feed the cat before dinner and water the garden plants out back"` with three proposed tasks — **proposed=3, filed=2**. `"feed the cat before dinner"` was never requested; it rides the `"add a task to"` marker that governs only `"buy milk"`. The third escapes only by falling outside 30 chars.

  This is round 1's accepted finding 2 reopened at a shorter but still realistic radius.

**Missing Tests** — no test exercises two or more *distinct* spans in one clause where only one has its own marker and the other falls within that marker's window. `test_marker_governs_only_the_nearby_span_in_an_unpunctuated_run_on` (`tests/test_pebble_capture.py:758-783`) uses long spans that push every bystander outside 30 chars, so it cannot catch a short bystander just inside the window.

**Status** — 157 passed. Verified negation/conditional/quoted/reported-speech still correctly rejected on the windowed path; the agent-schedule branch still uses the unwindowed `_explicit_scheduled_delegation`; casefold offset drift (`'ß'`) is real but ~1 char per expansion and not reachable at meaningful density in an English transcript — not reported as a defect.

**Summary** — The round-2 fix genuinely closes the marker-regex and `let me` holes and the delegated/agent-schedule paths remain sound, but the headline governance fix bounds *distance*, not *specificity*. Security posture on this gate is not yet sound.

#### Testing

**Defects** — None.

**Missing Tests**

- Same failure class, independently reproduced: when two or more short, syntactically-unrelated bystander spans sit close enough that both fall within ±30 chars of the *same* marker occurrence, both are filed.
  ```
  transcript = "remind me to check the mail, buy milk, walk the dog"
  spans = ["check the mail", "buy milk", "walk the dog"]  ->  filed: 2 ("check the mail", "buy milk")
  ```
  `"walk the dog"` escapes only by falling just outside the fixed radius from string length, not by any semantic distinction. Confirmed not an artifact: `"remind me to text bob and email jane about the trip"` files both. The `_explicit_task_request` docstring's claim that a marker must "govern this specific evidence span" is not fully true when spans are adjacent.

**Status** — 176 passed. **Five mutation experiments independently verified every one of the addresser's round-2 claims**, each reverted with a clean tree confirmed: (1) missing-`action_evidence` raise → bare `continue`: both new tests failed, and the classifier-level one failing with zero tasks created confirms it genuinely drives the two-call repair round rather than passing vacuously; (2) restoring the deleted bare `have` alternative → 4 tests failed; (3) removing `(?!me\b)` → the `let me know` calibration row failed; (4) widening the window to 999 → the run-on governance test failed; (5) removing the plain-task reuse guard → the delegation-scoping test failed. Confirmed "capture still completes with zero task actions" is pinned at the **consumer** level (`consumer.process(payload) == "complete"` with `tasks.list_tasks() == []`). Confirmed both modified pre-existing tests were forced by the reuse-guard change, not weakened: the old `actions[1].tags == ()` assertion is no longer reachable at all, and the title-bound-to-evidence property it also pinned is still asserted.

**Summary** — All four round-1 findings are genuinely fixed and each is now falsifiable by mutation, including the previously-unpinned repair-round raise. Both modified pre-existing tests are legitimate. The one real gap is that the governance fix bounds distance, not specificity.

### Referee Decisions

Each finding independently reproduced by the orchestrator against `f4df0d1b` before acceptance.

| 5 | Plain-task evidence consumption collides with the agent-schedule gate, which **raises** → capture permanently pending | general | Defect | **Valid — confirmed** | **Accept** | Reproduced exactly, and it is order-dependent: `task first, then schedule` → `RAISED agent schedule lacks explicit valid delegation`; `schedule alone` → OK; `schedule first, then task` → OK. This reintroduces round 1's trap 1 (a raise leaves the capture pending forever and retries identically at `temperature=0`) through a side door, and silently loses an authorized agent schedule. Smallest remedy: a plain, non-delegated task's consumption must not feed the set the schedule gate reads — keep the pre-round-2 cross-kind scope while retaining the intra-task one-span-one-action guard. Do not convert the schedule branch's raise into a drop. |
| 6 | Governance window bounds distance, not specificity — bystander spans within ±30 chars still file | security + testing (deduped) | Defect | **Valid — confirmed** | **Accept** | Two reviewers reproduced it independently with different transcripts; I reproduced both. `"add a task to buy milk and feed the cat…"` → proposed=3 **filed=2**; `"assign it to Codex then call the pharmacy for a refill"` → files a task from an unrelated segment. This is round 1 finding 2 at shorter range, and it is the operator's core complaint (one rambling note, several tasks). Smallest remedy — **proven by the orchestrator at 31/31 on the combined calibration + every reproduction from both rounds**: replace the character-distance window with a grammatical boundary. Segment each positive clause on coordinating conjunctions and commas (`and`, `and then`, `then`, `also`, `plus`, `as well as`, `,`), merge back any segment that begins with an infinitive complement (`to …`, `so that …`, `in order to …`) since that continues the preceding request, and require the evidence occurrence to **start** inside the marker's own segment. The merge rule is load-bearing: without it the real capture `"Add a task assigned to #claude, to review the synthetic report."` regresses to log-only. The start-inside rule is what keeps `"Add a task to buy milk and eggs."` filing as one task. |

**Rejected:** none.

**Convergence:** round 1 forwarded 4 accepted findings; round 2 forwards 2. Strictly fewer, no review-introduced architecture, both findings trace to the original acceptance criteria — converging.

**Findings forwarded to addresser:** 2

nbramia:
## Round 2 findings addressed

**Finding 5 — plain-task evidence collision aborting an independent agent schedule.** Split the single `used_action_evidence` set: a new `used_scheduled_delegation_evidence` set is written only by a *delegated* task (one that actually carries a proven assignee tag) or a filed schedule, and it's the only set the agent-schedule authority gate reads. A plain, untagged task still writes `used_action_evidence` (so it still blocks a later task from reusing its span, and a schedule filed first still blocks a later task — unchanged), but it no longer feeds the set the schedule branch's raise depends on. The schedule branch's raise itself is unchanged, per round 1.

Verified all three orderings from the repro (`Have #claude call the plumber tomorrow.` — plain task + independent agent schedule sharing one `action_evidence`):
- task-then-schedule: no raise; schedule files (`[('task', 0), ('schedule', 1)]`)
- schedule-alone: files (`[('schedule', 1)]`)
- schedule-then-task: schedule files, task dropped (`[('schedule', 0)]`)

**Finding 6 — governance window bounded distance, not specificity.** Replaced the ±30-char window in `_explicit_task_request` with a segmentation rule: split each positive clause on coordinating conjunctions/commas (`_task_request_segments`), merging an infinitive-complement segment (`to ...`, `so that ...`, `in order to ...`) back into the segment before it since it continues the same request. `action_evidence` must *start* inside a segment that itself carries a marker. `_TASK_REQUEST_GOVERNANCE_WINDOW` is removed — superseded, not kept alongside the new rule.

One gap found beyond the reference implementation's 31-case table: pure comma-based segmentation broke the pre-existing "`Codex, please handle this code repair.`" delegation shape (name-comma-please spans a comma by design in `_explicit_tags`). Fixed with a narrow lookahead — a comma directly followed by `please` is not a segment boundary — verified against the full 31-case table (still 31/31) and the full suite (all pre-existing tests still pass).

## Verification

- Reference 31-case table (adapted to call the real `_explicit_task_request`): **31/31**.
- `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py tests/test_no_change_narration.py -q` → **233 passed**.
- `ruff check` on both changed files → clean.
- Mutation-falsification (each mutated, confirmed the guarding test(s) fail, then reverted, confirmed `git status` clean):
  - Schedule branch reverted to read `used_action_evidence` instead of the new set → `test_plain_task_evidence_consumption_never_aborts_an_independent_agent_schedule` fails with the exact round-2 raise.
  - Infinitive-merge rule disabled → both `test_operator_delegated_task_capture_keeps_working` (pre-existing) and the new `test_comma_delegation_still_files_with_its_tag_non_regression` fail.
  - Segment rule reverted to the plain ±30-char window → all three new bystander-span tests fail (`test_marker_governs_only_its_own_short_bystander_span_within_the_window`, `test_comma_chained_spoken_list_files_only_the_marker_governed_span`, `test_unrelated_delegation_marker_does_not_govern_a_later_bystander_span`).
  - Comma-before-"please" lookahead removed → the pre-existing `test_task_execution_tags_require_positive_valid_delegation[Codex, please handle this code repair.-codex-True-True]` fails.

No changes to `docs/guides/pebble-capture.md`: its filing-rule description ("explicit filing request ... in the same positive clause as its action_evidence") remains accurate — the same-clause requirement is unchanged; only the within-clause governance mechanism moved from a character window to segmentation.

nbramia:

**Specialist rationale:** `review-security` retained — the delta rewrote the authority gate's core mechanism again (segmentation replacing the proximity window) and split an authority-bearing evidence set in two; that is precisely where a fix opens a new hole. `review-testing` **dropped** this round: its round-2 concern (unfalsifiable tests) was resolved, the addresser supplied per-mutation evidence, and the general reviewer covers test adequacy for the delta — it independently confirmed all four new tests fail against the pre-round-3 file.

Reviewed against `git diff f4df0d1b..e832f4d2`.

### Reviewer Findings

#### General

**Defects** — None. **Missing Tests** — None.

**Analysis** — `used_scheduled_delegation_evidence` is written in exactly two places (the `is_delegated_task` branch at `:859`, the `scheduled_action == "agent"` branch at `:909`) and read in exactly one (`:900`). A plain untagged task now writes only `used_action_evidence`, so it can no longer trip the schedule gate's raise — confirmed by reverting to `f4df0d1b` and watching the new test fail with exactly the round-2 `PebbleCaptureError`. The reverse direction still holds because a filed schedule writes to *both* sets. A task dropped by the filing gate is reached before either set is touched, so no set is left half-written. `_task_request_segments` returns `[]` for empty input, all-separator input, and leading/trailing separators; an infinitive-first clause is appended rather than merged into a nonexistent predecessor; segmentation is linear (a 120,000-char clause with 20,000 commas segments in ~0.02s). `_TASK_REQUEST_GOVERNANCE_WINDOW` has zero remaining references. `docs/guides/pebble-capture.md`'s "same positive clause" wording was already a simplification under the window mechanism and remains accurate (not false) under segmentation — leaving it unchanged is sound.

**Status** — 233 passed. Verified the four new tests (`test_plain_task_evidence_consumption_never_aborts_an_independent_agent_schedule`, `test_marker_governs_only_its_own_short_bystander_span_within_the_window`, `test_comma_chained_spoken_list_files_only_the_marker_governed_span`, `test_unrelated_delegation_marker_does_not_govern_a_later_bystander_span`) all fail against the pre-round-3 file, and the deliberate non-regression checks pass on both — so they genuinely pin the delta.

**Summary** — Clean converging round; correct, minimal, regression-free.

#### Security

**Defects**

- `_explicit_task_request`'s segment governance validates only where an `action_evidence` span **starts**, never where it **ends**, and the segment builder's own merge rules can stitch unrelated requests together. Three independently verified reproductions through `validate_plan`:
  1. **Span straddling.** `"Assign to Claude water the plants and then call the neighbor about the noise complaint."` with `action_evidence` spanning both halves and `tags:["claude"]` files a **delegated** task whose title is set directly from `action_evidence` (`:858-860`) — so `"…and then call the neighbor about the noise complaint"` lands on an agent-consumable title although it was never the object of "assign … to".
  2. **Backward continuation merge.** `_TASK_REQUEST_CONTINUATION_RE` merges a continuation segment into whatever precedes it without checking that the predecessor is part of the same request. `"Call the neighbor about the noise complaint, to remind me to water the plants."` becomes one segment, so the `remind me to` marker — which speaks only to "water the plants" — governs the unrelated clause-initial bystander, which files.
  3. **The `,(?!\s*please\b)` deviation.** `"Put milk on my list, please call the plumber about the leak."` stays one segment, so `"please call the plumber about the leak"` files by riding the unrelated `put … on my list` marker.

  All three reach the same public entry point, need no changes elsewhere, and reopen the exact defect class round 2's fix was meant to close, via mechanisms that fix didn't cover.

**Missing Tests** — one per mechanism: containment (not merely start-inside); the backward merge not retroactively extending a downstream marker over an unrelated preceding segment; and the `, please` carve-out not merging two independently-intended requests.

**Status** — 165 passed. Re-verified both round-2 reproductions are genuinely fixed. Independently re-walked the two-set split: a schedule cannot reuse a span spent by a delegated task, two agent schedules cannot share a span, and schedule-ordered-first is still respected; the one relaxation (a *plain* task's span no longer blocks a later agent schedule) is intentional and each side's authority is independently proven, so not treated as a defect. Negation/conditional/quoted-speech blocking and the no-raise fix both hold on the segmented path.

**Summary** — The evidence-set split is correctly scoped and opens no authority hole. The segmentation scheme, however, checks only that a marker occurs in the same delimited unit and that the span *starts* within it. Not clean; should not merge as-is.

### Referee Decisions

| 7 | Segment governance is too permissive about what a marker governs — span end unchecked, backward merge unconditional, `, please` carve-out too broad | security (+ orchestrator) | Defect | **Valid — all four reproductions confirmed** | **Accept** | I reproduced all three of security's mechanisms plus a fourth of my own (`"add a task to buy milk, please water the plants"` → `segments=[(0,47)]`, proposed=2 **filed=2**; the non-`please` variant correctly files 1). Mechanism 1 is the most serious: the smuggled text becomes the *title of a task tagged `claude`*, i.e. it reaches an agent's queue. Smallest remedy — **proven by the orchestrator at 37/37** across all four round-3 reproductions, every round-2 reproduction, and the full round-1 calibration table: give separators two tiers. **Hard** (`,` `then` `and then` `also` `plus` `as well as`) — a span may not cross one, which kills mechanism 1. **Soft** (bare `and`) — a span may cross one, which is what keeps `"Add a task to buy milk and eggs."` filing as a single task. Marker governance stays "evidence STARTS in a marker-bearing soft segment"; containment becomes "evidence lies entirely within one hard block". Then merge a continuation segment backward **only into a block that already contains a marker** (kills mechanism 2 while preserving `"Add a task assigned to #claude, to review the synthetic report."`), and treat `, please` as a boundary **unless the token before the comma is a valid assignee** (kills mechanisms 3 and 4 while preserving `"Codex, please handle this code repair."`). |
| — | General reviewer's probe of the `, please` deviation (`"Buy milk, please call the plumber."` → False, "no hole") | general | — | **Superseded** | **Not a finding** | Recorded for the audit trail: that probe used a clause containing *no filing marker at all*, so it could not exercise the shape that fails. The deviation is exploitable only when a marker appears earlier in the same clause. Noting it because it is the reason this round's two reviewers disagreed. |

**Rejected:** none.

**Convergence:** 4 accepted in round 1, 2 in round 2, **1** in round 3. Stated plainly so the count is not flattered: finding 7 bundles four reproductions, but they are one defect in one function (`_explicit_task_request` / `_task_request_segments`) closed by one coherent rule change, and forwarding them separately would invite four uncoordinated patches. Still strictly converging, no review-introduced architecture, and every finding traces to the original acceptance criteria.

nbramia:
Addressed round-3 finding #7 (segment governance is too permissive about what a marker governs).

**Root cause:** `_explicit_task_request` only checked where an evidence span *started* relative to a marker segment, never where it *ended*, and the single-level separator regex (`_TASK_REQUEST_SEGMENT_RE`) treated all commas/coordinators the same, letting a continuation merge and a blanket `,(?!\s*please\b)` exception stitch unrelated requests together.

**Fix — bounded two-tier separator rule (per the accepted remedy):**
- `_task_request_hard_blocks`: splits a clause on hard separators (`,`, "and then"/"then"/"also"/"plus"/"as well as"); an evidence span may not cross a hard block boundary (containment).
- `_task_request_soft_segments`: within a hard block, splits on bare "and" (soft); an evidence span may cross a soft separator, but a marker still only governs the soft segment its span starts in.
- Continuation merge (`^\s*(?:to|so that|in order to)\s+\w`) now only merges backward into a preceding hard block that already carries its own marker — a bare `", to remind me to ..."` no longer pulls an unrelated leading request under a later marker.
- The `, please` exception is narrowed: a comma is a boundary unless the token immediately before it is a valid task assignee ("Codex, please handle this" stays merged; "...on my list, please call the plumber" now splits).

**Verification:**
- Reproduced all four findings on head `e832f4d2` before fixing (all incorrectly filed/merged).
- Reference 37-case table run against the real `_explicit_task_request` — 37/37.
- Mutation-falsification: (i) removing the containment check re-breaks the straddling-span test, (ii) making the continuation merge unconditional re-breaks the backward-merge test, (iii) restoring the blanket `,(?!\s*please\b)` re-breaks both comma/please tests. All three reverted cleanly back to the fixed state.
- `tests/test_pebble_capture.py` (169 tests, incl. 5 new tests for this finding), `tests/test_journal_persona.py`, `tests/test_journal_capture.py`, `tests/test_journal_ring_ingest.py`, `tests/test_no_change_narration.py` — 237 passed.
- `ruff check` on changed files — clean.

https://claude.ai/code/session_01F98UyxdwoJRuUhb2iGDpLy

nbramia:
## Review Round 4 — Consolidated Review & Referee Decisions

Reviewed against `git diff e832f4d2..7a66a4d9` (the two-tier separator rule). Both reviewers confirmed the round-3 containment gap is genuinely closed and found no regression on the operator's real phrasings. Both then found a new defect in the same helper, `_comma_is_hard_separator`.

### Reviewer Findings

#### General

**Defects**

- `_comma_is_hard_separator` (`api/services/pebble_capture.py:114-118`) makes `_task_request_hard_blocks` **O(n²) in comma count**. For every comma it slices `clause[:comma_index]` — an O(n) copy — then runs an end-anchored search over that O(n) prefix. The superseded single-tier `_TASK_REQUEST_SEGMENT_RE.finditer(clause)` was one linear pass, so this is a complexity regression introduced by this delta, not pre-existing cost. `ready_result` places no length cap on `final_text`, and `_positive_clauses` splits only on `.!?;\n`, so a long punctuation-light dictated list lands in one clause. Reproduced through the public entry point: a 28.8 KB transcript with ~2400 commas takes 0.79s for a *single* action, scaling 0.015 → 0.05 → 0.2 → 0.79 → 3.2s as length doubles. cProfile attributes 763ms of 788ms to 2405 calls of `_COMMA_ASSIGNEE_BEFORE_RE.search`. Capture processing runs through one serial consumer, so a single comma-heavy capture stalls every queued capture behind it. Smallest fix: match against a bounded trailing window instead of the full prefix.

**Missing Tests** — none. All four new tests were verified to fail against `e832f4d2` and pass against `7a66a4d9`.

**Status** — 237 passed. Independently re-verified all six operator phrasings file correctly; probed degenerate inputs (empty, all-separator, long soft-segment chains) with no crashes; confirmed no dead references to the superseded `_TASK_REQUEST_SEGMENT_RE`/`_task_request_segments`.

**Summary** — The two-tier rule is correct and minimal for the problem it targets, with no regressions and tests that genuinely pin it. One new defect: the quadratic comma scan.

#### Security

**Defects**

- `_comma_is_hard_separator` treats a comma as non-hard whenever the token before it is **any** member of `_VALID_TASK_ASSIGNEES` — which includes `me`, `local`, and `cloud`, ordinary English words, not just the name-like `claude`/`codex`/`hermes` the exception was designed for. `"me"` in particular is in constant spoken use ("call me", "text me", "remind me"). This reopens the round-3 failure mode — un-requested text stitched onto a delegated task's agent-consumable title — through a narrower but still reachable trigger. Reproduced end-to-end:
  ```
  transcript = "Assign to Claude, call me, please water the plants while I'm away."
  -> ONE action filed: kind='task', tags=('claude',),
     title="call me, please water the plants while I'm away"
  ```
  Mechanism: the comma between "me" and "please" is judged non-hard, so the block survives intact; `_task_request_has_marker` then finds a marker only because `_explicit_tags`' pre-existing "TAG, please VERB" pattern captures `"me"` as a self-assignee — an accidental marker unrelated to watering plants. Containment then permits the whole straddling span, including the `"call me, "` prefix, to become the `#claude` task's literal title. Also reproduces with `"local"` (`"Save the file local, please back up the photos too."`). Smallest fix: exclude `"me"` (mirroring the `(?!me\b)` exclusion already used elsewhere in this file for the same reason), or restrict the exception to name-like tags.

**Missing Tests** — nothing exercises `me`/`local`/`cloud` as the pre-comma token in an ordinary, non-delegating sense; the new `test_comma_please_boundary_excludes_a_non_assignee_leading_span` covers only the non-assignee case.

**Status** — 169 passed. Confirmed legitimate `"Codex, please …"` delegation still files; negation/conditional/quoted-speech blocking holds; no production files or server touched.

**Summary** — The delta correctly closes the round-3 containment gap for the general case, but the new assignee exception is under-discriminating and reopens the same class on the highest-severity path.

### Referee Decisions

Both findings independently reproduced by the orchestrator against `7a66a4d9` before acceptance.

| 8 | The `, please` assignee exception accepts ordinary words (`me`, `local`, `cloud`), reopening bystander-smuggling onto a delegated title | security (+ orchestrator) | Defect | **Valid — confirmed** | **Accept** | I found the same hole independently before security reported it (`"add a task to buy milk for me, please water the plants"` → files **both** spans; `"put the oil change on my list for me, please call the vet"` → files **both**). Security's reproduction is the more severe one, landing `"call me, "` in a `#claude` task title. `"for me,"` is entirely ordinary voice-note phrasing, so this clears the normal-path bar. |
| 9 | `_comma_is_hard_separator` is O(n²) in comma count, stalling the serial capture consumer | general | Defect | **Valid — confirmed** | **Accept** | Reproduced the scaling exactly: 0.016 → 0.063 → 0.202 → 0.838 → 3.235s across five doublings — a clean ×4 per doubling. A genuine regression versus the linear pass this delta replaced. |

**One combined remedy — proven by the orchestrator, 9/9 on discrimination and linear on complexity.** Both findings live in `_comma_is_hard_separator`, so one change closes both: restrict the exception to **name-like** tags (drop `me`, `local`, `cloud`), require the tag to be in **vocative** position (opening its own segment, as in `"Codex, please …"`), and evaluate against a **bounded trailing window** rather than a full prefix slice. Measured after the change: `0.0010s` at 57 KB / 4800 commas, versus `3.235s` before — and 9/9 on the discrimination table including every `me`/`local` reproduction, with `"Codex, please handle this code repair."` and `"Buy the milk and then Codex, please handle the code repair."` still filing.

**Rejected:** none.

### Convergence Audit

| Round | Accepted | Sub-problem |
|---|---|---|
| 1 | 4 | marker regex breadth (2), run-on clause binding, repair-round coverage |
| 2 | 2 | schedule-gate collision, governance specificity |
| 3 | 1 | segment governance (containment, merge, `, please`) |
| 4 | 2 | `_comma_is_hard_separator` — discrimination + complexity |

**Finding density is concentrated in one sub-problem:** the task-request governance helper has produced a finding in all four rounds. Flagging it explicitly as a convergence-trap signal. Countervailing evidence that this is hardening rather than thrash: the count fell 4 → 2 → 1 before this round; **no fix has ever been reverted**; every finding traces to the original acceptance criterion (the filing gate) rather than to review-introduced architecture; and both of this round's findings share one small, already-proven remedy at a single call site.

Round 4 accepting 2 after round 3's 1 is a non-decrease — one strike, not a stall (which requires two consecutive). **Round 5 is the ceiling.** If round 5 is not clean, this stops and goes to the operator with a recommendation rather than continuing.

**Findings forwarded to addresser:** 2 (one combined fix)

nbramia:
## Round 4 addressed — both findings, one combined fix

Both findings were in `_comma_is_hard_separator`; fixed with one change.

### Finding 8 (discrimination) — Applied
The `, please` exception now fires only when **both** hold:
1. **Name-like tag** — derived from `_VALID_TASK_ASSIGNEES` minus `{"me", "local", "cloud"}` (i.e. `claude`, `codex`, `hermes`, `cloud-sonnet`, `cloud-haiku`), not a second hardcoded list.
2. **Vocative position** — the tag must open its own segment (clause start or immediately after a hard-separator token: `.!?;\n,` / "and then" / "then" / "also" / "plus" / "as well as"), distinguishing `"Codex, please…"` (address) from `"…for me, please…"` (ordinary object pronoun).

### Finding 9 (complexity) — Applied
Both the "please" lookahead and the vocative lookbehind now run over a fixed 40-char window around the comma (`_COMMA_SEPARATOR_WINDOW`) instead of a full clause slice, making the per-comma cost O(1) instead of O(n).

### Reproduced before fix (head `7a66a4d9`)
```
"Assign to Claude, call me, please water the plants while I'm away."
  -> [(('claude',), "call me, please water the plants while I'm away")]
"add a task to buy milk for me, please water the plants" -> ['Buy milk', 'Water the plants']  (both filed)
"put the oil change on my list for me, please call the vet" -> both filed
"add a task for me, please water the plants" -> ['Water the plants']  (want log-only)
"Save the file local, please back up the photos too." -> ['Back up the photos']  (want log-only)

Timing (please-dense adversarial, through validate_plan):
  commas=  100   0.017s
  commas=  200   0.042s   x2.5
  commas=  400   0.167s   x4.0
  commas=  800   0.651s   x3.9
  commas= 1600   2.497s   x3.8
```

### After fix
```
9/9 discrimination table correct (through the real _comma_is_hard_separator).
Timing at commas=2400 (please-dense): 0.087s (was multi-second quadratic).
37/37 on the round-3 governance table (through the real _explicit_task_request) — no regression.
```

### New tests (tests/test_pebble_capture.py)
- `test_comma_please_exception_excludes_ordinary_object_pronoun_me`
- `test_comma_please_exception_does_not_smuggle_bystander_text_onto_a_delegated_title`
- `test_comma_please_exception_excludes_ordinary_word_local`
- `test_comma_please_exception_excludes_ordinary_word_even_in_vocative_position`
- `test_comma_please_exception_requires_vocative_position_not_any_mention`
- `test_comma_please_vocative_exception_non_regression`
- `test_comma_hard_separator_check_is_bounded_not_quadratic` (2400-comma, "please"-dense adversarial transcript; budget 2.0s — old code took ~5.8s at this size, fixed code ~0.09s)

Each new test was falsified by mutating the production line(s) it guards (name-like filter, vocative anchor, and the bounded-window slicing, individually and combined), confirmed to fail, then reverted with `git diff` matching the clean fix exactly.

### Tests
`HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py tests/test_no_change_narration.py -q` → **244 passed**.
`ruff check` on changed files → all checks passed.

https://claude.ai/code/session_01F98UyxdwoJRuUhb2iGDpLy

nbramia:
## Review Round 5 — Consolidated Review & Referee Decisions

Reviewed against `git diff 7a66a4d9..a35144c1` (the narrowed `, please` exception: name-like tag + vocative position + bounded window).

### Reviewer Findings

#### General

**Defects** — None. **Missing Tests** — None.

**Status** — 244 passed. Confirmed `_TASK_REQUEST_NAME_LIKE_ASSIGNEES` is a live comprehension over `_VALID_TASK_ASSIGNEES` (`:99`) with no second hardcoded list, so a future tag added to `agent_board.py` flows through automatically. Confirmed no dead code survives the superseded full-slice version (`_ASSIGNEE_ALTERNATION_RE` / `_COMMA_ASSIGNEE_BEFORE_RE` fully removed, no stray references). The new quadratic-regression guard runs in ~0.5s against its 2.0s budget — a ~20x margin, not flake-prone. Probed boundary cases directly (comma at index 0, comma at end of clause, tag exactly at the 40-char window edge, a long preamble pushing the real separator against the boundary) — no crashes and no false-positive vocative match. Explicitly declined to report the one theoretical window-edge misfire: it requires 28+ consecutive whitespace characters before the tag, is unreachable from transcribed speech, and its failure direction is toward splitting (under-filing), which the contract accepts.

**Summary** — Clean. The whole-PR contract (strong presumption of log-only, explicit filing or delegation required) holds through this final change.

#### Security

**Defects** — None. **Missing Tests** — None; the delta ships direct coverage for every case probed.

**Status** — 176 passed. Independently exercised `validate_plan` / `_explicit_task_request` / `_comma_is_hard_separator` against realistic voice phrasings: the exact round-4 reproduction (`"Assign to Claude, call me, please water the plants…"` — now correctly drops the un-requested clause); non-vocative name-like mentions before `", please"` (`"I emailed codex, please…"`, `"Give the file to Codex, please…"` — correctly hard-split); the accidental `_explicit_tags` "TAG, please VERB" marker firing on "me" without a comma (`"Text me please water the plants…"` — files a plain *untagged* task, not an authority escalation); and a deliberate attempt to force the 40-char window's `^` anchor to fire on truncation rather than a real separator (mathematically unreachable — would need ~34 characters of literal whitespace between tag and comma). No case produced a false-positive authority grant.

**Summary** — The round-5 delta closes the round-4 authority defect cleanly. The O(1) window bound is real and tested against a 2400-comma transcript. Security posture is sound; this round is clean.

### Referee Decisions

No findings were reported by either reviewer. Independently re-verified against `a35144c1` before closing the loop:

- Round-3 governance table through the real `_explicit_task_request`: **37/37**.
- All five round-4 reproductions fixed; `"Codex, please handle this code repair."` and `"Buy the milk and then Codex, please handle the code repair."` still file with their `codex` tag.
- Complexity through `validate_plan`: ×2.0, ×2.1, ×1.9, ×2.0 across five doublings — linear, versus the ×4 quadratic scaling before.
- End-to-end against the operator's real capture shapes (synthetic substitutions, grammar preserved): all four task-creating shapes still file, including the delegated one with its `claude` tag; all five non-task shapes are log-only; and a rambling voice note in which the classifier proposed 4 tasks files exactly 1 — the one actually requested.

### Convergence

| Round | Accepted | Sub-problem |
|---|---|---|
| 1 | 4 | marker regex breadth (2), run-on clause binding, repair-round coverage |
| 2 | 2 | schedule-gate collision, governance specificity |
| 3 | 1 | segment governance (containment, merge, `, please`) |
| 4 | 2 | `_comma_is_hard_separator` — discrimination + complexity |
| 5 | **0** | — |

Nine findings accepted across four rounds, none rejected, none reverted. The round-4 convergence-trap flag is resolved: the sub-problem that drew findings in every round produced zero in this one.

nbramia:
## End-to-End Verification — PR #1135

### Verdict: PASS

verification-record: v1
verification-head: a35144c19ff19bbb1c1d7e32f6f4d6f950735813
suite-result: pass
suite-command: ["python trusted-runner/scripts/verify_candidate.py pushed-ref --repository \"$GITHUB_WORKSPACE/candidate\" --sha a35144c19ff19bbb1c1d7e32f6f4d6f950735813 --base 7ec80e6add95cfff57794ac7cf33cbeadc4a6555 --evidence-root \"$RUNNER_TEMP/lifeos-verification-evidence\" --lanes fast-unit --workers 4 --capacity-max-run-workers 4 --lane-log-dir \"$RUNNER_TEMP/lifeos-lane-logs\" --part-index 0 --part-count 4 --parallel-browser-free", "... --part-index 1 --part-count 4 ...", "... --part-index 2 --part-count 4 ...", "... --part-index 3 --part-count 4 ..."]
suite-executions: 1
suite-exit-status: 0
suite-command-results:
  - command: verify_candidate.py pushed-ref --part-index 0 --part-count 4 (fast-unit)
    result: pass
    exit-status: 0
  - command: verify_candidate.py pushed-ref --part-index 1 --part-count 4 (fast-unit)
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-2"
  - command: verify_candidate.py pushed-ref --part-index 2 --part-count 4 (fast-unit)
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-3"
  - command: verify_candidate.py pushed-ref --part-index 3 --part-count 4 (fast-unit)
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-4"
retry-reason: none

**Lane provenance:** consumed the hosted `candidate-verification-shadow` run (`.github/workflows/candidate-verification.yml`, `pull_request_target`, GH Actions run 35253232076) that GitHub had already executed automatically for this exact head when it synced — not a fresh local or re-triggered run. That run invoked the repository's own `scripts/verify_candidate.py pushed-ref` in 4 module-scope-balanced parts, on the isolated CPU-only trusted-runner image (torch 2.14.0+cpu, HIP/ROCR/CUDA disabled), covering the full `fast-unit` lane (2140+2209+2099+2373 = 8821/8821 selected tests, matching `lane_totals`). The workflow's own fail-closed diff check narrowed lanes to `fast-unit` only (`browser-free` excluded) because this PR touches no `web/` file — that narrowing is the trusted workflow's own logic, not mine. I did not run a local gate: a real, complete, passing hosted execution already existed for this exact head (confirmed via `gh pr view 1135 --json headRefOid` == `a35144c1...` both before and after this check), so a duplicate local run would have been pure re-execution risk (OOM/GPU-freeze exposure this host is documented to have) for no additional evidence. The literal App-bound *required* `candidate-verification` check (via `workflow_dispatch`) is published later by `candidate_publisher.py` as part of merge, outside this phase's scope.

Durations (from the run's own timestamps): job wall time (checkout + CPU-torch/requirements/Playwright install + test execution) ranged 5m54s–7m35s across the 4 parallel parts; within that, the `verify_candidate.py` step itself (pure test execution, excluding checkout/dependency setup) ranged 4m26s–5m40s per part. Setup/dependency-install time (~1.5–2min per part) is separated out above as queue/setup, not execution, per each part's own step timestamps.

### System Flow Verified

Pebble capture → `PebbleCaptureConsumer.process()` → local classifier proposes actions → `validate_plan()` (the real, unstubbed authority gate this PR changes) → `CaptureLedger` / `TaskManager` / `SchedulerStore` effects → ledger reaches a terminal receipt state. This is the production entry point used by `PebbleCaptureWatcher` against the real vault-backed stores; only the local-LLM classifier is stubbed (it is not the code this PR changes).

### Evidence

#### Scenario A — Explicit-ask capture files exactly the expected task
**Flow:** ready-result payload with `final_text = "Add a task to buy synthetic light bulbs for the garage."` → classifier proposes one task bound to `action_evidence = "buy synthetic light bulbs for the garage"` → `validate_plan` proves the marker governs that span → `TaskManager.create_or_find_by_operation` → ledger records the receipt.
**Result:** `process()` returned `"complete"`; `tasks.list_tasks()` == `["Buy synthetic light bulbs"]`; ledger's `-1` receipt state == `"applied"`. PASS — an explicit filing request is still honored end-to-end through the real stores.

#### Scenario B — Log-only capture drops the task but still completes (does not stick pending)
**Flow:** `final_text = "Went for a synthetic walk today and thought about the garden."` (no filing marker) → classifier over-eagerly proposes a task anyway, with `action_evidence` that is real transcript text but ungoverned by any marker → `validate_plan`'s `_explicit_task_request` fails to prove it → the action is dropped (not raised).
**Result:** `process()` returned `"complete"`; `tasks.list_tasks()` == `[]`; ledger's `-1` receipt state == `"applied"`. PASS — this is the PR's core behavior change: previously this transcript had no gate at all and would have filed the task; now it is silently dropped and the capture still reaches a terminal, non-pending state.

#### Scenario C — Mixed proposal files exactly the one requested task
**Flow:** `final_text = "Add a task to water the synthetic office plants. I was thinking about calling the plumber and maybe repainting the fence."` → classifier proposes 3 tasks (water plants / call plumber / repaint fence) → only "water the synthetic office plants" sits inside a marker-governed clause.
**Result:** `process()` returned `"complete"`; `tasks.list_tasks()` == `["Water the office plants"]` (exactly one, not three); ledger applied. PASS — proves the per-action gate, not an all-or-nothing capture-level gate.

### Issues Found
None.

### Holistic Assessment
The hosted `fast-unit` execution (8821/8821 tests, 4/4 parts green) and the three executable scenarios above together demonstrate the authority-boundary change works correctly against the real running stores: an unproven task is dropped (never raised) and the capture still reaches `complete`; a proven task still files with the correct title; and a mixed proposal files exactly the requested subset. No regression observed. `browser-free` was correctly not exercised (no `web/` changes) per the trusted workflow's own lane selection.

nbramia:
### Round: simplification-1

| 1 | AGENTS.md's unconditional "no data leaves the machine" privacy claim was false after this PR's routing change | Applied | Replaced the bullet in "6. Privacy Is Non-Negotiable" with: "LLM inference runs locally by default and on a keyless install; a remote or cloud provider is used only when the operator explicitly configures one (`LIFEOS_REMOTE_LLM_*`, `LIFEOS_LLM_BACKEND`, or an Anthropic key) — see [ADR-024](docs/adr/024-remote-llm-backend.md) and, for Pebble specifically, [guides/pebble-capture.md](docs/guides/pebble-capture.md). Sending data off-box is always an explicit operator choice, never a silent default." Only this one line changed; the rest of the section is untouched. |
| 2 | `scripts/eval_pebble_filing.py`'s 13 cases overlap the prompt's few-shot examples, so a green score partly reflects the model echoing its own examples | Applied | Added a held-out section (`_HELD_OUT_CASES`, `_HELD_OUT_MULTI_ITEM_CASES`, `_HELD_OUT_SCHEDULE_CASES`) whose 12 cases appear nowhere in `JOURNAL_BEHAVIOR_CASES` or the original table, plus a `_score_schedule_case` helper (the expectation model needed a "notify schedule, not a task" check for one case). The two sections are scored and printed separately, with an overall combined score/threshold check retained for the exit code. A comment explains why the section must stay out of the prompt. |

**Eval run against the real configured provider** (Fireworks DeepSeek v4-flash, `settings.remote_llm_configured`):
- Table (overlaps prompt examples): 13/13 (100%)
- Held-out (absent from prompt): 12/12 (100%)
- Overall: 25/25 (100%); threshold 80% — PASSED
- No failing cases.

**Tests:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/pytest tests/test_pebble_capture.py tests/test_journal_persona.py tests/test_journal_capture.py tests/test_journal_ring_ingest.py tests/test_no_change_narration.py -q` → 208 passed.

**Lint:** `ruff check` on changed files (`AGENTS.md`, `scripts/eval_pebble_filing.py`) → all checks passed (also confirmed by the pre-push hook's own ruff pass).

**Commit:** `fix: address review round simplification-1 — privacy claim and eval held-out set` (e368ce96)

nbramia:
## End-to-End Verification — PR #1135

### Verdict: PASS

verification-record: v1
verification-head: e368ce963232d5f6af6de7cc275e318c55ff6846
suite-result: pass
suite-command: ["python scripts/verify_candidate.py pushed-ref --repository /home/runner/work/LifeOS/LifeOS/candidate --sha e368ce963232d5f6af6de7cc275e318c55ff6846 --base 7ec80e6add95cfff57794ac7cf33cbeadc4a6555 --evidence-root $RUNNER_TEMP/lifeos-verification-evidence --lanes fast-unit --workers 4 --capacity-max-run-workers 4 --lane-log-dir $RUNNER_TEMP/lifeos-lane-logs --part-index 0 --part-count 4 --parallel-browser-free", "... --part-index 1 --part-count 4 ...", "... --part-index 2 --part-count 4 ...", "... --part-index 3 --part-count 4 ..."]
suite-executions: 1
suite-exit-status: 0
suite-command-results:
  - command: verify_candidate.py pushed-ref --lanes fast-unit --part-index 0 --part-count 4
    result: pass
    exit-status: 0
  - command: verify_candidate.py pushed-ref --lanes fast-unit --part-index 1 --part-count 4
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-2"
  - command: verify_candidate.py pushed-ref --lanes fast-unit --part-index 2 --part-count 4
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-3"
  - command: verify_candidate.py pushed-ref --lanes fast-unit --part-index 3 --part-count 4
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-4"
retry-reason: none

Evidence source: the App-bound required `candidate-verification` check has not
run for this head yet (it is only dispatched by the publisher at merge time).
The `pull_request_target`-triggered `candidate-verification-shadow` run
(diagnostic, same underlying `scripts/verify_candidate.py` mechanism and
lane-selection logic) completed for this exact head before I would otherwise
have started a redundant execution, so I consumed it rather than duplicate it
(GH Actions run 35266776631, all 4 `candidate-execution` parts + the
publisher job green, headSha confirmed = e368ce963232d5f6af6de7cc275e318c55ff6846).
Lane selection correctly narrowed to `fast-unit` only (no `web/` file changed
by this PR, so `browser-free` cannot be affected) — full coverage confirmed:
2419+2181+2171+2014 = 8785/8785 node IDs, all 4 parts `"result": "success"`.
My own local `pushed-ref` reproduction of the same command stalled at 99%
(confirmed genuinely stuck via zero CPU-tick progress over 45s, not just
slow) after ~75 minutes on this heavily-contended host (load avg 6+, two
long-running unrelated busy-loop processes pinning 2 cores since Sep 12) and
was killed; the local `EvidenceStore` correctly recorded that attempt as
`infrastructure_failure` (not a false pass) — capacity slots were not
exhausted (0/1 of 4 busy throughout), ruling out the documented
run-width-equals-capacity-total hang. The hosted shadow run's clean, fast
(~9 min), fully-green completion on an uncontended runner for the identical
candidate corroborates that the local stall was host contention, not a
defect exposed by this PR.

### System Flow Verified

Pebble ready-result payload -> `PebbleCaptureConsumer.process()` -> classifier
(stubbed, standing in for the LLM) -> `validate_plan()` authority gate ->
`CaptureLedger` (SQLite receipt ledger) + `TaskManager` (Markdown Inbox) +
`SchedulerStore` (Markdown Scheduler Inbox), exercised end-to-end in a
temporary isolated directory against real store implementations (no mocks
below the classifier boundary, no network calls).

### Evidence

#### Scenario (a): Explicit-ask capture creates exactly the expected task
**Flow:** ready-result payload ("Add a task to buy synthetic printer paper.")
-> stub classifier proposes one task -> `validate_plan` -> `TaskManager.create_or_find_by_operation`
-> ledger reaches `complete`.
**Command:** `~/.venvs/lifeos/bin/python e2e_scenario.py` (scenario_a)
**Result:** PASS — `result='complete'`, exactly 1 task created (`Buy synthetic printer paper`), 0.023s.

#### Scenario (b): Log-only capture creates zero tasks, still completes
**Flow:** ready-result payload ("I noticed a synthetic bird in the garden.")
-> stub classifier proposes nothing (log-only per policy) -> ledger still
reaches `complete`, not stuck pending.
**Result:** PASS — `result='complete'`, 0 tasks created, 0.010s.

#### Scenario (c): Multi-item capture files only the requested item
**Flow:** transcript mentions three things ("buy synthetic milk and feed the
synthetic cat before dinner and water the synthetic garden plants");
classifier (behaving per the log-only-default policy) proposes only the one
actually-requested item.
**Result:** PASS — exactly one task filed (`buy synthetic milk`); "cat" and
"garden" do not appear in any filed task title. (This demonstrates the
system doesn't add extra items beyond what the classifier proposed — the
item-selection judgment itself is the classifier's, per prompt policy, not
an application-code gate; that judgment is independently scored by
`scripts/eval_pebble_filing.py`, already run by the orchestrator: 25/25.)

#### Scenario (d): Execution-authority gates still block unproven delegation
**(d1) Executor-tagged task, no real delegation evidence in transcript:**
classifier proposes `tags:["codex"]` with fabricated `delegation_evidence`/
`action_evidence` that do not appear in the transcript. `validate_plan` drops
the tag (not raise) — capture completes with a plain, untagged task filed
(`review the synthetic login bug`, `tags: []`), zero codex-tagged tasks.
PASS.

**(d2) Agent schedule, no real delegation evidence in transcript:**
classifier proposes a `cloud-sonnet` agent schedule with fabricated evidence.
`validate_plan` raises `PebbleCaptureError("agent schedule lacks explicit
valid delegation")`; zero agent schedules created (confirmed via
`SchedulerStore.list_all()`). PASS.

**(d3) Not permanently pending after rejection:** same capture_id/revision
redelivered with a corrected (log-only) classifier output reaches `complete`
normally — 0 tasks, 0 agent schedules, no ledger corruption from the earlier
rejection (nothing was persisted before the raise, so the retry reprocesses
cleanly). PASS.

**Command:** `HIP_VISIBLE_DEVICES="" ROCR_VISIBLE_DEVICES="" CUDA_VISIBLE_DEVICES="" ~/.venvs/lifeos/bin/python e2e_scenario.py`
**Result:** 6/6 scenarios PASS (full stdout in the handoff artifact's supplementary evidence).

### Issues Found
None.

### Holistic Assessment
The system works correctly end-to-end after this change. Plain-task filing
is now genuinely governed only by the classifier prompt (no application-code
gate on plain tasks), verified by driving the real consumer/ledger/store
stack with a stubbed classifier standing in for that judgment. The
execution-authority gates that matter most — proving delegation for an
executor-tagged task or an agent schedule from an exact transcript span —
remain fully intact and code-enforced: an unproven tag is dropped (task still
files, capture completes), an unproven agent schedule is rejected outright
(no schedule created, capture stays retriable rather than corrupted). Full
fast-unit lane evidence (8785/8785 tests) confirms no regression across the
rest of the suite for this exact head.

Issue #1141, opened 2026-09-18
fix: agent session replies get lost, misattributed, or misreported


nbramia:
## Problem

Telegram is the operator's primary channel for staying in the loop with
background agent sessions, and in several distinct ways that channel is not
trustworthy right now.

Replying to a finished session's message is supposed to wake it back up so
the conversation can continue. Instead, that reply can silently strand the
session: the bot acknowledges the reply as if the session is resuming, but
the session never actually restarts, and nothing ever tells the operator
that it didn't. The operator has no way to notice this short of asking again
and getting a stale, second "I've noted it" reply — because the bot itself
doesn't know the first resume never took, so it can't warn about it either.
There is also no ongoing safety net: nothing periodically checks whether a
session that should be running again is actually running again, so a
silently orphaned session stays orphaned indefinitely.

Separately, when several sessions are active at once, their Telegram
messages land as one flat, unlabeled stream in the same chat. A heartbeat
from one task, a failure from another, and a completion from a third can
appear back-to-back with nothing distinguishing which message belongs to
which piece of work — so the operator has no reliable way to tell, at a
glance, which session just spoke.

A completed session's summary message is also not always the session's
actual final answer. A stray closing thought the agent has after it already
reported its results can overwrite the stored summary and get sent to the
operator moments later as if it were a second, separate completion — even
though it is just a leftover fragment, not new information.

Finally, when a session fails outright, the operator is told almost
nothing: a generic "exited with an error" message, with no indication of
what went wrong, which task it was, or where the agent's in-progress work
went. The task is silently marked cancelled and its unfinished work is
effectively abandoned, with no path back to it from the failure notice
itself.

Two real sessions are stuck in this orphaned state today as a direct result
of the first problem: an operator reply reopened each one, was acknowledged
as if it worked, and neither has run since.

## Solution

The agent worker's session dispatcher gains a rule that recognizes an
already-claimed top-level code/command-line session as eligible for pickup
whenever it is carrying a message queued for delivery — which is exactly
the signature every reopen mechanism (a Telegram reply on a finished
session's message, a threaded reply to a completion/failure notice, or an
automatic reopen triggered by a reply that arrives mid-run) leaves behind.
Today that signature is invisible to the dispatcher for a top-level session,
so a reopened session simply sits without being picked back up.

The Telegram reply handler that reopens a finished session adopts the same
reopen sequence the worker's own internal reopen paths already use —
putting the task's board status and tag back in sync with the session
[...]

PR #1145, opened 2026-09-18
fix: make operator replies reliably resume agent sessions

nbramia:
## Summary

Part of #1141.

- `Worker._dispatch_spawned_sessions` now admits a parentless `claude_code`/`codex` session carrying an undelivered `pending_messages` row — the signature every reopen path (Telegram followup reply, Telegram status-anchor reply, the worker's own `code_reopened_for_pending_messages`) leaves behind. Previously this shape was unconditionally skipped, so a reopened session sat at `claimed` forever once its one-shot fresh-claim dispatch had already run. The admission is safe against double-dispatch: a fresh top-level claim never produces a `pending_messages` row (`_dispatch` hands its first-turn payload straight to the CLI pool in memory), and `_dispatch_spawned_sessions` runs before the tick's own claim loop, so it can never see a session that loop is about to claim for the first time this tick.
- `telegram._handle_status_anchor_reply`'s terminal-session reopen now swaps the backing vault task's terminal tag back to `#agent-running` and its status to in-progress — the same sequence `_resume_as_followup`/`code_reopened_for_pending_messages` already perform — before claiming the session. If the swap fails (the card was retagged/reassigned since the session finished), the session is left alone and the operator is told the resume couldn't be completed, instead of the previous "waking the session" ack that quietly did nothing.
- A new `Worker._reconcile_stuck_claimed_sessions()` sweep runs alongside `_reconcile_lifecycle_drift()` and alerts once per stuck episode (idempotent via a transcript marker keyed to the reopen's `last_activity_at`) when a reopened top-level CLI session sits unresumed past `LIFEOS_AGENT_STUCK_SESSION_TIMEOUT_MINUTES` (default 15). It deliberately alerts rather than auto-dispatches — the ordinary path above already dispatches every tick, and re-driving it here would risk a second concurrent attempt on a session whose executor may genuinely still be starting up.
- Truthful resume acknowledgments: both the Telegram followup-reply ack and the status-anchor reopen ack now say the note is queued, not that the session has already resumed. A separate `"▶️ Resumed — continuing from your note."` confirmation is sent from the dispatch's resume branch, immediately before the executor's `resume()` call — the first point a resume is actually true.

## Test plan

- [x] New `tests/test_agent_worker_reopen_dispatch.py` (11 tests): reopened top-level claude_code/codex sessions get dispatched; a fresh top-level claim (no pending message) is not double-dispatched; spawned children and operator root-spawns are unaffected (regression); the stuck-sweep alerts exactly once per episode and stays silent inside its window and for sessions with no pending message; a resumed episode after a resolved alert gets a fresh alert; the "Resumed" confirmation fires before the turn runs and is withheld for spawned children.
- [x] New/updated tests in `tests/test_session_thread_replies.py`: the status-anchor reopen's tag/status swap on both the success and failure branches; the followup-reply ack's truthful "queued" text.
- [x] Every new/changed assertion was proven to fail against the pre-fix code (temporarily reverted the corresponding production hunk, watched the test fail, restored) before being counted as done.
- [x] 666 targeted tests green in the foreground (`test_agent_worker_reopen_dispatch.py`, `test_session_thread_replies.py`, and every existing test module touching worker dispatch, Telegram reply handling, session_store, and lifecycle drift) — no regressions.
- The full `./scripts/test.sh` broad gate was not run from this worktree — the shared `test.lock` queue was taking ~80 min/run under concurrent load from sibling implementers on this milestone; the coordinator is running one full gate on the integration branch tip after merging instead.

https://claude.ai/code/session_01W8FaSgWLaoyg36bYGEKYZK


nbramia:
Addressed the adversarial review findings (all 6 accepted):

- A resume no longer marks its queued note delivered, or sends "Resumed", until a subprocess is confirmed to have actually launched (`_confirm_resume_or_requeue`, using the same launch-count signal the re-execute-averted guard already relies on). A pre-start failure — whether the resume attempt returns without launching or raises before it does — leaves the note queued and reports the failure truthfully instead.
- The status-anchor reopen now performs the guarded tag/status transition *before* enqueueing the operator's note, and only claims the session once both the tag swap and the status write succeed; a status-write failure after a successful swap restores the original tag (fails closed) rather than leaving the card inconsistent.
- The stuck-session sweep now re-fetches the backing card and requires it still be actively claimed (`#agent-running`) before alerting; a candidate whose card was resolved some other way is retired silently.
- `LIFEOS_AGENT_STUCK_SESSION_TIMEOUT_MINUTES` is now documented in `docs/guides/agent-worker-setup.md`.
- Renamed the test file's "PR1"/"PR3" section labels to describe behavior instead.

Also rebased onto the merged #1142/#1143 (worktree/branch/PR + BLOCKED_TAG-resumable Human-queue reopen) — no code conflicts; docs conflict resolved by keeping both sections. Confirmed the new BLOCKED→answered→claimed reopen path (`_resume_as_followup`'s BLOCKED_TAG addition) produces the identical CLAIMED+pending-message shape this PR's dispatcher fix already discriminates on, so a Human-queue/blocked card answered via Telegram also resumes through the corrected dispatch path without further changes.

672 pre-rebase / 746 post-rebase targeted tests pass (including `test_agent_worker_clarifications.py` and every `test_agent_worker_git_*.py` file); `ruff check` clean.

nbramia:
Fixed the round-2 finding: the \`claude_code_spawn\`/\`codex_spawn\` transcript marker is written unconditionally before \`Popen\` (deliberate crash-safety margin), but only \`FileNotFoundError\` wrote the compensating \`..._binary_not_found\` marker — any other spawn-time \`OSError\` (e.g. \`PermissionError\`) was miscounted as a real launch, letting a resume confirm and mark its note delivered when no subprocess actually started.

Both executors now catch \`OSError\` broadly and write the same compensating marker for any spawn-time failure. \`_confirm_resume_or_requeue\`'s failure reporting is split: the \`_execute_resume\` exception path still reports its own "didn't start" notice (nothing else would); an ordinary non-raising FAILED outcome now falls through to the existing shared FAILED-outcome handling instead of a separate notice, so the operator gets the executor's own accurate reason without duplication.

Added \`test_permission_error_on_spawn_is_not_counted_as_a_launch\` (real \`ClaudeCodeExecutor\` with a \`PermissionError\`-raising \`spawn_fn\`) — mutation-proven to fail without the fix. The two accepted residual semantics (at-least-once delivery on a crash between launch and confirmation; a pre-launch failure needs an operator-triggered reopen, no auto-retry) are now documented in \`docs/specs/technical/agent-worker.md\`.

836 targeted tests pass; \`ruff check\` clean.

PR #1146, opened 2026-09-18
feat: notify by Telegram when an agent board snooze ends

### LINKED ISSUE #1136: feat: notify by Telegram when an agent board snooze ends
## Problem

An operator snoozes an agent board card to set it aside until a chosen moment — for example, "surface this again tomorrow morning." Today the only thing that happens at that moment is that the card quietly becomes eligible to reappear in its normal lane the next time the board happens to be open. Nothing tells the operator the wake-up occurred if they aren't already looking at the board, so a snooze that was meant to resurface a card at a specific, chosen time behaves indistinguishably from one that just expires unnoticed. This defeats the purpose of picking a wake-up time in the first place — the operator has to remember to go check, which is exactly the babysitting a snooze is supposed to replace.

## Solution

When a card's snooze wake-up time passes, LifeOS sends a Telegram message naming the card, through the same channel the system already uses for scheduler and reminder alerts, and clears the wake-up time so the card is never announced twice. Delivery happens even if no one has the board open, on a short, regular check independent of any board tab or API request — the same shape as the board's other passive watchers (the file watcher that picks up vault edits, the scheduler's own delivery thread). If the message can't be delivered, the wake-up time is left in place so the next check retries it, and if the operator re-snoozes or unsnoozes the card while a notification for its earlier wake-up is in flight, that newer state always wins — the notifier never clobbers a decision the operator made after the fact.

This stays a small, dedicated watcher alongside the board's other background threads rather than an entry in the Markdown-driven scheduler: that subsystem fires from schedule entries the operator authors ahead of time in the vault, not from a snooze picked interactively per card, so folding this in would mean writing and deleting a throwaway schedule entry on every snooze instead of reading the wake-up time that's already stored on the card.

## Acceptance Criteria

- [ ] When a card's `snoozed_until` time is in the past at a poll, the system shall send exactly one Telegram message that includes the card's description, via `SnoozeNotifier.check_once`.
- [ ] Given a card whose natural lane (status/tags alone) is no longer one of the snooze-eligible lanes by the time its wake-up time passes, the system shall NOT send a wake-up notification for it.
- [ ] Given the Telegram send for an expired snooze fails, the system shall leave `snoozed_until` unchanged so the next poll retries, and shall NOT count it as delivered.
- [ ] Given a card is re-snoozed to a new time, or unsnoozed entirely, while its earlier wake-up notification is being delivered, the system shall NOT clear or overwrite the newer value once the send completes.
- [ ] Given the API process was not running at a card's wake-up time, the system shall still deliver the notification once the process is running again (on its next poll or startup) rather than losing it permanently.
- [ ] The wake-up comparison shall use timezone-aware UTC arithmetic, so a `snoozed_until` value carrying any UTC offset fires correctly regardless of the operator's local timezone.
- [ ] The `/health` endpoint's `checks` object shall report the notifier's own thread liveness (e.g. `snooze_notifier`), the same way it already reports `reminder_scheduler`'s liveness — both are Telegram-gated background senders.
- [ ] The notifier shall send from the same Telegram bot/chat used for other system-generated alerts (the scheduler's `notify` action, reminder delivery) — not a persona-specific bot.
- [ ] If Telegram is not configured (`settings.telegram_enabled` is false), the notifier shall not start a background thread at all.

## Verification

- **Automated tests:** unit tests against `SnoozeNotifier.check_once` with an injected `TaskManager` and a fake send callable, covering: successful send + field clear, failed send + retry, future snooze (no-op), natural-lane-ineligible snooze (no-op), concurrent re-snooze/unsnooze during delivery (no clobber), and `/health` reporting the notifier's liveness (an integration-style test against the FastAPI app, following the existing pattern for `reminder_scheduler`/`scheduler_watcher`).
- **Manual checks:** none required — behavior is fully covered by injecting a fixed `now` and a fake sender, matching how the board's snooze-expiry lane derivation is already tested.
- **Existing test suite:** `tests/` for `api/services/agent_board.py` (lane derivation, `SNOOZABLE_LANES`) and `api/services/task_manager.py` (`update` with `_precondition`) must continue to pass unchanged.

## Scope

**In scope:**
- A background watcher, started and stopped with the API process lifespan, that polls for expired snoozes and delivers one Telegram message per expiry.
- Clearing `snoozed_until` only after a successful, still-current delivery.
- A `/health` liveness entry for the new watcher.

**Out of scope:**
- A clickable deep link back to the specific card in the notification message. The board already resolves a `?card=<id>` URL client-side, but nothing in the backend currently constructs a public URL for the board (no base-URL/hostname setting exists in configuration), so wiring a link would mean adding that setting as its own piece of work. The notification includes the card's description text only.
- Any change to how a snooze is set or cleared through the board UI or its API routes — those are unchanged.
- Delivering the notification through a specific persona bot rather than the shared system-alert channel.

## Technical Context

A prior session (Codex, task 4431c8ca) implemented this and left it uncommitted; the salvaged diff is the starting point for this issue, not a finished implementation. It lives at `~/.claude/lifeos-milestone/agent-reliability/snooze-salvage/` (`tracked.patch` touching `api/main.py`, `api/services/agent_board.py`, `docs/specs/product/agent-viz.md`, `docs/specs/technical/agent-viz.md`, plus new `api/services/snooze_notifier.py` and `tests/test_snooze_notifier.py`). Its shape is largely right — a `SnoozeNotifier` class with `check_once`/`start`/`stop`, wired into `api/main.py`'s `lifespan()` next to `_task_watcher` and `_pebble_capture_watcher`, using `TaskManager.update(..., _precondition=...)` to avoid clobbering a concurrent re-snooze — but has these specific gaps relative to the target behavior above:

1. **No `/health` liveness entry.** `api/main.py`'s `/health` handler's `checks` dict reports `reminder_scheduler` and `scheduler_watcher` liveness but the patch never adds a `snooze_notifier` entry, even though `SnoozeNotifier` follows the exact same "Telegram-gated background thread" shape as `reminder_scheduler` (`_reminder_scheduler.is_alive() if _reminder_scheduler else False`).
2. **Untested unsnooze-during-delivery case.** The salvaged tests cover a concurrent re-snooze (`test_concurrent_resnooze_is_not_cleared`) but not a concurrent full unsnooze (field removed rather than changed) during delivery — the `_precondition` callback should already handle it (`latest.fields.get(...) != raw_until` is true when the field is gone), but there's no test proving it.
3. **No test that a snooze which expired while the process was down still fires.** `check_once` naturally handles this (it has no "since when" state, it just re-scans all tasks), but nothing exercises it explicitly.
4. **No architectural note on why this is a standalone poller rather than a `SchedulerScheduler` entry.** `api/services/scheduler_store.py`'s `SchedulerScheduler` already runs a 60-second background loop with a `notify` action that sends Telegram messages and has its own crash-recovery/backoff — it's the more mature piece of "fire Telegram at a future time" infra in this codebase. The patch doesn't reference it or explain the decision to build a second, separate poller; the Solution section above records the reasoning (schedule entries are operator-authored in the vault, not per-card dynamic state) so a reviewer doesn't have to reconstruct it.
5. **Card identification is description-only**, per the Out of Scope note above — confirm this is still the intended scope before implementing a link.

Deployment is single-process (`scripts/server.sh` launches a single `uvicorn.run(...)`, no `--workers`), matching every other in-process watcher here (`TaskWatcher`, `SchedulerWatcher`, `SchedulerScheduler`) — no multi-worker duplicate-send risk to design around.

- **Key files:**
  - `api/main.py` — `lifespan()` (watcher start/stop, mirror `_task_watcher`) and the `/health` handler's `checks` dict.
  - `api/services/agent_board.py` — `SNOOZED_UNTIL_FIELD`, `parse_snoozed_until`, `natural_lane`, `SNOOZABLE_LANES` (reuse, don't reimplement).
  - `api/services/task_manager.py` — `TaskManager.update(..., _precondition=...)` (already supports the exact precondition pattern the salvage uses).
  - `api/services/telegram.py` — `send_message(text, chat_id=None, bot=None)`; omit `bot` for the primary/system channel.
  - `api/services/scheduler_store.py` — `SchedulerScheduler` (the existing 60s Telegram-delivery watcher; reference for the "why not this" note, not code to reuse).
  - New: `api/services/snooze_notifier.py`, `tests/test_snooze_notifier.py` (salvaged copies are the starting point).
- **Patterns to follow:** `_reminder_scheduler`'s start/stop/`is_alive()` wiring in `api/main.py` for both the lifespan hook and the `/health` entry.
- **Constraints:** no new third-party dependencies; no new configuration/settings beyond what already exists (`settings.telegram_enabled`).

## References

- `docs/specs/product/agent-viz.md` and `docs/specs/technical/agent-viz.md` — Snooze section (the salvaged patch's doc edits are consistent with this issue's target behavior and can be reused).

## History

Deferred from an uncommitted implementation (task 4431c8ca) discovered during the agent-session-reliability milestone's forensics; salvaged rather than lost.

nbramia:
## Summary
- Adds `SnoozeNotifier`, a background thread polling for expired agent-board snoozes every 30 seconds; for each expired card whose natural lane is still snooze-eligible, it sends one Telegram message naming the card through the primary bot and clears `snoozed_until` only after a successful send, guarded by an exact-value precondition so a concurrent re-snooze or unsnooze during delivery is never clobbered.
- Wires the notifier into the API lifespan (start/stop alongside the other background watchers) and reports its own thread liveness at `/health` as `snooze_notifier`, matching `reminder_scheduler`'s shape.
- A failed send, and a failed field-clear write for any other reason, both leave `snoozed_until` in place so the next poll retries rather than losing the notification.
- Updates `docs/specs/product/agent-viz.md` and `docs/specs/technical/agent-viz.md` to describe the notifier and why it's a standalone poller rather than a `SchedulerScheduler` entry.

Closes #1136

## Test plan
- `~/.venvs/lifeos/bin/python -m pytest tests/test_snooze_notifier.py tests/test_agent_board.py tests/test_task_manager.py tests/test_full_health_check.py -q` → 1043 passed
- `~/.venvs/lifeos/bin/python -m pytest tests/test_e2e_flow.py::TestHealthCheck -q` (deselecting the two tests that require an owned candidate server) → 4 passed
- `ruff check` on all touched files → clean
- Each new production-line change was mutation-proven: the field removed, then the change re-applied and tests re-confirmed green.

https://claude.ai/code/session_01W8FaSgWLaoyg36bYGEKYZK


nbramia:
Addressed the Codex review findings:

1. **Delivery semantics (HIGH):** rebased onto current origin/feat/agent-session-reliability, confirming finding 2 (logrotate) was a stale-base artifact — no logrotate files in this branch's diff. Reworded to at-least-once delivery: `SnoozeNotifier` now keeps an in-process record of `(task_id, snoozed_until)` pairs whose message already sent, so a clear that keeps failing is retried without ever resending for the process's lifetime. A duplicate is possible only if the process itself dies between a successful send and its clear. Docs updated in `docs/specs/technical/agent-viz.md`.
2. **Ineligible expiry (MEDIUM):** an expired snooze whose natural lane is ineligible now has `snoozed_until` cleared immediately (same exact-value precondition), so a later unrelated edit back to an eligible lane can't replay a stale wake-up. New test: `test_ineligible_at_expiry_does_not_replay_after_becoming_eligible_later`.
3. **Shutdown truthfulness (MEDIUM):** `stop()` keeps the thread reference until it has actually exited and joins for up to 35s (covers an in-flight Telegram send + its clear) instead of unconditionally discarding it after 5s. New test: `test_stop_does_not_falsely_report_stopped_during_an_in_flight_send`.
4. **Markdown injection (LOW):** the card description is now escaped for Telegram's legacy Markdown parse mode before being sent. New test: `test_description_with_markdown_special_characters_is_escaped`.

Each fix was mutation-proven (reverted, confirmed its test fails, restored). Targeted suite: `tests/test_snooze_notifier.py` (15), `tests/test_agent_board.py` (847), `tests/test_task_manager.py` (179), `tests/test_full_health_check.py` (7) — 1048 passed. `ruff check` clean.

Head SHA: 4d81aeaee7efa2bdee03601ffb0dbc87325267a0

Issue #1150, opened 2026-09-18
feat: support secure Instinct access to the LifeOS MCP


nbramia:
## Problem

LifeOS has a remote MCP transport, but Instinct cannot use it yet.

Today, `mcp_server.py --transport http` exposes JSON-RPC at `/mcp`, binds to `127.0.0.1:8765` by default, and refuses to start unless `LIFEOS_MCP_BEARER_TOKEN` is set. The HTTP transport checks `Authorization: Bearer ...`, and the test suite covers missing and incorrect credentials.

The remaining gap is the connection between Instinct and that endpoint:

- The existing Tailscale Serve setup only makes LifeOS reachable to devices on the same tailnet. Instinct is not a tailnet peer.
- Instinct does not currently have a supported way to register a custom remote MCP server, store its bearer token as a secret, and attach that MCP to an agent.
- Connecting the GitHub repo gives Instinct access to the code, not to the running LifeOS service.

Do not expose the endpoint or send a bearer token in chat until the Instinct-side connector exists.

### LifeOS side

- Keep the MCP server bound to localhost.
- Generate a dedicated, revocable credential for the Instinct connection. Do not reuse a credential used by another client.
- Publish only the MCP endpoint through a stable HTTPS, outbound-only tunnel, following the existing Cloudflare Tunnel pattern in `docs/guides/agent-worker-setup.md`.
- Do not expose the main LifeOS API or add an inbound firewall rule.
- Keep the MCP service disabled unless its credential is configured.
- Start with the minimum tool set Instinct needs, preferably read-only tools first. Expand access deliberately because the current MCP includes tools that can change LifeOS state.
- Redact credentials and authorization headers from application, proxy, and tunnel logs.
- Add a documented rotation and revocation procedure for this client credential.

### Instinct side

Instinct needs a supported custom remote-MCP connection that can:

- Register an exact HTTPS MCP URL ending in `/mcp`.
- Collect and store the bearer token through a secure secret flow rather than chat.
- Send the token in the `Authorization` header without exposing it to the conversation or model context.
- Call `initialize`, `tools/list`, and `tools/call` over the LifeOS HTTP transport.
- Keep the connection disabled by default and let the user remove or rotate it.
- Report authentication, network, and tool-call failures clearly without leaking secrets.

This is an external platform dependency. The LifeOS repo can prepare and document its side, but the connection should not be activated until Instinct supports the connector and secret handling above.

## Acceptance criteria

[...]

Issue #12, opened 2026-02-28
Add data consistency health endpoint


nbramia:
## Problem

\`/health/full\` checks service availability (ChromaDB, API key, reminder scheduler) but not data consistency. No way to know if interactions reference valid person_ids, if stats are stale, or if last sync left inconsistencies — without manually running scripts or tests.

**Existing health endpoints:** Defined in \`api/routes/health.py\` (search for \`/health\` routes)
**Existing data checks:** \`verify_person_stats()\` at \`api/services/person_stats.py:211-282\`, data integrity tests at \`tests/test_p91_data_integrity.py\`

## Proposed Direction

Add \`/health/data-integrity\` endpoint returning JSON:
\`\`\`json
{
  "orphaned_interaction_count": 0,
  "stale_merged_id_count": 0,
  "stats_discrepancy_count": 0,
  "orphaned_source_entity_count": 0,
  "last_check_timestamp": "2026-02-27T03:00:00Z",
  "status": "healthy"
}
\`\`\`

Cache results with 1-hour TTL (these checks hit the DB). Non-zero counts → status "degraded". Reuse verification logic from #7's Phase 7 and from \`verify_person_stats()\`.

## Acceptance Criteria

- [ ] \`/health/data-integrity\` endpoint returns JSON with all counts
- [ ] Non-zero counts trigger warning-level status
- [ ] Results cached (1-hour TTL) for performance
- [ ] Shares verification logic with Phase 7 (#7) — no duplication

## Key Files

- \`api/routes/health.py\` — existing health endpoints (add new route here)
- \`api/services/person_stats.py:211-282\` — \`verify_person_stats()\` (reuse)
- \`scripts/cleanup_interactions.py\` — orphan detection queries to extract
- \`tests/test_p91_data_integrity.py\` — integrity checks to reuse

## Dependencies

#7 (shares verification logic — implement together or extract shared module).

PR #1151, opened 2026-09-19
feat: prepare LifeOS-side infrastructure for secure Instinct MCP access

nbramia:
## TL;DR
LifeOS can now run a second MCP HTTP process, alongside the existing one, that exposes only an operator-chosen, verified read-only subset of tools and authenticates with its own credential — so a new remote client (Instinct) can be connected later without sharing a token with any existing client, and without widening what any client can call beyond what the operator explicitly allows.

Refs #1150

## Design
A per-instance tool allowlist is now opt-in on `LifeOSMCPServer`: an unset allowlist (the existing `:8765` instance's configuration) leaves every registered tool callable, unchanged from today. A configured allowlist restricts the tool set at construction time and fails closed on an unrecognized name (a typo), the same posture the transport already applies to a missing credential. Disallowed tools are hidden from `tools/list` and rejected by `tools/call` before any LifeOS API call.

A named-instance mechanism (`--instance <name>`) lets a second HTTP process read its credential and allowlist from instance-prefixed environment variables (`LIFEOS_MCP_<NAME>_BEARER_TOKEN` / `LIFEOS_MCP_<NAME>_ALLOWED_TOOLS`) instead of the default instance's `LIFEOS_MCP_BEARER_TOKEN` — so rotating or revoking one instance's credential never touches another's. A named instance additionally requires its allowlist to be set; the default instance does not.

A backstop log filter, following the repo's existing Telegram-token redaction pattern, redacts any `Authorization: Bearer <token>` shape that reaches the MCP HTTP process's logs — defense in depth, since nothing on the current auth-failure path logs the token itself.

The dedicated Instinct instance ships as a disabled-by-default systemd unit, mirroring how the existing MCP HTTP unit is only enabled once its credential is configured — this one additionally requires the allowlist. A new operator guide covers setup, rotation, revocation, and teardown with synthetic placeholders only.

## Implementation Notes
- `mcp_server.py`: `LifeOSMCPServer.__init__` gains an `allowed_tools: frozenset[str] | None` kwarg; `_apply_tool_allowlist` filters `self.tools` and raises `ValueError` on an unrecognized name. `dispatch()`'s `tools/call` branch checks the built `_tool_names` before calling `_call_api`. `_load_http_config(instance)` resolves the token/allowlist env-var names and enforces the fail-closed rules; `main()` wires it via a new `--instance` CLI flag. `build_http_app` installs the new redaction filter.
- `api/services/log_redaction.py`: adds `BearerTokenRedactionFilter` / `install_bearer_token_redaction_filter()`, alongside the existing Telegram and query-string filters.
- `config/systemd/lifeos-mcp-instinct.service`: new unit, port 8766, `--instance instinct`.
- `scripts/setup-systemd.sh`: reads `LIFEOS_MCP_INSTINCT_BEARER_TOKEN` / `LIFEOS_MCP_INSTINCT_ALLOWED_TOOLS`, enables the new unit only when both are set, and adds it to the passwordless-systemctl sudoers list alongside the existing MCP HTTP unit.
- `docs/guides/instinct-mcp-access.md` (new): setup, rotation, revocation, teardown. `docs/guides/configuration.md` and root `AGENTS.md`/`docs/guides/AGENTS.md` gain reference entries.
- No new dependencies, no new LifeOS API endpoints, no changes to the existing `:8765` instance's runtime configuration.

### Automated tests
```
HIP_VISIBLE_DEVICES= ROCR_VISIBLE_DEVICES= CUDA_VISIBLE_DEVICES= ~/.venvs/lifeos/bin/python -m pytest tests/test_mcp_http_transport.py tests/test_mcp_server.py -q
```
83 passed.

New coverage: unset-allowlist parity with the existing instance (every tool still listed and callable end to end), a disallowed tool rejected by `tools/call` with a mocked `_call_api` that raises if reached (proving the boundary holds before any API call), construction-time allowlist filtering and typo-guard, `_load_http_config`'s fail-closed rules for both the default and a named instance, the redaction filter's rewrite behavior in isolation, and that `build_http_app` actually wires that filter onto the root logger (verified from a filter-stripped clean state so the assertion can't pass on a leftover install from an earlier test).

Each new behavior was confirmed to fail when its production line was reverted: the dispatch allowlist guard, `_apply_tool_allowlist`'s filtering/typo-guard, the named-instance allowlist-required check, the named-instance credential-env-var derivation, the `build_http_app` → `install_bearer_token_redaction_filter` wiring, and the filter's own substitution logic — each mutation produced the expected test failure, then was reverted back to a clean diff.

### Manual verification
N/A — no runnable artifacts changed in a way that needs a live server: no token was generated or installed, no tunnel was created, `setup-systemd.sh` was not run, and no service was enabled, started, stopped, or restarted, per the task's constraints. The existing `:8765` instance's runtime behavior is unaffected (proven by the automated parity test above, not a manual restart).

## Review focus
- Whether the chosen env-var naming (`LIFEOS_MCP_<INSTANCE>_BEARER_TOKEN` / `_ALLOWED_TOOLS`, `--instance` CLI flag) reads as a sensible generic mechanism for any future dedicated instance, not just Instinct-specific naming (`mcp_server.py`'s `_load_http_config`, `main()`).
- Whether the suggested starter allowlist in `docs/guides/instinct-mcp-access.md` correctly excludes everything state-changing, financial, or home-control (cross-check against `CURATED_ENDPOINTS` in `mcp_server.py`).
- Whether `getattr(server, "_tool_names", None)` in `dispatch()` is the right way to keep a `LifeOSMCPServer.__new__`-constructed test double's pre-allowlist behavior intact (`mcp_server.py` dispatch tools/call branch).


nbramia:

**Specialist rationale:** security — this PR creates a second MCP HTTP instance intended for internet exposure via tunnel to an external client; the tool allowlist and credential isolation are an authentication/privacy boundary.

### Reviewer Findings

#### General

**Defects**
- Bearer-token redaction backstop does not reach uvicorn's own loggers. `build_http_app` installs the filter on root + its current handlers, then `uvicorn.run(..., log_level="info")` applies uvicorn's `LOGGING_CONFIG`, which creates fresh `default`/`access` handlers on `uvicorn`, `uvicorn.error`, `uvicorn.access` with `propagate=False`. Reproduced: after install + `dictConfig(LOGGING_CONFIG)`, `mcp_server` and `httpx` records were redacted but a `uvicorn.error` record emitted `Bearer SYNTHETICTOKEN2` unredacted. Smallest fix: attach the filter to the uvicorn loggers' handlers once uvicorn has configured logging (pass `log_config` with the filter, or install after config).
- New unit is never restarted by auto-deploy. `scripts/auto-deploy.sh:431` drift loop covers only `lifeos-api lifeos-agent-worker lifeos-mcp-http`, so a code change to `mcp_server.py`/`api/` or a `.env` edit (e.g. narrowing `LIFEOS_MCP_INSTINCT_ALLOWED_TOOLS`) never reaches the externally exposed instance. Smallest fix: add `lifeos-mcp-instinct` to that loop (inactive units already skipped; sudoers restart verb already added) and to the unit list in `docs/guides/operations.md:124`.

**Missing Tests**
- No test covers `main()` wiring `allowed_tools` into `LifeOSMCPServer`; reverting to `LifeOSMCPServer()` leaves all 83 tests green while exposing every tool. Add: argv `--transport http --instance instinct` + env token + one-tool allowlist, monkeypatch `run_http`, assert `server.tools` is exactly that tool; plus unknown name → `SystemExit(2)`.
- Redaction wiring test only checks `root.filters`; the per-handler install is what actually redacts propagated child-logger records. Add a capturing-handler test logging `Authorization: Bearer <token>` via `logging.getLogger("mcp_server")`.
- Error-path half of the token-echo criterion untested: only the wrong-token 401 is covered. Add a `_call_api`-raises case asserting the configured token is absent from response and captured logs, plus a no-`Authorization` 401 case.
- "A read-only tool call works end to end" untested on a restricted instance. Add: `allowed_tools=frozenset({"lifeos_health"})` app, stubbed `_call_api`, assert the call reaches the stub and returns no `isError`.

**Status:** Ran `pytest tests/test_mcp_http_transport.py tests/test_mcp_server.py tests/test_settings_documented.py -q` → 84 passed; `bash -n scripts/setup-systemd.sh` OK; read-only probes confirmed default instance registers all 62 curated + 9 inter-agent tools, so the new `_tool_names` guard is behavior-neutral there.

**Summary:** Core mechanism sound and matches operator decisions; all 7 starter tools verified GET or query-only POST. Gaps: redaction misses uvicorn loggers, auto-deploy omission, and missing tests for `main()` wiring, handler-level redaction, error-path token echo, and an allowed call on a restricted instance.

#### Security

**Defects**
- **Path-parameter traversal bypasses the allowlist.** `mcp_server.py:1387` substitutes path params verbatim (`url.replace(f"{{{param}}}", str(arguments.pop(param)))`) and httpx normalizes `..`. With the guide's starter allowlist, `lifeos_person_profile` with `person_id="../../monarch/accounts"` issued `GET /api/monarch/accounts`; `../../gmail/search` and `../../imessage/search` likewise reached those routes. With `lifeos_photos_person` (raw-JSON formatter) the upstream JSON is returned verbatim. Side-effecting GET routes (e.g. `GET /api/crm/slack/callback`) are also reachable. Fix: reject or percent-encode path-param values (`urllib.parse.quote(value, safe="")`).
- **`lifeos_agent_*` tools are accepted into a named-instance allowlist, where the external client can forge caller proofs.** Proof = `HMAC(bearer_token, session_id)`; on the Instinct instance the key is the client-held token, so adding e.g. `lifeos_agent_spawn` lets Instinct impersonate any session and spawn a full-catalog agent. Guide only says these "aren't meant for an external client". Fix: `_load_http_config` refuses a named instance whose allowlist contains `lifeos_agent_*`.
- **Instinct unit does not pin its bind host.** `ExecStart` passes `--port 8766` but no `--host`; `main()` falls back to shared `LIFEOS_MCP_HTTP_HOST` from `.env`. If set to `0.0.0.0` for `:8765`, the Instinct instance listens on all interfaces. Not the case on this host today. Fix: `--host 127.0.0.1` in `ExecStart`.
- **Redaction backstop misses uvicorn loggers and exception tracebacks.** Probe: `uvicorn.error` logged `Authorization: Bearer deadbeefSECRET2` unredacted after uvicorn's `LOGGING_CONFIG`; `logger.exception` traceback text also unredacted (filter rewrites only `record.msg`). Fix: attach to uvicorn handlers; also redact formatted `exc_text`.
- **Revocation step does not persist; auto-deploy never restarts the unit.** Guide revokes with `systemctl stop`, but the unit stays enabled and restarts on reboot (this host self-reboots via its network watchdog) or on `setup-systemd.sh` re-run, with the same token. `auto-deploy.sh:431` omits the unit. Fix: revocation = `disable --now` + remove/rotate the token; add unit to auto-deploy list.

**Missing Tests**
- Path-parameter escape: assert an upstream request captured via `httpx.MockTransport` stays under `/api/crm/people/` for `person_id="../../monarch/accounts"`.
- Batch requests and the `/` route: nothing pins that a JSON-RPC batch or POST `/` cannot reach a non-allowlisted tool (probe shows both rejected today).
- Token echo on error paths: missing-header 401 and dispatch-exception response untested.
- Named instance rejects `lifeos_agent_*` (depends on fix above).

**Status:** Ran `pytest tests/test_mcp_http_transport.py tests/test_mcp_server.py -q` → 83 passed. Throwaway probes (mocked httpx transport, fallback catalog, nothing bound) in the session scratchpad. Verified: starter tools are GET or query-only POST; empty/whitespace token cannot authenticate; named instance never falls back to `LIFEOS_MCP_BEARER_TOKEN`; non-string tool name errors before any call; result cache unused on HTTP path; sudoers entries match existing pattern.

**Summary:** Tool-name allowlist itself holds (fail-closed on missing token/allowlist/unknown names; holds for batch and `/`), but the boundary doesn't: path-param traversal reaches arbitrary GET routes from the starter set, and `lifeos_agent_*` inclusion would allow session impersonation. Bind, redaction, and revocation each have a reachable gap. Path-escape fix must land before exposure.

### Referee Decisions

Referee verified the traversal independently (reproduced `GET /api/monarch/accounts` via allowlisted `lifeos_person_profile`) and the auto-deploy omission (`scripts/auto-deploy.sh:431`). Duplicates collapsed: redaction-vs-uvicorn (general + security) and auto-deploy omission (general + security) are each one finding.

| 1 | Path-param traversal bypasses allowlist | security | Defect | Valid (reproduced) | Accept | Percent-encode every substituted path-param value with `urllib.parse.quote(str(value), safe="")` in `_call_api`, **and** reject a value that is empty, `.` or `..` (quote leaves dots unescaped, so `person_id=".."` would still normalize `/people/../facts` → `/api/crm/facts`). Applies to all instances — also closes the same hole on `:8765`. |
| 2 | `lifeos_agent_*` allowed on named instance → forgeable caller proofs | security | Defect | Valid | Accept | Named instance with any `lifeos_agent_*` in its allowlist fails startup (exit 2) in `_load_http_config`. |
| 3 | Instinct unit doesn't pin `--host 127.0.0.1` | security | Defect | Valid | Accept | Add `--host 127.0.0.1` to the unit's `ExecStart`; fix the guide's "shared host" note. |
| 4 | Redaction backstop misses uvicorn loggers | general + security (dup) | Defect | Valid | Accept (partial) | Make the filter reach uvicorn's `uvicorn`/`uvicorn.error`/`uvicorn.access` handlers (e.g. `run_http` passes a `log_config` copy of uvicorn's `LOGGING_CONFIG` with the filter attached, or equivalent). **Reject** the `exc_text`/traceback-redaction extension: no path puts the token into an exception; speculative hardening. |
| 5 | Auto-deploy never restarts the new unit | general + security (dup) | Defect | Valid (verified) | Accept | Add `lifeos-mcp-instinct` to the drift loop in `scripts/auto-deploy.sh` and to the unit list in `docs/guides/operations.md`. |
| 6 | Revocation via `stop` doesn't survive reboot / setup re-run | security | Defect | Valid | Accept | Guide: revoke = `sudo systemctl disable --now lifeos-mcp-instinct` + remove the token from `.env` (and revoke/rotate); doc-only. |
| 7 | `main()` allowlist wiring untested | general | Missing Test | Valid | Accept | Test via argv + env + monkeypatched `run_http`: tools restricted; unknown name → `SystemExit(2)`; `lifeos_agent_*` on named instance → `SystemExit(2)` (covers #2's test). |
| 8 | Redaction test doesn't pin handler-level / uvicorn coverage | general | Missing Test | Valid | Accept | One behavioral test: capture output of a child logger and of `uvicorn.error` under the logging config `run_http` uses; assert token absent. |
| 9 | Error-path + missing-header token echo untested | general + security (dup) | Missing Test | Valid | Accept | Missing-`Authorization` 401 case; `_call_api`-raises case asserting the configured token is absent from the response body. |
| 10 | Allowed call on restricted instance untested | general | Missing Test | Valid | Accept | `allowed_tools={"lifeos_health"}` app, stubbed `_call_api` reached, no `isError`. |
| 11 | Batch and `/` route rejection untested | security | Missing Test | Valid | Accept | Parametrize the disallowed-tool test over a batch body and POST `/`. |
| 12 | Path-param escape test | security | Missing Test | Valid | Accept | `httpx.MockTransport`-captured upstream URL stays under `/api/crm/people/` for `person_id="../../monarch/accounts"`. |

**Findings forwarded to addresser:** 12 (item 4 partial)

nbramia:

**Specialist rationale:** security — the round-1 fix delta changed the allowlist boundary itself (path-param handling, agent-tool refusal, host pin, uvicorn redaction); re-verification of those closures was needed.

### Reviewer Findings

#### General

**Defects**
- **Finding #1 (path-param traversal) only partially fixed — sibling routes still reachable.** `_call_api` percent-encodes `/` → `%2F`, but uvicorn decodes `%2F` back to `/` in `scope["path"]` before Starlette routes, so the encoded value still adds segments server-side. Cross-prefix escape is closed (`../../monarch/accounts` → 404; httpx has nothing to normalize and Starlette doesn't resolve `..`). But with the starter allowlist, `lifeos_person_profile` with `person_id="<id>/timeline"` reaches `GET /api/crm/people/{id}/timeline` (the route behind the non-allowlisted `lifeos_person_timeline`); likewise `/facts`, `/connections`, `/contact-sources`, `/source-entities`, `/strength`, `/timeline/aggregated`. Reproduced on a real uvicorn: `new 'abc123/timeline' (200, {"route":"timeline","person_id":"abc123"})`. Today's leak is limited by the profile formatter, but the "a tool maps to exactly one endpoint" boundary still doesn't hold, and `test_path_param_traversal_is_percent_encoded_not_normalized` only checks the httpx raw path, not the decoded route. Smallest fix: also reject any value containing `/` (every route uses single-segment `{param}`, so a `/` value already 404s before and after this PR); test that `person_id="x/timeline"` errors with no request.
- **Fix delta breaks an existing test:** `tests/test_mcp_human_queue.py::TestCallApiDispatch::test_resolve_dispatches_put_with_key_in_url_and_note_in_body` — `quote(value, safe="")` now encodes `:`, so `sync:gmail` becomes `sync%3Agmail`. Behavior is still correct (API decodes `%3A`), but the test is in the fast-unit gate. Fix: update the assertion with justification, or add `:` to `quote`'s `safe` set.
- **`tests/test_no_change_narration.py::test_no_change_narration_beyond_baseline` fails** on two hits from the original commit: `docs/guides/instinct-mcp-access.md:17` ("…unchanged from before this mechanism existed") and `mcp_server.py:2375-2376` (same phrase). Fix: reword to describe current behavior.

**Missing Tests:** none beyond the stronger path-param test folded into the first defect.

**Status:** 273 passed / 2 failed across `test_mcp_http_transport.py`, `test_mcp_server.py`, `test_deploy_drift.py`, `test_settings_documented.py`, `test_mcp_human_queue.py`, `test_gmail_draft_send_gate.py`, `test_workout_mcp_route.py`, `test_tasks_api.py`, `test_log_redaction.py`, `test_no_change_narration.py` (the 2 failures above); `test_scheduler_bot_validation.py`, `test_tool_filter.py`, `test_agent_worker_mcp_exposure.py` pass. Toy-FastAPI/uvicorn probes: old vs new encoding route/decode identically for normal values (memory-search queries with spaces/unicode/`+`/`&`/`%`, event ids with `_`/`@`); `uvicorn.Config(log_config=_uvicorn_log_config_with_redaction()).configure_logging()` on uvicorn 0.41.0 keeps level/format and doesn't disturb root/`mcp_server` redaction. No services touched.

**Summary:** Round-1 findings #2–#12 resolved. #1 partially resolved (sibling-route reach via decoded `%2F`). Two fast-unit gate failures (human-queue URL assertion; narration) must be fixed.

#### Security

**Defects:** none.

**Missing Tests**
- Nothing pins that `run_http` actually passes the redacting `log_config` to uvicorn; the new test calls `_uvicorn_log_config_with_redaction()` + `dictConfig` directly, so reverting `run_http` to `uvicorn.run(app, host=..., port=..., log_level="info")` leaves the suite green. Fix: monkeypatch `uvicorn.run`, call `run_http`, assert every handler in the received `log_config` carries `bearer_redact`.

**Status:** 95 passed (`test_mcp_http_transport.py`, `test_mcp_server.py`). Probes (scratchpad): toy FastAPI mirroring real `/api/crm/people/*`, `/api/photos/*`, `/api/monarch/accounts` under real uvicorn on 127.0.0.1/ephemeral port, driven by the real `_call_api` (httptools and h11); `uvicorn.Config(log_config=...)` redaction probe. No repo files or services touched.

**Summary:** All five accepted round-1 security findings closed against the listed bypasses (`%2e%2e`, backslash, `?`/`#`, non-str values, padded `..`, U+2024 lookalikes, `{` injection, query/body channels; agent-tool refusal un-bypassable; host pinned; uvicorn handlers redact; revocation docs correct). Residual noted (not raised as a finding): decoded `%2F` lets `person_id="p1/timeline"` reach sibling handlers under `/api/crm/people/p1/`, though it cannot escape the prefix and no raw-JSON tool has sibling routes; the code comment claiming a value "can never introduce a new path segment" is inaccurate; one-line hardening is to reject `/`.

### Referee Decisions

Duplicates collapsed: the general reviewer's sibling-route defect and the security reviewer's residual note are one finding.

| 1 | Allowlisted tool reaches non-allowlisted sibling routes via decoded `%2F` | general (+ security residual) | Defect | Valid (reproduced on real uvicorn) | Accept | The operator decision is that a disallowed tool's endpoint is never reached; `lifeos_person_profile` → `/timeline` reaches `lifeos_person_timeline`'s endpoint. Reject any path-param value containing `/` in the existing guard (no route accepts one). Correct the comment to match. Test: `person_id="x/timeline"` → error, no upstream request. |
| 2 | Encoding `:` breaks `test_mcp_human_queue` URL assertion | general | Defect | Valid | Accept | Tests-are-sacred default is to fix the code: add `:` to `quote`'s `safe` set so the existing wire format (`sync:gmail`) is unchanged. Leave the existing test untouched. |
| 3 | Narration phrases fail `test_no_change_narration` | general | Defect | Valid | Accept | Reword both sites to current-behavior phrasing. |
| 4 | `run_http` → uvicorn `log_config` wiring untested | security | Missing Test | Valid | Accept | Monkeypatch `uvicorn.run`, call `run_http`, assert each handler in the received `log_config` has `bearer_redact`. |

**Findings forwarded to addresser:** 4 (round 1: 12 — converging)

nbramia:
## Convergence Audit — after Round 2

| Sub-problem | Round 1 findings | Round 2 findings | Traces to |
|---|---|---|---|
| Path-param traversal past the tool allowlist | 2 (defect + test) | 1 (sibling-route reach via decoded `%2F`) | Original criterion: disallowed tools never reach the API |
| Path-param encoding side effect on existing tools (`:` in human-queue keys) | — | 1 | Review-introduced (round-1 fix); fixed by preserving the existing wire format |
| Log redaction backstop | 2 | 1 (wiring test) | Original criterion: token absent from logs |
| Agent-tool allowlist, host pin, auto-deploy, revocation docs, main() wiring, error-path/batch/`/` tests | 10 | 0 | Original criteria |
| Narration wording | — | 1 | Repo rule (pre-existing text from the original commit) |

**Density note:** path-param traversal is the only sub-problem with findings in both rounds; round 2's finding was a genuine gap in the round-1 remedy (encoding without rejecting `/`), now closed by rejecting `/`, `""`, `.`, `..`.

**Assessment:** converging (12 → 4 accepted). Remaining work concerns the original task's boundary, not review-introduced architecture. Proceeding to round 3 on the round-2 fix delta (`73c357e9..68583ed2`).

nbramia:

**Reviewers run:** general

**Specialist rationale:** none — the round-2 delta strictly tightened the already-cleared path-param guard (adds rejection of `/`) and adjusted encoding/tests/wording; the general reviewer was directed at the boundary guard and confirmed it by mutation.

### Reviewer Findings

#### General
**Defects:** None.
**Missing Tests:** None.
**Status:** 428 passed, 0 failed — the 11 focused files plus other path-param-exercising files (`test_scheduler_bot_validation.py`, `test_agent_worker_tools.py`, `test_agent_worker_local_executor.py`, `test_audit_e2e.py`, `test_agent_worker_working_dir.py`); both round-2 gate failures (`test_mcp_human_queue`, `test_no_change_narration`) now pass. Mutation checks in a throwaway copy: removing `or "/" in value` fails both new path-param tests; dropping `log_config=` from `run_http` fails the wiring test; reverting `safe=":"` fails the human-queue resolve test. No repo files or services touched.
**Summary:** All 4 round-2 findings resolved with no regressions. `/`-containing values rejected before any request with an accurate comment; no legitimate value lost (every target route is single-segment `{param}`; the only `:path` converters are SPA/voice catch-alls no curated tool uses). `safe=":"` restores the wire format without reopening traversal (values always land in a non-first segment after `API_BASE`). Narration removed. `run_http` wiring pinned.

### Referee Decisions

No findings.

nbramia:
## End-to-End Verification — PR #1151

### Verdict: PASS

verification-record: v1
verification-head: 68583ed2a99567d6a08e3676f41ef26ee7e9e5d6
suite-result: pass
suite-command: ["~/.venvs/lifeos/bin/python -m pytest tests -q --tb=short --durations=25 --durations-min=1.0 --ignore=tests/archive -p no:cacheprovider -p scripts.test_lane_plugin -m \"unit and not browser and not requires_server and not integration and not slow\"", "~/.venvs/lifeos/bin/python -m pytest tests -q --tb=short --durations=25 --durations-min=1.0 --ignore=tests/archive -p no:cacheprovider -p scripts.test_lane_plugin -m \"browser and not requires_server\""]
suite-executions: 1
suite-exit-status: 0
suite-command-results:
  - command: "fast-unit lane (8812 selected)"
    result: pass
    exit-status: 0
  - command: "browser-free lane (979 selected)"
    result: pass
    exit-status: 0
    evidence-pointer: "#/suite-evidence/command-2"
retry-reason: none

Both lanes were executed once, fresh (no prior evidence existed for this exact candidate — `reason: missing`), via this repo's documented local mirror of the hosted gate: `scripts/verify_candidate.py pushed-ref --repository . --sha 68583ed2a99567d6a08e3676f41ef26ee7e9e5d6 --base bb0fe4f74990ec49fb14c01b7539a0d742c433b4 --lanes fast-unit,browser-free`. Persisted evidence key `3c730efb1ee755e22dbdccafc3b7c8f64316c9012f5ba1930be50b7a5eb4d48c`, candidate `8001db50bdd13b08752f0db7`, single attempt, both `provenance.merge_identity` (candidate) and `base_identity` confirmed matching. `fast-unit`: 8811 passed, 1 skipped (the documented no-real-`.env` privacy-audit `not_applicable`), 1383 deselected, exit 0. `browser-free`: 979 passed, 9216 deselected, exit 0.

### System Flow Verified

This PR adds a second, named MCP HTTP transport instance (`--instance <name>`) with a required, fail-closed tool allowlist and its own bearer credential, alongside the existing unrestricted default `:8765` instance. Beyond the authoritative test lanes, I drove the **real** `mcp_server.py --transport http` process as a separate child (never the production `lifeos-mcp-http`/`lifeos-api` services, never `.env`, nothing enabled/restarted) against the real running LifeOS API at `localhost:8000`, with synthetic bearer tokens only.

### Evidence

#### Dedicated "instinct" instance — bind, auth, allowlist boundary, live tool call
**Flow:** spawn `mcp_server.py --transport http --host 127.0.0.1 --port <ephemeral> --instance instinct` with `LIFEOS_MCP_INSTINCT_BEARER_TOKEN=synthetic-instinct-token-9f13ac6e2b4d` and `LIFEOS_MCP_INSTINCT_ALLOWED_TOOLS=lifeos_health,lifeos_search` → JSON-RPC over `/mcp`.
- Bind: `ss -tlnp` local-address column shows `127.0.0.1:<port>` only (no `0.0.0.0`/wildcard listener).
- No `Authorization` header → HTTP 401 `{"detail":"missing bearer token"}`.
- Wrong bearer → HTTP 401 `{"detail":"invalid bearer token"}`.
- Correct bearer: `initialize` → 200, `serverInfo.name == "lifeos"`.
- `tools/list` → exactly `["lifeos_health", "lifeos_search"]`.
- `tools/call lifeos_health` → real end-to-end result (`LifeOS Status: degraded ... chromadb: healthy ...`), `isError` absent.
- `tools/call lifeos_monarch_accounts` (not allowlisted) → `isError: true`, text `"Error: Unknown tool: lifeos_monarch_accounts"` — the allowlist-rejection message, not a Monarch API response, proving `_call_api`/upstream was never reached.
- Process stopped (SIGTERM); captured stdout+stderr (1217 bytes) does not contain the synthetic token.
**Result:** PASS

#### Path-param escape attempt (separate instance, allowlist `lifeos_health,lifeos_person_profile`)
**Flow:** `tools/call lifeos_person_profile` with adversarial `person_id` values.
- `person_id="sibling/timeline"` → `isError`, `"Error: Invalid person_id: 'sibling/timeline'"` (rejected before any request; would otherwise decode-route into `.../timeline`).
- `person_id` of `".."`, `"."`, `""` → same guard, same rejection.
- Control: `person_id="synthetic-nonexistent-entity-id-000"` (normal shape) → reaches the real API and gets a genuine `404 Person '...' not found` — proving the guard targets escape shapes specifically, not blanket-blocking.
- Synthetic token absent from captured process log.
**Result:** PASS

#### Startup refusals (exit 2), one short-lived process each
- No `LIFEOS_MCP_INSTINCT_BEARER_TOKEN` → exit 2, `"HTTP transport requires LIFEOS_MCP_INSTINCT_BEARER_TOKEN..."`.
- Token but no `LIFEOS_MCP_INSTINCT_ALLOWED_TOOLS` → exit 2, `"Named instance 'instinct' requires LIFEOS_MCP_INSTINCT_ALLOWED_TOOLS..."`.
- Typo in allowlist (`lifeos_helth`) → exit 2, `"Unknown tool name(s) in allowlist: lifeos_helth"`.
- `lifeos_agent_spawn` in allowlist → exit 2, `"Named instance 'instinct' cannot allowlist inter-agent tools: lifeos_agent_spawn..."`.
**Result:** PASS

#### Default instance parity (no `--instance`, allowlist unset)
**Flow:** spawn with only `LIFEOS_MCP_BEARER_TOKEN=synthetic-default-instance-token-4d8e` (no `--instance`, no allowlist var).
- `tools/list` → 71 tools (every registered tool, including `lifeos_monarch_accounts` and `lifeos_agent_spawn`) — unchanged from today's `:8765` behavior.
- No bearer → 401.
- Synthetic token absent from captured process log.
**Result:** PASS — confirms the existing unrestricted instance's behavior is unaffected by this change.

### Issues Found
None.

### Holistic Assessment
The dedicated-instance allowlist mechanism holds end to end on the real system: bind is 127.0.0.1-only, auth fails closed (missing/wrong token → 401, missing token/allowlist/typo/agent-tool-in-allowlist at startup → exit 2), `tools/list` and `tools/call` both honor the allowlist (disallowed tools never reach `_call_api`), a path-param escape attempt is rejected before any upstream request while normal values still reach the real API, the bearer token never appeared in any captured process log, and the pre-existing unrestricted default instance is provably unchanged. Combined with the fully-passing fast-unit + browser-free authoritative lanes (9,790 tests, 0 failures) for this exact candidate, the system works correctly as a whole after this change.

PR #1178, opened 2026-09-20
fix: close leaked SQLite connections in agent worker + viz cache stores

nbramia:
## Summary
Several stores use `with sqlite3.connect(...) as conn:` — sqlite3's connection context manager only commits/rolls back on exit, it never closes the connection, so every call leaks a file descriptor. Fixed here: `SessionStore` and `UsageLedger` (agent worker), `agent_viz_summary.py`'s disk cache (5 call sites), `SpendTracker`, and `CaptureLedger` (`pebble_capture.py`). Observed in production: 303 fds on a macOS worker after ~2 days (256 limit), 44 fds on the Linux worker after 79 minutes, and 88 fds on `data/agent_viz_summaries.db` from a live API process after 23 minutes.

Each `_connect()`/`_disk_*` call site now opens through a `contextlib.contextmanager` that nests the original `with conn:` (preserving the existing commit/rollback semantics) and closes the connection in a `finally` block. No call sites changed.

Adds `tests/test_sqlite_connections_close.py`: a ratchet scanning `api/` for the raw `with sqlite3.connect(...)` pattern, with a shrinking allowlist for 7 modules not yet migrated (28 sites, tracked as a followup). A match outside the allowlist fails the scan; an allowlisted file that no longer matches also fails, so the allowlist can't go stale.

## Test plan
- [x] Witness test for the worker stores (`tests/test_agent_worker_stores.py::test_session_store_connect_does_not_leak_file_descriptors`): fails on unmodified `session_store.py` (7 → 110 fds over 50 iterations), passes after the fix.
- [x] Witness test for the viz cache (`tests/test_agent_viz_api.py::test_disk_cache_helpers_do_not_leak_file_descriptors`): fails on unmodified `agent_viz_summary.py` (5 → 139 fds over 50 iterations), passes after the fix.
- [x] Full worker/session-store test suite (70 files, 1962 tests) and the agent_viz_summary/spend_tracker/pebble_capture suite (24 files, 804 tests) pass.
- [x] `ratchet` tests pass.

https://claude.ai/code/session_01PUAZNx8fu7azpXPbfHKKiY


nbramia:

**Specialist rationale:** the change alters SQLite connection lifetime across five stores used from worker threads and API request threads — the lifecycle standard's "resource ownership" trigger — so an independent adversarial correctness review of connection/transaction semantics was required in addition to the general review.

### Reviewer Findings

#### General
**Defects:** none.

**Missing Tests:** `tests/test_sqlite_connections_close.py`'s regex (`with\s+(?:_DB_LOCK,\s*)?sqlite3\.connect\(`) only matches the inline shape that `agent_viz_summary.py` had. The other four modules' bug shape was a `_connect()` method doing a bare `return conn`, with callers writing `with self._connect() as conn:`; that text never contains `with sqlite3.connect(`. Reviewer confirmed concretely that reintroducing `SessionStore._connect`'s original body with callers unchanged produces zero ratchet matches. Proposed smallest fix: also flag a `_connect`-style method whose body opens `sqlite3.connect(` and returns/yields the connection without an enclosing `with conn:` / `finally: conn.close()`.

**Status:** ran both witness tests (pass, skip cleanly off Linux, exercise the changed helpers), the ratchet, `tests/test_no_change_narration.py`, and the full `test_agent_worker_stores.py`, `test_agent_viz_api.py`, `test_pebble_capture.py`, `test_usage_ledger.py`, `test_session_store_test_isolation.py` files: 267 passed. `ruff check` clean on all 8 files. Traced ~140 `_connect` call sites: no connection, cursor, or Row is used after its `with` block; `.rowcount`/`.lastrowid` reads after the block are post-close-safe. Nested `with conn:` preserves commit/rollback timing including the explicit `BEGIN IMMEDIATE` sites; `_DB_LOCK` still releases after close.

#### Correctness
**Defects:** none.

**Missing Tests:** none.

**Status:** ran `test_sqlite_connections_close.py`, `test_agent_worker_stores.py`, `test_agent_viz_api.py` (105 passed) plus `test_pebble_capture.py`, `test_usage_ledger.py` (160 passed) on head 15ff363d. Reproduced the leak against main's unmodified `session_store.py` with the witness loop: 7 → 101 fds over 50 iterations; 105/105 pass on the fix. Verified by execution: cursor `.rowcount`/`.lastrowid` survive `close()` while `.fetchone()` after close raises; on an `isolation_level=None` connection `commit()`/`rollback()` with no open transaction are no-ops and an exception inside an explicit `BEGIN IMMEDIATE` block is rolled back by the nested `with conn:`; no `except sqlite3.*` retry-on-same-connection pattern exists in the changed files; no connection is cached on `self` across threads; `_DB_LOCK` in `agent_viz_summary.py` now spans a strict superset of its prior scope (through `close()`).

### Referee Decisions

| 1 | Ratchet regex cannot detect a reintroduced bare-`return conn` `_connect()` in four of the five fixed modules | general | Missing Test | Valid | Reject for this PR; remedied in #1182 | The concern is real: the regex pins only the inline shape. Two of the four modules (`SessionStore`, `agent_viz_summary`) are independently pinned by the witness tests in this PR. A regex over method bodies is fragile; the robust rule is "only `api/services/sqlite_connect.py` may call `sqlite3.connect(`", which becomes possible in the already-open stacked #1182 where every module moves onto that single helper. #1182 will carry the strengthened ratchet (allowlisting only the helper and any module with its own verified closing helper). Not a deferral to an unplanned follow-up: #1182 is queued to merge immediately after this PR. |

nbramia:
## End-to-End Verification — PR #1178

### Verdict: PASS

verification-record: v1
verification-head: 15ff363df52818f719a09866da622c674967feb5
suite-result: pass
suite-command: "python scripts/verify_candidate.py pushed-ref --repository . --sha 15ff363df52818f719a09866da622c674967feb5 --lanes fast-unit,browser-free --workers 1"
suite-executions: 1
suite-exit-status: 0
suite-command-results:
  - command: "python scripts/verify_candidate.py pushed-ref --repository . --sha 15ff363df52818f719a09866da622c674967feb5 --lanes fast-unit,browser-free --workers 1"
    result: pass
    exit-status: 0
retry-reason: none

### System Flow Verified

`SessionStore`/`UsageLedger`/`SpendTracker` back the agent worker's per-tick read/write cycle (`api/services/agent_worker/worker.py` and its executors); `CaptureLedger` backs Pebble capture ingestion; `agent_viz_summary`'s disk cache backs `/agents` viz summaries. All five previously leaked one fd per `_connect()`/`with sqlite3.connect(...)` call because sqlite3's own context manager never calls `.close()`. The fix wraps each in a `contextlib.contextmanager` that nests the original `with conn:` (preserving commit/rollback) and closes in `finally`. Every call site already used `with self._connect() as conn:` (verified by grep across all 5 modules), so no caller changed shape.

### Evidence

#### Authoritative lane evidence (fast-unit + browser-free, per AGENTS.md's documented hosted-gate contract)
**Command:**
```
python scripts/verify_candidate.py pushed-ref --repository . --sha 15ff363df52818f719a09866da622c674967feb5 --lanes fast-unit,browser-free --workers 1
```
**Result:** PASS — both lanes exit 0, 8987 fast-unit + 985 browser-free node IDs selected and passed (one `not_applicable` privacy-audit node, expected off a real `.env`), against an isolated worktree of exact head `15ff363d`, non-reused (fresh execution, no prior receipt existed for this candidate).

#### Concurrency/resource-ownership evidence — independent live-process check
Beyond the in-repo witness tests, ran a fresh interpreter process importing `SessionStore` directly from the checked-out head and driving 200 real worker-tick-shaped calls (4x the witness test's 50 iterations) against a temp db:
```
baseline_fds_on_db=0 after_200_ticks_fds_on_db=0 total_process_fds=4
PASS: no fd growth across 200 real SessionStore calls in a fresh process
```
**Result:** PASS — confirms zero fd growth from a real, non-pytest process exercising the exact production code path, corroborating the in-repo witness tests (`test_session_store_connect_does_not_leak_file_descriptors`, `test_disk_cache_helpers_do_not_leak_file_descriptors`).

#### Review status
Round 1 (general + correctness, required by the concurrency/resource-ownership risk trigger) converged with zero accepted findings; one valid ratchet-coverage gap (regex only matches the inline `with sqlite3.connect(` shape, misses the 4 modules' `_connect()`-method shape) was explicitly deferred to stacked PR #1182, which is queued to merge immediately after this one and carries the strengthened ratchet.

### Issues Found
None.

### Holistic Assessment
The fix closes a real, production-observed fd leak (303 fds/2 days on the macOS worker against a 256 limit, 88 fds/23 min on the Linux API's viz cache) across all five affected stores, with no call-site behavior change, full authoritative-lane evidence at exit 0, and independent live-process confirmation of the resource-ownership fix beyond unit tests. Safe to merge.

Issue #1179, opened 2026-09-20
Close SQLite connections in the remaining stores that use sqlite3's context manager without closing


nbramia:
## Summary
`with sqlite3.connect(...) as conn:` only commits or rolls back on exit — it never closes the connection, so every call leaks a file descriptor. PR #1178 fixed this pattern in `SessionStore`, `UsageLedger`, `agent_viz_summary.py`'s disk cache, `SpendTracker`, and `CaptureLedger`, and added a ratchet test (`tests/test_sqlite_connections_close.py`) that scans `api/**/*.py` for the raw pattern. That ratchet currently allowlists 7 modules that still need the same fix:

| File | Sites |
|---|---|
| `api/services/perf_trace.py` | 5 |
| `api/services/usage_store.py` | 8 |
| `api/services/agent_viz_label_override.py` | 4 |
| `api/services/gsheet_sync.py` | 4 |
| `api/services/hermes_persona_thread_store.py` | 3 |
| `api/services/hermes_question_thread_store.py` | 3 |
| `api/services/job_queue.py` | 1 |

28 sites total. Each of these modules' `_connect`-style method (or inline `sqlite3.connect(...)` call) needs the same closing-contextmanager treatment: open the connection, nest the original `with conn:` for unchanged commit/rollback behavior, close it in `finally`.

## Acceptance criteria
- All 28 sites across the 7 files above close the connections they open.
- `tests/test_sqlite_connections_close.py`'s allowlist is empty.
- Each touched module's existing tests still pass; `ruff check` is clean.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01PUAZNx8fu7azpXPbfHKKiY

nbramia:
## Delivered

**PR:** #1203 — fix: close SQLite connections in the remaining stores that leak them

### Changes delivered
- Added `api/services/sqlite_connect.py`'s `connect_closing` contextmanager, nesting the original `with conn:` for identical commit/rollback behavior and closing the connection in a `finally` block.
- Migrated the remaining 28 bare `sqlite3.connect(...)` call sites across `perf_trace.py`, `usage_store.py`, `hermes_persona_thread_store.py`, `hermes_question_thread_store.py`, `gsheet_sync.py`, and `agent_viz_label_override.py` onto `connect_closing`.
[...]

PR #1182, opened 2026-09-20
fix: close SQLite connections in the remaining stores that leak them

nbramia:
## Summary
`with sqlite3.connect(...) as conn:` only commits or rolls back on exit — it never closes the connection, so each of the 28 call sites across `perf_trace.py`, `usage_store.py`, `hermes_persona_thread_store.py`, `hermes_question_thread_store.py`, `job_queue.py`, `gsheet_sync.py`, and `agent_viz_label_override.py` leaks a file descriptor per call. Adds `api/services/sqlite_connect.py`'s `connect_closing`, a small contextmanager that nests the original `with conn:` for the same commit/rollback behavior and closes the connection in a `finally` block, and uses it at every one of those sites. `job_queue.py`'s one violation (`_init_db`) instead calls the class's own `JobQueue._conn`, which already provides the same guarantee — no need for a second mechanism in that file.

`SessionStore._connect`, `UsageLedger._connect`, `agent_viz_summary._connect`, `SpendTracker._connect`, and `CaptureLedger._connect` (fixed in #1178) now delegate their open/close boilerplate to the same `connect_closing` helper too, keeping their extra per-store setup (`row_factory`, PRAGMAs) local — one pattern across all 14 modules.

`tests/test_sqlite_connections_close.py`'s scan is broadened to catch any `sqlite3.connect(` under `api/`, not just the `with` form (a regression could reintroduce the leak as a bare `return sqlite3.connect(...)`, which the narrower pattern would miss). Every module outside `sqlite_connect.py` that still calls `sqlite3.connect(` directly is now on an explicit, individually-verified allowlist: `_ALLOWLIST_VERIFIED_SAFE` closes every connection it opens (`try`/`finally` or `contextlib.closing`, or a single connection held for the object's life with its own explicit close), and `_ALLOWLIST_KNOWN_GAP` names six modules with a narrower leak-only-on-an-exception-path gap. Six modules with exception-path-only leaks are allowlisted as a known gap, tracked in #1184.

Base branch is `fix/session-store-connection-leak` (#1178) since this stacks on it; rebase onto `main` once #1178 merges.

Closes #1179

## Test plan
- [x] `tests/test_sqlite_connections_close.py` (no violations outside the allowlist, no stale allowlist entries) passes; verified it fails when a bare `return sqlite3.connect(...)` is reintroduced into a migrated `_connect` method (then reverted).
- [x] The witness tests from #1178 (worker stores, viz cache) still pass after migrating their `_connect` methods to `connect_closing`.
- [x] Each touched module's existing test suite (32 files found via grep, ~700+ tests) passes; the only errors are 7 pre-existing `test_perf_benchmark.py` cases that require a live candidate server instance (`LIFEOS_TEST_BASE_URL` etc.), unrelated to this change and reproducible on unmodified `main`.
- [x] `tests/test_no_change_narration.py` passes.

https://claude.ai/code/session_01PUAZNx8fu7azpXPbfHKKiY


nbramia:
Superseded by #1203: GitHub would not let this PR be reopened after its base branch (`fix/session-store-connection-leak`) was deleted on merge and the head branch was subsequently force-pushed during the rebase onto `main` ("state cannot be changed... branch was force-pushed or recreated"). #1203 has the identical commits, description, and issue references, retargeted directly at `main`.

Issue #1191, opened 2026-09-20
feat(ci): publish an explicit success for docs-only candidates without running the lanes


nbramia:
## Problem

A candidate that changes only documentation runs the full fast-unit lane in hosted CI, twice, for about 66 runner-minutes and ten minutes of wall time, and no test can observe the change. The local pre-push path already treats a docs-only change as nothing to run, but the hosted lane-selection step has no such case. Three of the last forty merged PRs were docs-only.

## Solution

The trusted runner's lane-selection step classifies the changed set with the same docs-only rule the local plan uses, including the carve-out that keeps dependency manifests in the code class. A docs-only candidate skips the environment install and the lane run, and the execution job records `docs-only` as its verification mode. The publisher publishes the required check as success with a summary stating that no retained lane could be affected. The decision is an explicit trusted outcome, not the absence of a job. Any failure to compute the changed set keeps today's behavior and runs every lane.

For shadow runs the changed set is the diff against the merge base, so a branch that is behind the base still qualifies when its own changes are docs-only.

## Acceptance Criteria

- [ ] Given a candidate whose changed files all match `\.(md|txt|rst)$` or live under `docs/`, and none match `(requirements|constraints)[^/]*\.txt$`, when lane selection runs, then the install and verify steps shall be skipped and the job output `verification_mode` shall be `docs-only`.
- [ ] Given a candidate that changes `requirements.txt` and nothing else, the run shall execute the fast-unit lane.
- [ ] Given a candidate that changes a file under `docs/` and one under `api/`, the run shall execute the fast-unit lane.
- [ ] If the changed set cannot be computed, the run shall execute every retained lane, as today.
- [ ] When a docs-only candidate is published, the check summary shall state that the candidate was docs-only and name the rule applied.
- [ ] The docs-only rule shall live in one trusted-runner script shared by the workflow step and covered by unit tests, and `scripts/test.sh`'s `decide_plan` shall keep passing its existing plan tests.
- [ ] The publisher summary shall distinguish `executed`, `docs-only`, and `reused` verification modes.

## Verification

- **Automated tests:** unit tests for the classification function with the four cases above; `tests/test_candidate_ci_boundary.py` asserts the selection step precedes the install step and that the verify step is conditioned on the mode.
- **Manual checks:** a docs-only PR publishes green `candidate-verification` in under two minutes.
- **Existing test suite:** must continue to pass, including `tests/test_test_sh_auto.py`.

## Scope

**In scope:**
[...]

PR #1196, opened 2026-09-20
chore(ci): retain lane receipts from every verification run

nbramia:
Each execution part now uploads its lane-execution receipts (node IDs, outcomes, per-module durations; never test output) as `lane-receipts-<sha>-part<n>` with 30-day retention, whatever the part's result. The failure-only lane-log upload is unchanged and the publisher still consumes no artifact. The CI guide documents downloading a run's receipts and regenerating `scripts/lane_scope_durations.json` from them.

Closes #1192

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Hxgcj6oZvtDKY1neU5B3Bz

PR #1198, opened 2026-09-20
feat(ci): publish an explicit success for docs-only candidates

nbramia:
Stacked on #1196; retargets to `main` once that merges.

The hosted runner classifies the changed set with `scripts/candidate_lanes.py` (trusted-runner checkout) before any environment is built. A docs-only candidate installs and executes nothing and records `docs-only` as the job's verification mode; every other candidate records `executed` and runs the lanes its changes can affect. The publisher publishes success only when the execution job passed *and* recorded a mode, and its check summary names the mode. The docs-only rule mirrors `test.sh`'s `decide_plan`, and `tests/test_candidate_lanes.py` holds the two to the same answers. The changed set prefers the merge-base diff, falling back to the two-commit diff as a superset.

Closes #1191


nbramia:

**Reviewers run:** general

**Specialist rationale:** none — the trust boundary is unchanged (the classifier runs from the trusted-runner checkout; the publisher gains a stricter success condition).

### Reviewer Findings

#### General
**Defects:** The rewritten "Select the lanes" step diffs `$TRUSTED_RUNNER_SHA..$CANDIDATE_SHA` inside the `candidate` checkout, but that checkout is a depth-1 clone of the candidate alone; the step on `main` first ran `git -C candidate fetch --no-tags --depth=1 origin "$TRUSTED_RUNNER_SHA"` and this PR dropped it. Both diff forms therefore fail, `CHANGED` is always empty, the mode is always `executed`, and the existing `web/` narrowing regresses to running both lanes.

**Missing Tests:** nothing executed the step's shell against two real shallow clones, so the missing fetch was invisible to the substring assertions in `tests/test_candidate_ci_boundary.py`.

**Status:** focused checks passed (78 tests, ruff clean, YAML parses). Matrix-output → publisher wiring confirmed correct; empty `LANES=` for docs-only confirmed harmless because every consumer is gated on `mode == 'executed'`; the fail-closed and trusted-checkout invariants are still pinned.

### Referee Decisions

| 1 | Base commit never fetched into the candidate clone; changed set always empty | general | Defect | Valid — reachable on every hosted run | Accept | Restore the fetch ahead of the diff chain (applied in the follow-up commit). |
| 2 | No test runs the step's shell against two shallow clones | general | Missing Test | Valid | Accept | Two tests extract the step's `run:` from the YAML and execute it against a base clone and a head clone laid out like the workflow's, pinning docs-only, `web/`, and api-only outcomes (applied). |

**Findings forwarded to addresser:** 2 (applied inline; routine change outside the six risk triggers).

nbramia:

**Reviewers run:** general

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none. **Status:** focused checks passed; the round-1 remedies were verified against the real mechanics (shell semantics reproduced by hand, step gating and key propagation traced, guide prose current-behavior-only).

### Referee Decisions

No findings to evaluate.

nbramia:
Superseded by #1202: this PR was auto-closed when its stacked base branch was deleted on merge of #1196. The branch is unchanged apart from a rebase onto main; the review record above stands for #1202.

PR #1199, opened 2026-09-20
feat(ci): reuse a passing shadow verification for an identical candidate tree

### LINKED ISSUE #1190: feat(ci): reuse a passing shadow verification for an identical candidate tree
## Problem

Every merged pull request is verified twice by the same trusted runner, and the second run usually re-tests a tree the first run already passed. The shadow check runs on every push to the PR head. The required check runs again on the candidate commit the publisher builds from the base and that head. When the branch is already up to date with the base, the candidate's tree is byte-for-byte the head's tree, so the required run repeats work with no new information. Over the last 80 merges the two trees were identical 49 times. Each repeat costs about 33 runner-minutes and adds roughly ten minutes to the time between "green" and "merged."

## Solution

The trusted runner, before installing anything, checks whether a passing shadow check already covers the exact tree and runner it is about to verify. It reads the check runs on the candidate's second parent, keeps only those published by the dedicated App under the shadow check name, and accepts one whose recorded tree and trusted-runner commit both match. On a match the execution job records that the verification was reused, skips the environment install and the lane run, and the publisher publishes the required check as success with a summary naming the reused check. On no match, or any doubt, the run proceeds exactly as today.

The publisher records the verified tree and trusted-runner commit as structured fields on every check it publishes, so the match is against data the App wrote, never against candidate-controlled text.

## Acceptance Criteria

- [ ] When the publisher creates a check run, the check's output text shall carry a JSON object with `candidate`, `tree`, `trusted_runner`, and `lanes`, in addition to the existing summary line.
- [ ] Given a `workflow_dispatch` run whose candidate tree equals its second parent's tree, and a successful `candidate-verification-shadow` check on that second parent published by the dedicated App with matching `tree` and `trusted_runner`, when the execution job starts, then it shall skip the install and verify steps and expose `reused_check_id` as a job output.
- [ ] Given the same run but a shadow check whose `trusted_runner` differs, or whose conclusion is not `success`, or that was published by any other actor, then the run shall execute the lanes as today.
- [ ] Given a `workflow_dispatch` run whose candidate tree differs from its second parent's tree, the run shall execute the lanes as today.
- [ ] If the check-run query fails, returns malformed JSON, or the parent cannot be fetched, the run shall execute the lanes as today.
- [ ] When a run reuses a shadow check, the published `candidate-verification` summary shall name the reused check run id and the shadow's candidate SHA.
- [ ] A `pull_request_target` run shall never attempt reuse.
- [ ] `tests/test_candidate_ci_boundary.py` shall assert the reuse step precedes the install step, runs only under `workflow_dispatch`, and that the publisher still never checks out or consumes an artifact from the candidate.

## Verification

- **Automated tests:** extend `tests/test_candidate_ci_boundary.py` for the workflow structure; unit-test the reuse decision as a pure function in a new trusted-runner script (`scripts/candidate_reuse.py`) fed synthetic check-run JSON: matching, wrong runner, wrong app, wrong conclusion, malformed.
- **Manual checks:** one PR whose branch is rebased on `main` shows the required check published within about a minute of dispatch with a "reused" summary; one PR behind `main` still runs the lanes.

## Scope

**In scope:**
- The reuse step and job output in the execution job; the structured check output in the publisher; the pure decision script and its tests; a paragraph in the CI guide.

**Out of scope:**
- Reusing anything from a run on a different trusted-runner commit, even if the verifier files are unchanged. Keep the match exact.
- Reusing a shadow result from a different tree. No "close enough."
- Changing the publisher's dispatch or wait logic in `scripts/candidate_publisher.py`; the wait already accepts any check the App publishes for the candidate SHA.

## Technical Context

- **Key files:** `.github/workflows/candidate-verification.yml` (execution job steps at lines ~84–133; publisher script at lines ~245–270 where `github.rest.checks.create` is called), `scripts/candidate_publisher.py:444` (`_is_required_check`, the App-id filter to mirror), `scripts/candidate_publisher.py:291` (`build_candidate`: candidate is `commit-tree <merged tree> -p base -p head`, so the second parent is the PR head), `tests/test_candidate_ci_boundary.py`, `docs/guides/candidate-verification-ci.md` § "The actual gate".
- **Patterns to follow:** the existing "Bind dispatched runner to protected workflow provenance" step reads the candidate's parents with `git cat-file commit`; fetch the second parent with `git -C candidate fetch --depth=1 origin <sha>` and compare `rev-parse <sha>^{tree}` values. The check-run query uses `gh api repos/{repo}/commits/{sha}/check-runs` as the publisher does. Grant the execution job `checks: read` only.
- **Constraints:** the execution job must not gain any write permission or any secret. The App id comes from `vars.LIFEOS_CANDIDATE_APP_ID`; when it is empty, reuse is disabled. All shell in the trusted runner comes from the `trusted-runner` checkout, never from `candidate`.

## References

- #1153 — parent issue
- `docs/guides/candidate-verification-ci.md`

nbramia:
Stacked on #1198; retargets to `main` once that merges.

A dispatched candidate whose tree is byte-for-byte its head's tree has already been verified by this same runner in the head's shadow run (49 of the last 80 merges). Before building any environment, the runner reads the App-published check runs on the head with a read-only token and, through `scripts/candidate_reuse.py`, reuses a `candidate-verification-shadow` verdict only when the App published it, its conclusion is success, its recorded `trusted_runner` and `tree` equal the run's own, its mode is `executed`, and its lanes cover every selected lane. The run records `reused` as its verification mode and the publisher names the reused check. Every published check now carries a structured record (`candidate`, `tree`, `trusted_runner`, `mode`, `lanes`, `conclusion`) in its output text, with tree and lanes computed by the selection step before any candidate code runs. Any mismatch executes the lanes as usual; a shadow run never reuses anything.

Closes #1190


nbramia:

**Specialist rationale:** security — the change lets the required check be satisfied without executing the lanes, a new decision on the merge-gate trust boundary.

### Reviewer Findings

#### General
**Defects:** none. The `gh api | python3 | tee` pipeline in the reuse step fails open as required: no step sets `shell: bash`, so the implicit `bash -e {0}` has no `pipefail` and the step's status is `tee`'s (reproduced locally both ways). Fetching an arbitrary reachable SHA into a depth-1 clone works on GitHub.com (verified against this repo); the list check-runs endpoint returns `output.text` (verified); a skipped reuse step leaves `steps.reuse.outputs.mode` empty so `!= 'reused'` holds; newest-verdict and lanes-superset rules are pinned by tests.

**Missing Tests:** the criterion that the summary names the reused check id and shadow candidate is implemented (job outputs → publisher env → summary) but no test pins that wiring.

**Status:** focused checks passed (47 tests, ruff clean, YAML parses).

#### Security
**Defects:** none. Tree, lanes and mode are computed by the trusted checkout before candidate code runs, and per-step output files mean a later step cannot rewrite an earlier step's resolved outputs. `trusted_runner` equality pins the verifier, registry and selection rules, hence the lanes. The read-only token lives only in the reuse step's env. `mode == "reused"` is itself never reusable. The second-parent extraction reads header lines that git places before the message body, so a message line starting with `parent <hex>` cannot redirect it. `RESULT` is success only when every part succeeded.

**Missing Tests:** nothing exercises the second-parent extraction against a real two-parent commit whose message carries an adversarial `parent <hex>` line.

**Status:** focused checks passed (46 tests).

### Referee Decisions

| 1 | Reused id / candidate wiring not pinned | general | Missing Test | Valid | Accept | Assertions on the job outputs, the publisher env, and the summary template (applied). |
| 2 | Second-parent extraction untested against a forged message | security | Missing Test | Valid | Accept | One test builds a two-parent commit via `commit-tree` with `parent ffff…` lines in its message and runs the step's own extraction line (applied). |
| — | Pipeline fragility should `shell: bash` ever be added | general | (note, not a finding) | Not reachable | Reject | Speculative hardening; recorded here so a future `shell:` change knows to re-check. |

**Findings forwarded to addresser:** 2 (test-only additions, applied inline; no substantive change, so no confirmation round).

PR #1200, opened 2026-09-20
feat(ci): cache the installed CPU test environment across candidate runs

### LINKED ISSUE #1193: feat(ci): cache the installed CPU test environment across candidate runs
## Problem

Each of the four execution parts spends about 100 seconds of its roughly 510 seconds installing the same CPU torch wheel, the same requirements, and the same Chromium before it can run a test. The pip download cache is already on, so the remaining cost is unpacking and installing, which the cache does not cover. That is about a fifth of every part and roughly 400 runner-seconds per run.

## Solution

The execution job restores a prebuilt test environment keyed on everything that determines it: runner image, Python version, the torch CPU version, the hash of the requirements file, and a schema version. On a hit, the install step verifies the environment and continues. On a miss, the install step runs as today and saves the environment immediately after installing and before any candidate test code runs, so the saved environment is the product of wheel installation only. The identity checks that torch is the declared CPU build stay in place on both paths.

## Acceptance Criteria

- [ ] When the execution job starts, it shall attempt to restore a cache whose key includes `runner.os`, the Python version, `TORCH_CPU_VERSION`, `hashFiles('candidate/requirements.txt')`, and a literal schema version.
- [ ] Given a cache hit, the install step shall skip pip and Playwright installation and shall still run the torch identity assertions and the dependency fingerprint.
- [ ] Given a cache miss, the install step shall install as today and a save step shall run before the verify step; no step after the verify step shall write to the cache.
- [ ] The install shall keep `--only-binary=:all:` so no source distribution executes during environment construction.
- [ ] Measured on a hit, the combined restore-plus-install steps shall take under 45 seconds on a hosted runner; record the before and after timings in the PR description.
- [ ] `tests/test_candidate_ci_boundary.py` shall assert the save step precedes the verify step and that the workflow still pins the cache action by SHA.

## Verification

- **Automated tests:** workflow-structure assertions in `tests/test_candidate_ci_boundary.py`.
- **Manual checks:** two consecutive runs of the same PR; the second shows the hit and the timing target.

## Scope

**In scope:**
- Restore and save steps, the venv layout that makes the environment cacheable, the guide note.

**Out of scope:**
- Caching apt packages that Playwright's `install-deps` pulls; keep that step and measure it separately.
- A prebuilt container image.

## Technical Context

- **Key files:** `.github/workflows/candidate-verification.yml:111-127` (setup-python and the install step), `tests/test_candidate_ci_boundary.py:60-77` (pinned actions and the torch identity strings the test expects; keep those exact strings), `docs/guides/candidate-verification-ci.md`.
- **Patterns to follow:** create the environment in a venv under `$RUNNER_TEMP/lifeos-venv` and put its `bin` on `GITHUB_PATH`, so one directory holds everything cacheable; cache `~/.cache/ms-playwright` alongside. Use `actions/cache/restore` and `actions/cache/save` pinned by SHA like the other actions.
- **Constraints:** never save after the verify step. The pip freeze fingerprint on a hit must equal the fingerprint the miss path produced for the same key; if it differs, fail the step.

## References

- #1153 — parent issue, "Adjacent" section

nbramia:
Stacked on #1199; retargets to `main` once that merges.

Each execution part restores a venv (CPU torch + `requirements.txt`) and Playwright's Chromium from a cache keyed on the runner image, interpreter version, `TORCH_CPU_VERSION`, the requirements hash, and a schema version. On a miss the install step builds it with `--only-binary=:all:` and records a package fingerprint inside the venv; the save step runs immediately after, before any candidate code, and nowhere later. On a hit the install activates the venv, installs Chromium's system packages (never cached), and fails unless the restored fingerprint equals the recorded one. The torch identity assertions run on both paths.

Timing before/after will be recorded here once the workflow is live on `main` (the YAML only takes effect after merge).

Closes #1193


nbramia:

**Specialist rationale:** security — the change introduces a cache that other candidates restore, a cross-PR surface the workflow did not have.

### Reviewer Findings

#### General
**Defects:** none. All eight GitHub Actions semantics questions verified against upstream source: `cache-hit`/`cache-primary-key` are real restore outputs; `~` expands in `path:`; `hashFiles` resolves under the workspace; `steps.python.outputs.python-version` is real; the venv's absolute interpreter paths are stable because the key carries the exact resolved Python version; `playwright install-deps` is valid; no legitimate fingerprint drift on a hit; the torch install strings survive verbatim; `GITHUB_PATH` makes the verify step and its `sys.executable`-spawned pytest use the venv. Concurrent parts saving the same key only warn.

**Missing Tests:** the boundary test never asserts the `echo "$VENV/bin" >> "$GITHUB_PATH"` line that makes the cached venv take effect downstream.

**Status:** focused checks passed (20 tests, YAML parses). The sub-45 s hit timing criterion is unmeasurable before merge because the workflow YAML only takes effect on `main`; disclosed in the PR body.

#### Security
**Defects:** the "pure function of the requirements file" premise is false: nearly every entry in `requirements.txt` is an open range and there is no `--require-hashes`, so a hit, which skips resolution entirely, pins every PR sharing the key to whatever resolved when the key was first built, for as long as the cache stays warm. A dependency release, security fixes included, would not be exercised by the gate until something busted the key. Suggested: state the tradeoff and bound the key's lifetime.

**Missing Tests:** nothing checks that `requirements.txt` is fully pinned, which the stated invariant relied on.

**Status:** focused check passed (19 tests). No new cache-poisoning path: exact-key restore with no `restore-keys`, the backend refuses to overwrite an existing key, and any attacker-controlled content changes the key.

### Referee Decisions

| 1 | `GITHUB_PATH` line not pinned | general | Missing Test | Valid | Accept | One assertion in the existing cache test (applied). |
| 2 | Hit freezes open-range resolution indefinitely | security | Defect | Valid — reachable on every hit | Accept | Add the ISO week to the key (one `date` step, bounded to a week's staleness) and correct the guide's framing; the fingerprint is described as a consistency check (applied). |
| 3 | No test that `requirements.txt` is fully pinned | security | Missing Test | Valid concern, wrong remedy | Reject | Pinning the manifest is a dependency-policy change outside this issue; the accepted remedy in #2 removes the invariant the test would have guarded. |

nbramia:

**Reviewers run:** general

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none. **Status:** focused checks passed; the round-1 remedies were verified against the real mechanics (shell semantics reproduced by hand, step gating and key propagation traced, guide prose current-behavior-only).

### Referee Decisions

No findings to evaluate.

PR #1201, opened 2026-09-20
feat(ci): report shadow-mode test-impact selection without changing the gate

### LINKED ISSUE #1194: feat(ci): report shadow-mode test-impact selection without changing the gate
## Problem

Nobody knows which tests a given change can actually affect, so every code candidate runs the whole fast-unit lane. The parent issue proposes a static import-graph selector, but an ast-level measurement shows the suite is an import ball: a change to almost any service module reaches about three quarters of the lane through the application entry point, and across the last forty merged PRs such a selector would have kept 66% of lane time. That number is an estimate from one snapshot, and whether the selector would have caught every real failure is unknown. Narrowing the required check on an unmeasured selector is not acceptable.

## Solution

The trusted runner computes what an import-graph selector would have chosen for each candidate and records it next to the real result, without changing what runs. After the lanes execute, the summary states the selected module count and duration share, and whether any failing module fell outside the selection. Once enough runs accumulate, the data decides whether selection is worth shipping and with which fail-closed rules.

## Acceptance Criteria

- [ ] A trusted-runner script shall build a static import graph over `api/`, `config/`, `scripts/`, `tests/`, and `mcp_server.py` and, given a changed-file list, return the set of test modules whose import closure reaches any changed file, plus every changed test module.
- [ ] Given a change to `tests/conftest.py`, any non-`test_` file under `tests/`, a dependency manifest, a workflow file, or any verifier input, the script shall return `full` with the triggering path.
- [ ] Given a changed file with an extension the script does not model, it shall return `full` with that path.
- [ ] When the verify step finishes, the step summary shall record `impact_selection` with `mode`, `selected_modules`, `total_modules`, `selected_duration_share`, and `missed_failures` (failing modules outside the selection).
- [ ] The set of tests that execute shall be identical with and without the report.
- [ ] The script shall run from the `trusted-runner` checkout against the candidate tree and never import candidate code.

## Verification

- **Automated tests:** unit tests over a synthetic package tree for closure, the fail-closed triggers, and the unmodeled-extension case; one test asserting that the report never alters the verifier's node-id selection.
- **Manual checks:** three recent PRs replayed locally show plausible selections.

## Scope

**In scope:**
- The selector script, its tests, the report in the step summary and in the receipt artifact.

**Out of scope:**
- Narrowing the gate. Any narrowing is a later issue that cites this report's data.
- Coverage-based mapping; it adds a dependency and belongs to a later decision.

## Technical Context

- **Key files:** new `scripts/test_impact.py`; `.github/workflows/candidate-verification.yml:134-183` (where the changed set is computed and the verify step runs); `scripts/verify_candidate.py:44` (`RUNNER_INPUTS`, the verifier-input list to reuse as a full trigger), `scripts/verify_candidate.py:147-152` (summary JSON fields), `scripts/test_lane_plugin.py:103-123` (per-nodeid outcomes to derive failing modules), `scripts/lane_scope_durations.json` (for the duration share).
- **Patterns to follow:** resolve imports with `ast` only; treat `from pkg import name` as reaching both `pkg` and `pkg/name.py` when the latter exists; ignore relative imports with a level (none exist in this tree).
- **Constraints:** graph construction must fail closed to `full` on any exception. Keep the script dependency-free.

## References

- #1153 — parent issue, item 3
- https://claude.com/blog/agentic-coding-is-straining-ci-heres-how-we-scaled-test-impact-analysis-at-anthropic

nbramia:
Stacked on #1200; retargets to `main` once that merges.

After the lanes run, each execution part records what a static import-graph selector (`scripts/test_impact.py`, run from the runner's checkout, parsing the candidate tree with `ast` and never importing it) would have selected: the mode, the selected modules and their share of recorded duration, the changed paths the model cannot see, and which failing modules in that part fell outside the selection. The selector fails closed to `full` for the conftest or any helper under `tests/`, dependency manifests, the workflow, verifier inputs, unmodeled file types, an empty changed set, or any error. The report lands in the receipts artifact (its own directory, so the duration record never globs it) and in the step summary. Nothing reads it; the verifier's arguments are unchanged.

Measured on the current tree, a change to almost any `api/` module reaches ~76% of lane time through `api/main.py`, so this data decides whether selection is worth shipping before anything narrows the gate.

Closes #1194


nbramia:

**Reviewers run:** general

**Specialist rationale:** none — measurement-only change; the selector runs from the trusted checkout and nothing consumes its output.

### Reviewer Findings

#### General
**Defects:** adding `lifeos-impact/impact_selection.json` as a second path on the receipts artifact changes that artifact's root to the paths' common ancestor (`$RUNNER_TEMP`), nesting the real receipts under `lifeos-lane-logs/`. The documented `record-scope-durations` command globs non-recursively and would find zero receipts on every ordinary run.

**Missing Tests:** the boundary assertions were textual and could not see the artifact layout change.

**Status:** focused checks passed (47 tests, ruff clean, YAML parses). Verified: the `__main__` path insert makes `scripts.verify_candidate` resolve to the trusted-runner copy and its transitive imports are stdlib-only; the `always()` guard degrades gracefully when receipts are missing; a `.py` outside the modeled roots is dropped rather than failing closed but is out of the issue's stated graph scope and still covered by `missed_failures`; renames and empty duration records are handled.

### Referee Decisions

| 1 | Second artifact path moves the receipts' root | general | Defect | Valid — every executed run | Accept | Upload the report as its own artifact `impact-selection-<sha>-part<n>`; receipts artifact keeps its single path (applied). |
| 2 | Layout regression invisible to textual assertions | general | Missing Test | Valid | Accept | Pin the receipts artifact to exactly one `path:` line and the report to its own artifact (applied). |

nbramia:

**Reviewers run:** general

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none. **Status:** focused checks passed (47 tests, YAML parses, ruff clean). Verified against upload-artifact v4.6.2 semantics that a single file path roots the new artifact at that file and the receipts artifact's root stays the lane-log directory; `if-no-files-found: ignore` is correct for the docs-only and reused modes; the test slices are scoped to the right steps; guide prose is current-behavior-only.

### Referee Decisions

No findings to evaluate.

PR #1205, opened 2026-09-21
fix(ci): save the cached test environment from a trusted job only

nbramia:
The execution job's token carries no cache-write scope (the first live run logged `cache write denied: token has no writable scopes`), so its save step never persisted anything and every run missed. Granting that scope to the job that runs candidate code would reopen the poisoning path the security review closed, so the save moves to a new `prepare-environment` job: it checks out the protected runner alone, builds the venv wheels-only from the runner's `requirements.txt`, and saves under the same key shape the execution job restores with. It is the only job with `actions: write`; the execution job only restores. A candidate whose requirements file equals the protected one restores the shared environment; one that changes the file misses, installs fresh from its own file, and saves nothing.

Refs #1193


nbramia:

**Specialist rationale:** security — a new job writes a cache that every later run restores, and the change alters which job holds write access.

### Reviewer Findings

#### General
**Defects:** `permissions: actions: write` does not govern the Actions cache token. GitHub scopes it by trigger trust (read-only for `pull_request_target`, writable for `workflow_dispatch`) and by the `cache-mode` key at workflow or job level ([changelog 2026-06-26](https://github.blog/changelog/2026-06-26-read-only-actions-cache-for-untrusted-triggers/), [changelog 2026-09-10](https://github.blog/changelog/2026-09-10-control-github-actions-cache-access-with-cache-mode/), [workflow syntax](https://docs.github.com/actions/reference/workflows-and-actions/workflow-syntax#cache-mode)). Without `cache-mode`, the new job's save fails on every shadow run exactly as the execution job's did.

**Missing Tests:** none (server-side token behavior).

**Status:** focused checks passed (24 tests, YAML parses, ruff clean). Keys are byte-identical across the two jobs when the requirements files match; `lookup-only` populates `cache-hit` and `cache-primary-key`; a week-boundary straddle degrades to a plain miss.

#### Security
**Defects:** the new cache-writing job consumes the dispatch input `trusted_runner_sha` as its checkout ref without the provenance binding the execution job applies (`test "$TRUSTED_RUNNER_SHA" = "$WORKFLOW_SHA"` under `workflow_dispatch`). A caller able to dispatch the workflow could point it at a controlled commit and persist that commit's requirements into the shared cache. Bounded today because the publisher is the only dispatcher and derives the value from `main` itself.

**Missing Tests:** no test pins that the builder rejects a forged `trusted_runner_sha` the way `test_provenance_binding_step_rejects_mismatched_or_forged_runner_sha` does for the execution job.

**Status:** focused check passed (23 tests). Moving the save out of the candidate-running job correctly closes the original poisoning path; cache scoping and the fingerprint-as-consistency-check wording are accurate.

### Referee Decisions

| 1 | Cache write is governed by `cache-mode`, not `permissions` | general | Defect | Valid — every `pull_request_target` run | Accept | Top-level `cache-mode: read`; the builder job alone declares `write`; `actions: write` removed (applied, commit 396b6702). |
| 2 | Builder job trusts the dispatch input without provenance binding | security | Defect | Valid — reachable by any dispatcher | Accept | Same binding step as the execution job, first in the builder job, before checkout (applied, commit ae5443e2). |
| 3 | No test for the builder's binding | security | Missing Test | Valid | Accept | Executes the step's own script with matching, mismatched, empty, and `pull_request_target` inputs (applied). |

**Findings forwarded to addresser:** 3 (applied inline).

nbramia:

**Reviewers run:** general

**Specialist rationale:** none — round 2 covers the two fix deltas from round 1 (cache-mode pinning; the builder's provenance binding).

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none. **Status:** focused checks passed (25 tests, YAML parses, ruff clean). Verified against GitHub's workflow-syntax reference and the 2026-09-10 changelog that `cache-mode` is a real top-level and job-level key with job overriding workflow, that `pull_request_target` defaults to `read` and `workflow_dispatch` to `write`, that `write` (not `write-only`) is required because the builder restores with `lookup-only` before saving, and that a denied `setup-python` pip-cache save warns rather than fails. The binding step added in ae5443e2 mirrors the execution job's check and is exercised by its own test.

### Referee Decisions

No findings to evaluate.

Issue #1153, opened 2026-09-19
ci: adopt test impact analysis + per-test history in candidate verification


nbramia:
## Summary

Anthropic's post [*Agentic coding is straining CI — here's how we scaled test impact analysis*](https://claude.com/blog/agentic-coding-is-straining-ci-heres-how-we-scaled-test-impact-analysis-at-anthropic) says agent-written code multiplied their CI load 25× in six months. They argue that running every test on every change doesn't scale. Their answer is **test impact analysis (TIA)**, built from two parts: a *listener* that records every test result into a per-test history, and a *selector* that uses that history plus package relevance to pick which tests each PR runs. The other lessons:

- keep state out of the process
- avoid singleton CI services
- instrument "jobs in == jobs out" so an agent can spot a stall
- plan for another large jump in load

LifeOS has the same shape of problem at a smaller scale: almost every PR here is written by an agent. This issue lists the principles LifeOS doesn't apply yet and what adopting them would take.

## Current state (measured 2026-09-19)

- **Selection is per lane, not per test.** `candidate-verification.yml` always runs the whole `fast-unit` lane (**~8,800 tests**, ~4,180 s of recorded scope time in `scripts/lane_scope_durations.json`). The only narrowing is dropping `browser-free` when no `web/` file changed.
- **Docs-only PRs still run the full `fast-unit` lane in hosted CI.** Locally, `test.sh`'s `decide_plan` already maps docs-only changes to `skip`. The hosted lane-selection step has no such case.
- **Cost per gate:** a 4-part matrix at ~7–10 min wall time each. About 1m45s of every part is the uncached `torch` CPU + `requirements.txt` + Playwright Chromium install. That's ~30–40 runner-minutes per candidate.
- **Each merged PR is verified at least twice:** the `pull_request_target` shadow run on every push, plus the `workflow_dispatch` required run. Over the last 200 runs (2026-09-11 → 09-19): 136 shadow runs (59 cancelled, 14 failed) and 64 dispatch runs (10 cancelled, 6 failed). Each run holds 5 jobs, so a handful of concurrent candidates saturates the account's concurrent-job cap.
- **No per-test history.** Results aren't kept across runs. `lane_scope_durations.json` is a static snapshot committed once (`4f0df8c3`), and partition balance drifts as tests change.
[...]

PR #1206, opened 2026-09-21
feat(ci): add a flow check that every gate run finished with a published check

nbramia:
`scripts/gate_flow_check.py` reads the verification workflow's runs from the last day and the check runs on each run's commit, and reports every run the gate started but did not finish with a published check: **stalled** (still queued or in progress past the 25-minute ceiling) or **unpublished** (completed, but no check of the expected name published by the dedicated App on its commit). Cancelled runs are superseded ones and are not counted. Exit 0 when balanced, 2 on an imbalance, 1 when the API could not be read; `--notify telegram|email` delivers the report only on an imbalance. It changes nothing. Guide section added; a recurring run belongs in the host's crontab or a systemd timer.

Implements item 5 of #1153 (the capacity deadlock behind the earlier gate hang looked like slowness from the outside; this is the cheap signal that distinguishes the two).


nbramia:

**Reviewers run:** general

**Specialist rationale:** none — a read-only reporting script; no trust boundary, schema, or running-application behavior changes.

### Reviewer Findings

#### General
**Defects:** the workflow-runs call fetches one page of 100 and never paginates; a live 24 h run already counted 84 runs, so a busy day would silently drop the oldest runs in the window and could report "balanced" over a truncated list.

**Missing Tests:** no test pins pagination; every fixture returns one short page.

**Status:** focused checks passed (15 tests, ruff clean); live read-only run reported balanced (62 counted, 22 cancelled). Confirmed empirically that a `workflow_dispatch` run's `head_sha` is the candidate SHA the check is published on; the lazy Telegram import has no import-time side effects; exit codes and never-notify-when-balanced are pinned.

### Referee Decisions

| 1 | Runs list truncated at one page | general | Defect | Valid — reachable at observed volume | Accept | Follow pages while a page comes back full, bounded at 20 pages with an error beyond that (applied). |
| 2 | Pagination unpinned | general | Missing Test | Valid | Accept | Tests for a full first page followed by a second carrying the stalled run, a short single page, and the bound (applied). |

nbramia:

**Reviewers run:** general

**Specialist rationale:** none — round 2 covers only the pagination delta.

### Reviewer Findings

#### General
**Defects:** none. **Missing Tests:** none. **Status:** focused checks passed (18 tests, ruff clean). `page`/`per_page` combine with the `created` filter per GitHub's docs; an exact multiple of 100 yields one empty page the loop returns on; the fixture change keeps every earlier test meaningful. Note for awareness only: the `created` filter caps results at 1000 server-side, far above observed volume (~84 runs/day).

### Referee Decisions

No findings to evaluate.

